PR #4, opened 2020-06-30
Ignore maven-remote-resources-plugin:bundle goal

demobox:
m2eclipse does not know how to map the maven-remote-resources-plugin:bundle goal, which causes a warning on project import

Since the more common "process" goal is already ignored, it would seem useful to ignore this goal too.

### DIFF
--- a/org.eclipse.m2e.lifecyclemapping.defaults/lifecycle-mapping-metadata.xml
+++ b/org.eclipse.m2e.lifecyclemapping.defaults/lifecycle-mapping-metadata.xml
@@ -79,12 +79,13 @@
         <artifactId>maven-remote-resources-plugin</artifactId>
         <versionRange>[1.0,)</versionRange>
         <goals>
+          <goal>bundle</goal>
           <goal>process</goal>
         </goals>
       </pluginExecutionFilter>
       <action>
         <ignore>
-          <message>maven-remote-resources-plugin (goal "process") is ignored by m2e.</message>
+          <message>maven-remote-resources-plugin is ignored by m2e.</message>
         </ignore>
       </action>
     </pluginExecution>


mickaelistria:
I'm not that enthusiast about adding such stuff in m2e. Efforts would rather be directed towards the "offending" Maven plugins to work better inside Eclipse IDE (and other incremental builders) https://www.eclipse.org/m2e/documentation/m2e-making-maven-plugins-compat.html

demobox:
> I'm not that enthusiast about adding such stuff in m2e. Efforts would rather be directed towards the "offending" Maven plugins to work better inside Eclipse IDE (and other incremental builders)

I can understand that. Without having looking into it, I suspect it would be quite a lot more work, though, and the current situation - the warning - definitely feels suboptimal to users.

I agree that this change doesn't "fix" the underlying issue, but I think it would at least create a better user experience, especially since m2e *already* skips some goals (although I'm not aware of the underlying reasons for that, which may be different).

demobox:
> I'm not that enthusiast about adding such stuff in m2e. Efforts would rather be directed towards the "offending" Maven plugins to work better inside Eclipse IDE (and other incremental builders)

I can understand that. Without having looking into it, I suspect it would be quite a lot more work, though, and the current situation - the warning - definitely feels suboptimal to users.

I agree that this change doesn't "fix" the underlying issue, but I think it would at least create a better user experience, especially since m2e *already* skips some goals (although I'm not aware of the underlying reasons for that, which may be different).

demobox:
> I'm not that enthusiast about adding such stuff in m2e. Efforts would rather be directed towards the "offending" Maven plugins to work better inside Eclipse IDE (and other incremental builders)

I can understand that. Without having looking into it, I suspect it would be quite a lot more work, though, and the current situation - the warning - definitely feels suboptimal to users.

I agree that this change doesn't "fix" the underlying issue, but I think it would at least create a better user experience, especially since m2e *already* skips some goals (although I'm not aware of the underlying reasons for that, which may be different).

demobox:
> I'm not that enthusiast about adding such stuff in m2e. Efforts would rather be directed towards the "offending" Maven plugins to work better inside Eclipse IDE (and other incremental builders)

I can understand that. Without having looking into it, I suspect it would be quite a lot more work, though, and the current situation - the warning - definitely feels suboptimal to users.

I agree that this change doesn't "fix" the underlying issue, but I think it would at least create a better user experience, especially since m2e *already* skips some goals (although I'm not aware of the underlying reasons for that, which may be different).

demobox:
> I'm not that enthusiast about adding such stuff in m2e. Efforts would rather be directed towards the "offending" Maven plugins to work better inside Eclipse IDE (and other incremental builders)

I can understand that. Without having looking into it, I suspect it would be quite a lot more work, though, and the current situation - the warning - definitely feels suboptimal to users.

I agree that this change doesn't "fix" the underlying issue, but I think it would at least create a better user experience, especially since m2e *already* skips some goals (although I'm not aware of the underlying reasons for that, which may be different).

PR #5, opened 2020-07-01
Bug 564787 - NPE when resolving remote Filesystem project with Eclips…

ivy-rew:
…eWorkspaceArtifactRepository

- Use EFS cache to create real local java.io.File representations of remotely available POM.xml files
- Avoid conversion to java.io.File at all and prefer InputStream reading since that works also for remote FS.

Change-Id: I0000000000000000000000000000000000000000

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/MavenModelManager.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/MavenModelManager.java
@@ -16,6 +16,7 @@ package org.eclipse.m2e.core.embedder;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
+import java.io.IOException;
 import java.io.InputStream;
 import java.util.Collection;
 import java.util.Collections;
@@ -93,12 +94,25 @@ public class MavenModelManager {
     return maven.readModel(reader);
   }
 
+  /**
+   * @deprecated use {@link #readMavenModel(InputStream)} instead.
+   */
+  @SuppressWarnings("deprecation")
+  @Deprecated
   public org.apache.maven.model.Model readMavenModel(File pomFile) throws CoreException {
     return maven.readModel(pomFile);
   }
 
+  /**
+   * @deprecated use {@link #readMavenModel(InputStream)} instead.
+   */
+  @Deprecated
   public org.apache.maven.model.Model readMavenModel(IFile pomFile) throws CoreException {
-    return maven.readModel(pomFile.getLocation().toFile());
+    try (InputStream is = pomFile.getContents()) {
+      return maven.readModel(is);
+    } catch(IOException ex) {
+      throw new CoreException(new Status(IStatus.ERROR, IMavenConstants.PLUGIN_ID, -1, null, ex));
+    }
   }
 
   public void createMavenModel(IFile pomFile, org.apache.maven.model.Model model) throws CoreException {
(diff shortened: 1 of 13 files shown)


ivy-rew:
contains the same patch as gerrit change 4: https://git.eclipse.org/r/plugins/gitiles/m2e/m2e-core/+/3d0483677a293ab948cccc658c794732fe1c1a98

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/MavenModelManager.java
```diff
@@ -93,12 +94,25 @@ public MavenModelManager(IMaven maven, IMavenProjectRegistry projectManager) {
     return maven.readModel(reader);
   }
 
+  /**
+   * @deprecated use {@link #readMavenModel(InputStream)} instead.
+   */
+  @SuppressWarnings("deprecation")
+  @Deprecated
   public org.apache.maven.model.Model readMavenModel(File pomFile) throws CoreException {
     return maven.readModel(pomFile);
```
mickaelistria, line 103: Let's use
```java
try (InputStream is = new FileInputStream(pomFile)) {
  return maven.readModel(is);
} catch (IOException ex) {
  throw new CoreException(new Status(IStatus.ERROR, IMavenConstants.PLUGIN_ID, -1, null, ex));
}
```
and get rid of the @SuppressWarnings.

mickaelistria: actually, it's ok like this, let's not bother more than necessary ;)

mickaelistria:
This change looks good. I hope I'll be able to fix the CI job today to get it running the tests and so on to validate it before we can merge.

mickaelistria:

ivy-rew:
Yes sure, I've just rebased on top of the m2e-core/master should be up to date now.
Hopefully the CI will now like my change too :)

mickaelistria:
That's better ;) However, as described in https://github.com/eclipse-m2e/m2e-core/blob/master/CONTRIBUTING.md#%EF%B8%8F-version-bump , you need to increase the version of the bundle you modified (and of including references) in case no change did happen since last release.
So please move version of org.eclipse.m2e.core to 1.16.1, and you'll probably need to bump the version of the features that include it by the way.
Running `mvn clean verify` focally should reproduce the same issue and help you to remediate them.

mickaelistria:
The last commit is mostly to be discarded. Please first rebase on top of master, and then *only* update the version of the bundles you modified.
What specifically in https://github.com/eclipse-m2e/m2e-core/blob/master/CONTRIBUTING.md#%EF%B8%8F-version-bump makes you assume that all versions need to be bumped? Why are you changing dependency ranges and so on for bundles that don't require it?

ivy-rew:
The thing is that almost every bundle has a dependency to m2e.core .... which I have changed. So I thought they must be adjusted too. 
But I'll gladly drop that commit and bump only the one that I've actually touched. Since most bundles have a range dependency this should be an easy change.

ivy-rew:
I have one additional test that did not fail on master, but on my branch (MissingSchemaMarkerTest). Locally it runs just fine .... so I hope its just a flaky test. Maybe someone with rights on the build node can trigger the build once again ?
MissingSchemaMarker-localRun

mickaelistria:
Yes, it's a flaky test and can be ignored.
Thanks for your contribution, I'm going to merge it.

ivy-rew:
Thanks a lot for your help and guidance @mickaelistria 
One last thing, I absolutely like the switch to Github as SCM/Review platform. It makes contributions way easier for newcomers. 
Keep up the good work.

PR #8, opened 2020-07-02
Longer timeout

mickaelistria:

PR #9, opened 2020-07-02
Bug 564880 - Mal-formed template is ignored

mbooth101:
The hyphen "-" is not a valid identifier part for template
variables, so instead we use the special value "${cursor}"
to place the cursor next to "UTF-8" in case the user wishes
to change it

Also adds a template for the similar but different property
for setting the encoding of report plugin output.

Change-Id: I20a269b787f8569c8af5909b69cbc0d148fab6a8
Signed-off-by: Mat Booth <mat.booth@redhat.com>


mbooth101:
Hi @mickaelistria,

The CI build failed with:

```
[ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:1.6.0:compare-version-with-baselines (check-no-version-regression) on project org.eclipse.m2e.editor.xml: Only qualifier changed for (org.eclipse.m2e.editor.xml/1.16.0.20200702-2216). Expected to have bigger x.y.z than what is available in baseline (1.16.0.20200514-2159) -> [Help 1]
```

I guess m2e has not had post-release version-bump yet?

mickaelistria:
> I guess m2e has not had post-release version-bump yet?

m2e doesn't systematically bump versions any longer, it uses the same approach as Platform: keep versions the same unless there is a change. So please bump version for this plugin with this same patch.

mbooth101:
> m2e doesn't systematically bump versions any longer, it uses the same approach as Platform: keep versions the same unless there is a change. So please bump version for this plugin with this same patch.


Alright, cool. I bumped some versions, and now it fails like this:

```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-surefire-plugin:1.6.0:test (default-test) on project org.eclipse.m2e.tests:
An unexpected error occurred while launching the test runtime (return code 143).
See log /home/jenkins/agent/workspace/m2e_PR-9/m2e-core-tests/org.eclipse.m2e.tests/target/work/data/.metadata/.log for details.
```

https://ci.eclipse.org/m2e/job/m2e/view/change-requests/job/PR-9/8/console

I can't read the log in the jenkins workspace and all tests pass locally for me. Any hints?

mickaelistria:
> I can't read the log in the jenkins workspace and all tests pass locally for me. Any hints?

This is an issue is a frozen test triggering the timeout kill and happens on all builds lately, it's not related to your patch. I've been trying to troubleshoot it with various hacks you can see on #8 , but without success so far. The annoying part is that tests work like a charm locally.
I'm currently struggling to have the build leaving some logs to inspect...

mickaelistria:

mbooth101:
Hi @mickaelistria I rebased my change. However, earlier you said this:

> m2e doesn't systematically bump versions any longer, it uses the same approach as Platform

So I bumped the service component by ".100" in accordance with platform guidelines. But I noticed you bumped the service component by only ".1" in your changes... So which do you prefer? And should I change my patch to bump only by ".1"?

mickaelistria:
I think increasing with +0.0.1 is enough. The Platform rule comes from the fact that Platform used to maintain multiple streams simultaneously IIRC, which is something we do not plan to do for m2e; so we go for the more straightforward +0.0.1

mbooth101:
> I think increasing with +0.0.1 is enough. The Platform rule comes from the fact that Platform used to maintain multiple streams simultaneously IIRC, which is something we do not plan to do for m2e; so we go for the more straightforward +0.0.1

Ok thanks, I updated my change one more time 1.16.100 -> 1.16.1 :-)

mbooth101:
Well, CI runs to completion now \ó/

FWIW the 2 test failures also occur on the master branch, so are unrelated to my change.

mickaelistria:
Thanks!

PR #10, opened 2020-07-03
Bug 478647: Refactor MavenPomSelectionComponent and callers to enable search providers

mpiggott:
I had previously opened this a long time ago but it got lost somewhere.

Basically the PR refactors the component search dialog to create a search extension point. The current index based search is refactored as one of these providers, I've also written an implementation based on https://search.maven.org REST APIs which does not require downloading indexes.

Eventually I'll release an external plugin with a Nexus Repository Manager based implementation.


mpiggott:
Its not clear to me how the version bump is supposed to work.

mickaelistria:
I added some doc about Version Bumps in https://github.com/eclipse-m2e/m2e-core/blob/master/CONTRIBUTING.md#%EF%B8%8F-version-bump . Hope this helps! If not, please add specific question and I'll try to improve doc accordingly.

mpiggott:
> I added some doc about Version Bumps in https://github.com/eclipse-m2e/m2e-core/blob/master/CONTRIBUTING.md#%EF%B8%8F-version-bump . Hope this helps! If not, please add specific question and I'll try to improve doc accordingly.

From the Jenkins log leaving the version as 1.16.0 Jenkins complains that the version needs to be increased (though its not clear why this should be the case, AFAICT 1.16.0 is a nightly and isn't released).

Bumping to 1.16.1 (from your doc perhaps it ought to be 1.17.0, except that 1.16.0 doesn't have a release?) results in jenkins complaining because the test plugins rely on 1.16.0.

Something doesn't seem quite right to me.

mickaelistria:
> AFAICT 1.16.0 is a nightly and isn't released).

You can see the current release at https://download.eclipse.org/technology/m2e/releases/latest/ . It's 1.16.0.

> Bumping to 1.16.1 (from your doc perhaps it ought to be 1.17.0, except that 1.16.0 doesn't have a release?) results in jenkins complaining because the test plugins rely on 1.16.0.

What's the specific error? I don't think this should happen.
Also, please try to build m2e locally to reproduce the issue. The same doc as I linked earlier gives a description of how to build.

mickaelistria:

mpiggott:
> You can see the current release at https://download.eclipse.org/technology/m2e/releases/latest/ . It's 1.16.0.

Ah, when I was looking for tags I couldn't find it - looks like it used a different naming scheme for the tag.

mpiggott:
On Jenkins

Without version bump:
```
[ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:1.6.0:compare-version-with-baselines (check-no-version-regression) on project org.eclipse.m2e.core: Only qualifier changed for (org.eclipse.m2e.core/1.16.0.20200706-0249). Expected to have bigger x.y.z than what is available in baseline (1.16.0.20200605-0817) -> [Help 1]
```

With a version bump:

```
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[FATAL] Non-resolvable parent POM for org.eclipse.m2e:m2e-core-tests:1.16.0-SNAPSHOT: Could not find artifact org.eclipse.m2e:m2e-core:pom:1.16.0-SNAPSHOT and 'parent.relativePath' points at wrong local POM @ line 7, column 10
 @ 
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]   
[ERROR]   The project org.eclipse.m2e:m2e-core-tests:1.16.0-SNAPSHOT (/home/jenkins/agent/workspace/m2e_PR-10/m2e-core-tests/pom.xml) has 1 error
[ERROR]     Non-resolvable parent POM for org.eclipse.m2e:m2e-core-tests:1.16.0-SNAPSHOT: Could not find artifact org.eclipse.m2e:m2e-core:pom:1.16.0-SNAPSHOT and 'parent.relativePath' points at wrong local POM @ line 7, column 10 -> [Help 2]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
```

mickaelistria:
That seems to say you did bump version for the parent pom. This is neither necessary nor desired.

mpiggott:
> That seems to say you did bump version for the parent pom. This is neither necessary nor desired.

To me it seems a more than a little crazy to version bundles separately:
* understanding what makes up a particular release becomes a headache
* minor bumps to core bundles will require updates to dependent bundles anyway due to version ranges
* feature versions will quickly become arbitrary and detached from plugin versions
* every contributor needs to understand the feature hierarchy to ensure they are updated
* more effort into `@since` 
* wonky building of snapshots for already released artifacts

mpiggott:
For reference 2d4f9dc which deals with a minor bump to `core` and `core.ui` took about an hour and is 2000 lines.

mpiggott:
It still errors out with this:

```
[ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:1.7.0:compare-version-with-baselines (check-no-version-regression) on project org.eclipse.m2e.model.edit: Only qualifier changed for (org.eclipse.m2e.model.edit/1.16.0.20200707-2038). Expected to have bigger x.y.z than what is available in baseline (1.16.0.20200514-1409) -> [Help 1]
```

This bundle isn't changed as part of my PR.

mickaelistria:
> For reference 2d4f9dc which deals with a minor bump to core and core.ui took about an hour and is 2000 lines.

This is caused by version ranges beng too string in m2e. Ideally, versions ranges should be like `[1.16.0,2.0.0)` so they wouldn't need to be changed for such changes. We'll get to improve it eventually, and then the amount of things to change will get very small and things like baseline tools or CI will help more precisely and more efficiently.

mpiggott:
The build also seems to be poorly merging master into the branch.

mpiggott:
Still failing on the version check for an unchanged bundle.

mickaelistria:
The main reason why you face issue is that you changed `<m2e.version>` backwards.
What did suggest to change that?

mickaelistria:
For the build to work, I simply took 4a67e7f and then changed
```diff
diff --git a/org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF b/org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF
index 50800187..e447ca03 100644
--- a/org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.m2e.core.ui;singleton:=true
-Bundle-Version: 1.16.0.qualifier
+Bundle-Version: 1.16.1.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Bundle-Name: %Bundle-Name
 Bundle-Vendor: %Bundle-Vendor
diff --git a/org.eclipse.m2e.core.ui/pom.xml b/org.eclipse.m2e.core.ui/pom.xml
index 9d8d466d..c7a96aa9 100644
--- a/org.eclipse.m2e.core.ui/pom.xml
+++ b/org.eclipse.m2e.core.ui/pom.xml
@@ -17,5 +17,6 @@
 
   <artifactId>org.eclipse.m2e.core.ui</artifactId>
   <packaging>eclipse-plugin</packaging>
+  <version>1.16.1-SNAPSHOT</version>
   <name>Maven Integration for Eclipse UI Plug-in</name>
 </project>
diff --git a/org.eclipse.m2e.core/META-INF/MANIFEST.MF b/org.eclipse.m2e.core/META-INF/MANIFEST.MF
index 1d68b914..08a0b290 100644
--- a/org.eclipse.m2e.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.m2e.core;singleton:=true
-Bundle-Version: 1.16.0.qualifier
+Bundle-Version: 1.16.1.qualifier
 Bundle-Activator: org.eclipse.m2e.core.internal.MavenPluginActivator
 Bundle-Vendor: %Bundle-Vendor
 Bundle-Localization: plugin
diff --git a/org.eclipse.m2e.core/pom.xml b/org.eclipse.m2e.core/pom.xml
index 13903bca..8a021123 100644
--- a/org.eclipse.m2e.core/pom.xml
+++ b/org.eclipse.m2e.core/pom.xml
@@ -17,6 +17,7 @@
 
   <artifactId>org.eclipse.m2e.core</artifactId>
   <packaging>eclipse-plugin</packaging>
+  <version>1.16.1-SNAPSHOT</version>
 
   <name>Maven Integration for Eclipse Core Plug-in</name>
```

I'm looking at the code (I will add details later), and it seems like only the org.eclipse.m2e.core.ui bundle needs to get to 1.17 as it introduces new API; other ones wouldn't need to change (although the version ranges need to be more open in several places, ideally moved from `org.eclipse.m2e.core.ui;version="[1.16.0,1.17.0)"` to `org.eclipse.m2e.core.ui;version="[1.17.0,2.0.0)"` if we assume m2e properly maintains API and versions.

m2e-maven-runtime/pom.xml
```diff
@@ -35,7 +35,7 @@
 
 	<properties>
 		<maven-bundle-plugin.version>4.2.1</maven-bundle-plugin.version>
-		<m2e.version>1.16.1</m2e.version>
+		<m2e.version>1.16.0</m2e.version>
```
mickaelistria, line 38: Unnecessary change, causing errors.

org.eclipse.m2e.binaryproject.tests/META-INF/MANIFEST.MF
```diff
@@ -2,9 +2,9 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.m2e.binaryproject.tests;singleton:=true
-Bundle-Version: 1.16.0.qualifier
+Bundle-Version: 1.16.1.qualifier
 Bundle-Vendor: %Bundle-Vendor
-Require-Bundle: org.eclipse.m2e.core;bundle-version="[1.16.0,1.17.0)",
+Require-Bundle: org.eclipse.m2e.core;bundle-version="[1.17.0,1.18.0)",
```
mickaelistria, line 7: See general comment at the end

org.eclipse.m2e.core.ui/schema/m2e_component_search.exsd
```diff
@@ -0,0 +1,125 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
mickaelistria, line 1: Seems to me like an older version of the schema from org.eclipse.m2e.core.
Shouldn't it be removed?

org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF
```diff
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.m2e.core.ui;singleton:=true
-Bundle-Version: 1.16.0.qualifier
+Bundle-Version: 1.17.0.qualifier
```
mickaelistria, line 4: As this change doesn't introduce new APIs, you could just stick to 1.16.1.qualifier, and that would require less changes in consumers.

pom.xml
```diff
@@ -34,7 +34,7 @@
 
     <!-- Current target version of m2e, used to deduce some URL at verify and publish.
     This should usually be sync'd with the main feature version -->
-    <m2e.version>1.16.1</m2e.version>
+    <m2e.version>1.16.0</m2e.version>
```
mickaelistria, line 37: Erroneous change, stick to 1.16.1.

m2e-maven-runtime/pom.xml
mpiggott: > Unnecessary change, causing errors.

I think this came in one of the times I rebased. I believe when I submitted the PR originally this was the version and the build didn't work with the same errors at the time.

org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF
mpiggott: > As this change doesn't introduce new APIs, you could just stick to 1.16.1.qualifier, and that would require less changes in consumers.

The extension point is part of the UI bundle, wouldn't that be considered API?

mpiggott:
A thought I was reminded of, around the time I was working on this Igor was working on the sourcelookup code. Perhaps it would make sense for me to pull the search logic out of the sourcelookup plugin and add a hash based query option.

org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF
mickaelistria: In practice, extension points are all considered as API unless the documentation states otherwise. I don't think the plugin that hosts it has an infuence.

mickaelistria: Sorry, I misunderstood. Yes, you're right that extension point is API, so 1.17.0 is correct.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/CentralSearchPreferencePage.java
```diff
@@ -0,0 +1,48 @@
...
+import org.eclipse.m2e.core.internal.preferences.MavenPreferenceConstants;
+import org.eclipse.m2e.core.ui.internal.M2EUIPluginActivator;
+import org.eclipse.m2e.core.ui.internal.Messages;
+
+
+/*
+ * @author Matthew Piggott
+ * @since 1.17.0
+ */
+public class CentralSearchPreferencePage extends FieldEditorPreferencePage implements IWorkbenchPreferencePage {
```
mickaelistria, line 28: Why isn't this preference page inside the search.central bundle?

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/search/util/SearchResultAdaptableFactory.java
```diff
@@ -0,0 +1,128 @@
...
+import org.eclipse.m2e.core.search.ISearchResultGA;
+import org.eclipse.m2e.core.ui.internal.M2EUIPluginActivator;
+
+
+/**
+ * SearchResultAdaptableFactory
+ *
+ * @author Matthew Piggott
+ */
+public class SearchResultAdaptableFactory implements IAdapterFactory {
```
mickaelistria, line 33: What's the purpose of this adapter? Couldn't we achieve similar results with a more explicit API?

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/wizards/WarningComposite.java
```diff
@@ -0,0 +1,69 @@
...
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Label;
+import org.eclipse.swt.widgets.Text;
+
+/**
+ * Composite which can be used to show an IStatus message & image
+ *
+ * @author Matthew Piggott
+ */
+public class WarningComposite extends Composite {
```
mickaelistria, line 28: This doesn't seem to be used. It would be good to remove it then.

org.eclipse.m2e.core/plugin.xml
```diff
@@ -22,6 +22,7 @@
    <extension-point id="conversionEnabler" name="%extension-point.conversionEnabler.name" schema="schema/conversionEnabler.exsd"/>
    <extension-point id="incrementalBuildFrameworks" name="%extension-point.incrementalBuildFrameworks.name" schema="schema/incrementalBuildFrameworks.exsd"/>
    <extension-point id="workspaceClassifierResolvers" name="%extension-point.workspaceClassifierResolvers.name" schema="schema/workspaceClassifierResolvers.exsd"/>
+   <extension-point id="m2eComponentSearch" name="m2e component search" schema="schema/m2eComponentSearch.exsd"/>
```
mickaelistria, line 25: What are the "core" cases that may make search useful here? I would expect the component search to be a pure-UI operation.

mpiggott: > What are the "core" cases that may make search useful here? I would expect the component search to be a pure-UI operation.

At the moment there is also the [org.eclipse.m2e.sourcelookup](https://github.com/eclipse-m2e/m2e-core/tree/master/org.eclipse.m2e.sourcelookup) code where it may make sense to replace the existing code (which also talks to S.M.O). Once upon a time there was also the thought that some of the m2e bundles might be headless use cases.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/wizards/WarningComposite.java
mpiggott: > This doesn't seem to be used. It would be good to remove it then.

Its included in `MavenPomSelectionComponent` to show warnings, its now been a couple years since I wrote it but I believe I was making an attempt to simplify the class and it seemed like a potentially re-usable piece.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/CentralSearchPreferencePage.java
mpiggott: > Why isn't this preference page inside the search.central bundle?

At the moment there is also the org.eclipse.m2e.sourcelookup code where it may make sense to replace the existing code (which also talks to S.M.O). Once upon a time there was also the thought that some of the m2e bundles might be headless use cases.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/search/util/SearchResultAdaptableFactory.java
mpiggott: > What's the purpose of this adapter? Couldn't we achieve similar results with a more explicit API?

Its been long enough that I don't really know - I recall doing some digging into the the available options for displaying trees and chose what seemed like the most appropriate given the use case.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/CentralSearchPreferencePage.java
mickaelistria: so there should be a search.central.ui bundle for that page. Or -unless there is already a compelling use-case for headless bundles-, just make it really all about UI and get rid of all the complications brought by the attempt to work headless.

org.eclipse.m2e.core/plugin.xml
mickaelistria: As mentioned in another comment, is the headless use-case a realistic one? It noticeably increases complexity and development/maintenance cost compared to make it fully about UI; so we shouldn't force such constraints unless we're sure of an ROI.

fbricon: Lemminx-maven would definitely benefit from this search. Ideally we'd make the lucene based index optional in m2e, as it brings too much pain ( too big to download, too many OOMEs, dead weight for jdt.ls)
So yeah headless based search is 100% needed

mickaelistria: Lemminx-Maven will never depend on m2e, nor use OSGi, nor the extension registry.
A search engine would be useful for Lemminx-Maven, but it would be misplaced in m2e. It would better be a separate plain Maven artifact, independent of m2e. So from both m2e and lemminx-maven perspective, having it headliess doesn't seem useful.

mickaelistria:
@mpiggott Are you aware of a Central Search API available that m2e and others could reuse?

mickaelistria:
@mpiggot Did you see my last question? I'm particularly curious about there is an existing Eclipse-independent API one can use to query Central or even random repos or Nexus instance. If this were existing beforehand, it would be reusable across multiple tool and we could then make m2e 1 consumer of it.

mickaelistria:
@mpiggott Could this leverage the org.sonatype.nexus:nexus-rest-client ? Is Maven Central leveraging the same search API so this could be used for Central as well? Is nexus-rest-client really maintained?

mickaelistria:
I think we'll include direct support for Central and Nexus in LemMinX-Maven: https://github.com/eclipse/lemminx-maven/issues/232 and https://github.com/eclipse/lemminx-maven/issues/233 .
FWIW, I don't think it's worth trying to make m2e smarter regarding discovery of artifacts, it's better to place this in the edition part, ie the language server; so it's more reusable.
Once LemMinX-Maven has such feature and m2e has adopted it, it will be worth considering dropping support for index entirely.

akurtakov:
@mickaelistria @vrubezhny Is this one still needed?

mickaelistria:
I personally could leave without it, and since there is no feedback from @mpiggott towards further intent to have it part of m2e and since there is a clear direction to just remove everything indexer-based, I think we can leave without it.
So let's close.

PR #17, opened 2020-09-09
Bug 566822 - NullPointerException in SourceLocationHelper.findLocation

PPazderski:
If the Maven plugin has uncovered lifecycle SourceLocationHelper tries
to find the best location to show the error message.
If the plugin is defined with variable as groupId current Maven fails to
find the plugin when using the resolved groupId.

### DIFF
--- a/org.eclipse.m2e.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.m2e.core;singleton:=true
-Bundle-Version: 1.16.1.qualifier
+Bundle-Version: 1.16.2.qualifier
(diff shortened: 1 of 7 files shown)


mickaelistria:
the change looks good.
However, as m2e uses semantic versioning with baseline replacement in order to minimize the number of artifacts with same payload, it requires that the version of the modified bundle get updated on 1st change after last release; and you're lucky enough that your change is the 1st change in org.eclipse.m2e.core since last release ;)
So can you please bump version of this bundle (in pom and MANIFEST.MF) by +0.0.1 ? You may also need to update the feature version by +0.0.1. Running a local build as described in https://github.com/eclipse-m2e/m2e-core/blob/master/CONTRIBUTING.md#%EF%B8%8F-build should allow you to detect the issue locally (but feel free to push directly to CI to externalize the tests, it's fine as well to make several iterations until CI is happy).

PPazderski:
> So can you please bump version...

Done.

mickaelistria:
Perfect, thanls Paul! (the test failures are unrelated and are a regression in WTP)

PR #19, opened 2020-09-16
Bug 567014 - Avoid newPlexusContainer(..) at startup

mickaelistria:
Delay creation of container (which is expensive) later in m2e lifecycle,
only when it's necessary.

Change-Id: I00f3bc645e0b962bf5c914293b001869ae73e928

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenPluginActivator.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenPluginActivator.java
@@ -14,6 +14,8 @@
 package org.eclipse.m2e.core.internal;
 
 import java.io.File;
+import java.util.Collection;
+import java.util.HashSet;
 
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.BundleListener;
@@ -51,12 +53,9 @@ import org.apache.maven.artifact.factory.ArtifactFactory;
 import org.apache.maven.artifact.metadata.ArtifactMetadataSource;
 import org.apache.maven.artifact.resolver.ArtifactCollector;
 import org.apache.maven.execution.MavenSession;
-import org.apache.maven.index.ArtifactContextProducer;
-import org.apache.maven.index.NexusIndexer;
 import org.apache.maven.index.updater.IndexUpdater;
 import org.apache.maven.plugin.LegacySupport;
 import org.apache.maven.project.DefaultProjectBuilder;
-import org.apache.maven.project.MavenProject;
 import org.apache.maven.repository.legacy.WagonManager;
 
 import org.eclipse.m2e.core.MavenPlugin;
@@ -98,14 +97,7 @@ public class MavenPluginActivator extends Plugin {
   // The shared instance
   private static MavenPluginActivator plugin;
 
-  /**
-   * @deprecated see {@link #getPlexusContainer()} for more info
-   */
-  private DefaultPlexusContainer plexus;
-
-  private DefaultPlexusContainer indexerContainer;
-
-  private DefaultPlexusContainer archetyperContainer;
+  private Collection<PlexusContainer> toDisposeContainers = new HashSet<>();
 
   private MavenModelManager modelManager;
 
@@ -183,6 +175,7 @@ public class MavenPluginActivator extends Plugin {
   /**
    * This method is called upon plug-in activation
    */
+  @Override
   public void start(final BundleContext context) throws Exception {
     super.start(context);
 
@@ -206,25 +199,8 @@ public class MavenPluginActivator extends Plugin {
     }
 
     this.mavenConfiguration = new MavenConfigurationImpl();
-
-    // this is suboptimal. ideally, we want single "root" container that exposes maven core components
-    // and two "child" containers that expose indexer and archetyper. root container should also
-    // be used by MavenImpl. this is not currently possible due to sisu limitations, so we create
-    // four separate containers for now and hopefully clean this up further later.
-    this.plexus = newPlexusContainer(MavenProject.class.getClassLoader());
-    this.indexerContainer = newPlexusContainer(IndexUpdater.class.getClassLoader());
-    this.archetyperContainer = newPlexusContainer(ArchetypeGenerationRequest.class.getClassLoader());
-
     File stateLocationDir = getStateLocation().toFile();
 
-    // TODO this is broken, need to make it lazy, otherwise we'll deadlock or timeout... or both 
-    this.archetypeManager = newArchetypeManager(archetyperContainer, stateLocationDir);
-    try {
-      this.archetypeManager.readCatalogs();
-    } catch(Exception ex) {
-      String msg = "Can't read archetype catalog configuration";
-      log.error(msg, ex);
-    }
 
     this.mavenMarkerManager = new MavenMarkerManager(mavenConfiguration);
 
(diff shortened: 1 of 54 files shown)


org.eclipse.m2e.feature/feature.xml
```diff
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.m2e.feature"
       label="%featureName"
-      version="1.16.1.qualifier"
+      version="1.16.2.qualifier"
```
fbricon, line 5: it's worth bumping the version to 1.17

fbricon:
If you plan on updating the all plugin versions anyways, please do it in a consistent manner, set everything to 1.17.
thx

mickaelistria:
I don't plan to, but I have to because the dependency range were too narrow and update versions had to propagate. I'll update all changed versions to 1.17.

mickaelistria:
Done.

mickaelistria:
@fbricon any progress with impact assessment on m2e-wtp and JBoss Tools?

fbricon:
So m2e-wtp is fine, but JBoss Tools is not. There are 2 calls to MavenPluginActivator.getDefault().getPlexusContainer()
- https://github.com/jbosstools/jbosstools-central/blob/c8b3ebb7876f5383a1fc125fe65ed3e25a6f63cb/maven/plugins/org.jboss.tools.maven.core/src/org/jboss/tools/maven/core/MavenUtil.java#L232
- https://github.com/jbosstools/jbosstools-central/blob/c8b3ebb7876f5383a1fc125fe65ed3e25a6f63cb/maven/plugins/org.jboss.tools.maven.core/src/org/jboss/tools/maven/core/internal/resolution/ArtifactResolutionService.java#L166

mickaelistria:
Thanks @fbricon for checking. I've submitted https://github.com/jbosstools/jbosstools-central/pull/463 to JBoss Tools so we can safely apply the patch to m2e after JBoss Tools does its release.

fbricon:
we can still apply the patch without the removal of the deprecated method (it's an orthogonal issue anyways), without waiting for JBoss Tools to be fixed downstream.

mickaelistria:
As you prefer, it depends on how urgent it is for JDT-LS to consume it. I personally prefer not missing an opportunity to remove deprecated code even if it costs some delay until the merge; but it's something I'm totally flexible with.

fbricon:
I'd rather prefer we merge sooner, even without removing the deprecated methods

mickaelistria:
> I'd rather prefer we merge sooner, even without removing the deprecated methods

OK, I just pushed a new version of the commit on this PR.

org.eclipse.m2e.profiles.core/src/org/eclipse/m2e/profiles/core/internal/management/ProfileManager.java
```diff
@@ -248,9 +248,12 @@ private String findSource(Profile profile, List<Model> modelHierarchy) {
...
       //TODO find an alternative way to get the Maven RepositorySystem, or use Aether directly to resolve models??
-      return MavenPluginActivator.getDefault().getPlexusContainer().lookup(RepositorySystem.class);
-    } catch(ComponentLookupException e) {
-      throw new NoSuchComponentException(e);
+      return MavenPluginActivator.getDefault().getMaven().lookup(RepositorySystem.class);
+    } catch(CoreException e) {
+      if(e.getStatus().getException() instanceof ComponentLookupException) {
+        throw new NoSuchComponentException((ComponentLookupException) e.getStatus().getException());
+      }
+      throw new NoSuchComponentException(null);
```
fbricon, line 256: we completely lose the exception information/context that way. At least log the cause of the failure

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenPluginActivator.java
```diff
@@ -384,6 +372,24 @@ public MavenRuntimeManagerImpl getMavenRuntimeManager() {
...
+          this.archetypeManager = newArchetypeManager(archetyperContainer, getStateLocation().toFile());
+          try {
+            this.archetypeManager.readCatalogs();
+          } catch(Exception ex) {
+            String msg = "Can't read archetype catalog configuration";
+            log.error(msg, ex);
+          }
+        } catch(PlexusContainerException ex1) {
+          // TODO Auto-generated catch block
+          log.error(ex1.getMessage(), ex1);
```
fbricon, line 390: log.error("Failed to initialize the ArchetypeManager", ex1);

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenPluginActivator.java
```diff
@@ -384,6 +372,24 @@ public MavenRuntimeManagerImpl getMavenRuntimeManager() {
...
+          // TODO this is broken, need to make it lazy, otherwise we'll deadlock or timeout... or both 
+          this.archetypeManager = newArchetypeManager(archetyperContainer, getStateLocation().toFile());
+          try {
+            this.archetypeManager.readCatalogs();
+          } catch(Exception ex) {
+            String msg = "Can't read archetype catalog configuration";
+            log.error(msg, ex);
+          }
+        } catch(PlexusContainerException ex1) {
+          // TODO Auto-generated catch block
```
fbricon, line 389: remove TODO

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenPluginActivator.java
```diff
@@ -376,6 +348,23 @@ public ProjectRegistryManager getMavenProjectManagerImpl() {
...
+          toDisposeContainers.add(indexerContainer);
+          this.indexManager = new NexusIndexManager(indexerContainer, getMavenProjectManager(), getRepositoryRegistry(),
+              getStateLocation().toFile());
+          getMavenProjectManager().addMavenProjectChangedListener(indexManager);
+          getMaven().addLocalRepositoryListener(new IndexingTransferListener(indexManager));
+          ((RepositoryRegistry) getRepositoryRegistry()).addRepositoryIndexer(indexManager);
+          ((RepositoryRegistry) getRepositoryRegistry())
+              .addRepositoryDiscoverer(new IndexesExtensionReader(indexManager));
+        } catch(PlexusContainerException ex1) {
+          log.error(ex1.getMessage(), ex1);
```
fbricon, line 364: log.error("Failed to initialize the NexusIndexManager", ex1);

pom.xml
```diff
@@ -34,14 +34,14 @@
 
     <!-- Current target version of m2e, used to deduce some URL at verify and publish.
     This should usually be sync'd with the main feature version -->
-    <m2e.version>1.16.2</m2e.version>
+    <m2e.version>1.17.0</m2e.version>
     <m2e-maven-runtime.version>1.16.0-SNAPSHOT</m2e-maven-runtime.version>
```
fbricon, line 38: 1.17.0-SNAPSHOT

all versions have been bumped elsewhere, so let's stay consistent

fbricon (changes requested):

pom.xml
mickaelistria: Please see comment, this version is used for releng scripts and -SNAPSHOT is not desired in repository URLs (so far).

fbricon: the comment is for L38, which contains SNAPSHOT already

mickaelistria: In that case, it's not something we want to do either. If we change that, then it means we have to build newer versions of the maven runtime while it actually didn't change at all, leading to distribution of 10MB of redundant content to the millions of users. It's something we now avoid by decoupling runtime version from consuming m2e version.

PR #20, opened 2020-09-18
Tycho pomless

mickaelistria:


fbricon:
given that m2e doesn't support mvn extensions, this change makes it impossible to import some projects as Maven projects.

mickaelistria:
OK, let's abandon and try it again when m2e supports Maven extensions.

PR #28, opened 2020-11-10
Guard about null pointer access if image was never created

laeubi:
On shutdown an NPE can occur if the image was never created.


mickaelistria:
Test failures are not related, let's merge.

PR #30, opened 2020-11-14
[m2e-pde] Fix race-condition by using a synchronized + file locking CacheManager

laeubi:
When creating wrapped artifacts these are currently stored at the
original location of the artifact (e.g. the current maven local repository) independent from the target context, this can lead to
the following race-conditions:
- different eclipse installations using the same target file
- different target files are referencing the same artifact

This can then lead to strange intermediate errors (bundle can't be
wrapped or similar errors).

This commit proposes a solution using a CacheManager that provides
target aware locations, is synchronized to prevent concurrent access to
cached files and using filelocks to prevent other eclipse instances to
access the file at the same time.

### DIFF
--- a/org.eclipse.m2e.pde/src/org/eclipse/m2e/pde/CacheManager.java
+++ b/org.eclipse.m2e.pde/src/org/eclipse/m2e/pde/CacheManager.java
@@ -1,0 +1,127 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Christoph Läubrich
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ * 
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *      Christoph Läubrich - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.m2e.pde;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.RandomAccessFile;
+import java.nio.channels.FileChannel;
+import java.nio.channels.FileLock;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.commons.codec.digest.DigestUtils;
+import org.apache.commons.io.FileUtils;
+import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.pde.core.target.ITargetHandle;
+import org.eclipse.pde.core.target.TargetBundle;
+
+public class CacheManager {
+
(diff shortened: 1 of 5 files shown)


mickaelistria:
I don't fully understand what's the "original location" you mention in the commit message. Can you please elaborate? Shouldn't this location be internal to the workspace instance?

laeubi:
the "original location" was/is the local maven repository (mostly ~/.m2/repository or alike) and is shared across all instances (eclipse, cli, ...)

mickaelistria:
What does it bring to have "target aware locations" here. What's not enough with regular filesystem lock/release approach to concurrency?

org.eclipse.m2e.pde/src/org/eclipse/m2e/pde/CacheManager.java
```diff
@@ -0,0 +1,127 @@
...
+	public CacheManager(File folder) {
+		this.folder = folder;
+		try {
+			FileUtils.touch(new File(LASTACCESS_MARKER));
+		} catch (IOException e) {
+			// can't mark last access then...
+		}
+	}
+
+	public synchronized <R> R accessArtifactFile(Artifact artifact, CacheConsumer<R> consumer) throws Exception {
```
mickaelistria, line 52: Why not just `Funtion<File, R>` instead of CacheConsumer here?

laeubi: Functions are not allowed to throw (checked) exceptions.

laeubi:
File Locks only work with different processes not inside the JVM itself. as each target can be used/resolved independent it was the choice to have independent locations for them so there is no interference.
This is especially useful for another feature I'm working on that will allow to specify BND instructions alongside with the wrapped artifact stored inside the target file itself.

mickaelistria:
OK, thanks for clarifying. Do you think you can parts of your answers to my question in the Git commit message and/or in the code as comments?

laeubi:
I'll try to add some comments on the code as separate commit, about the commit message, when you merge a PR on github you can choose "Squash" (instead of merge) then all commits will be commenced into one commit and you can adjust the (original) commit message as you like to have a nice git history on master.

laeubi:
@mickaelistria jsut wondering, is m2e included in the daily I-Builds / release train and will this be then included in the 2020-12 release? Or does m2e has its own release cycle?

mickaelistria:
Nope, m2e releases when desired by m2e, and contributes latest release to SimRel: https://git.eclipse.org/c/simrel/org.eclipse.simrel.build.git/tree/m2e.aggrcon
Eg we could release m2e next week if we think it's worth it.

laeubi:
Is there some kind of "snapshot" repro so people can test this before release? I'm currently trying to get a patch into PDE for advanced editing capabilities, do we need to wait for the next release until m2e can use it?

mickaelistria:
Sure, there is https://download.eclipse.org/technology/m2e/snapshots/1.17.0/latest/

PR #33, opened 2020-11-18
Add support for the user to influence the generated manifest.

laeubi:
Currently there is no way for the user to influence the generated
manifest. With this change a user can edit the template that is used to generate
manifest headers from maven artifacts.

If the "generate" mode is chosen, a new button is displayed
grafik

A dialog pops up that allows to either use the defaults, or edit the BND template
grafik

The instructions are then stored inside the target and will be used to generate missing manifests.

**Just a note** because one might wonder why the code allows multiple templates while the UI only allows one per location: due to [current PDE limitations](https://bugs.eclipse.org/bugs/show_bug.cgi?id=568865) it is not possible to edit individual dependent items but that is planned when this feature hopefully is available in the 2021-03 release.

### DIFF
--- a/org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/MavenTargetLocationWizard.java
+++ b/org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/MavenTargetLocationWizard.java
@@ -30,13 +30,15 @@ import org.eclipse.pde.core.target.ITargetLocation;
 import org.eclipse.pde.ui.target.ITargetLocationWizard;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.custom.CCombo;
+import org.eclipse.swt.events.ModifyEvent;
+import org.eclipse.swt.events.ModifyListener;
 import org.eclipse.swt.events.SelectionEvent;
 import org.eclipse.swt.events.SelectionListener;
 import org.eclipse.swt.layout.GridData;
 import org.eclipse.swt.layout.GridLayout;
-import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Label;
+import org.eclipse.swt.widgets.Link;
 import org.eclipse.swt.widgets.Text;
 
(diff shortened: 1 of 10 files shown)


org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/MavenTargetLocationWizard.java
```diff
@@ -95,14 +93,63 @@ public String getText(Object element) {
...
+					@Override
+					public String getText(Object element) {
+						if (element instanceof MissingMetadataMode) {
+							return ((MissingMetadataMode) element).name().toLowerCase();
+						}
+						return super.getText(element);
+					}
+
+				});
+				editInstructionsButton = new Button(composite, SWT.PUSH);
```
mickaelistria, line 124: IMO, this would be better be a link, as the button besides the combo looks heavy and -unless I'm mistaken- the new dialog pops-up on top of the current one and doesn't affect the "parent" dialog. Links IMO better represent such "navigate without changing state" approach.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/MavenTargetLocationWizard.java
```diff
@@ -95,14 +93,63 @@ public String getText(Object element) {
...
+				editInstructionsButton.setText("edit manifest generation instructions");
```
mickaelistria, line 125: Ideally, strings should be externalized with NLS.
Also, labels often need to start with an upper case letter; and have a mnemonic on 1 character to enable more keyboard accessibility.

laeubi: grafik
What do you think? It seems mnemonic can only be used in the text outside the clickable area and only focus it so it looks a bit odd to me.

Maybe just "Edit Instructions" without mnemonic?

laeubi: grafik

At least one can use tab to activate the link even without mnemonic.

PR #34, opened 2020-11-20
Add advanced editing capabilities based on upcoming PDE API

laeubi:
This shows the new features that would be possible once the [advanced PDE API](https://bugs.eclipse.org/bugs/show_bug.cgi?id=568865) is available.

It is possible to select an item and exclude it from the target resolution (e.g. one wants to exclude a dependency that is not a bundle, an older version or similar)
grafik
 
of course it is possible to include the item afterwards again.

It is now also possible to edit individual items and edit the instructions for this particular item
grafik

This also includes some more enhancements, e.g. if the clipboard contains a maven-dependency-xml fragment and a new item is added, the GAV fields are automatically filled.


akurtakov:
What is the status of this PR?

laeubi:
@akurtakov I assume that this PR needs to wait for the next release of PDE, then I need to update it to reflect latest code changes.

akurtakov:
If there is not gonna be m2e release prior to the March Platform release, m2e can build against and require latest PDE. @mickaelistria wdyt?

mickaelistria:
I think it's better to leave room for a potential release before March.

laeubi:
This is not crucial so I think its fine to wait until the next PDE release.

mickaelistria:
Should this be revived?

laeubi:
Yep I'll try to update the PR base don latest API.

laeubi:
The source has just diverged too much in the meantime so I decided to create a new on top of current master see: https://github.com/eclipse-m2e/m2e-core/pull/121

PR #35, opened 2020-11-21
Relax version constraint on wwd

kdvolder:
See: https://github.com/eclipse/wildwebdeveloper/pull/543#issuecomment-731477895


mickaelistria:
Sorry, we can't do that because the underlying version of LemMinX are not API compatble. So extensions (lemminx-maven) for 0.9.0 won't work 0.10.0. This will have to be done with #31 , when lemminx-maven provides a release compatible with latest lemminx and wild web developer 0.10.0.
I've opened https://github.com/eclipse/wildwebdeveloper/issues/567 on the Wild Web Developer side which is related. It would allow to update at least the "web" language servers even with a restriction on the XML LS.
If you have the free cycles to spend here, some help would be really welcome to work on lemminx-maven and troubleshoot a few bugs, mainly about the artifact resolution; those are a must-have before we can release the new lemminx-maven, and consider using it in m2e. and then allow to move to latest Wild Web Developer.

kdvolder:
> If you have the free cycles to spend here...

Not much cycles for anything at the moment, maybe after new year.

PR #36, opened 2020-11-25
provide url-handler to support maven depolyed updatesites

laeubi:
Currently maven deployed update-sites can not be used in target-files because PDE requires that an udpate-site is an URL with individual files to query data.

This adds an URLHandler that allows to use URLs in the format: `mvn:<groupId>:<artifactId>:<version>[:type]`.

grafik


org.eclipse.m2e.pde/META-INF/urlhandler.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mickaelistria, line 1: I think this would better be part of org.eclipse.m2e.core.

laeubi:
See https://github.com/eclipse-m2e/m2e-core/pull/37

PR #37, opened 2020-11-25
provide url-handler to support maven depolyed updatesites

laeubi:
Currently maven deployed update-sites can not be used in target-files because PDE requires that an udpate-site is an URL with individual files to query data.

This adds an URLHandler that allows to use URLs in the format: `mvn:<groupId>:<artifactId>:<version>[:type]`.

grafik


mickaelistria:
Thanks for this good and powerful addition!

PR #41, opened 2020-11-30
Bug 569298 - NPE when computing maven classpath

laeubi:
add null checks


fbricon:
null checks are not sufficient. The current bundle loading is inherently broken. See https://bugs.eclipse.org/bugs/show_bug.cgi?id=563069#c0

PR #42, opened 2020-12-01
Bug 569294 - Bundle exception on start after installing m2e latest

laeubi:
Move from DS to activator


mickaelistria:
Perfect, thanks!

PR #44, opened 2020-12-07
Bump version for org.eclipse.m2e.maven.indexer

mickaelistria:
As there is some API change, fix version and requirements.

### DIFF
--- a/m2e-maven-runtime/org.eclipse.m2e.maven.indexer/pom.xml
+++ b/m2e-maven-runtime/org.eclipse.m2e.maven.indexer/pom.xml
@@ -19,6 +19,7 @@
   </parent>
 
   <artifactId>org.eclipse.m2e.maven.indexer</artifactId>
+  <version>1.17.1-SNAPSHOT</version>
   <packaging>bundle</packaging>
 
   <name>Nexus Indexer Bundle</name>
--- a/org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.m2e.core.ui;singleton:=true
-Bundle-Version: 1.17.0.qualifier
+Bundle-Version: 1.17.1.qualifier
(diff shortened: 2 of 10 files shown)


m2e-maven-runtime/org.eclipse.m2e.maven.indexer/pom.xml
```diff
@@ -19,6 +19,7 @@
   </parent>
 
   <artifactId>org.eclipse.m2e.maven.indexer</artifactId>
+  <version>1.17.1-SNAPSHOT</version>
```
fbricon, line 22: 1.17.0 has not been released, why is this necessary?

org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF
```diff
@@ -28,7 +28,7 @@ Require-Bundle: org.eclipse.m2e.core;bundle-version="[1.16.0,2.0.0)",
  org.eclipse.m2e.model.edit;bundle-version="[1.16.0,2.0.0)";visibility:=reexport,
  org.eclipse.m2e.maven.runtime;bundle-version="[1.16.0,2.0.0)",
  org.eclipse.m2e.archetype.common;bundle-version="[1.16.0,2.0.0)",
- org.eclipse.m2e.maven.indexer;bundle-version="[1.16.0,2.0.0)",
+ org.eclipse.m2e.maven.indexer;bundle-version="[1.16.0.20201124-0221,2.0.0)",
```
fbricon, line 31: [1.17.0,2.0.0)

fbricon (changes requested):

m2e-maven-runtime/org.eclipse.m2e.maven.indexer/pom.xml
mickaelistria: 1.17.0 has been tagged and published, and then, I found this issue to fix...

mickaelistria:
@mbooth101 FYI, this is an important change we forgot about when upgrading the indexer version. I would expect the build to detect such issues, but apparently the "compare-version-with-baselines" mojo isn't executing properly here to fail the build in such erroneous case.

mbooth101:
@mickaelistria Sorry about this omission :-(

mickaelistria:
@mbooth101 No big deal, it's not only yours and it actually reveals a build infra problem, so that's not all bad!

PR #45, opened 2020-12-17
Add suitable description/license information to PDE feature

laeubi:
I noticed that the feature does not include appropriate copyright information, this is updated now.

beside this, it seems that the update for the new version (1.17.1-SNAPSHOT) did not happened, do I need to add this feature somewhere for this to take place automatically?


mickaelistria:
> it seems that the update for the new version (1.17.1-SNAPSHOT) did not happened

We only update artifacts that actually change.
As we're not starting a 1.17.2 stream, it may be worth calling it 1.17.2 from now on.

mickaelistria:
Can you please update versions of this bundle and including feature to 1.17.2 ?

PR #48, opened 2021-01-07
Bug 570080 - Module name for test launch config

paper-tiger:
Hi,

to be able to launch applications from the test folder in a modularized project the given modification could be applied.

This is probably a corner case as those applications are not picked up by maven. My use case are some evaluations not intended to be part of the main application. I guess those could just be moved to another project and probably the visibility of the classes used must be changed but if felt weird that you can not launch an application from the test folder as this was previously (without modules) possible.

Anyway the change sets the IJavaLaunchConfigurationConstants.ATTR_MODULE_NAME attribute of the launch configuration for those running a main-class from the test folder to the module name derived from the project. It does this on any launch-config change and thus needs a safeguard against stack overflow by recursion implemented by the equals in line 88.

So when org.eclipse.jdt.debug.ui.launchConfigurations.JavaApplicationLaunchShortcut sets the same attribute to " " as the test folder has no module-info, this code overrides it.

### DIFF
--- a/org.eclipse.m2e.feature/feature.xml
+++ b/org.eclipse.m2e.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.m2e.feature"
       label="%featureName"
-      version="1.17.1.qualifier"
+      version="1.17.2.qualifier"
       provider-name="%providerName"
       plugin="org.eclipse.m2e.core"
       license-feature="org.eclipse.license"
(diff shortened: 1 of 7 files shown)


mickaelistria:
Please see https://github.com/eclipse-m2e/m2e-core/blob/master/CONTRIBUTING.md#%EF%B8%8F-version-bump as some versions need to be bumped for the build to go on.
@brychcy would you like to review this? I think you're the best one around here when it comes to that family of issues.

mickaelistria:
Fixed with https://github.com/eclipse-m2e/m2e-core/commit/d61bcd5d05532310b0f93095707bc6e6a5ddf595 , thanks @paper-tiger  !

PR #50, opened 2021-01-15
Update m2e-core-tests submodule

akurtakov:
To get some test fixes.

Signed-off-by: Alexander Kurtakov <akurtako@redhat.com>


akurtakov:
2 warnings less but there is still 1 failing the build. @laeubi can you help with the remaining one https://ci.eclipse.org/m2e/job/m2e/job/master/lastSuccessfulBuild/testReport/org.eclipse.m2e.tests/BuildPathManagerTest/testDownloadSources_006_nonMavenProject/

laeubi:
Is there a complete StackTrace for the error?

OverlappingFileLockException usually happens when one tries to lock the same file (region) twice int the same jvm. So usually the lock/unlock needs to be guarded by some code that prevents for different threads and/or reentrancy.

akurtakov:
I can reproduce locally so probably you can get it too.

laeubi:
I'll try to take a look, sadly my m2e dev-ide is currently broken what makes debugging/testing a mess ...

laeubi:
Any hint how I can enable remote debugging or run the test from within eclipse? The usual mvnDebug does not work as it seems the itests are spawning a new (eclipse) process.

mickaelistria:
https://www.eclipse.org/tycho/sitedocs/tycho-surefire-plugin/test-mojo.html#debugPort

laeubi:
@mickaelistria thanks for the hint, I was able to start a debug session. For the problem itself I have not found the cause yet but hope to get a deeper look into it. At the moment it seems that maybe the index-server is not shutdown properly or started more than once, if that's a problem of m2e or indexer is currently not clear to me, but if that rings a bell for someone...

PR #52, opened 2021-01-15
Bug 563742 - Cache finalName

vtintillier:
To improve performance of classpath configuration with modules, which
otherwise need to read all pom files of projects in the classpath.

Signed-off-by: Vivien TINTILLIER <vivien.tintillier@sap.com>

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
@@ -88,6 +88,8 @@ public class MavenProjectFacade implements IMavenProjectFacade, Serializable {
 
   private final IPath testOutputLocation;
 
+  private final String finalName;
+
   private final Set<ArtifactRepositoryRef> artifactRepositories;
 
   private final Set<ArtifactRepositoryRef> pluginArtifactRepositories;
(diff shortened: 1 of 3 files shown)


vtintillier:
- More details in https://bugs.eclipse.org/bugs/show_bug.cgi?id=563742
- I don't know how to add a test for this, any idea?
- My Eclipse user (https://accounts.eclipse.org/users/vmisingnamei5c) uses a personal email address, any idea how I can link it to my work email?

mickaelistria:
> I don't know how to add a test for this, any idea?

Ideally, a test that reproduce the issue (slowness in the launch config) would have been great. But here, the fix seems safe and doesn't introduce a regression. I wouldn't mind merging it without a test if writing a test seems a blocker for you. Your decide ;)

> My Eclipse user (https://accounts.eclipse.org/users/vmisingnamei5c) uses a personal email address, any idea how I can link it to my work email?

The simplest seems to be that you go on your Eclipse account page to change your email address.

mickaelistria:
ECA check is now passing. Thanks a lot for updating that. We can now merge this very good patch, thanks!

vtintillier:
> Ideally, a test that reproduce the issue (slowness in the launch config) would have been great. But here, the fix seems safe and doesn't introduce a regression. I wouldn't mind merging it without a test if writing a test seems a blocker for you. Your decide ;)

The (first) issue I have is that for example the projects I reproduced the issue with (attached in the bug report) would take around 30s to launch e.g. a junit test, **but** it takes 15 to 30 minutes to import those projects in the workspace in the first place, so the test would be very slow. Or maybe you were not thinking about such an integration test?

By the way this is something I'd like to understand why this is so slow to import those 36 projects (and so slow to update when e.g. changing the parent pom). It is orders of magnitude slower than the Maven build itself, or from Intellij maven import/refresh.

> The simplest seems to be that you go on your Eclipse account page to change your email address.

Ok I did that 👍

mickaelistria:
> By the way this is something I'd like to understand why this is so slow to import those 36 projects (and so slow to update when e.g. changing the parent pom). It is orders of magnitude slower than the Maven build itself, or from Intellij maven import/refresh.

I suggest you open a bug with an example and steps to reproduce and we can investigate. Spoiler alert: most perf issues in m2e are always about the same thing, facade.getMavenProject(monitor) called too often.

PR #54, opened 2021-01-19
Bug 570473: M2e Plugins still uses low resolution icons

BeckerWdf:
Signed-off-by: Matthias Becker <ma.becker@sap.com>


mickaelistria:
Did you close intentionally? Isn't this PR valid?

BeckerWdf:
It's invalid. Pullrequest 55 is the correct one.

PR #57, opened 2021-01-21
Bug 570456 - m2e PDE integration clipboard integration does not work if scope is also copied

laeubi:
Scope is now considered and filled, if the xml can't be parsed a warning is displayed for 10 seconds
grafik

### DIFF
--- a/org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/MavenTargetLocationWizard.java
+++ b/org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/MavenTargetLocationWizard.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2018, 2020 Christoph Läubrich
+ * Copyright (c) 2018, 2021 Christoph Läubrich
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v2.0
  * which accompanies this distribution, and is available at
@@ -12,9 +12,11 @@
  *******************************************************************************/
 package org.eclipse.m2e.pde.ui.editor;
 
+import java.text.MessageFormat;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.concurrent.TimeUnit;
 
 import org.eclipse.jface.resource.ImageDescriptor;
 import org.eclipse.jface.viewers.ArrayContentProvider;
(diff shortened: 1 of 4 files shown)


org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/MavenTargetLocationWizard.java
```diff
@@ -115,10 +117,18 @@ public void createControl(Composite parent) {
...
 					classifier.setText(clipboardParser.getClassifier());
 					type.setText(MavenTargetLocation.DEFAULT_PACKAGE_TYPE);
-					scope.setText(MavenTargetLocation.DEFAULT_DEPENDENCY_SCOPE);
+					scope.setText(clipboardParser.getScope());
 					metadata.setSelection(new StructuredSelection(MavenTargetLocation.DEFAULT_METADATA_MODE));
 					bndInstructions = new BNDInstructions("", null); //$NON-NLS-1$
 					includeSource.setSelection(true);
+					String clipboardError = clipboardParser.getError();
+					if (clipboardError != null) {
+						setMessage(MessageFormat.format(Messages.MavenTargetLocationWizard_11, clipboardError),
```
mickaelistria, line 126: Please just show a MessageDialog.openWarning in that case.

laeubi: I think that's too disturbing to the user. As auto-filling is really optional and we can't assume that users always have valid content a Dialog might really annoy users if triggered by random snippets in the clipboard.

mickaelistria: The proposed approach is totally alien, it's not used in any other dialog and I'm not certain about how accessible it would be to understand for people using a screenreader. It is probably better to just not show anything then if auto-filling is not applicable, as users are probably not even expecting auto-fill to be happening in that case, a warning is actually more a disturbance than nothing.

laeubi: jgit uses autofilling for the "add repository" dialog, but if you think its better to be silent I'll remove the message, it just showed that if filling is not performed but the user tries to use the feature it is considered a bug: https://bugs.eclipse.org/bugs/show_bug.cgi?id=570456 but I think its okay to simply tell people then to check their xml fragment.

mickaelistria: Auto-filling is nice; however what I don't find good is the temporary warning on startup. As far as I can see, the "Clone Repository" wizard doesn't warn if clipboard content isn't a Git repo URL, it just does nothing, which is good enough.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/MavenTargetLocationWizard.java
```diff
@@ -121,13 +121,10 @@ public void createControl(Composite parent) {
...
 					includeSource.setSelection(true);
-					String clipboardError = clipboardParser.getError();
+					Exception clipboardError = clipboardParser.getError();
 					if (clipboardError != null) {
-						setMessage(MessageFormat.format(Messages.MavenTargetLocationWizard_11, clipboardError),
-								WARNING);
-						parent.getDisplay().timerExec((int) TimeUnit.SECONDS.toMillis(10), () -> {
-							setMessage(null);
-						});
+						Platform.getLog(MavenTargetLocationWizard.class).warn(MessageFormat
```
mickaelistria, line 126: In most cases, the clipboard wouldn't contain a XML document; which means in most cases, this warning will be shown, even for regular usage. I don't think this should be logged at all, a clipboard that doesn't contain GAV is a normal thing.

laeubi: I do a poor-man-precheck by requiring that the clipboard:
a) must contain text
b) the text must start with an '<' char
that should guard against most common non-xml clipboard content. E.g. my error log is filled with many "UI freeze", "duplicate keybinding" stuff so I would be happy to at least see somewhere something useful if I add a new target entry and it does not happen as expected :-)

mickaelistria: OK, I missed the preliminary check. I think it's OK to log in this restricted cases.

PR #62, opened 2021-01-29
Declare mvn URL protocol handler to ECF

mickaelistria:


mickaelistria:
cc @laeubi : p2 uses ECF, so we need the protocol registered to ECF as well. Maybe this allows to get rid of the dynamic registration in activator?

laeubi:
From my testing there is no special need for ecf as it falls back to "normal" url handlers. Anyways the dynamic registration is still usefull for usecase outside P2.

PR #65, opened 2021-02-12
Bug 570315 - The latest contribution of m2e to SimRel 2021-03 M1 has bad licenses and unsigned bundles

laeubi:
This should sign all unsigned artifacts thanks to the change in cbi https://github.com/eclipse-cbi/org.eclipse.cbi/pull/3 but needs verification on the actual build server (do PR get signed/build the site and can we get access to these artifacts of the build without a deploy?)

### DIFF
--- a/org.eclipse.m2e.site/pom.xml
+++ b/org.eclipse.m2e.site/pom.xml
@@ -36,6 +36,62 @@
   </build>
 
   <profiles>
+  	<!-- Eclipse SimRel requires all artifacts to be jar-signed. So we sign the (non eclipse) artifacts and update the p2 metadata accordingly -->
+	<profile>
+		<id>eclipse-sign</id>
+		<pluginRepositories>
+			<pluginRepository>
+				<id>cbi-snapshots</id>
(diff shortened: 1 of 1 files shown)


akurtakov:
Can you please rebase on master? That should fix the build failure.

laeubi:
hm... I have created the patch from the master already and git tells me there are no new changed and also github says 'This branch is 1 commit ahead of eclipse-m2e:master. " so everything seems to be in sync here...

akurtakov:
I see @mickaelistria workarounded split package issue and thought that's the fix but seems like another one is needed in tests.

org.eclipse.m2e.site/pom.xml
```diff
@@ -36,6 +36,61 @@
   </build>
 
   <profiles>
+	<profile>
```
mickaelistria, line 39: Please comment here about why this is necessary (ie "Eclipse SimRel requires all artifacts to be jar-signed. So we sign the BND artifacts and update the p2 metadata accordingly").

mickaelistria:
Thanks Christoph!

PR #79, opened 2021-03-19
Cleanups

mickaelistria:


cal101:
If you want those changes to be combined I should be able to do that.
Currently the size of the single PRs is set to a maximum of 20 files to be easily reviewable
and to a maximum of 3 PRs per Type.
(See the Diamond-Operator-PRs.)
Otherwise the PRS can get quite big: For example the full set of Diamond-Operator-Changes would be 160 and some files.

cal101:
@mickaelistria Hi Mickael! When I provide the next batch of Cleanups: What shall I change besides bumping the relevant versions?
One PR containing commits for multiple change types?
What should be the rough file/size limit of a single commit?
I would really like to be able to open the whole m2e project in eclipse but some sub projects/sub modules fail. How can I get help on setting that up?

mickaelistria:
Really do as you prefer and as it's most productive to you (while remaining a "safe" change); 30 or 1 PR is fully up to you and we'll deal with it.

laeubi:
@cal101 the version should only be bumped **once** within a release so not for every change.  So if the last version was 1.7.2 and now we have a change it must bum (at laest) micro version to 1.7.3, then the next change do not need to bump the version again. Don't know if you think your tool can detect this? If you like to experiment a bit and like to make your tool superior, take a look at [BND baselining](https://bnd.bndtools.org/chapters/180-baselining.html) and [the baseline-mojo](https://github.com/bndtools/bnd/blob/master/maven/bnd-baseline-maven-plugin/src/main/java/aQute/bnd/maven/baseline/plugin/BaselineMojo.java) what can detect from a given released artifact if a micro, minor or major version increase is required from the binary compatibility point of view.

cal101:
@mickaelistria I try to provide only "safe" changes. The Cleanups do contain bugs and I typically report them and skip the change or the cleanup depending on the issue WHEN I SEE IT IN MY REVIEW.
I am playing with different pipeline configurations and workflows so feel free to make suggestions how to make YOUR life easier and I take the freedom to use or ignore them ;-)
There will also be change types that are a matter of taste or project policy. If you don't want them in m2e-core just reject them with a note and I won't feel offended. That is typically easier with separated PRs.
I try to avoid it but it may happen that separated PRs have merge/rebase conflicts if combined. The easiest way is letting them being re-created.

mickaelistria:
@laeubi for m2e, things are slightly different than from Platform (at the moment but I intend to change that, and it's actually changing discretely): the rule is more or less that anytime we need to change the version of bundle, we bump it to upcoming release version. So a bundle can bump from 1.16.1 to 1.18.0 even for micro fixes. It's aimed at making things easier for consumers; but I don't know if it's making things easier for contributors... In any case, version bumps are always a tricky thing if we want to keep reproducible builds, whichever strategy we use.
I have some hope that one day, we'll be able to work on a bot that would automatically see the build result and push a review with suggested edits contributors could apply just by clicking them. I've seen this workflow (with a human) recently and it's very comfortable as a contributor. If we can make it a bot, we may have the best of both worlds.

@cal101 OK for separate PRs. To be honest, I very quickly review such changes, I look at a few edits on a PR and if I feel no risk seems introduced and all tests pass as usual, I just merge. Overall, it make the code better, maybe introducing a bug, but on the long run, nice code is more important; and it's easy-ish if some bug happen to bisect the change and find what can be the cause to maybe revert some change later.

cal101:
@laeubi Thanks for the info, I understood it before to bump on every change.
Sigh, until now I was able to avoid BND ;-)  I take a look.
I don't expect hacking the pipeline to detect and handle that to be the problem.
The bigger problem is getting me to understand the mixture of maven, m2, p2, tycho, bnd, m2e, eclipse, osgi and whatever else.

cal101:
@mickaelistria The kind of workflow having a bot providing quality fixes ready to be reviewed and merged with a minimum of manual work is my vision, too.

@mickaelistria @laeubi You leave me confused regarding bumping the versions. You seem to have different opinions.
Bumping minor version of touched sub modules on each PR: Yes or no?
Using BND baseline approach: Yes or no?

laeubi:
@cal101 Mickael is the lead of m2e so in doubt simply follow his advice :-)

@mickaelistria sorry for confusion. it would be really cool to have m2e (core) as an example using semantic versioning and BND already offers a tool for that. If I remember right last time I used it it gives very clear guidance how to bump (major, minor, micro) and this generally could lead to smaller deployments as one only needs to deploy changed items. Also, if using proper version ranges on package imports) there is less things to adjust for consumers and its always clear if its just a reformatting/bugfixin, API adjustments for consumer/provider or even an API break...

I found PDE baselining hard to setup and mainly different by eclipse needs than for actual semantic versioning. e.g your example bumping 1.16.1 to 1.18.0 should really be 1.16.2 regardless of the "release-version-number". That would even allow a rolling release where only single bundles are released.

mickaelistria:
>  Mickael is the lead of m2e so in doubt simply follow his advice :-)

I'm not the project lead , see https://projects.eclipse.org/projects/technology.m2e . The PL is @fbricon but these days he's not deeply involved in m2e so maybe it'd be worth officializing stuff. That's to be discussed on the project mailing-list.

> The bigger problem is getting me to understand the mixture of maven, m2, p2, tycho, bnd, m2e, eclipse, osgi and whatever else.

Sure, feel free to give it a try and report on the mailing-list whether it gives you satisfaction, what are the requirements to use it and so on. If there is a good ROI, we can make the project opt-for it.

> eclipse needs than for actual semantic versioning. e.g your example bumping 1.16.1 to 1.18.0 should really be 1.16.2 regardless of the "release-version-number"

That's the current practice in m2e. This can be discussed also on the mailing-list. I also favor semantic versioning as defined by OSGi and implemented in PDE API Tools, and -apparently even better- by BND.

cal101:
@mickaelistria Hopefully the last question on version bumping. To detect the last release is it sufficient to search for the last version tag like "tag: 1.17.2"? If yes combining that with +0.0.1 seems to be a simple sufficient rule at the moment which should be easy to implement.

mickaelistria:
>  To detect the last release is it sufficient to search for the last version tag like "tag: 1.17.2"? If yes combining that with +0.0.1 seems to be a simple sufficient rule at the moment which should be easy to implement.

That's one way of doing things a priori. Currently, what's happens is more that the validation build will fail and give instructions of what to bump. I was imagining that instead of re-building 1 tool for that, reusing the build output from Tycho/PDE/BND... would allow better factorization and separation of concerns.

cal101:
In theory I agree with you that using the tools that know the data and report the conflicts should be used to fix them.
In practice the maven/tycho duo frightens me and I fear a "mars attacks" like explosion of my brain.
Spending an hour on some bash scripting to get a "good enough" solution to fix this extra-on-top-non-functional-requirement solved sounds far less dangerous :-)

mickaelistria:
> Spending an hour on some bash scripting to get a "good enough" solution to fix this extra-on-top-non-functional-requirement solved sounds far less dangerous :-)

OK, please keep us posted of your progress via the m2e-dev@eclipse.org mailing-list. Either it's all good and then everyone is happy, or there will be some issues detected, but at least, the track for improvement would be open. I'm particularly interested in the integration with Pull Requests (as a kind of review bot), building that workflow is IMO even more important that the quality of the bot itself, that we can then improve.
One difficulty you may face is with the `jgit.ignore` Maven configuration that says (dynamically) what are the files to be ignored for the version change. That typically, but not always, include the pom.xml files for instance.

PR #85, opened 2021-03-21
Jdt convert for loop to enhanced clean up

cal101:
@mickaelistria Are there any plans on integrating a maven+eclipse (e.g. spotless) based formatter into the m2e project?
Some other cleanups need a formatter. 
Typically I would suggest doing it for changed files only, but i m2e erefactor case that would me still big.
Anyway to format only changed lines?
BTW: Bumping versions is a pain in the ass. The current script is a whole m2e specific hack.


mickaelistria:
@cal101 some conflicts need a manual rebase. Can you please take care of it?

PR #91, opened 2021-03-22
Jdt use diamond operator clean up 7

cal101:


akurtakov:
I really like cleaning the codebase like that. Do you plan to enable the warnings in the project settings and the save actions to auto fix these issues from now on?

cal101:
I didn't thought about that yet.
Things to consider:
- I don't know if all cleanups are valid as save actions.
- I would not want all of those actions to run automatically:
    - The erefactor pipeline automatically filters out cleanups per file that lead to compile errors. The dashboard dev lags
       behind so I don't have numbers how many that is.
    - Manual review shows errors and they are typically reported and excluded manually.
    - Running the test suite of the project sometimes shows interesting cleanup bugs.
There are a few actions I would consider save enough.
I won't want most of them doing changes without manual review.
I am following AutoRefactor for quite some years now (A good part of the JDT Cleanups are migrated by the current AutoRefactor maintainer from AutoRefactor.)
Big thanks to @JnRouvignac and @Fabrice-TIERCELIN for the fantastic work!
So I saw far to many edge cases that fail.

Something different:
I can't provide cleanups for all of m2e-core because I am not able to manually load the whole project into a current eclipse.
Some sub modules fail to load and so they are disabled from erefactor, too.
That is currently a precondition for execution of the cleanups.
Where can I get help on that? The current documentation does not work for me.

mickaelistria:
> Where can I get help on that? The current documentation does not work for me.

Please open specific bugs for issues you face on that matter, and we'll try to either fix them, or improve documentation accordingly.

PR #95, opened 2021-03-22
Jdt use diamond operator clean up 5

cal101:


cal101:
@mickaelistria I can't see how this change can have any effect for this test. How much of the waiting for jobs to finish is exact and how much is heuristics? Is it possible that something changed in eclipse platform so that the wait does not wait enough and it is a race condition now?

mickaelistria:
Looks like this change needs a rebase (this happens when changes are happening in a short time frame and merged in a different order than the one they were authored).

PR #142, opened 2021-03-31
[cleanup] erefactor/EclipseJdt - Remove redundant modifier

cal101:
Note that this breaks formatting slightly by replaing

```
@Deprecated
public void somethod()
```

with

```
@Deprecated void somethod()
```

If that's not acceptable we either need some formatter or get the cleanup changed.

EclipseJdt cleanup 'RemoveRedundantModifier' applied by erefactor.

For EclipseJdt see https://www.eclipse.org/eclipse/news/4.19/jdt.php
For erefactor see https://github.com/cal101/erefactor

Signed-off-by: Carsten Heyl <cal-1@web.de>

### DIFF
--- a/org.eclipse.m2e.editor/src/org/eclipse/m2e/editor/pom/PomTemplateContext.java
+++ b/org.eclipse.m2e.editor/src/org/eclipse/m2e/editor/pom/PomTemplateContext.java
@@ -594,11 +594,11 @@ public enum PomTemplateContext {
 
   private final String contextSuffix;
 
-  private PomTemplateContext(String nodeName) {
+  PomTemplateContext(String nodeName) {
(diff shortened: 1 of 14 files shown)


mickaelistria:
Actually, I merged the other one too fast: why is `private` sometimes removed to go to package-protected? I don't think it's equivalent and introduce more risk of APIs leaking by inadvertance.

cal101:
Btw: your epp tip is gold! Check out,set path,mvn and ... ... ... have tar. Great!

mickaelistria:
> I noticed that only on enums where I had to look it up that for enums the constructor without modifiers means private because outside the enum no instances are possible.

OK thanks!

Issue #120, opened 2021-03-30
Provide a Setup file + Button to instantly start coding


laeubi:
@merks has described a great opprotunity to have a quick-start fro contributors, we should offer a quick-link as this is great for new contributors and current instructions are hard to follow.

Copied from the mailing list:

> I think I recently read about that's it is possible to have a "button" in a html page that links to an Oomph setup file and simply creates the IDE+checkouts and so on. I think that should be the very first thing any project that wishes to attract developers should setup. 

Yes, this page has a link with such a button:

  https://www.eclipse.org/setups/installer/?url=https://git.eclipse.org/c/oomph/org.eclipse.oomph.git/plain/setups/interim/PlatformSDKConfiguration.setup&show=true

I.e., like this for those who don't follow links:
![grafik](https://user-images.githubusercontent.com/1331477/112962554-cb54c280-9146-11eb-93b4-14f9d9b57372.png)

The actual link URL on the button is this:
  eclipse+installer:https://git.eclipse.org/c/oomph/org.eclipse.oomph.git/plain/setups/interim/PlatformSDKConfiguration.setup

Most people use the installer in a disposable way, but of course one must keep it around more permanently and register it as a link handler to make direct use of such a single-click link.  That's why I provide a generic page through which to indirect the configuration link such that the user has documentation about different ways to use the configuration link.

The embedded documentation in the Configuration itself is also included on this indirection page.  In this case that documentation has a link to the tutorial documentation that I wrote specifically for the overall platform's configuration to make it easy for anyone to get started in order contribute to any part of the platform or all parts of the platform:

https://wiki.eclipse.org/Eclipse_Platform_SDK_Provisioning

So mostly one just needs a Project setup (I've already written ones for all the platform projects, so that's done), and a Configuration (which essentially just selects which Product Version to install and which Project Streams to provision):

https://wiki.eclipse.org/Eclipse_Oomph_Authoring#Automation_and_Specialization_with_Configurations

mickaelistria:
That would be a good addition to CONTRIBUTING.md , before the "Clone in Eclipse" button.
The issue you'll see is that GitHub doesn't allow links to non http/https URLs. So you'll probably need to build a redirect service for that, similarly to what's done in https://github.com/mickaelistria/redirctToEclipseIDECloneCommand . This repo could even be extended/forked to directly add the necessary magic for Oomph in it.

laeubi:
[...]

Issue #119, opened 2021-03-30
Bug 465483 - Support polyglot maven


mickaelistria:
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=465483

laeubi:
This relates to #118 as polyglot is integrated as a maven extension.

laeubi:
Requires
- https://github.com/eclipse-m2e/m2e-core/issues/546

laeubi:
It is now possible to import polyglot projects.

mickaelistria:
Awesome!

PR #150, opened 2021-04-01
Add some .project files

mickaelistria:
Import from Eclipse IDE is not functional with pom-less and without
.project.
So we add .project files to allow easy import in the Eclipse IDE.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,4 @@
 target/
 bin/
-.project
-.classpath
-.settings/
 .DS_Store
 /org.eclipse.m2e.binaryproject/
--- a/.project
+++ b/.project
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>m2e-core</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.m2e.core.maven2Builder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.m2e.core.maven2Nature</nature>
+	</natures>
+</projectDescription>
--- a/org.eclipse.m2e.core.tests/.project
+++ b/org.eclipse.m2e.core.tests/.project
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.eclipse.m2e.core.tests</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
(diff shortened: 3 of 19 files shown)


fbricon:
you need to restore the pom.xml too

mickaelistria:
> you need to restore the pom.xml too

Why?
Dealing with pom.xml version updates is causing difficulties to contributors. With .project + pom-less, versions can be updated only with a change in MANIFEST, which is much more convenient.

fbricon:
because you need to be able to import as Maven project.

mickaelistria:
> because you need to be able to import as Maven project.

I don't think so. All other Eclipse projects we're working on don't need that. Why would m2e have that constraint?

fbricon:
1 - dogfooding: m2e should be able to import itself
2 - import as maven project works better in general (actually imports the relevant maven projects instead of everything found under the tree - hello test projects)

mickaelistria:
I'm not convinced about the value provided by pom files for that case, while I experienced a significant maintenance cost.
I suggest we merge the addition of .project so people can easily contribute using standard workflows of plugin development, and consider addition of pom files in separate issue/PR.

fbricon:
If you ask me, the significant maintenance cost was introduced by the 1 version / per project thing. It's been a PITA for every contributor, trying to figure out what version to update.

Now, without a pom.xml, good luck updating the classpath automatically after changing the JavaSE version in the MANIFEST.MF. 
We get half-assed Maven projects for the project about Maven and that's pretty sad.

laeubi:
@fbricon honestly what are we talking about? Is it just about frustration/prefrences or is there a real technical (!) issue to solve here, 

> If you ask me, the significant maintenance cost was introduced by the 1 version / per project thing

One version for all artifacts is the real PITA as it completely contradict semantic versioning ... that versions need to match between manifest+pom is a complete different story anyways.

> Now, without a pom.xml, good luck updating the classpath automatically

How often do the JavaSE version changes where it is not possible to solve this by a simple search+replace and how often in contrast there is a change in the manifest version where it is required to change a specific artifact? 

> import as maven project works better in general (actually imports the relevant maven projects instead of everything found under the tree - hello test projects)

Only as long as these test projects are not pom.xml based itself and according to your dogfooding-theory we should instantly force to have al test project to have one and could be imported right away ;-)

As mentioned on the mailinglist we should better invest time and effort into #120 and #119

mickaelistria:
I merged a more complete patch with .project/.classpath added for all projects. Contributor can now import m2e as any other Eclipse plugin project.

fbricon:
So you tested it on a clean new workspace with a recent Eclipse IDE?

cal101:
I tried as far as I could.

See https://github.com/eclipse-m2e/m2e-core/issues/153 for log of actions.

Issue #163, opened 2021-04-08
Missing project-files and inconsistent .classpath configurations


HannesWell:
The plugin `org.eclipse.m2e.binaryproject` has no `.project` file and is therefore not imported when importing all projects from the m2e-core repo into a clean workspace, which results in compile errors.

For all feature projects the `.project` file is missing too. No feature project is therefore imported.

All these statements are based on the assumption that all projects from this m2e-core repo should be imported, so please correct me if I'm wrong.

Furthermore the `.classpath` files are configured differently. Some plugin-projects have the entry `<classpathentry kind="output" path="target/classes"/>` is used, but for plugin-projects the output folder `bin` is more common (and also used by some projects). Second some classpath-files in plugin-projects use `<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>` while I assume it should be `<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11"/>`

To me, it looks like the clean-up in PR #69 respectively commit 569b246d3a34f1bf43fad704a676d4a40d9d733e was too aggressive respectively it was forgotten to convert the former Maven-projects into pure PDE plugin/feature-projects. I assume it just worked at other peoples workspace because they imported the projects before the mentioned clean-up so the .project and .classpath files were generated at that import.
I found that some `.project` files are already added in #150 but the additions seem to be incomplete.

If you agree with the suggestions I can provide a PR to fix those issues.

HannesWell:
I just found that the entire project `org.eclipse.m2e.binaryproject` is included in the root `.gitignore` which makes me wonder if the project is to be ignored. But from were should the `org.eclipse.m2e.binaryproject.*` packages come from?

HannesWell:
It looks like the "src" paths are also not set up correctly for some projects and there exist general inconsistencies regarding the out-path between the .classpath and the build.properties file.

mickaelistria:
The main issue is that the .project and .classpath were part of .gitignore when pom-less was adopted, so their addition did fall under the radar. And the further addition was quick and dirty.
Anyway, sure, fixes to .project and .classpath are really welcome!

HannesWell:
Alright, I'm about to prepare a fix.

HannesWell:
Do you also want the `module=true` attribute set for the JRE container in all .classpath files?
[...]

PR #164, opened 2021-04-09
Fix and improve project configurations

HannesWell:
This PR fixes the inconsistent project configurations outlined in #163 and does some more minor clean up:
- add missing .project files
- fix inconsistencies between `.classpath` and `build.properties` and unify `.classpath` files in the repo
- change all output-directories of plug-in projects that were `target/classes` to `bin` and use 'src-gen' instead of `target/generated-sources/modello` as target for source code generation.
- upgrade the remaining projects to Java-11 (the depend on plug-ins that require Java-11 anyway).
- more minor clean up in project-configurations (fix warnings)

With this changes all projects from the m2e-core repo can be imported into a clean workspace and compile without errors (after the target-platform is resolved), which wasn't the case before.


.gitignore
```diff
@@ -1,5 +1,5 @@
 target/
 bin/
+src-gen/
 .DS_Store
-/org.eclipse.m2e.binaryproject/
```
HannesWell, line 4: I think this is critical since now changes in project `org.eclipse.m2e.binaryproject` are considered again.
I wonder if the exclusion was intentional?

org.eclipse.m2e.core/pom.xml
```diff
@@ -50,6 +50,7 @@
               <goal>xpp3-writer</goal>
             </goals>
             <configuration>
+              <outputDirectory>${project.basedir}/src-gen</outputDirectory>
```
HannesWell, line 53: At this point I'm not 100% certain if the generated .java-files are still compiled and included in the jars during Maven build?

mickaelistria: I think it's fine. Did you try a local build to verify that?

HannesWell: My local build is just running (fetching the dependency for the first time takes awhile).
I will check and report the results.

HannesWell: I verified that the classes from the now called `src-gen` folder in m2e.core (for example `org.eclipse.m2e.core.internal.lifecyclemapping.model.io.xpp3.LifecycleMappingMetadataSourceXpp3Reader`) are in `m2e.core.jar` as compiled class files and in `m2.core-sources.jar` as java source files.

mickaelistria:
Thank you!

PR #166, opened 2021-04-10
Include sources in development target-platform

HannesWell:
Having the sources of dependencies is handy during development.


mickaelistria:
ThanksQ

PR #168, opened 2021-04-10
Remove more unnncessary POMs and avoid creation of target-folder in m2e.core

HannesWell:
With this PR POMs are removed were not necessary anymore since Tycho's pomless capability is used.
Furthermore the resources section is removed from m2e.core's pom.xml since it is not necessary anymore and causes the Maven builder to generate a `target` folder in the project's root containing the `messages.properties` file.

I did a test build and the resulting jar files (regular and source) for `me2.core`, `m2e.lifecyclemapping.defaults`, `m2e.model.edit` and `m2e.scm` are binary equal to those jars created before. Additionally I verified that the `messages.properties` file is contained in the `m2e.core` jars, that `lifecycle-mapping-metadata.xml` is in the `m2e.lifecyclemapping.defaults` jar and the xsd files are in the `m2e.model.edit` jars.


mickaelistria:
Thanks!

PR #176, opened 2021-04-24
Add initial Oomph setup for m2e

HannesWell:
This PR adds an Oomph setup for m2e as part of issue #120 

At the moment the setup only contains the basic tasks based on [CONTRIBUTING.md](https://github.com/eclipse-m2e/m2e-core/blob/master/CONTRIBUTING.md)
- clone the m2e repo
- install all required m2e features (for dogfooding the latest snapshots are used)
- install those m2e connectors I could identify to be necessary (if there is a more elegant way please let me know)
- Import projects directly nested in m2e-core, m2e-maven-runtime, and m2e-core-tests and assign them to corresponding working-sets
- load the defined target-platform and perform refresh, clean, build
- Run an install build for m2e-maven-runtime
- set a JRE for JavaSE-11 to be used

Besides that more user preferences could be set (like activating save-actions) or more launch configs could be add (e.g. one launcher for an ordinary mvn clean verify build and one to build all, like the buildall,sh just startable from the IDE on Windows).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,19 +39,31 @@ Latest builds, for testing, can usually be found at https://download.eclipse.org
 
 ## 🧑‍💻 Developer resources
 
-### Source repositories
+### Prerequisites
 
-Clone this repository <a href="https://mickaelistria.github.io/redirctToEclipseIDECloneCommand/redirect.html"><img src="https://mickaelistria.github.io/redirctToEclipseIDECloneCommand/cloneToEclipseBadge.png" alt="Clone to Eclipse IDE"/></a> for m2e-core.
+Java 11 and Maven 3.6.3 (only if you want to build from the command-line), or newer.
 
+### ⌨️ Setting up the Development Environment automatically, using the Eclipse Installer (Oomph)
+
+1. Download the [Eclipse Installer](https://wiki.eclipse.org/Eclipse_Installer).  
+	1. If you are already in the workspace of an Eclipse provisioned by Oomph, go to *File > Import... > Oomph > Projects from catalog* and continue with step 6.
+2. Start the installer using the `eclipse-inst` executable.
+3. On the first page (product selection), click the preference button in the top-right corner and select the *Advanced Mode* .
+4. If you are behind a proxy, at this point you might want to double check your network settings by clicking in the *Network Proxy Settings* at the bottom.
+5. Select *Eclipse IDE for Eclipse Committers* . Click *Next* .
+6. Under *Eclipse Projects* , double-click on *m2e-core* (single click is not enough!). Make sure that *m2e-core* is shown in the table on the bottom. Click *Next*.
+7. You can edit the *Installation Folder* , but you do not have to select the *Target Platform* here, this will be set later automatically. By choosing *Show all variables* at the bottom of the page, you are able to change other values as well but you do not have to. Click *Next* .
+8. Press *Finished* on the *Confirmation* page will start the installation process. 
+9. The installer will download the selected Eclipse version, starts Eclipse and will perform all the additional steps (cloning the git repos, etc...). When the downloaded Eclispe started, the progress bar in the status bar shows the progress of the overall setup.
+10. Once the *Executing startup task* job is finished you should have all the *m2-core*, *m2-core-tests* and *m2e-maven-runtime* projects imported into three working sets called *m2-core*, *m2-core-tests* and *m2e-maven-runtime*.
+11. Happy coding!
+
+### ⌨️ Setting up the Development Environment manually
+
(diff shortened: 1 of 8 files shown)


mickaelistria:
Can you please include into this PR the changes to CONTRIBUTING.md with instructions about how to use Oomph to provision the IDE?

laeubi (commented):
I think a launch config that launches an Eclipse-Dev-IDE would be handy.

HannesWell:
> Can you please include into this PR the changes to CONTRIBUTING.md with instructions about how to use Oomph to provision the IDE?

Sure. Am I correct in assuming that the m2e-setup will be included into the `Eclipse projects` catalog of Oomph? Otherwise it had to be imported as user-project.

If yes I can also submit a bug against Oomph as described in the [Eclipse Oomph Authoring article](https://wiki.eclipse.org/Eclipse_Oomph_Authoring#Contributing_to_a_Project_Catalog) in order to initiate its inclusion.

HannesWell:
> I think a launch config that launches an Eclipse-Dev-IDE would be handy.

Sure. But for this it could be better to use the running Eclipse as Target-Platform or do you think the features/plugins in the dev-target-platform are sufficient for such an Eclipse?

laeubi:
The target should always be sufficient to run otherwise its not really a target :-)
But you can always add the current eclipse install as a target by adding 
`<location path="${eclipse_home}" type="Profile"/>`
to the target

HannesWell:
I've added the Launch-Config for an Eclipse-IDE and more launch configs for the different builds and updated the CONTRIBUTION.md mentioning the Oomph-setup and the launch-configs for build and IDE.

laeubi (approved):
I think this is good enough as a first step.

laeubi:
@HannesWell this currently has failing test, I'll restart the build to see if it is an intermediate failure, probably you need to rebase on master if that dosen't help.

HannesWell:
> I think this is good enough as a first step.

Agree. I have some more things to suggest, but as initial setup this should be sufficient. Especially because the setup should be included into Eclipse-project container. I think that should be done before we are going to fine-tune the setup.

keilw:
@HannesWell Which Eclipse versions are supported this way? And for at least the last 12 months or whatever  you can cover, are there or will there be "Available Updates" (I just spotted **m2e 1.17.2.20210219**) in the Dev IDE to fix those problems?

HannesWell:
> @HannesWell Which Eclipse versions are supported this way? And for at least the last 12 months or whatever you can cover, are there or will there be "Available Updates" (I just spotted **m2e 1.17.2.20210219**) in the Dev IDE to fix those problems?

Do you mean which Eclipse version is required to use this Oomph setup? It should be usable with older versions of Eclipse as long as the used Oomph Tasks are supported by the corresponding version of Eclipse-Oomph. But with that restriction I think you could use quite old versions of Eclipse.

> Actually **m2e 1.17.2.20210219** is useless because it would force an upgrade to Java 9 or 11 and we are bound to Java 8 in our environment. I have to experiment, if we could at least try 9 for the development environment while keeping the JRE at 8, hope all the other necessary plugins do not force e.g. using Java 11 because that would be a total no-go at this point?

The Eclipse platform requires Java-11 since version 4.17 (IIRC) which corresponds to Eclipse 2020-09. So yes the required plugins enforce Java-11 and I'm pretty sure it wont be possible to run Eclipse 2020-09 or later with an Java-9 JRE.
But of course you can use a Java-11 JRE to run Eclipse and develop against Java-8. Some Eclipse-products provided by Oomph are already shipped with a suitable JRE (see project JustJ/Oomph), so you don't have to install one on your system.

akurtakov:
Should we push this one too? Improvements can continue after that

laeubi:
Yes please if your are also fine with that go ahead.

keilw:
We luckily found a workaround for our Eclipse 2020-06 environments installing some of those plugins differently (also without Ooph there I assume ;-) I also saw JCenter would stay, but I guess the problem was because of a Bintray hosting and those won't come back.

akurtakov:
@HannesWell do you have other work pending? If not I would like to push the current patches

HannesWell:
> @HannesWell do you have other work pending? If not I would like to push the current patches

Not at the moment. There are more things that could be done in the setup, but I fully agree that this is fine as a first initial setup.

After this is merged I suggest to open a Bug report against Oomph to ask for inclusion of the m2e setup into the Eclipse-project container, so m2e appears on the Oomph Installer's Project selections page. I think this should be done before we talk about further improvements of the setup because then more people can already use the setup.
I can take care of this at the beginning of next week (today I'm busy with my every-day job).

akurtakov:
Thanks Hannes.

PR #177, opened 2021-04-24
Rename remaining 'src/main/java' src-folders to 'src'

HannesWell:
This is the last PR for issue #163 with the goal to rename the source folders of the following projects from `src/main/java` to `src` in order to be consistent to all other projects and comply with Eclipse's plug-in project convention:
- org.eclipse.m2e.editor.xml
- org.eclipse.m2e.logback.appender
- org.eclipse.m2e.model.edit

I run a build with this PR and with the previous state and I noticed that the Manifest.MF is different (some elements changed their order and the SHA-256-Digest Headers are missing) and the `META-INF/ECLIPSE_.RSA` respectively `ECLIPSE_.SF` are missing now. So I think the signing now fails but I did not find the location to fix it.

Another thing that makes me wonder in `org.eclipse.m2e.model.edit` are the files `Class.javajet`, `Header.javajetinc` and `emf-merge.xml`. They are located in the folder `src/main/templates` and I'm not certain were to place them when `src` is used as source-folder. Furthermore I don't understand their purpose but tend to think that they serve no purpose and could be left-overs of the initial code generation. If those files are not used or only required during code generation I suggest to remove them (from git).


mickaelistria:
> the META-INF/ECLIPSE_.RSA respectively ECLIPSE_.SF are missing now. So I think the signing now fails but I did not find the location to fix it.

You cannot produce Eclipse signatures from your local machine, this can only be done from Eclipse infrastructure. But usually, this kind of things work well and you don't need to care about them if they're different locally.

> They are located in the folder src/main/templates and I'm not certain were to place them when src is used as source-folder.

You could put them in a dedicated folder such as "templates".

> Furthermore I don't understand their purpose but tend to think that they serve no purpose and could be left-overs of the initial code generation. If those files are not used or only required during code generation I suggest to remove them (from git).

Code customizing of code generation is code. When one is using a generator, just like a compiler, the customization/source code is what will need to be maintained in some cases. We expect that if we need to make an update the the model (eg a new pom xsd appears), we should be able to just regenerate the code from it instead of having to modify things manually in the generated code.

HannesWell:
> You cannot produce Eclipse signatures from your local machine, this can only be done from Eclipse infrastructure. But usually, this kind of things work well and you don't need to care about them if they're different locally.

That makes sense, but what makes me wonder is that for the 'old' state (the current m2e-core master) signing seems to be done.
But I if you are fine with it, I'm fine with it.

> You could put them in a dedicated folder such as "templates".

Done.

> Code customizing of code generation is code. When one is using a generator, just like a compiler, the customization/source code is what will need to be maintained in some cases. We expect that if we need to make an update the the model (eg a new pom xsd appears), we should be able to just regenerate the code from it instead of having to modify things manually in the generated code.

Fully agree. When I open the `Class.javajet` I get a warning that the file cannot be compiled because the JET nature is missing. So it could be the case that the project is not properly configured for that task, but this can be addressed when necessary. My concern at the moment is just that everything will work fine for users of m2e.

HannesWell:
Do you have any more remarks?
For me this PR is complete and can be merged, unless you request more changes.

mickaelistria:
See output of the CI check, some versions must be bumped in feature.xml apparently.

HannesWell:
The `org.eclipse.m2e.feature` contains only lazy versions (0.0.0) but somehow the the site got the wrong build qualifier. But I don't know were it is comming from.
I've bumpded the service version segment of all three affected plug-ins hoping that this will solve the problem. If it does not help I will revert this bump.

Or can you give me instructions how to solve the problem?

HannesWell:
Because the build is still successful after reverting the version bumps I think it was the rebase that fixed it.

mickaelistria:
Thanks!

Issue #180, opened 2021-04-28
m2e feature is incomplete


laeubi:
I have added the m2e-core feature to a product but eclipse complains about missing requirements

![grafik](https://user-images.githubusercontent.com/1331477/116398639-513c5a00-a828-11eb-8ee8-c892c408efea.png)

I don't know if a feature can support package requirements, but at least the required bundles should be included in the feature so it is self-contained.

mickaelistria:
If you think about it that way, then it means that the m2e feature should also include the platform feature and all dependencies. In practice, features are not really meant to be self-contained, but are more a grouping of functionalities to help users in picking some interesting subsets. One issue with including dependencies is that features will force a specific version of each deps, which may then later cause conflicts because some other bundle require a different version.
I see this more as an issue of product launch failing at including necessary dependencies.

laeubi:
>  then it means that the m2e feature should also include the platform feature

it should actually list perquisites in the "dependency" section (currently empty).

>  One issue with including dependencies is that features will force a specific version of each deps, which may then later cause conflicts because some other bundle require a different version

But that's actually a limitation of the feature file-format and I suppose that "how platform works"... but it seems dependencies could specify version ranges in some kind.

My product currently includes pde, jdt nad platform but m2e requires guava 27 and platform does not require this by default, so I'm forced to write a feature my own that includes "a very specific version" and that's just contrary to "grouping of functionalities to help users in picking some interesting subsets".

laeubi:
It seems also some feature dependencies are not declared, if I use "add required" I still get missing bundles:

![grafik](https://user-images.githubusercontent.com/1331477/116403417-d24a2000-a82d-11eb-90e7-f1425aec3fb1.png)

mickaelistria:
> it should actually list perquisites in the "dependency" section (currently empty).

That the purpose of p2 to resolve transitive dependencies and allow features to not explicitly declare every dependency.

laeubi:
but whats the purpose of the dependency section then? Fact is, it is currently not possible to use the m2e feature without some magic knowledge....

mickaelistria:
> but whats the purpose of the dependency section then?

It's more or less to add extra requirements on things that do not come as transitive deps by lack of Require-Bundle/Import-Packages (eg more general Require-Capability that p2 isn't able to process).

m2e feature is included in several EPP products successfully. So I wouldn't say "not possible".
[...]

PR #181, opened 2021-04-28
m2e feature is incomplete #180

laeubi:
Signed-off-by: Christoph Läubrich <christoph@laeubi-soft.de>

### DIFF
--- a/org.eclipse.m2e.feature/feature.xml
+++ b/org.eclipse.m2e.feature/feature.xml
@@ -22,6 +22,11 @@
          version="0.0.0"
          optional="true"/>
 
+   <requires>
+      <import plugin="com.google.gson" version="2.2.4" match="equivalent"/>
+      <import plugin="com.google.guava" version="27.1" match="compatible"/>
+   </requires>
+
    <plugin
          id="org.eclipse.m2e.archetype.common"
          download-size="0"
--- a/org.eclipse.m2e.feature/feature.xml
+++ b/org.eclipse.m2e.feature/feature.xml
@@ -23,7 +23,7 @@
          optional="true"/>
 
    <requires>
-      <import plugin="com.google.gson" version="2.2.4" match="equivalent"/>
+      <import plugin="com.google.gson" version="2.2.4" match="compatible"/>
       <import plugin="com.google.guava" version="27.1" match="compatible"/>
    </requires>
 
--- a/org.eclipse.m2e.sourcelookup/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.sourcelookup/META-INF/MANIFEST.MF
@@ -17,9 +17,9 @@ Require-Bundle: org.eclipse.m2e.launching;bundle-version="[1.16.0,2.0.0)",
  org.eclipse.debug.ui;bundle-version="3.10.0",
  org.eclipse.core.resources,
  com.google.guava;bundle-version="[27.1,28.0)",
- com.google.gson;bundle-version="2.2.4",
  org.eclipse.core.variables;bundle-version="3.2.0"
-Import-Package: org.slf4j;version="1.6.2"
+Import-Package: com.google.gson;version="[2.2.4,3.0.0)",
+ org.slf4j;version="1.6.2"
 Bundle-RequiredExecutionEnvironment: JavaSE-11
 Bundle-ActivationPolicy: lazy
 Eclipse-BundleShape: dir


mickaelistria:
p2 is supposed to resolve dependencies from Import-Package. What's the benefit of adding transitive deps in the feature.xml ?

laeubi:
Maybe one can use P2 in that way, but in general this will not happen:

1. Eclipse PDE Product launch will complain about missing requirements
2. Tycho will either complain or build an incomplete product (depend on the type of missing features)
3. Update-sites will simply miss additional features (even if included the required **bundles**)

So for sure if you just install m2e into an existing eclipse and allows to pull in everything from everywhere it might work, but adding the feature m2e-core for example is not enough as transitive dependent features/bundles are not declared and need to be provided by hand.

And even though it seems hard to declare everything, that's how it is supposed to work. The feature.xml and its dependencies are the counterpart of the MANIFEST.MF and its declared dependencies for a bundle (e.g. if your bundle does not declare it requires DS it will happily install but most ptobably not fulfill the desired work).

The problem is, that most features are not crafted very well and the tool-support is bad (e.g. there is no validate-feature button afaik) but to the bottom line a feature should be self-contained that if I install it either everything is well defined (include) or declared (required) as a dependency so it could be looked up in the local or remotes.

mickaelistria:
> Eclipse PDE Product launch will complain about missing requirements

Is there a bug open to PDE on this topic (transitive deps not resolved on product launch)?

> Tycho will either complain or build an incomplete product (depend on the type of missing features)

That's surprising as Tycho does dependency resolution using p2. I think it's a bug in Tycho and should be reported and fixed there.

> a feature should be self-contained

So why not including all platform artifacts in it as well? Since inception of p2, a feature is just a "top-level" IU that's not a bundle and that allows to list a consistent set of bundles to provide some support for X or Y, it's not really intended to list all the transitive dependencies required to work.

laeubi:
> Is there a bug open to PDE

This is not a bug

> I think it's a bug in Tycho

Not really, a product is not required to contain everything as one can install (optional) components later on, beside that there is a validate-product goal if you don't like that feature (as well as PDE allwos you to still launch such a product)

> So why not including all platform artifacts in it as well?

In general, m2e should have one feature-dependecy to `eclipse-jdt` (don't know if there is actually such a single feature so from now on let us assume these features exits) and that's enough for the basics as `eclipse-jdt` should (transitively) include everything that's required there (e.g. platform itself), and one for each other not maintained feature (e.g `maven-embedded` might be such a feature, or `lemmix-xml-editor`) and include all m2e bundles that make up "the-core" or are not part of any stand-alone feature (e.g. google-guava, apache-common...).

So if all features are well defined, it is enough to create an **empty**, feature-based product and just add m2e core to get a full running ide with JDT+m2e that could be build/run right away!

Now we have the m2e-pde feature, that would have a requirement to `m2e-core` and `eclipse-pde`, if we use that (or install it later) our IDE will get JDT+PDE+m2e

mickaelistria:
> This is not a bug

Yet it's something that cause issue and push the need fo workarounds in
other places.

> Not really, a product is not required to contain everything as one can
install (optional) components later on

As far as I understand, we're not talking about optional componentsin this
PR, but of full requirements that are already explicited in MANIFEST.MF,
hence why the redundancy?

> In general, m2e should have one feature-dependecy to eclipse-jdt (don't
know if there is actually such a single feature so from now on let us
assume these features exits) and that's enough for the basics as
eclipse-jdt should (transitively) include everything that's required there
(e.g. platform itself), and one for each other not maintained feature (e.g
maven-embedded might be such a feature, or lemmix-xml-editor) and include
all m2e bundles that make up "the-core" or are not part of any stand-alone
feature (e.g. google-guava, apache-common...).
>
> So if all features are well defined, it is enough to create an empty,
feature-based product and just add m2e core to get a full running ide with
JDT+m2e that could be build/run right away!
>
> Now we have the m2e-pde feature, that would have a requirement to
m2e-core and eclipse-pde, if we use that (or install it later) our IDE will
get JDT+PDE+m2e

I don't get the relationship with this PR.
Also, I believe you miss the history behind p2: what you suggest is almost
exactly how things worked before p2, with features that were expected to be
perfectly defined, include all dependencies and so on. The result was
horrible: features were rarely correct, installing a feature required lots
of documentation reading, and then a lot of troubleshooting to find out the
missing dependency and so on. Basically, this illustrates that any strategy
that assume features are perfectly defined is doomed to failure; and that
we should rely on p2 for this kind of stuff.

laeubi:
A product (or OSGi runtime) is not required to include everything that is declared in the manifest. That's a common misunderstanding, OSGi allows installing thinks that are not yet fully satisfied (bundles, configuration, services). That's why a product, even though not all included bundles are fully satisfied, is valid and could be useful.

Having features that describe 'useful configurations' is very common even outside eclipse, every package-manager works that way (declaring whats required, whats optional and even whats recommend), Apache Karaf also has a much more straight feature-model that allows people to define good features even with configurations. Having a feature just for the sake of bundling some useful bundles is a really poor model...

If you say 'how things worked before p2' i can only repeat that maybe P2 was meant to solve such issues but obviously either is not integrated into the eclipse-eco system until now or does a bad job.

Just try out the example described above, empty feature based product, add m2e feature to it and it will fail in all cases (tycho, pde-launch, product export). Even adding 'obvious' dependencies like platform does not solve the issue... even adding "add required" makes the story even worse as it adds arbitrary features (some would call them useless ones), and still not ending with a valid product.

PR #187, opened 2021-05-06
Prevent NPE in NexusIndexManager

vrubezhny:
Fixes an NPE in NexusIndexManager:

!MESSAGE Reindexing error
!STACK 0
java.lang.NullPointerException
	at org.eclipse.m2e.core.internal.index.nexus.NexusIndexManager.purgeCurrentIndex(NexusIndexManager.java:529)
	at org.eclipse.m2e.core.internal.index.nexus.NexusIndexManager.reindexWorkspace(NexusIndexManager.java:561)
	at org.eclipse.m2e.core.internal.index.nexus.NexusIndexManager.updateIndex(NexusIndexManager.java:1072)
	at org.eclipse.m2e.core.internal.index.nexus.NexusIndex.updateIndex(NexusIndex.java:147)
	at org.eclipse.m2e.core.ui.internal.preferences.MavenSettingsPreferencePage$2.runInWorkspace(MavenSettingsPreferencePage.java:263)
	at org.eclipse.core.internal.resources.InternalWorkspaceJob.run(InternalWorkspaceJob.java:42)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)

Signed-off-by: Victor Rubezhny <vrubezhny@redhat.com>


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/index/nexus/NexusIndexManager.java
```diff
@@ -526,7 +526,9 @@ protected String getArtifactFileKey(String group, String artifact, String packag
   }
 
   private void purgeCurrentIndex(IndexingContext context) throws IOException {
```
laeubi, line 528: I wonder what the use of purgeCurrentIndex anyways if it just calls another method, maybe it would be better to inline the call?

vrubezhny: Honestly - I don't know. This method is `private` and used in `reindexLocalRepository()` and `reindexWorkspace()` methods of the same class, so we probably can easily get rid of it directly calling `context.purge()` instead (but imho we still need to have a check for null `context` value)

vrubezhny: @laeubi I have updated the PR - `reindexLocalRepository()` method is removed which made yet another possible NPE to be fixed in `reindexLocalRepository()`. 

There are still 4 JUnit Tests failing, however I don't know if all of them are related to the change:
- The one with `OverlappingFileLockException`, for sure, is related, but it's good to have it happen actually (this means that NPE doesn't happen during the test), `OverlappingFileLockException` error is a bug to be solved in https://github.com/eclipse-m2e/m2e-core/issues/169#issuecomment-833504762 
- 3 other tests fail with the same error: `org.junit.ComparisonFailure: expected:<commons-io-2.[5].jar> but was:<commons-io-2.[7].jar>` - I don't think it's something  related to the current issue. There are quite a lot of PR with this kind of failures in tests - I haven't investigated yet, but it looks like some common problem.

laeubi (approved):

PR #196, opened 2021-05-13
Mark test source folders as such

akurtakov:
As test harness plugin already does it and this fixes compilation.

Signed-off-by: Alexander Kurtakov <akurtako@redhat.com>

### DIFF
--- a/org.eclipse.m2e.editor.lemminx.tests/.classpath
+++ b/org.eclipse.m2e.editor.lemminx.tests/.classpath
@@ -6,6 +6,10 @@
 		</attributes>
 	</classpathentry>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
-	<classpathentry kind="src" path="src"/>
+	<classpathentry kind="src" path="src">
+		<attributes>
+			<attribute name="test" value="true"/>
+		</attributes>
+	</classpathentry>
 	<classpathentry kind="output" path="bin"/>
 </classpath>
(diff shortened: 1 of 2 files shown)


mickaelistria:
I think this change is actually not bringing best nor even expected outcome and that for OSGi bundles, "tests" scope should never be used for the exported packages. I let @laeubi provide more details while I'm on PTO ;)

laeubi:
It depends a bit what one wants to archive with that, so whats the goal of this change? Mark source folders as test makes mostly sense if there is also some "non test folders".

akurtakov:
So the dependency in platform marked it as test dependency https://git.eclipse.org/c/platform/eclipse.platform.ui.git/tree/tests/org.eclipse.ui.tests.harness/.classpath . Thus these bundles will not find tests.harness classes as they are available only for other test source folders and fail compilation if o.e.ui.tests.harness bundle is imported from git in the same workspace. IMHO this is good enough reason to put this change.

laeubi:
@akurtakov yeah then it is required here also even though its strange that platform do it that way as there is no option to mark a bundle as "test-only" the code will behave different if imported in a workspace or consumed from a target platform.

akurtakov:
So should this one be pushed? I don't see how it will behave different as all normal and test dependencies are available to test source folders.

laeubi:
This can be pushed, my comment was related to the org.eclipse.ui.tests.harness ...

laeubi:
I have created https://bugs.eclipse.org/bugs/show_bug.cgi?id=573534

PR #201, opened 2021-05-18
Improve Oomph-setup, adjust doc and add auto-launch Installer button

HannesWell:
This PR improves the m2e-Oomph-setup in the following way:
- add the variable eclipse.target.platform=None so it is not prompted by the installer (it is unused).
- Swap order of Launch of m2e-maven-runtime-installion and refresh/clean/build all. This way the maven-runtime installation is done before all projects are build, which can avoid a manual refresh/rebuild or restart.

In the CONTRIBUTING.md some names are corrected, settings of SSH keys are mentioned and the text is improved in general.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,20 +42,26 @@ Latest builds, for testing, can usually be found at https://download.eclipse.org
 ### Prerequisites
 
 Java 11 and Maven 3.6.3 (only if you want to build from the command-line), or newer.
+Furthermore a local git installation is required and the git executable must be in the PATH environment variable.
 
 ### ⌨️ Setting up the Development Environment automatically, using the Eclipse Installer (Oomph)
 
 1. Download the [Eclipse Installer](https://wiki.eclipse.org/Eclipse_Installer).  
 	1. If you are already in the workspace of an Eclipse provisioned by Oomph, go to *File > Import... > Oomph > Projects from catalog* and continue with step 6.
 2. Start the installer using the `eclipse-inst` executable.
-3. On the first page (product selection), click the preference button in the top-right corner and select the *Advanced Mode* .
-4. If you are behind a proxy, at this point you might want to double check your network settings by clicking in the *Network Proxy Settings* at the bottom.
-5. Select *Eclipse IDE for Eclipse Committers* . Click *Next* .
-6. Under *Eclipse Projects* , double-click on *m2e-core* (single click is not enough!). Make sure that *m2e-core* is shown in the table on the bottom. Click *Next*.
-7. You can edit the *Installation Folder* , but you do not have to select the *Target Platform* here, this will be set later automatically. By choosing *Show all variables* at the bottom of the page, you are able to change other values as well but you do not have to. Click *Next* .
-8. Press *Finished* on the *Confirmation* page will start the installation process. 
-9. The installer will download the selected Eclipse version, starts Eclipse and will perform all the additional steps (cloning the git repos, etc...). When the downloaded Eclispe started, the progress bar in the status bar shows the progress of the overall setup.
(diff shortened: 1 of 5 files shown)


HannesWell:
Additionally I removed the unintended second 'install' argument in the `m2e-maven-runtime--install` launch.

HannesWell:
From Ed Merks' testing of the m2e setup I now that a local git installation is required to run the maven build for the m2e-maven-runtime. Since I did not quickly find a replacement for using the exec-maven-plugin with the git.exe (I tried the pl.project13.maven::git-commit-id-plugin but this only provides the timestamp of the latest commit in the repo) I just updated the documentation accordingly.

HannesWell:
With the second commit I added a configuration setup that enables the use of a `Auto-launch Eclipse installer`- button as intended in #120. Besides that I added such a button to the `README.md` and the `CONTRIBUTING.md`.

The description of the configuration (which is displayed in the resulting auto-launch installer web-page) is based on the configuration of the [Eclipse Platform SDK Configuration](https://www.eclipse.org/setups/installer/?url=https://git.eclipse.org/c/oomph/org.eclipse.oomph.git/plain/setups/interim/PlatformSDKConfiguration.setup&show=true) and the [EMF Configuration](https://www.eclipse.org/setups/installer/?url=http://git.eclipse.org/c/emf/org.eclipse.emf.git/plain/releng/org.eclipse.emf.releng/EMFDevelopmentEnvironmentConfiguration.setup&show=true).
The button design is just copied from [EMF's CI page](https://ci.eclipse.org/emf/).

Thanks to @merks, who showed me the mentioned examples.

The button and the descriptions are a first shot, please don't hesitate to suggest improvements.
The alignment with the `Clone-to-Eclipse` button is not very nice, but I have only very few to none experience with markdown using html elements.

HannesWell:
@mickaelistria , @akurtakov or @laeubi do you have any remarks on this PR?

For me this PR is fine and I have no more plans to change it, unless you request changes.

merks:
It all looks good to me.  I'd like to see it action and try it out when it's committed...

merks:
It mostly works except the documentation doesn't show up.

The page uses Javascript but loads the XML using PHP.

And while this link works:

https://www.eclipse.org/setups/installer/fetch.php?url=http://git.eclipse.org/c/emf/org.eclipse.emf.git/plain/releng/org.eclipse.emf.releng/EMFDevelopmentEnvironmentConfiguration.setup

The m2e link does not:

https://www.eclipse.org/setups/installer/fetch.php?url=https://raw.githubusercontent.com/eclipse-m2e/m2e-core/master/setup/m2eDevelopmentEnvironmentConfiguration.setup

Probably it's some cross-site loading restriction...

laeubi:
As the URL loads fine in a browser it might be at the php level, can you share how the data is loaded in PHP?
I found that using [cURL](https://www.php.net/manual/de/book.curl.php) often works well while things like [fopen](https://www.php.net/manual/de/function.fopen.php) are often blocked/restricted by providers.

HannesWell:
Thanks for reviewing and testing!
I also tested it and came up with the same result. In the end the setup works fine.
But there are are a two more cosmetic flaws I recognized, besides the documentation/icons that is not showing up.

- The button is only a link on github, maybe the html tags (like that) are not supported?
Do you know if buttons in markdown are supported on github? If not I would just change it to be a link. No need to make it complicated.

- After the setup is completed, there are errors in the workspace. But they vanish if I refresh and clean and rebuild all projects.
Maybe a second refresh, clean, build task would fix that.

laeubi:
What works for the wiki probably also works in readme.md as well: https://gist.github.com/TT--/14260aef6c0e31fca5b37e7cb3c53020

PR #215, opened 2021-06-03
Remove unnecessary pom.xml and update pomless extension to Tycho 2.3.0

HannesWell:
With this PR I suggest to simplify some test-projects by removing their pom.xml files that are not absolutely required and to update the tycho pom-less extension to 2.3.0 (Tycho's build plug-ins are already used in version 2.3.0.)

The removed pom files were only used to specify if the UIHarness respectively -Thread should be used in surefire tests and to specify a test-suite and -class. The latter can be direclty replaced by a `pom.model.property.`-entry in the build.properties.
To make use of UIHarness/-Thread configurable by a `pom.model.property.`-entry I introduced corresponding variables in the root pom that are overwritten in the build.properties if necessary. This way the pom.xml files are not needed anymore and can be removed.

The surefire element in the poms of `org.eclipse.m2e.editor.lemminx.tests` and `org.eclipse.m2e.editor.xml.sse.tests` referred to a property named `requiresUIHarness` and `requiresUIThread` which does not exist (anymore). So they had no effect and can be removed without any further changes.

Besides that the pom.xml in the target-patlform project also had no effect and can also be removed without any further changes (tycho pom-less creates an equivalent pom).


mickaelistria:
Although I like pom-less in general to avoid redundant information, I'm not really enthusiast about the idea of moving settings that are in pom.xml (so well documented, with good tools to provide assistance, documentation, validation...) to some way less documented, tooled and standard format. IMO, keeping the configuration in pom files make the build more approachable to the average Maven users.

HannesWell:
Understand, that's a reasonable point.

I've updated the commit accordingly and now the pom's that were replaced by a `pom.model.property.`-entry in the build.properties are left unchanged.
However, I think the other pom's mentioned in my initially comment can be removed.

laeubi:
@HannesWell if you simply push (instead of force push) it much more easier to see what was adjusted, so no idea why pom.model.property was used by maybe these where unnecessary anyways?

mickaelistria:
@laeubi there is a discrete link on the "force-pushed" text in the GitHub line that allows to diff the commits. I think it's actually pretty recent, I remember asking this in some GitHub support ticket some months ago.

HannesWell:
> @HannesWell if you simply push (instead of force push) it much more easier to see what was adjusted, so no idea why pom.model.property was used by maybe these where unnecessary anyways?

I know this is not nice, but I wanted to avoid two commits, where the first one removes files and the second one re-adds them.
At this point Gerrit makes thinks easier than GitHub, in my opinion.
But as Mikael said you can still find the previous commit, even tough comparing the patch-sets is not that easy.

laeubi:
> At this point Gerrit makes thinks easier than GitHub, in my opinion.

Its easy in github as-well, simply use "sqash-and-merge" instead of "merge" or "rebase and merge" when merging commits.

HannesWell:
> Its easy in github as-well, simply use "sqash-and-merge" instead of "merge" or "rebase and merge" when merging commits.

That's right. But sometimes you want to have more than one commit in a PR.
But for future PRs I can act as you prefer it.

Issue #214, opened 2021-06-03
Synchronize and maybe unify target definitions into one


HannesWell:
With commit bbea072d694a386269ce20e0c85dce7e98f92581 the target definition `target-platform.target` and `dev-workspace.target` went out of sync (i.e. in the locations for `https://download.eclipse.org/eclipse/updates` and `https://download.eclipse.org/tools/orbit`).

This made me think about if those to target-definitions could be merged into one.
The difference between these to is that the `dev-workspace.target` contains sources, the maven-runtime bundles and the   `org.eclipse.sdk` feature additionally to the content of `target-platform.target`.
Having those in the target-platform used for builds should not have a negative impact on the out-come of the build and could only slightly increase the runtime of dependency resolution (I did not measure a difference at my computer, both took 3:06min, when all dependencies are fetched).

In favor of avoiding problems due to not synchronized target definitions and the synchronization effort, I suggest to merge those to target definitions into one.

Furthermore the maven-runtime bundles could be removed from the target-platform used in the workspace because with the Oomph setup they are imported into the workspace anyways. So if you consider using the Oomph setup as default, I suggest to remove the runtime-bundles from the TP.

I can provide a PR if you agree.

laeubi:
> This made me think about if those to target-definitions could be merged into one.

Good idea, tycho can be configured to ignore sources even if the are enabled in the target. Instead of the org.eclipse.sdk a reference to the current eclipse can be used (`<location path="${eclipse_home}" type="Profile"/>`) what will then be ignored by tycho as well...

HannesWell:
> Good idea, tycho can be configured to ignore sources even if the are enabled in the target. 

Thanks for that hint. I didn't know yet that this is possible. You mean the [`target-platform-configuration` plugin's `filter` parameter](https://www.eclipse.org/tycho/sitedocs/target-platform-configuration/target-platform-configuration-mojo.html#filters)?

>Instead of the org.eclipse.sdk a reference to the current eclipse can be used (`<location path="${eclipse_home}" type="Profile"/>`) what will then be ignored by tycho as well...

That would be one solution. But I'm not a big fan of using the current eclipse as part of the target-platform because it makes the target-platform depended of your running eclipse. And this in turn can lead to problems if your development eclipse is not the one the m2e in workspace target to. So I would prefer to filter out the `org.eclipse.sdk`, if necessary.
[...]

PR #228, opened 2021-06-09
[TP] Unify targets into one and improve Maven-runtime inclusion (#214)

HannesWell:
As suggested in #214 the two target definition files are merged into one to avoid the need for synchronization.

Besides that the maven-runtime bundles are now referenced via Maven-targets. This has the advantage that the artifacts installed in the local maven-repo of the developer are used and don't have to be published to a p2 repo first and one can test changes to the Maven-runtime locally.

As a side-note: those bundles (respectively m2e at all) are not published to maven-central, at least I did not found them. So this only works if the maven-runtime components are installed locally. But it is done in the build and the Oomph setup does it as well.

Also some cross-referencing comments were added which versions have to be kept synchronized.


mickaelistria:
Thanks a lot, that's yet another great improvement in making m2e easier to contribute to and maintain!

HannesWell:
Thanks and you're very welcome. It's great that my contributions are appreciated.

Issue #224, opened 2021-06-07
[Update-site] include all versions in the 'releases' and 'snapshots' updatesites


HannesWell:
The composite repositories `releases` and `snapshots` in `https://download.eclipse.org/technology/m2e/` do not specify all of the nested version folders as their child.
The releases repo contains only versions up to 1.15, the snapshot repo only contains the version 1.12 and 1.13.
If they would contain all nested repos as their child it would have the advantage that one can reference the composite and get the latest release or snapshot updates from the site automatically.

For example the snapshots repo is used in the m2e-core setup and its URL has to be updated with each release at the moment.

Furthermore I ask my self if it is necessary to have snapshot repos for each released version?
I know (p2) repos should never change their content, but they do not add real value because for all versions but the latest the content should be equal to the corresponding releases repo?
From that I think there should only be one snapshot repo that contains the latest master branch build and is replaced with each new build. So users that want the latest snapshot can use that repo.

A scenario were you want the latest snapshot but only until the the next release seems to be not very relevant to me, but please correct me if I'm wrong. I would use snapshots only for prototyping/testing/dog-feeding, so in an environment were I can change to the corresponding release, that contains the brand new feature/bug fix I only wanted, once it is published.

mickaelistria:
IMO, we should simply remove the composite repositories.

HannesWell:
For the snapshot repo I agree with you. Especially since m2e does not require them anymore since #228 is merged.

But for the `releases` repo, I think the situation is different and I find composite-repos very handy. For example I've include m2e in several Oomph setups in conjunction with Eclipse products that don't include m2e. For the eclipse products I've selected latest-released as version so the installations are updated automatically. But for m2e these setups have to be adjusted with each release. If I forget it, m2e stucks with an old release. I think others are probably in comparable situations.
Having a composite repo with all releases (especially the latest) would be much more convenient.

From a maintainers point of view it would be best if a new release would be added automatically when it is deployed. 
I'm not fully aware of the deployment process, but if you guide me, I can work on a solution.

Alternatively a `releases/latest` sub-repo could be provided, which either is a copy of the latest release repo or could be a composite with a single child pointing to the latest version. Then at each release only the pointer would have to be updated.

mickaelistria:
> From a maintainers point of view it would be best if a new release would be added automatically when it is deployed.
I'm not fully aware of the deployment process, but if you guide me, I can work on a solution.

[...]

PR #230, opened 2021-06-09
Add snapshot/latest to m2e's software repository and use it in m2e's Oomph setup (#224)

HannesWell:
With this PR I suggest to add a `snapshot/latest` software repository for m2e like proposed in #224.
The final URL of the resulting repo should be `https://download.eclipse.org/technology/m2e/snapshots/latest`.

I'm not an expert Linux's sh shell and also have no corresponding Linux system easily available were I could test the change. Therefore I ask you to double check if the change in the Jenkins file is valid. Thanks in advance.

Additionally the new `snapshot/latest` repo is used in m2e's Oomph setup.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -37,14 +37,21 @@ pipeline {
 			steps {
 				sshagent ( ['projects-storage.eclipse.org-bot-ssh']) {
 					sh '''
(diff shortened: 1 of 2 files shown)


mickaelistria:
Thanks!

PR #227, opened 2021-06-08
[cleanup] erefactor/EclipseJdt - Add missing @override annotation

cal101:
Not sure if you care for @Override annotations and if it is useful to clean up this way? It's a lot of changes.

EclipseJdt cleanup 'AddMissingOverrideAnnotation' applied by erefactor.

For EclipseJdt see https://www.eclipse.org/eclipse/news/4.19/jdt.php
For erefactor see https://github.com/cal101/erefactor

PR #231, opened 2021-06-09
Re-normalize line-endings in git of all files to Linux style ("\n")

HannesWell:
Some ASCII-files were checked-in into the m2e-core git having windows style line endings (`\r\n`).
But the vast majority of files in m2e-core is checked in into git using linux style line-endings (which is complies to the recommended git check-in policy: check-in linux style and on windows check-out windows-style).

On windows usually git should be set up to use auto-crlf to enforce the mentioned check-in/out policy.

Having mixed line-endings can lead to strange situations in the EGit staging-view. If a file with mixed line-endings is touched it appears as changed file in the staging-view but in the compare-editor, no difference is indicated. Additionally these files are often hard to reset to the head's state.
Therefore mixed checked-in line-endings should be avoided.

This PR re-normalizes the line-endings of all differing files to Linux-style.


HannesWell:
Since PR #227 is merged, the scenario described happened to occur, at least on my Windows computer.
For example the `org.eclipse.m2e.core.ui.internal.markers.MavenProblemResolution` now has mixed Windows- and Linux-style line-endings checked in into git.

This happened because Carsten Heyle either works on Linux or has his Windows git set-up correctly with enabled auto-crlf.

When you look into the commit message you can see the mixed line-endings:
grafik


Now I have `MavenProblemResolution` and `EditorAwareMavenProblemResolution` in my Staging-View as supposed changed files, but the Compare Editor shows no changes.
Unfortunately most of the time I can neither replace the changed files with their head-revision nor reset hard to the current head.
I can only commit the changes in a temp-commmit (which shows that the line-endings are unified, back to `\r\n`).

laeubi:
> Therefore mixed checked-in line-endings should be avoided.

While this would prevent the problems, it might be good to reprot these problems to Egit too so it could help the user (e.g. with a 'normalize whitespace' button).

mickaelistria:
Thanks!

HannesWell:
Thanks Mickael for merging!

> > Therefore mixed checked-in line-endings should be avoided.
> 
> While this would prevent the problems, it might be good to reprot these problems to Egit too so it could help the user (e.g. with a 'normalize whitespace' button).

You are right. I raised a bug against EGit: https://bugs.eclipse.org/bugs/show_bug.cgi?id=574192

Maybe you also have an idea how to solve this issue more permanently and participate in the discussion.

PR #233, opened 2021-06-10
Update to Guava 30.1.0

rgrunber:
- Bump minimum Guava version to 30.1.0 and eliminate 27.1.0 from target
  platform

Signed-off-by: Roland Grunberg <rgrunber@redhat.com>

@mickaelistria , anything I'm missing preventing such a change ? I recall Guava was updated in the maven-{runtime,indexer} bundles that embedded it, but I don't see why 27.1.0 was kept around for m2e.core.


mickaelistria:
CI seems as good as usual with that change, so let's merge. Thanks!

PR #237, opened 2021-06-15
Remove Maven builder and nature from projects without a pom.xml

HannesWell:
The .project file of the `org.eclipse.m2e.editor.xml.sse.tests` and `target-platform` projects contains a Maven builder command and the Maven nature.
But both projects don't have a pom.xml file, so I think those builders and natures can be removed.
It looks like I have been the one who forgot to remove the builders and natures when removing the pom.xml files of those projects with PR #215.


mickaelistria:
Thanks!

Issue #188, opened 2021-05-07
Bug 441978 - Snapshot sources/javadoc not updated with "Force Update of Snapshots/Releases"


rhoehener:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=441987 (2014)
https://bugs.eclipse.org/bugs/show_bug.cgi?id=367758 (2012)

mickaelistria:
Fixed with #238

mickaelistria:
I merged a related change which is available in latest snapshots. Please give it a try and report here if you see any issue regarding "force updates"

Issue #207, opened 2021-05-28
Bug 546184 - Update Maven Project Dialog: Remember last used settings, especially "Force Update of Snapshots / Releases"


rhoehener:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=546184

still suffering from this

mickaelistria:
Such change would be welcome. Are you willing to submit a patch for it?

rhoehener:
I am not, sorry.

PR #238, opened 2021-06-15
Bug 441978 - Snapshot sources/javadoc not updated with "Force Update …

vrubezhny:
…of Snapshots/Releases" #188

Issue: #188
Signed-off-by: Victor Rubezhny <vrubezhny@redhat.com>

### DIFF
--- a/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/BuildPathManager.java
+++ b/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/BuildPathManager.java
@@ -75,6 +75,7 @@ import org.apache.maven.artifact.repository.ArtifactRepository;
 import org.apache.maven.artifact.resolver.filter.ArtifactFilter;
 import org.apache.maven.artifact.resolver.filter.ScopeArtifactFilter;
 import org.apache.maven.project.MavenProject;
+import org.apache.maven.repository.RepositorySystem;
 
 import org.eclipse.m2e.core.MavenPlugin;
 import org.eclipse.m2e.core.embedder.ArtifactKey;
@@ -330,10 +331,14 @@ public class BuildPathManager implements IMavenProjectChangedListener, IResource
 
         ArtifactKey aKey = desc.getArtifactKey();
         if(aKey != null) { // maybe we should try to find artifactKey little harder here?
-          boolean downloadSources = desc.getSourceAttachmentPath() == null && srcPath == null
-              && mavenConfiguration.isDownloadSources();
-          boolean downloadJavaDoc = desc.getJavadocUrl() == null && javaDocUrl == null
-              && mavenConfiguration.isDownloadJavaDoc();
+          // We should update a sources/javadoc jar for a snapshot in case they're already downloaded.
+          File plainFile = desc.getPath() != null ? desc.getPath().toFile() : null;
+          File srcFile = srcPath != null ? srcPath.toFile() : null;
+          boolean downloadSources = (srcPath == null && mavenConfiguration.isDownloadSources())
+              || (plainFile != null && plainFile.canRead() && srcFile != null && srcFile.canRead());
+          File javaDocFile = javaDocUrl != null ? getAttachedArtifactFile(aKey, CLASSIFIER_JAVADOC) : null;
+          boolean downloadJavaDoc = (javaDocUrl == null && mavenConfiguration.isDownloadJavaDoc())
+              || (plainFile != null && plainFile.canRead() && javaDocFile != null && javaDocFile.canRead());
 
           scheduleDownload(facade.getProject(), facade.getMavenProject(monitor), aKey, downloadSources,
               downloadJavaDoc);
@@ -342,9 +347,34 @@ public class BuildPathManager implements IMavenProjectChangedListener, IResource
     }
   }
 
+  private static final String ARTIFACT_TYPE_JAR = "jar";
+
   private boolean isUnavailable(ArtifactKey a, List<ArtifactRepository> repositories) throws CoreException {
-    return maven.isUnavailable(a.getGroupId(), a.getArtifactId(), a.getVersion(), "jar" /*type*/, a.getClassifier(), //$NON-NLS-1$
-        repositories);
+    return maven.isUnavailable(a.getGroupId(), a.getArtifactId(), a.getVersion(), ARTIFACT_TYPE_JAR /*type*/,
+        a.getClassifier(), repositories);
+  }
+
+  /*
+   * Returns 'true' if a snapshot sources/javadoc jar doesn't exist locally or it exists and is outdated (its lastModificationTime 
+   * is less than lastModificationTime of according plain jar). 
+   * Returns 'false' only if a snapshot sources/javadoc jar exists locally and is not outdated (no need to be re-downloaded)
+   */
+  private boolean isOutdatedSnapshot(ArtifactKey a, List<ArtifactRepository> repositories) throws CoreException {
+    Artifact artifact = maven.lookup(RepositorySystem.class).createArtifactWithClassifier(a.getGroupId(),
+        a.getArtifactId(), a.getVersion(), ARTIFACT_TYPE_JAR, a.getClassifier());
+    ArtifactRepository localRepository = maven.getLocalRepository();
+
+    File artifactFile = new File(localRepository.getBasedir(), localRepository.pathOf(artifact));
+
+    if(artifactFile.canRead()) {
+      // artifact is available locally
+      Artifact artifactBin = maven.lookup(RepositorySystem.class).createArtifactWithClassifier(a.getGroupId(),
+          a.getArtifactId(), a.getVersion(), ARTIFACT_TYPE_JAR, null);
+      File artifacBintFile = new File(localRepository.getBasedir(), localRepository.pathOf(artifactBin));
+      return artifactFile.lastModified() < artifacBintFile.lastModified();
+    }
+
+    return true;
   }
 
 //  public void downloadSources(IProject project, ArtifactKey artifact, boolean downloadSources, boolean downloadJavaDoc) throws CoreException {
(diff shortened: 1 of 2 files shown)


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
```diff
@@ -862,7 +865,17 @@ public boolean isUnavailable(String groupId, String artifactId, String version,
 
     if(artifactFile.canRead()) {
       // artifact is available locally
-      return false;
+      // Need to compare last version to the artifact's one - if they're different then need to re-download
```
mickaelistria, line 868: This seems to be an unrelated optimization; that would better fit in the implementation of Maven itself (and I guess it's already present). Can you please move it to a separate this part in another commit?

mickaelistria (commented):
I'm failing to see the piece of code that does trigger update of sources/javadoc. Can you please point it to me?

vrubezhny:
> I'm failing to see the piece of code that does trigger update of sources/javadoc. Can you please point it to me?

When a Maven Project is being updated with forcing Snapshots/Release checkbox set (we're forced to download here by having downloadSources/downloadJavaDoc arguments set to true). According to Lifecycle the Maven project is cleaned, then configuration is updated, then it resoled and so on.   
At some point an `MavenProjectChangedEvent` is sent for the project. In `org.eclipse.m2e.jdt` plugin, as reaction on this event, `BuildPathManager` updates project's classpath and during this update the attached sources and javadocs are to be downloaded/updated for every library existing on classpath. But here we know NOTHING on are we forced to download sources or we are not forced, so `mavenConfiguration.isDownloadSources()` and `mavenConfiguration.isDownloadJavaDoc()` values used instead of initial "forced" arguments. 
By default mavenConfiguration disables the downloading. It should be OK in other kind of project updates as we should respect the configuration, but in case of forced update we must somehow override mavenConfiguration's values, but at this point we don't know if the update is to be  "forced".

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
vrubezhny: I don't understand what do you mean by "unrelated optimization" and why should we move it outside of this change...
This check gives a clear signal on that a sources snapshot jar needs to be updated (re-downloaded). Without this change the sources jar, once downloaded will never be updated.
I think making it to be up-to-date (when it's changed on remote repository) is the part of the fix for this exact issue.

mickaelistria: I mean that this line is in general about skipping update if artifact is already available, and your code is about doing the update whenever there is a newer version available remotely.
This is a good improvement in general, but I don't think it's a requirement for the story of "downloading sources/javadoc with force update".

mickaelistria (commented):
Thanks for the explanation, I know understand better what's going wrong.
So if the issue is in BuildPathManager, shouldn't the fix be in BuildPathManager (eg downloadSources would become 
```java
downloadSources = (desc.getSourceAttachmentPath() == null && srcPath == null && mavenConfiguration.isDownloadSources()) || (isSnapshot(desc) && lastUpdate(desc) > lastSourceUpdate(desc))
```
What do you think?

vrubezhny:
> Thanks for the explanation, I know understand better what's going wrong.
> So if the issue is in BuildPathManager, shouldn't the fix be in BuildPathManager (eg downloadSources would become
> 
> ```java
> downloadSources = (desc.getSourceAttachmentPath() == null && srcPath == null && mavenConfiguration.isDownloadSources()) || (isSnapshot(desc) && lastUpdate(desc) > lastSourceUpdate(desc))
> ```
> 
> What do you think?

Yes, I have the same idea. If sources jar exists and it's "older" then according plain jar - we should update the sources regardless of what 'mavenConfiguration.isDownloadSources()` returns. Then: - on first run (when no sources available yet) we will not download sources jar 
- after the source is requested (for example, if user opened a dependency class in editor) - sources jar is downloaded
- after the source is already downloaded it's always updated if it's older then its plain jar

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
```diff
@@ -853,16 +853,19 @@ public String getArtifactPath(ArtifactRepository repository, String groupId, Str
...
         classifier);
-
     ArtifactRepository localRepository = getLocalRepository();
 
     File artifactFile = new File(localRepository.getBasedir(), localRepository.pathOf(artifact));
 
     if(artifactFile.canRead()) {
       // artifact is available locally
-      return false;
+      Artifact artifactBin = lookup(RepositorySystem.class).createArtifactWithClassifier(groupId, artifactId, version,
```
vrubezhny, line 865: As the PR is updated the discussion is lost: https://github.com/eclipse-m2e/m2e-core/pull/238#discussion_r652737970

> I mean that this line is in general about skipping update if artifact is already available, and your code is about doing the update whenever there is a newer version available remotely.
> This is a good improvement in general, but I don't think it's a requirement for the story of "downloading sources/javadoc with force update".

This change is required. Just the fact of existence of the sources jar file before the change made this code to return false (=="artifact is available locally" == "no need to re-download"), thus skipping the sources jar from being updated.

We must return true here in case we really need to re-download the sources jar (true=="no artifact is available"=="need to re-download"), otherwise it will never be updated (regardless of if it "forced" or not) when its plain jar is updated by maven.

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/BuildPathManager.java
```diff
@@ -330,8 +330,15 @@ private void configureAttachedSourcesAndJavadoc(IMavenProjectFacade facade, Prop
...
         if(aKey != null) { // maybe we should try to find artifactKey little harder here?
-          boolean downloadSources = desc.getSourceAttachmentPath() == null && srcPath == null
-              && mavenConfiguration.isDownloadSources();
+          // We should update a sources jar for a snapshot in case it's already downloaded in case of
+          // plan jar is  updated by maven (its last modification time id older than the plain's one)
+          //
+          File srcFile = srcPath != null ? srcPath.toFile() : null;
+          File plainFile = desc.getPath() != null ? desc.getPath().toFile() : null;
+
+          boolean downloadSources = (srcPath == null && mavenConfiguration.isDownloadSources())
```
mickaelistria, line 339: same thing should be done for javadoc

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
mickaelistria: OK, I see, but this logic seems misplaced. MavenImpl shouldn't be touched for this work.
It should be in `BuildPathManager.isUnavailable()` or even better a new `BuildPathManager.isOutdatedSnapshot()` method, and then consumers in BuildPathManager would call `if (isUnavailable() || isOutdatedSnapshot())`

mickaelistria (changes requested):

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/BuildPathManager.java
vrubezhny: OK. Done

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/BuildPathManager.java
```diff
@@ -330,10 +331,14 @@ private void configureAttachedSourcesAndJavadoc(IMavenProjectFacade facade, Prop
...
-          boolean downloadJavaDoc = desc.getJavadocUrl() == null && javaDocUrl == null
-              && mavenConfiguration.isDownloadJavaDoc();
+          // We should update a sources/javadoc jar for a snapshot in case they're already downloaded.
+          File plainFile = desc.getPath() != null ? desc.getPath().toFile() : null;
+          File srcFile = srcPath != null ? srcPath.toFile() : null;
+          boolean downloadSources = (srcPath == null && mavenConfiguration.isDownloadSources())
+              || (plainFile != null && plainFile.canRead() && srcFile != null && srcFile.canRead());
+          File javaDocFile = javaDocUrl != null ? getAttachedArtifactFile(aKey, CLASSIFIER_JAVADOC) : null;
+          boolean downloadJavaDoc = (javaDocUrl == null && mavenConfiguration.isDownloadJavaDoc())
+              || (plainFile != null && plainFile.canRead() && javaDocFile != null && javaDocFile.canRead());
```
mickaelistria, line 341: Isn't this condition going to re-download sources even f the plainFile was not modified?

vrubezhny: Yes. But it doesn't mean that a snapshot jar will be actually re-downloaded.
I have removed the check for "outdated" snapshot from this method because it was a  duplicated one. So now it only enables the "download" and `BuildPathManager.isOutdatedSnapshot()` decides if we really need to re-download a snapshot jar.

mickaelistria: OK, I guess that the underlying Maven API do all the necessary optimization. Let's wait for CI feedback and we'll merge.

mickaelistria:
Thanks!

rhoehener:
Fantastic, thank you very much! Really looking forward to this.

Reading the comments... Do I still have to check that force update box or will the snapshots update if outdated even without this setting?

mickaelistria:
@rhoehener Please try the m2e snapshots from https://download.eclipse.org/technology/m2e/snapshots/latest/ to see this patch in action without having to wait for a release. With the patch, the sources&javadoc will be updates whenever "main" jar is updated. The "force updates" still decides whether to force updates of snapshots (main/sources/javadoc) or not.

rhoehener:
Thanks! I installed 1.18.1 and can confirm that the snapshot sources are updated correctly now :) 

The checkbox is still needed, I assume it corresponds to the `-U` maven option. Makes sense, and once [207](https://github.com/eclipse-m2e/m2e-core/issues/207) is done, I will never think about it again.

PR #239, opened 2021-06-15
Format all pom.xml files and unify license headers

HannesWell:
Some elements of the pom files were not formatted like other elements.
With this PR I suggest to unify the formatting of all `pom.xml` files.
I tried to use an XML Editor preference that modifies to pom files as few as possible and ended with the following settings:
- Line width: 120
- All boxes not checked except for
  - Format comments (but not join lines)
- Insert whitespace before closing empty end-tags
- Indent using spaces and indentation size 2

Do you think we should set these preferences in the Oomph setup, so everybody has the same setting **in the workspace**?

Besides that I added blank lines between all `dependency`, `plugin` and `profile` elements in order to increase the readability.

Additionally I unified the formatting of the license headers, updated them to the current year and added the missing header in the pom.xml of the `org.eclipse.m2e.editor.lemminx` project.
The initial commit for that pom file was authored by Andrew Obuchowicz using a redhat email address.
So I used the copy right header of pom's create by @mickaelistria as template.
I hope this is fine.


mickaelistria:
I'm not really finding this change particularly improves readability of pom.xml files; also I personally hate spaces as indent character: I like the width to be configurable and that's the whole purpose of tabs/columns. IMO, because it adds empty lines (thus takes more screen space) when there is no clear need for it, and because it reduce the indent width I see on my screen, I perceive this proposal as a regression.

HannesWell:
> I'm not really finding this change particularly improves readability of pom.xml files; also I personally hate spaces as indent character: I like the width to be configurable and that's the whole purpose of tabs/columns. IMO, because it adds empty lines (thus takes more screen space) when there is no clear need for it, and because it reduce the indent width I see on my screen, I perceive this proposal as a regression.

Alright, then no blank lines. Personally, I like to use some extra space to _group elements visually_. A blank line gives me a better immediate impression than indentation does. But that's a personal preference and we don't have to discuss that.

Regarding space as indent character: I just took it because it was used in most of m2e' pom files and I don't wanted to change more than necessary. I don't have a strong option about spaces or tabs as indent-character. So if you prefer tabs, I can change all poms to use tabs instead.

My main intention with this PR is to unify the format of all pom files.
I would like to be able to modify a pom with regards to content, then hit `Ctrl + Shift + F` to format my change (unfortunately there is no auto-formatting for pom files) and end up with my formatted change and not with five other whitespace modifications in the file. At the moment the latter is the case and it is a bit annoying, because then I have to sort out formatting changes manually in order to create minimal commits.

So if you give me a XML Editor setting, I would be happy to apply it. What do you think about?
```
- Line width: 120
- All boxes not checked except for
  - Format comments (but not join lines)
  - Insert whitespace before closing empty end-tags
- Indent using tabs and indentation size 1
```
And no extra blank lines.

HannesWell:
I added a modification-commit to this PR that removed the additional blank lines.

If you agree on the proposed formatter-setting of my previous comment (mainly use tabs instead of spaces), then I can apply them, add one more commit and this branch can be squashed and merged.

mickaelistria:
You're now a committer and can merge your own changes.
Formatting related changes are usually controversial as there is not 1 good formatting, it's very subject to personal tastes. I don't plan to get involved in that more than I already expressed; so it's up to you to decide what and when to merge; I personally will cope with your choices even if they're not my favorites.

HannesWell:
> You're now a committer and can merge your own changes.
> Formatting related changes are usually controversial as there is not 1 good formatting, it's very subject to personal tastes. I don't plan to get involved in that more than I already expressed; so it's up to you to decide what and when to merge; I personally will cope with your choices even if they're not my favorites.

Formatting is indeed a very controversial topic, that can lead to endless discussions.

Alright, thanks. As soon as commiter right's to this GitHub repository are granted to me, I will merge this PR, unless somebody has objections by then.

For the records: all project pom.xml files were formatted with the following XML Editor preferences set:
```
- Line width: 120
- All boxes not checked except for
  - Format comments (but not join lines)
  - Insert whitespace before closing empty end-tags
- Indent using tabs and indentation size 1
```

PR #254, opened 2021-06-29
Don't fail on empty goals list in plugin execution filter

atanasenko:
Some plugins might define empty sets of goals. One example is `ca.vanzyl.provisio.maven.plugins:provisio-maven-plugin`, it is built using takari-lifecycle which generates  lifecycle-metadata file automatically from takari-based mojos, but in this case the only mojo is plain maven mojo, so it leaves the set of goals empty.
The mojo in question is bound to package phase, so it's not supposed to be 'interesting', but this check breaks the whole build completely including unrelated dependent modules.

```
<lifecycleMappingMetadata>
  <pluginExecutions>
    <pluginExecution>
      <pluginExecutionFilter>
        <goals/>
      </pluginExecutionFilter>
      <action>
        <configurator>
          <id>io.takari.m2e.incrementalbuild.builderMojoExecutionConfigurator</id>
        </configurator>
      </action>
    </pluginExecution>
  </pluginExecutions>
</lifecycleMappingMetadata>
```

### DIFF
--- a/org.eclipse.m2e.core/mdo/lifecycle-mapping-metadata-model.xml
+++ b/org.eclipse.m2e.core/mdo/lifecycle-mapping-metadata-model.xml
@@ -164,7 +164,7 @@
   }
 
   private void checkAllFieldsSet() {
-    if (groupId==null || artifactId==null || versionRange==null || goals==null || goals.isEmpty() ){
+    if (groupId==null || artifactId==null || versionRange==null || goals==null ){
       throw new IllegalArgumentException(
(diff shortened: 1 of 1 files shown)


mickaelistria:
thanks, merged as https://github.com/eclipse-m2e/m2e-core/commit/cdd10dcd1f5f99263042b45e9f53bfcbb6c120c1

mickaelistria:
Merged as https://github.com/eclipse-m2e/m2e-core/commit/cdd10dcd1f5f99263042b45e9f53bfcbb6c120c1

Issue #223, opened 2021-06-06
[Maven-runtime projects] Generate Eclipse plug-ins meta-data and include them in m2e-core build


HannesWell:
At the moment the projects in `m2e-maven-runtime` are pure Maven projects. Because they aren't Eclipse plug-ins, they do not participate in a Eclipse Workspace's PDE build and therefore bundles like `org.eclipse.m2e.maven.runtime` are only considered from the target platform and not from the workspace. This has the down-side that the maven-runtime bundles have to be part of the target-platform and it is not sufficient to have them in the workspace.
Additionally if you modify the maven-bundles in the workspace and want to test them e.g. by launching an Eclipse-Application, you first have to get them into the target-platform (the known maven-install procedure and then you can include them for example using a Maven-target). This is quite cumbersome.

Therefore I suggest to generate the required Metadata files (driven by Maven) to make the maven-runtime projects full Eclipse-plugins too. 
This requires mainly two things: the nested jars need to be fetched and the MANIFEST.MF need to be generated into the META-INF folder.
For the latter a m2e-profile is already present in the `m2e-maven-runtime/pom.xml` but this does not seem to have any effect in my workspace. I suspect that's because the `bundle` goal of the maven-bundle-plugin is not specified explicitly as execution (does m2e even honor lifecycle goals automatically?). But even if it is covered my Eclipse did not find a connector for it.

However the jar fetching could be done, using the `copy-dependencies` goal of the `maven-dependency-plugin` and the manifest generation by using the `manifest` goal of the maven-bundle-plugin. I've already done some prototyping for the `org.eclipse.m2e.maven.runtime` in my [m2e-fork](https://github.com/HannesWell/m2e-core/commits/runtime).
I still have to experiment which project natures and builders are required (maybe the .project file needs to be commited) and if the jars need to be added solely to the `.classpath` in order to make it a fully functional Eclipse plugin that participates in the PDE workspace builds and can be used in launched Eclipse applications.

If you like this idea, I continue to fully implement it.

Besides that I want to suggest to integrate the maven-runtime projects into the maven build of m2e-core so that those projects don't have to be build separately. Are there general objections against that? It would have the advantage that only one build is required and configurations/versions/properties can be inherited.

mickaelistria:
That's a great goal, and the implementation seems to be on the right track.
[...]

PR #260, opened 2021-06-30
Unify Maven-runtime project Poms

HannesWell:
This is the first step towards #223, aiming to unify the poms of the Maven-runtime projects and to avoid duplication.

The only relevant change in the resulting jar files is that all embedded jar files are now contained in a folder named `jars`. Before this was only the case for `maven.runtime` and `maven.runtime.slf4j.simple`.
Additionally the execution of the `tycho-p2-plugin:p2-metadata` goal is moved to the `eclipse-sign` profile in the first commit. So that goal is only executed on Jenkins and not locally. This behavior is the same for the 'main'-build of the other m2e projects.

Using the `m2e-core` as parent allows to remove many elements from the maven-runtime pom.xml, including evolving configuration like the tycho- or m2e-version properties which needs to be synchronized manually.
Due to this change in the second commit the goals `jacoco-maven-plugin:prepare-agent` and `tycho-source-plugin:plugin-source` are now also executed in the maven-runtime build. But I didn't expect and notice any negative impact from that.

Furthermore the project's metadata in the poms, like the issue management, scm connections and mailing-list, are unified and updated. The old mailing-list links still work, but seem to redirect to the new values.

### DIFF
--- a/m2e-maven-runtime/org.eclipse.m2e.archetype.common/pom.xml
+++ b/m2e-maven-runtime/org.eclipse.m2e.archetype.common/pom.xml
@@ -98,10 +98,6 @@
 					</instructions>
 				</configuration>
 			</plugin>
-			<plugin>
-				<groupId>org.eclipse.tycho</groupId>
-				<artifactId>tycho-p2-plugin</artifactId>
-			</plugin>
 		</plugins>
 	</build>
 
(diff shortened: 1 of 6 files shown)


mickaelistria:
I think it's fine to merge.
However, I noticed an issue (which isn't a regression from this patch) is that the timestamp version would change whenever parent pom changes.

HannesWell:
I didn't observe that issue yet, but also had an eye on it. Is this because of the way the commitDate is computed for the Maven-components?

However if my plans for the Maven-runtime components works, they participate as eclipse-plugins in the regular build and use tycho's git timestamp provider, which works well from my experience.

Issue #208, opened 2021-06-01
Slow Maven import of deeply nested projects


nwwerum:
When importing a deeply nested multi module project (with many intermediate poms), the import into eclipse takes a very long time. This sort of structure happens frequently in (very) large projects.

A simple reproducer can be found here: https://github.com/nwwerum/m2eslow-test
Its a project with 10 subfolders containing poms, the last of which contains 30 jar projects.
Importing this into eclipse takes more than 3 minutes on my machine:
> 6/1/21 2:59:55 PM CEST: [INFO] Imported and configured 41 project(s) in 202 sec

A command line build of this project is fast:
> [INFO] Total time: 6.069 s

The problem goes away when leaving out the intermediate poms (a root pom listing the leaf project as \<module\>s), even when keeping the folder structure, see https://github.com/nwwerum/m2efast-test
> 6/1/21 2:55:45 PM CEST: [INFO] Imported and configured 31 project(s) in 7 sec

So the problem seems to be related to the number of parent/ancestor poms a project pom has.

As a hint for other people suffering from slow imports: as long as this issue is not fixed, reducing the number of intermediate/parent poms (especially if they are only there because of convention and do not add any dependencies or plugins) in your projects makes a _huge_ difference for eclipse import performance (but no difference for command line builds). If doing so, keep in mind to adapt the relative path of your poms (e.g. \<relativePath\>../../../../\</relativePath\>) to point to the correct parent.

nwwerum:
A few further notes:

This also impacts the performance of "update project", but not quite as strongly as for import.
Update time for the slow reproducer:
> [INFO] Update completed: 52 sec

Update for the fast variant:
> [INFO] Update completed: 6 sec

I had a quick look (by getting a few thread dumps) at what eclipse is doing during the slow import - most of the time seems to be taken for file IO - it seems that for each imported project, all ancestor poms are read from disk. Example dump:

```
"Worker-283: Importing Maven projects" #5632 prio=5 os_prio=0 tid=0x000000003d57e000 nid=0x1063c runnable [0x000000003713c000]
   java.lang.Thread.State: RUNNABLE
        at java.io.FileInputStream.readBytes(Native Method)
        at java.io.FileInputStream.read(FileInputStream.java:255)
        at java.io.BufferedInputStream.fill(BufferedInputStream.java:246)
        at java.io.BufferedInputStream.read(BufferedInputStream.java:265)
[...]

PR #261, opened 2021-06-30
[Setup] perform Maven-runtime installation before TP resolution

HannesWell:
Since the Maven-runtime bundles are included via the target-platform, it is necessary to install them before the target-platform is resolved! 
The m2e Maven-runtime bundles are not published to Maven-central and the intention is to obtain them from the local .m2 repository. But of course this this only works if they are installed into the local .m2 repository first.
This should fix the setup issues mentioned in #208

I didn't consider this when including the Maven-runtime bundles via the target-platform. For me it worked because the Maven-bundles were installed before, but it does not work for somebody who runs the setup for the first time.

Additionally the Maven-runtime installation now runs on every start-up to ensure that developers always have the latest Maven-runtime bundles installed.

### DIFF
--- a/setup/m2e.setup
+++ b/setup/m2e.setup
@@ -224,11 +224,12 @@
     </setupTask>
   </setupTask>
   <setupTask
+      xsi:type="launching:LaunchTask"
+      launcher="m2e-maven-runtime--install"
+      runEveryStartup="true"/>
+  <setupTask
       xsi:type="pde:TargetPlatformTask"
       name="m2e-target-platform"/>
-  <setupTask
-      xsi:type="launching:LaunchTask"
-      launcher="m2e-maven-runtime--install"/>
   <setupTask
       xsi:type="projects:ProjectsBuildTask"
       refresh="true"


laeubi:
@HannesWell I really would find it interesting to extend m2e-pde support to this scope, so if the target includes a reference to a project in the workspace, maybe I should take a deeper look how "resolve workspace artifacts" works :-)

HannesWell:
That would indeed be interesting. It would go towards being able to mix Maven- and Eclipse Plug-in projects in one build.
On one hand I think it could be useful, on the other hand the question arises why one does not convert the corresponding project into an Eclipse-Plug-in project if it is used in with others? I think for m2e to situation is quite special, but maybe there are other use cases.

laeubi:
> on the other hand the question arises why one does not convert the corresponding project into an Eclipse-Plug-in project

One major problem (from my POV) what holding Eclipse/PDE back for so long times is the assumption that there is always only "the eclipse way" and that this way is such superior that other projects are willing to adopt whatever strange "best practice" is advertised by Eclipse.

I would turn your question the other way round: Why should I ever force anyone to convert a (working) maven project into an  Eclipse-Plug-in project just for the sake of using that in a PDE context just because PDE/m2e fails to integrate with industry-standards?

We are wasting so much time convincing people that they are doing things "wrong" instead investing that time into making the tools smarter.

That's why I think, whenever we see a limitation (here my initial implementation of m2e-pde seem not understand workspace projects) we should check how this could be solved in a way that requires the smallest effort on the user side and best integrates with common patterns.

HannesWell:
> One major problem (from my POV) what holding Eclipse/PDE back for so long times is the assumption that there is always only "the eclipse way" and that this way is such superior that other projects are willing to adopt whatever strange "best practice" is advertised by Eclipse.

I don't think Maven or Eclipse-PDE is superior over the other. Both have their advantages and disadvantages and often things are as they are and you have to use one because of other circumstances and nothing is wrong about it.
But I see your point and agree that, if feasible with an reasonable effort, it should be simple for the user.

I just asked myself about use cases and thought if there is a simpler solution.
As I said being able to mix Maven and Eclipse projects in one build would indeed be a great feature but I think it would be very difficult because of the differences how Maven and Eclipse-PDE describe and handle dependencies. I don't see an obvious solution atm.
My main concern is about how this should be handled in Tycho? The Maven projects would have to be build before the Target-Platform resolution takes place, or at least they have to be pre-process until the point where all OSGi metadata are available that are required for the TP resolution, for example the available packages).

But if you want to say that this is only a convenient for Maven projects that have to be build in a previous Maven-run this should be no factor and should be possible within the Eclipse-IDE (while resolving the target the project just have to be build if it is not build yet).

In fact at the moment I have use-case where it would be handy, where I have imported the project of a Maven dependency I want to use but it does not work as expected and I'm about to test some possible changes that could fix my problems.
Furthermore this would also allow us to keep the Maven-runtime-components as pure Maven projects as proposed in #223 (nevertheless I will continue with to work for it so we have it earlier, we still can convert it back later). 

So yes, let's go for it!

One more thing: What to you think about an umbrella issue for m2e-pde where we can list and discuss the plans for the future of m2e-pde at a higher level? This could help to get a better overview and should reduce the chance for incompatible partial developments.

laeubi:
You can simply filter on the pde-integration label:

https://github.com/eclipse-m2e/m2e-core/issues?q=is%3Aopen+is%3Aissue+label%3Apde-integration

PR #263, opened 2021-07-01
[Target-Platform] include eclipse.sdk instead of running Eclipse

HannesWell:
With issue #214 we discussed to include the bundles of the running Eclipse instance into the target-platform in order to be able to launch Eclipse-Application for debugging. To include the running Eclipse instance was favoured over including the `org.eclipse.sdk` feature because the bundles of the running Eclipse are ignored during the Maven/Tycho build.

However, even I use the Oomph-setup with the `latest Eclipse for Commiters`, I get errors about multiple versions of the same plug-in when launching the debug Eclipse application. At the moment it is one of the `wildwebdeveloper` plugin-ins.
Of course the Run-Configuration for the Eclipse application could be adjusted but this would not be very convenient because it would have to be updated continuously.

Therefore I suggest to include the `org.eclipse.sdk` instead which solves the issue.
I attempted to exclude the `org.eclipse.sdk` during build, but have the impression that this does not work (at least I didn't get a print-out that it is filtered).
@mickaelistria and @laeubi, what do you think about this? Should we just include the sdk-feature and not filter it for the build at all? I think this wouldn't be too bad.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -138,12 +138,6 @@
 					<pomDependencies>consider</pomDependencies>
 					<filters>
 						<filter>
-							<!-- Remove as it is only included to be able to launch Eclipse-application for debugging -->
-							<type>p2-installable-unit</type>
-							<id>org.eclipse.sdk.feature.group</id>
-							<removeAll />
-						</filter>
-						<filter>
 							<!-- Remove as it may come as an undesired transitive dep from p2 under Java 11 -->
 							<type>p2-installable-unit</type>
 							<id>javax.xml</id>
--- a/pom.xml
+++ b/pom.xml
@@ -138,6 +138,12 @@
 					<pomDependencies>consider</pomDependencies>
 					<filters>
 						<filter>
+							<!-- Remove as it is only included to be able to launch Eclipse-application for debugging -->
+							<type>p2-installable-unit</type>
+							<id>org.eclipse.sdk.feature.group</id>
+							<removeAll />
+						</filter>
+						<filter>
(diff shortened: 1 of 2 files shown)


mickaelistria:
+1 for this approach. The reference to local Eclipse instance was a convenience workaround to get the m2e-maven-runtime bundles visible in the TP. But since they're now referenced as Maven artifacts in the TP, this workaround can be removed.
However, why having the whole SDK in, and then why filtering it out? Wouldn't using the Platform+JDT features be enough?

HannesWell:
>However, why having the whole SDK in, and then why filtering it out? Wouldn't using the Platform+JDT features be enough?

The filtering was just one alternative suggestion in #214 because bundles of the running Eclipse are also not considered during Tycho builds. But you are right, actually this does not make sense.

Regarding the question why the whole sdk:
The `org.eclipse.sdk` feature contains the features `org.eclipse.platform`, `org.eclipse.jdt`, `org.eclipse.pde`, `org.eclipse.help` and the corresponding source features.

For m2e we require `org.eclipse.platform`, `org.eclipse.jdt`, `org.eclipse.pde` and jdt contains help. So we end up with the same features anyways. Nevertheless pde also has a hard dependency to the jdt feature, so if you only include pde you will have pde and jdt and the platform feature also found it's way into the TP (I haven't traced the exact path, but probably some of the features in the other locations have dependencies to it).
In the end, the only difference is that the source features are also included in the TP. But since the `includeSources` attribute of all locations is true the including plug-ins should remain the same.

And with the SDK the pde feature and the platform.ide plugin can be removed.

PR #268, opened 2021-07-07
[Maven-Runtime] clean up, mainly BND instructions (to fix warnings) #223

HannesWell:
This PR is another preparation to resolve issue #223 which aims to fix warnings during the initial build, of which some later lead to errors in PDE when the runtime bundles are unpacked. Additionally some clean up is done.

The important changes are:
- ~~remove Export-Packages entries for all META-INF packages~~:
The `maven.runtime` and `archetype.common` bundles do not have the folder specified as exported-packages. This is an error in PDE when the bundle is unpacked.
Besides that I think it is not necessary to specify `META-INF.sisu` as exported package in the indexer bundle because it only contains the `javax.inject.Named` file which is an ordinary ASCII file and not a class file.
Or is it required for the indexer to work properly that the mentioned file is in an exported package?
_Edit: Those entries are necessary. See comment below._

- fix version of org.eclipse.aether version and made it a property hoping that it is more obvious that it needs to be synchronized.

The clean up changes are:
- ~~remove own bundles from Bundle-ClassPath (i.e. the entry '.') for modules that do not provide own classes (which is true for all except the indexer bundle)~~ EDIT: they still seem to be relevant relevant.
- include the dot-Bundle-Classpath entry in the parent maven-runtime/pom.xml and only override it in the archetype project to not include self/the dot.
- Delete .factorypath files since they seem to serve no purpose (at least I don't know any, if they are relevant please let me know).


HannesWell:
There are test errors (but somehow the build succeeds nevertheless...). Have to investigate those first.

HannesWell:
Contrary to my initial expectationthe Export package entries are required. [Plexus states that it uses `ClassLoader.getResources()` in order to find component.xml files](https://codehaus-plexus.github.io/guides/developer-guide/configuration/index.html). And if I saw it right the `EquinoxClassloader/ModuleClassLoader` respectively the `BundleLoader` check the exported packages before searching for resources on required-bundles But I'm not certain about that. So in order to be save I just leave all exported-packages as they are and have a look for another solution to avoid errors when the manifests are unpacked.

HannesWell:
Never change a running system...
It looks like the dot in the Bundle-Classpath is also required for the maven.runtime module. Otherwise the `org.eclipse.m2e.editor.lemminx.tests.EditorTest` fails, but it is not clear to me why.

However I changed the default Bundle-Classpath in the maven-runtime/pom.xml to include the dot (`<Bundle-ClassPath>.,{maven-dependencies}</Bundle-ClassPath>`, which is suitable for all but the archetype module) and only override it in the archetype module to not include the doc.

And I wonder if the dot should also be in the `Bundle-ClassPath` for the archetype module as well?

laeubi:
The `dot` represents the classes of the bundles itself. If you omit that, you will end up with an empty bundle (all compiled classes missing).
If the bundle itself does not define any classes but only consists of external jars, the `dot` can be omitted even though I won't recommend that as its confusing if some times later someone adds a class to the bundle..

HannesWell:
> The `dot` represents the classes of the bundles itself. If you omit that, you will end up with an empty bundle (all compiled classes missing).
> If the bundle itself does not define any classes but only consists of external jars, the `dot` can be omitted even though I won't recommend that as its confusing if some times later someone adds a class to the bundle..

Yes exactly, but the question is: Why is 'required' for the dot/the jar itself to be on the classpath for the maven.runtime bundle? It does not define any own classes. But if the dot is missing the test mentioned above fails.

For the runtime.slf4j.simple jar it makes sense because it contains the `simplelogger.properties` file in its root, which I assume is accessed via `ClassLoader.getResource()` or similar, which explains that the jar itself _needs to be on its own classpath_. The only explanation I have why it is also required for the runtime jar is, that the `bundle.properties` in the `OSGI-INF` are also accessed via the ClassLoader. I think that the `about_files` and the `about.html` are only contained for legal reasons and are not relevant at runtime.

If this is the right explanation I think the archtype jar should also have the dot on its classpath and we were just lucky that it did not yet cause any failures.
However, I agree with you that the dot should be there for reasons of convention as long as it is not harmful.

mickaelistria:
I'm not good at BND, I would only look at test results here and as long as there is not regression in tests, that looks good to me.
I let you and @laeubi decide of whether this patch makes things better or not, and merge or not according to your conclusions.

HannesWell:
From my point this can be merged.
To only real change with an effect on the build jars is that the archtype module now contains the dot in its `Bundle-Classpath`.
Deleting the `.factorypath` files seems to have no effect (they probably were not used) and fixing the versions of the eclipse.aether/maven-resolver packages is also for the better.

So if @laeubi has no objects I squash and merge this and continue with the next steps towards #223.

laeubi (approved):
Looks good to me

HannesWell:
Thanks.

PR #270, opened 2021-07-10
[PDE] fix bugs in properties-changed check, lastAccess file location and others

HannesWell:
This PR fixes some (minor) bugs in m2e-PDE:
- `MavenTargetBundle.propertiesChanged()` did not compare the new and old properties right
- CacheManager placed the `lastAccess` file in the root directory of the running Eclipse instead of the root of the cache folder
- consider artifact classifier only if not blank
- close resources in BNDInstructions by making them resources of the try-block and unify creation of the reader for the default instructions


laeubi (approved):
Looks good to me.

HannesWell:
Thanks for reviewing.

PR #276, opened 2021-07-17
[Maven-Runtime] make Maven-runtime projects Eclipse plug-ins (#223)

HannesWell:
This PR is the current state of the change to solve issue #223.
The Maven-runtime modules are converted into ordinary Eclipse Plug-ins and are integrated as such into the main m2e Maven build.

This has the main advantage that those bundles don't have to be integrated via the target platform, which simplifies their handling:
- no need to reload the target platform if the maven-runtime changes
- less versions to synchronise 

Besides that the requirement for a git installation can also be dropped.

However the build still requires two steps. In the first build step the MANIFEST.MF (and the .classpath) is generated and the jars embedded into the runtime bundles are fetched. As second step the actual m2e Maven build is performed.

The changes work for the IDE and the Maven build.

The current limitations are:
- MANIFEST.MF and .classpath cannot be generated on incremental builds since it would end in an endless loop (probably the usual problem). Therefore the meta-data are only regenerated when the corresponding build is performed, either explicitly as Maven build or as part of the Oomph setup.
- no sources are available (this isn't a regression but would be nice to have).

What do you think about it?

### DIFF
--- a/m2e-maven-runtime/org.eclipse.m2e.maven.indexer/pom.xml
+++ b/m2e-maven-runtime/org.eclipse.m2e.maven.indexer/pom.xml
@@ -19,8 +19,8 @@
 	</parent>
 
 	<artifactId>org.eclipse.m2e.maven.indexer</artifactId>
-	<version>1.18.0-SNAPSHOT</version><!--Keep it in sync with version in target-platform.target -->
-	<packaging>bundle</packaging>
+	<version>1.18.1-SNAPSHOT</version>
+	<packaging>eclipse-plugin</packaging>
 
 	<name>Nexus Indexer Bundle</name>
 
@@ -98,36 +98,43 @@
 				<version>0.3.4</version>
 				<executions>
 					<execution>
+						<?m2e execute onIncremental?>
 						<id>index-project</id>
 						<phase>prepare-package</phase>
 						<goals>
 							<goal>index</goal>
 						</goals>
+						<configuration>
+							<outputDirectory>${project.basedir}</outputDirectory>
+						</configuration>
 					</execution>
 				</executions>
 			</plugin>
-			<plugin>
-				<groupId>org.apache.felix</groupId>
-				<artifactId>maven-bundle-plugin</artifactId>
-				<configuration>
-					<instructions>
-						<Embed-Dependency>
-							*;scope=compile|runtime
-						</Embed-Dependency>
-						<_exportcontents>
-							META-INF.sisu;-noimport:=true,
-							org.apache.lucene.*;provider=m2e;mandatory:=provider,
-							org.apache.maven.*;provider=m2e;mandatory:=provider,
-						</_exportcontents>
-						<Import-Package>!*</Import-Package>
-						<Require-Bundle>
-							org.eclipse.m2e.maven.runtime;bundle-version="[1.18.0,1.19.0)",
-							org.eclipse.m2e.archetype.common;bundle-version="[1.18.0,1.19.0)"
-						</Require-Bundle>
-						<Include-Resource>{maven-resources},META-INF/sisu/javax.inject.Named=${project.build.outputDirectory}/META-INF/sisu/javax.inject.Named</Include-Resource>
-					</instructions>
-				</configuration>
-			</plugin>
 		</plugins>
+		<pluginManagement>
+			<plugins>
+				<plugin>
+					<groupId>org.apache.felix</groupId>
+					<artifactId>maven-bundle-plugin</artifactId>
+					<configuration>
+						<instructions>
+							<Embed-Dependency>
+								*;scope=compile|runtime
+							</Embed-Dependency>
+							<_exportcontents>
+								META-INF.sisu;-noimport:=true,
+								org.apache.lucene.*;provider=m2e;mandatory:=provider,
+								org.apache.maven.*;provider=m2e;mandatory:=provider,
+							</_exportcontents>
+							<Import-Package>!*</Import-Package>
+							<Require-Bundle>
+								org.eclipse.m2e.maven.runtime;bundle-version="[1.18.0,1.19.0)",
+								org.eclipse.m2e.archetype.common;bundle-version="[1.18.0,1.19.0)"
+							</Require-Bundle>
+						</instructions>
+					</configuration>
+				</plugin>
+			</plugins>
+		</pluginManagement>
 	</build>
 </project>
--- a/m2e-maven-runtime/org.eclipse.m2e.maven.runtime/metadataTemplates/.classpath
+++ b/m2e-maven-runtime/org.eclipse.m2e.maven.runtime/metadataTemplates/.classpath
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11">
+		<attributes>
+			<attribute name="module" value="true"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry kind="src" path="src/main/resources"/>
+	<classpathentry kind="output" path="target/classes"/>
+	${dependency.jars.lib.entries}
+	<!--
+	TODO: simply use the Maven-Classpath-Container entry. Then this does not need to be copied?! This should also include the sources?!  
+	And for the long term create a directory jars classpath container?! use JDTCore.newLibraryEntry to create an entry for each jar in a directory (think about the sources?!).
+	-->
+</classpath>
--- a/m2e-maven-runtime/pom.xml
+++ b/m2e-maven-runtime/pom.xml
@@ -25,6 +25,8 @@
 
 	<properties>
 		<archetype-common.version>2.4</archetype-common.version>
+		<dependency.jars.folder>jars</dependency.jars.folder>
+		<metadata.templates.folder>metadataTemplates</metadata.templates.folder>
 	</properties>
 
 	<modules>
@@ -35,79 +37,32 @@
 	</modules>
 
 	<build>
-		<plugins>
-			<plugin>
-				<groupId>org.codehaus.mojo</groupId>
-				<artifactId>exec-maven-plugin</artifactId>
-				<version>3.0.0</version>
-				<executions>
-					<execution>
-					<?m2e ignore?>
-						<goals>
-							<goal>exec</goal>
-						</goals>
-						<phase>initialize</phase>
-						<configuration>
-							<executable>git</executable>
-							<commandlineArgs>log -1 --format='commitDate=%cd' --date=format:'%Y%m%d-%H%M' -- .</commandlineArgs>
-							<outputFile>${project.build.directory}/timestamp.properties</outputFile>
-						</configuration>
-					</execution>
-				</executions>
-			</plugin>
-			<plugin>
-				<groupId>org.codehaus.mojo</groupId>
-				<artifactId>properties-maven-plugin</artifactId>
-				<version>1.0.0</version>
-				<executions>
-					<execution>
-						<?m2e ignore?>
-						<goals>
-							<goal>read-project-properties</goal>
-						</goals>
-						<phase>initialize</phase>
-						<configuration>
-							<files>
-								<file>${project.build.directory}/timestamp.properties</file>
-							</files>
-						</configuration>
-					</execution>
-				</executions>
-			</plugin>
-		</plugins>
-
 		<pluginManagement>
 			<plugins>
-				<plugin>
-					<groupId>org.apache.maven.plugins</groupId>
-					<artifactId>maven-compiler-plugin</artifactId>
-					<version>3.8.1</version>
-					<configuration>
-						<!-- http://maven.apache.org/plugins/maven-compiler-plugin/ -->
-						<source>11</source>
-						<target>11</target>
-					</configuration>
-				</plugin>
 				<plugin>
 					<groupId>org.apache.felix</groupId>
 					<artifactId>maven-bundle-plugin</artifactId>
 					<version>5.1.1</version>
 					<extensions>true</extensions>
 					<configuration>
+						<!-- PDE does not honor custom manifest location -->
+						<manifestLocation>META-INF</manifestLocation>
+						<supportedProjectTypes>eclipse-plugin</supportedProjectTypes>
 						<instructions>
 							<Embed-Transitive>true</Embed-Transitive>
-							<Embed-Directory>jars</Embed-Directory>
+							<Embed-Directory>${dependency.jars.folder}</Embed-Directory>
 
 							<_failok>true</_failok>
 							<_nouses>true</_nouses>
 							<_nodefaultversion>true</_nodefaultversion>
-							<_snapshot>${commitDate}</_snapshot>
+							<_snapshot>qualifier</_snapshot>
 
 							<Bundle-SymbolicName>${project.artifactId};singleton:=false</Bundle-SymbolicName>
 							<Bundle-RequiredExecutionEnvironment>JavaSE-11</Bundle-RequiredExecutionEnvironment>
 							<Bundle-Name>%Bundle-Name</Bundle-Name>
 							<Bundle-Vendor>%Bundle-Vendor</Bundle-Vendor>
 							<Bundle-ClassPath>.,{maven-dependencies}</Bundle-ClassPath>
+							<Automatic-Module-Name>${project.name}</Automatic-Module-Name>
 
 							<Eclipse-BundleShape>dir</Eclipse-BundleShape>
 						</instructions>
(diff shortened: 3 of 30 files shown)


HannesWell:
The problem of missing sources could be addressed if the source jars were fetched too. Then corresponding source attributes could be added to the classpath entries. Fetching the sources can be done with the [maven-dependency-plugin](https://maven.apache.org/plugins/maven-dependency-plugin/examples/using-dependencies-sources.html).

In order to avoid the need to specify dependencies twice (once for the maven-bundle-plugin to include them in the jars and once for the dependency plugin to fetch the sources), I suggest to exclude all not wanted transitive dependencies in the dependencies sections of the pom. Then we could simply fetch and include all dependencies.
With this I think it would be simpler to use the Maven-dependecy-plugin also for fetching the (main) jars and to use the maven-bundle-plugin only to generate the manifest and not to create a full bundle and to unwrap it in the next step.

Alternatively I thought about using m2e's Maven classpath container to add the jar dependencies to the JDT classpath. But this somehow did not work and could potentially lead to inconsistencies because things would be mixed up. A Classpath container to include all jars in a folder would be nice instead. Nevertheless excluding unwanted dependencies in the dependencies section of the pom would also be required here.

HannesWell:
Furthermore I just noticed that in the m2e.maven.indexer jar the other (original) class-files in the `org.apache.maven.index` package are missing and that the `META-INF/sisu/javax.inject.Named` file has missing entries

mickaelistria:
That's really good progress, I can't wait to include this change!

> Furthermore I just noticed that in the m2e.maven.indexer jar the other (original) class-files in the org.apache.maven.index package are missing and the corresponding entries in the META-INF/sisu/javax.inject.Named file are missing. I have to find out, if this is a problem or if they can be found in the originals nevertheless.

That's a fragile part of m2e. So I'd recommend that for a 1st iteration, you try to make this on par with the current state (ie keep the .class files in it and the javax.inject.Named entries) to reduce the risk.

HannesWell:
> That's really good progress, I can't wait to include this change!

Thanks, me too!

> > Furthermore I just noticed that in the m2e.maven.indexer jar the other (original) class-files in the org.apache.maven.index package are missing and the corresponding entries in the META-INF/sisu/javax.inject.Named file are missing. I have to find out, if this is a problem or if they can be found in the originals nevertheless.
> 
> That's a fragile part of m2e. So I'd recommend that for a 1st iteration, you try to make this on par with the current state (ie keep the .class files in it and the javax.inject.Named entries) to reduce the risk.

That's a good point. The thing is, I haven't removed them actively. They just vanished with the other changes, so I first have to figure out how I can bring them back.


With the latest changes I changed the way the way the manifest is created and the bundles are fetched. Instead of creating a complete bundle-jar with the bundle:bundle goal and unzipping that jar afterwards, the Manifest is now created with the bundle:manifest goal and the dependencies are fetched using the maven-dependencies-plugin.
With this the manve-runtime bundles now have sources available in the development workspace, what I consider a big advantage.
I also started to prepare proper source bundles, but at least when launching a debugged Eclipse-App from the m2e-workspaces the sources are not yet attached properly.

If this works fine, I will submit the commit "[Maven-Runtime] manage embedded jars fully by dependencies" in a separate PR.

HannesWell:
Because the Export-Package directrices for `META-INF/plexus` and `META-INF/sisu` causes errors in the MANIFEST and require some nasty workaround (add the dummy files), I spend some more time to better understand the details and to find out if the Export-Package directrices for `META-INF/plexus` respectively `META-INF/sisu` are really necessary and why. This has been in my mind for a while (e.g. see https://github.com/eclipse-m2e/m2e-core/pull/268#issuecomment-876764092)  and I think I found the final answer now:

The container/component frameworks used by Maven, namely Codehaus Plexus respectively Eclipse Sisu, use `ClassLoader.getResources(String)` to find resources like [`/META-INF/plexus/components.xml`](https://codehaus-plexus.github.io/guides/developer-guide/configuration/index.html) or `/META-INF/sisu/javax.inject.Named` from where they read their configuration.

When `getResources(String)` is called on a modular EquinoxClassloader it finds the resources contained in the bundle of that ClassLoader plus the the resources contained in all imported packages and (transitively) required bundles, **but only if the resource is in a package that is exported by the required bundle**.

So for example when the ClassLoader of the `org.eclipse.m2e.archetype.common` bundle is asked for all resources with name `/META-INF/plexus/components.xml` it finds the `components.xml` files in all jars nested in it plus the `components.xml` files in the jars nested in the bundle `org.eclipse.m2e.maven.runtime` (which m2e.archtype.common requires), but only if m2e.maven.runtime exports the package `META-INF.plexus`. If m2e.maven.runtime does not export package `META-INF.plexus` the classloader finds only the `components.xml` in the m2e.archtype.common plugin, but those in m2e.maven.runtime seem to be required as well.

So I assume the strategy of exporting META-INF/plexus and sisu mimics a flat classpath where all resources can be found.

If you are interested in the details, the following Stack excerpt is a good starting point:
	BundleLoader.addExportedProvidersFor(String, List<PackageSource>, Collection<BundleLoader>) line: 975	
	BundleLoader.addExportedProvidersFor(String, List<PackageSource>, Collection<BundleLoader>) line: 970	
	BundleLoader.findRequiredSource(String, Collection<BundleLoader>) line: 1221	
	BundleLoader.findResources(String) line: 742	
	EquinoxClassLoader(ModuleClassLoader).getResources(String) line: 250	
	ClassRealm(ClassLoader).getResources(String) line: 1463	

`ClassRealm` is the `org.codehaus.plexus.classworlds.realm.ClassRealm`.

So the bad news is that the exported packages `META-INF/plexus` and `META-INF/sisu` are really required. But the good news is, that I think it is a bug in PDE that those exports are marked as an error.
I will open a bug report against PDE and provide a fix for it. This will avoid the initially mentioned workarounds are not necessary.

laeubi:
> So the bad news is that the exported packages `META-INF/plexus` and `META-INF/sisu` are really required. But the good news is, that I think it is a bug in PDE that those exports are marked as an error.

I must confess that exporting a "package" META-INF.plexus sound really strange as it really it not a package.. if we really like to merge the class-path of org.eclipse.m2e.archetype.common we should instead either make those other items a fragment or use require-bundle instead (as it is done with org.eclipse.m2e.maven.runtime if I understand correctly).

laeubi:
If m2e even creates the ClassRealm it might even be better to create a more OSGI aware version of it...

HannesWell:
I have created [Bug 575419](https://bugs.eclipse.org/bugs/show_bug.cgi?id=575419) and submitted a fix.

HannesWell:
> 
> 
> > So the bad news is that the exported packages `META-INF/plexus` and `META-INF/sisu` are really required. But the good news is, that I think it is a bug in PDE that those exports are marked as an error.
> 
> I must confess that exporting a "package" META-INF.plexus sound really strange as it really it not a package.. if we really like to merge the class-path of org.eclipse.m2e.archetype.common we should instead either make those other items a fragment or use require-bundle instead (as it is done with org.eclipse.m2e.maven.runtime if I understand correctly).

I it indeed feels odd. Require-Bundle is already used by archtype.common, but for this to work the package must be exported. Transitive resources are only considered if their package/folder is exported.

Making one a fragment of the other could work, as long only classes from archtype.common are asked for resources like the components.xml. If classes that require e.g. archtype.common or maven.runtime are asked for those resources the chain is again broken and I don't know if it is required.

> If m2e creates the ClassRealm it might even be better to create a more OSGI aware version of it...

I cannot say if this is possible (until know I have almost done nothing with ClassRealm, so my knowledge is very limited atm.). But I assume it is woven quite deeply into Maven.

mickaelistria:
I also believe exporting META-INF.whatever package is a very bad smell and should be avoided.
What we expect from the Maven "wrapper" jars in m2e is that they become the single entry-point for things like classloading. IMO they should contain some specific META-INF/plexus or META-INF/sisu folders that happen to be a kind of merge of the various content from the children files; or -maybe better- some manually maintained file that are designed to work for the m2e/OSGi case.

HannesWell:
> I also believe exporting META-INF.whatever package is a very bad smell and should be avoided.
> What we expect from the Maven "wrapper" jars in m2e is that they become the single entry-point for things like classloading. IMO they should contain some specific META-INF/plexus or META-INF/sisu folders that happen to be a kind of merge of the various content from the children files; or -maybe better- some manually maintained file that are designed to work for the m2e/OSGi case.

I agree with you that this is definitely not nice and if would be feasible I would like to change this as well.
Nevertheless I'm afraid this will be hard.
Your suggestion would be one option but I cannot estimate the maintenance effort exactly, but I fear it could be high and error prone with hard to find problems. Furthermore it would be again the problem, if a class in a bundle that requires archetype.common or maven.runtime, is asked for e.g. all component.xml files they would again not be found.
So from the discoverability perspective the current approach is IMHO the safest and simplest one, that also works.

For now I would like to keep it as is for the mentioned reasons, but I we find a better solution I'm happy as well.

I could additionally mark the packages as `;x-internal:=true` but I think we don't gain much.
In the m2e workspace I cannot add an imported package META-INF.sisu/plexus, but in a launched Eclipse workspace with the running platform as TP it is possible.

m2e-maven-runtime/pom.xml
```diff
@@ -47,146 +50,212 @@
...
-							<configuration>
-								<!-- PDE does not honour custom manifest location -->
-								<manifestLocation>META-INF</manifestLocation>
-							</configuration>
+							<groupId>org.eclipse.tycho</groupId>
+							<artifactId>tycho-packaging-plugin</artifactId>
+							<version>${tycho-version}</version>
+							<executions>
+								<execution>
+									<id>default-build-qualifier</id>
```
mickaelistria, line 108: How is qualifier set then?

m2e-maven-runtime/pom.xml
```diff
@@ -47,146 +50,212 @@
...
+									<phase>none</phase>
+								</execution>
+								<execution>
+									<id>default-validate-id</id>
```
mickaelistria, line 112: Please document this is disabled to MANIFEST.MF doesn't exist that early.

m2e-maven-runtime/org.eclipse.m2e.maven.runtime/metadataTemplates/.classpath
```diff
@@ -0,0 +1,12 @@
...
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11">
+		<attributes>
+			<attribute name="module" value="true"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry kind="src" path="src/main/resources"/>
+	<classpathentry kind="output" path="target/classes"/>
+	${dependency.jars.lib.entries}
```
mickaelistria, line 11: I don't get it; aren't those deps supposed to be already made visible with the PDE container?

mickaelistria (commented):
That's really good progress! I have a few questions inline.

m2e-maven-runtime/pom.xml
HannesWell: The qualifier is set in the second build, when the runtime modules are build as ordinary Eclipse plug-ins that embed jars.
The profile `generate-osgi-metadata` is only activated for the first build.
The first Maven build, that is also executed in the IDE, is a 'Maven-mode' build and only downloads all dependency jars into the lib folder of each runtime plugin, generates the `MANIFEST.MF` and the `.classpath` files. So a build-qualifier is not necessary here.

I have tried to describe the relationship between the two build steps in `m2e-maven-runtime/README.md`.

m2e-maven-runtime/pom.xml
HannesWell: Yes, I will make the comment at the beginning of the plug-ins section more precise.

m2e-maven-runtime/pom.xml
mickaelistria: OK, thanks. Would you mind making adding a comment in the pom  to explains it?

mickaelistria:
Do I get it right that the issue is that "Export-Package" coming from a jar listed in "Bundle-ClassPath" are not visible to clients unless the jars are listed directly in .classpath ?

HannesWell:
> 
> 
> Do I get it right that the issue is that "Export-Package" coming from a jar listed in "Bundle-ClassPath" are not visible to clients unless the jars are listed directly in .classpath ?

The embedded have to be listed as in the `.classpath` as lib-classpath entry to be visible to JDT/PDE. For example the entry for `maven-core` looks like this:
```
<classpathentry exported="true" kind="lib" path="jars/maven-core-3.8.1.jar" sourcepath="jars/sources/maven-core-3.8.1-sources.jar"/>
```
There is one such entry for each embedded jar, because unfortunately there is no classpath-container to include a folder of jars at once. Because it would be very cumbersome and error-prone to create those entries manually, they are generated during the first _metadata-generating_-Maven-build.

But I reworked the creation of the .classpath files and found a way to simplify it so that the metadataTemplates/.classpath files are not necessary any more (it is even less and simpler code).

HannesWell:
> 
> 
> Furthermore I just noticed that in the m2e.maven.indexer jar the other (original) class-files in the `org.apache.maven.index` package are missing and the corresponding entries in the `META-INF/sisu/javax.inject.Named` file are missing. I have to find out, if this is a problem or if they can be found in the originals nevertheless.

I have verified that all class files that are not in the `org.apache.maven.index` package from the root of the `org.eclipse.m2e.maven.indexer` jar any more are in the `indexer-core-6.0.0.jar` which is embedded in the first one. So the classes should be accessible from the `org.eclipse.m2e.maven.indexer` bundle respectively its ClassLoader.

Regarding the reduced number of entries in `META-INF/sisu/javax.inject.Named`:
The `META-INF/sisu/javax.inject.Named` is read by `org.eclipse.sisu.space.IndexedClassFinder.indexedNames()` which creates a set of all lines (which denote class-names) in all `META-INF/sisu/javax.inject.Named` files it can find using `ClassLoader.getResources()`. Consequently the `META-INF/sisu/javax.inject.Named` in the embedded jars are found as well.

I have verified that the entries that are now missing in the `META-INF/sisu/javax.inject.Named` from the root of `org.eclipse.m2e.maven.indexer` jar are contained in the `META-INF/sisu/javax.inject.Named` in the `indexer-core-6.0.0.jar`, which is embedded in the first one. Furthermore I checked that the set of class-names returned by `org.eclipse.sisu.space.IndexedClassFinder.indexedNames()` has an equal content in all calls (I launched the pre-defined Eclipse-IDE from my m2e-workspace and started a little Maven build).

To make a long story short: I'm pretty sure it isn't a problem that the `META-INF/sisu/javax.inject.Named` in the root of `org.eclipse.m2e.maven.indexer` does not contain all the name of all classes relevant for sisu as long as another file on the bundle's classpath contains another file with the other names. The same applies for the classes.

HannesWell:
I will again verify that the content of the build jars has not changed in an unwanted manner, but besides that I think this PR is complete and ready to be merged.

However I think we should wait until the Eclipse 2021-12 M1 is available from Oomph (which I assume to happen in the next day).
It contains the fix of [PDE Bug 575419](https://bugs.eclipse.org/bugs/show_bug.cgi?id=575419) that Mickael kindly merged. Otherwise the m2e dev's get three errors like this:
`Package 'META-INF.plexus' does not exist in this plug-in`. 
Those errors are not harmful but I think we can wait those few more days to not confuse other developers.

HannesWell:
The build fails, but I think the problem is that the Jenkinsfile from the master is used, but I had to modify it to adopt it to the changed build-steps:
```
Loading trusted files from base branch master at 2b689c6295ed5e24218f5a73ba309ff3eb227b5f rather than 399de906c12061bc091221d1ec66296eadb7bbdb
Obtained Jenkinsfile from 2b689c6295ed5e24218f5a73ba309ff3eb227b5f
‘Jenkinsfile’ has been modified in an untrusted revision
```
Is there a way to tell Jenkins to trust my modifications and to take the Jenkinsfile from my branch?

laeubi:
I think you need to push your branch to the m2e repo instead of your fork in such a case. Another option would be to use the [replay](https://ci.eclipse.org/m2e/job/m2e/job/PR-276/21/replay/) (you need to be logged in to jenkins for this) for your build and copy the jenkins file content into the text-box.

HannesWell:
> 
> 
> I think you need to push your branch to the m2e repo instead of your fork in such a case. Another option would be to use the [replay](https://ci.eclipse.org/m2e/job/m2e/job/PR-276/21/replay/) (you need to be logged in to jenkins for this) for your build and copy the jenkins file content into the text-box.

The replay function is a great hint. Thanks for that! I didn't know it yet, but I have the impressions that it is very handy in such situations.

I noticed you have already triggered another build with an adjusted Jenkinsfile: https://ci.eclipse.org/m2e/job/m2e/job/PR-276/22/
This build looks good (just the usual test-failure).

m2e-maven-runtime/pom.xml
HannesWell: I added a reference to the co-located README.md in the beginning of the profile.

laeubi:
>  I didn't know it yet, but I have the impressions that it is very handy in such situations

Very powerful, just keep in mind with great power there must also come great responsibility ;-)

> I noticed you have already triggered another build with an adjusted Jenkinsfile

Yep, feel free to trigger another build if required, one must just notice that if you use replay it does normally do not fetch new git-refs so sadly when you change the code it requires one failed build before you can replay with changed code...

HannesWell:
> 
> 
> > I didn't know it yet, but I have the impressions that it is very handy in such situations
> 
> Very powerful, just keep in mind with great power there must also come great responsibility ;-)

That's right. Of course I will use it with care and caution. :-)

> 
> > I noticed you have already triggered another build with an adjusted Jenkinsfile
> 
> Yep, feel free to trigger another build if required, one must just notice that if you use replay it does normally do not fetch new git-refs so sadly when you change the code it requires one failed build before you can replay with changed code...

That's indeed not ideal, but the situations where you need it are probably rare.

HannesWell:
I have just removed the sources plug-ins from the m2e-sdk feature because I assume they will not work (at least I would be very surprised if they would, because the sources jars of the embedded jars are in a sub-folder and I found no reference to them).
IMO the inclusion of the runtime-module sources is a good topic for a follow-up PR.

m2e-maven-runtime/org.eclipse.m2e.maven.indexer/pom.xml
```diff
@@ -19,13 +19,14 @@
...
-	<version>1.18.0-SNAPSHOT</version><!--Keep it in sync with version in target-platform.target -->
-	<packaging>bundle</packaging>
+	<version>1.18.1-SNAPSHOT</version>
+	<packaging>eclipse-plugin</packaging>
 
 	<name>M2E Maven/Nexus Indexer</name>
 
 	<properties>
 		<guava.version>30.1-jre</guava.version>
+		<extraClasspathEntries><![CDATA[<classpathentry kind="src" path="src/main/java"/>]]></extraClasspathEntries>
```
mickaelistria, line 29: This one would ideally be inferred from the build.properties `source..` entry. Maybe add a comment here to kepe it in mind for future work.

m2e-maven-runtime/pom.xml
```diff
@@ -115,7 +68,7 @@
 							<_failok>true</_failok>
 							<_nouses>true</_nouses>
 							<_nodefaultversion>true</_nodefaultversion>
-							<_snapshot>${commitDate}</_snapshot>
+							<_snapshot>qualifier</_snapshot>
```
mickaelistria, line 71: so how does this happen at build-time?
The _snapshot is 1st set to qualifier in the 1st `mvn` command and then in the 2nd, the `qualifier` is replaced with the last git commit timestamp using the usual tycho version computation?

mickaelistria (approved):
That's really good!
I've put a couple of questions/comments inline.
If you're confident about this enough, feel free to merge; and then announce on the m2e-dev that the way to setup a dev environment got considerably improved and point to the documentation.

m2e-maven-runtime/org.eclipse.m2e.maven.indexer/pom.xml
HannesWell: Yes, actually you are right. And you gave me the idea, that the `source..` element for the projects `build.properties` could be read using a `LoadProperties` Ant-Task and a subsequent additional for-loop could create src-classpathentry for each element. Then that special treatment is not required now and in case of future changes.
I'll try that.

Furthermore I think I should add another remark in the new README.md to state that the generated files should not be modified directly but only through the pom.xml.

m2e-maven-runtime/pom.xml
HannesWell: Exactly. The 1st build creates a MANIFEST.MF that contains a header like `Bundle-Version: 1.18.1.qualifier`, which is then processed as usual by Tycho.

HannesWell:
> 
> 
> That's really good! I've put a couple of questions/comments inline. If you're confident about this enough, feel free to merge; and then announce on the m2e-dev that the way to setup a dev environment got considerably improved and point to the documentation.

Thank you, I'm looking forward to merge this.
I've checked the results multiple times and expect this change to work to the best of my knowledge.
I'm planning to merge this as soon as the Oomph product catalogue contains 2021-12 M1 (because of the PDE-fix), which expect to happen on Friday from the [2021-12 Release schedule](https://wiki.eclipse.org/Category:SimRel-2021-12). I will gladly announce this on the dev-mailing list, also to inform the other devs, that the latest (not yet released) Eclipse version is recommended.

m2e-maven-runtime/org.eclipse.m2e.maven.indexer/pom.xml
HannesWell: > 
> Yes, actually you are right. And you gave me the idea, that the `source..` element for the projects `build.properties` could be read using a `LoadProperties` Ant-Task and a subsequent additional for-loop could create src-classpathentry for each element. Then that special treatment is not required now and in case of future changes. I'll try that.

It worked.

HannesWell:
Actually it would also be possible to commit the generated `MANIFEST.MF`, `.classpath` and the content of the `jars` folders into git. Then the first build step would not be necessary and one only has to run the generate-OSGi-metadata build locally in case of changes to the runtime-modules. The manual and generated changes could then be committed together.

But I'm not in favour of this option because it makes the repo larger (the jars combined are roughly calculated about 45MB) and has the chance that the pom.xml and the generated files go out-of-sync. And the Oomph setup takes care of generating the metadata.

mickaelistria:
Build shows a version inconsistency
```
[...]
[INFO] --- tycho-packaging-plugin:2.5.0:build-qualifier (default-build-qualifier) @ org.eclipse.m2e.maven.runtime ---
[INFO] The project's OSGi version is 1.18.1.20211002-1936
[...]
[INFO] --- tycho-p2-repository-plugin:2.5.0:assemble-repository (default-assemble-repository) @ org.eclipse.m2e.site ---
[INFO] {osgi.os=linux, osgi.ws=gtk, org.eclipse.update.install.features=true, osgi.arch=x86_64}
[ERROR] Cannot resolve project dependencies:
[ERROR]   Software being installed: org.eclipse.m2e.site 1.16.0.qualifier
[ERROR]   Missing requirement: org.eclipse.m2e.feature.feature.group 1.18.2.20211005-0719 requires 'org.eclipse.equinox.p2.iu; org.eclipse.m2e.maven.runtime [1.18.1.20211004-1243,1.18.1.20211004-1243]' but it could not be found
[ERROR]   Cannot satisfy dependency: org.eclipse.m2e.site 1.16.0.qualifier depends on: org.eclipse.equinox.p2.iu; org.eclipse.m2e.feature.feature.group 0.0.0
```
So it's building 1.18.1.20211002-1936 and then expecting later 1.18.1.20211004-1243.

HannesWell:
> 
> However I think we should wait until the Eclipse 2021-12 M1 is available from Oomph (which I assume to happen in the next day). It contains the fix of [PDE Bug 575419](https://bugs.eclipse.org/bugs/show_bug.cgi?id=575419) that Mickael kindly merged. Otherwise the m2e dev's get three errors like this: `Package 'META-INF.plexus' does not exist in this plug-in`. Those errors are not harmful but I think we can wait those few more days to not confuse other developers.

The required PDE version is now available in Oomph's Eclipse 'latest' stream.
So I'm going to squash-and-merge this PR as soon as the currently running build is successful.

PR #282, opened 2021-07-19
[Maven-Runtime] unify findbugs-jsr305 exclusion and cleanup dependencies

HannesWell:
The findbugs-jsr305 is now excluded globally in the `m2e-maven-runtime/pom.xml`.

Additionally unnecessary `<optional>` elements are removed. The m2e artifacts are not further used in other Maven builds, so there is no need to mark dependencies optional. However this first had the effect that the imported slf4j-API packages of m2e.maven.runtime were not marked with `;resolution:=optional` any more. While I think this would not have be a problem because slf4j.api bundle is usually present, I added the resolution directive to the BND-instructions. 

Furthermore the not required provided `org.eclipse.sisu.plexus` dependencies are removed.

I compared the content of the four jars build for the m2e-maven-runtime bundles and their content is effectively equal to the content before.


mickaelistria:
Thanks a lot!

PR #284, opened 2021-07-20
[WIP] Test indexer 6.1.0-SNAPSHOT

mickaelistria:


akurtakov:
@mickaelistria should this one be closed now that indexer is not used anymore?

HannesWell:
Isn't the indexer still used by the old editor, is it?
But in case we remove the old editor one day we could probably also remove the entire indexer plugin?

akurtakov:
Even if used by old editor, it's not advancing to testing with snapshot version of maven-indexer is beyond what anyone planned to do AFAICT.

HannesWell:
> Even if used by old editor, it's not advancing to testing with snapshot version of maven-indexer is beyond what anyone planned to do AFAICT.

That's right. I thought you were referring to the indexer in general. Sorry, my mistake.

akurtakov:
Close.

mickaelistria:
I think it's fine to close it now that the overall goal is just to stop relying on indexer at all. We won't support newer versions or any further work for indexer.
We should even deprecate all code that relies on it already.

PR #292, opened 2021-07-26
Build with Tycho 2.4.0

HannesWell:
Build using the latest Tycho version 2.4.0.

@laeubi or @mickaelistria do you know why the execution of the `org.eclipse.tycho:target-platform-configuration:target-platform` goal is now not covered by the lifecycle configuration any more.
Has some lifecycle-configuration been removed in Tycho?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -37,7 +37,7 @@
 			This should usually be sync'd with the main feature version. -->
 		<m2e.version>1.18.2</m2e.version>
 
-		<tycho-version>2.3.0</tycho-version>
+		<tycho-version>2.4.0</tycho-version>
 
 		<tycho.testArgLine>-Xmx800m</tycho.testArgLine>
 		<tycho.surefire.timeout>7200</tycho.surefire.timeout>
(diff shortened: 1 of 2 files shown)


laeubi:
@HannesWell what specifically you are referring to? A lifecycle mapping is generally bound to a specific packagin-type so do you have anything in mind (as the build here seem to succeed beside the usual test-failure)?

HannesWell:
Sorry for being not clear: I'm referring to `Plugin execution not covered by lifecycle configuration: org.eclipse.tycho:target-platform-configuration:2.4.0:target-platform (execution: default-target-platform, phase: initialize)` errors I get from m2e in the IDE in the m2e-workspace as soon I change the `tycho-version` property from 2.3.0 to 2.4.0.
Therefore I marked the goal as ignored for m2e. I already observed this behaviour in other projects that use Maven/Tycho.

Of course the actual build is not affected, but I think it will disturb developers.

mickaelistria:
The uncovered mojo warning is more the topic of https://github.com/eclipse/tycho/issues/146 and can be ignored for the moment.

pom.xml
```diff
@@ -153,6 +153,16 @@
 						</filter>
 					</filters>
 				</configuration>
+				<executions>
```
mickaelistria, line 156: I think we should just not include such hacks in the pom, especially not at part of this commit, to give exposure to https://github.com/eclipse/tycho/issues/146

mickaelistria (changes requested):

pom.xml
HannesWell: I didn't know there is already an issue for this, so I wanted to prevent corresponding errors in the m2e poms.
But if you are fine with them, I'm also.
Thanks for the remark.

PR #300, opened 2021-08-18
[Maven-Runtime] control embedded jars completely by dependencies (#223)

HannesWell:
This PR is one more step to prepare the resolution of issue #223 and is extracted from PR #276.

It aims to simplify the inclusion of the jars nested into m2e's maven runtime modules. To achive this only only those dependencies that should be emebedded into the module-jars are specified and not wanted (transitive) dependencies are excluded, either per dependency or 'globablly' by setting their scope to provided. Then all dependencies of a m2e maven-runtime-modules can simply be nested into the modules jar.

The only relevant change in the build jars (I performed a diff on the extracted jars) is that for the archetype.common bundle the element order in the `Bundle-ClassPath` has changed, but it contains still the same elements.
The same applies for the `Embedded-Artifacts` header, but as far as I now this is not a header consider by OSGi (I think this is a bnd specific one) and is therefore not relevant.


HannesWell:
There are additional test failures:

```
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 8.113 s <<< FAILURE! - in org.eclipse.m2e.core.ui.tests.AllTests
testConsoleHasOutput(org.eclipse.m2e.core.ui.tests.ConsoleTest)  Time elapsed: 8.058 s  <<< FAILURE!
java.lang.AssertionError: missing output in console
	at org.eclipse.m2e.core.ui.tests.ConsoleTest.testConsoleHasOutput(ConsoleTest.java:61)
```

I have to investigate.
They also fail on my Windows computer with this PR but also with the state before, when I run it as JUnit plug-in test.
IIRC they worked in a local Maven build with all integrated tests. But also other contributes that use Windows reported that this test fails. So it is interesting.

mickaelistria:
`.ConsoleTest.testConsoleHasOutput()` was added recently, when we add issue about slf4j integration in m2e. As I see you modified some slf4j related stuff, I think it's where you could start investigating.

HannesWell:
Since the same test failed in the builds of PR #319 and #320 I suspect that the failure could be unrelated. Do you have an idea why the test failed in those PRs?

From the changes in the build jars I actually don't expect any changes with respect to slf4j, but I will investigate it.

mickaelistria:
That's right; failure was not related. Merged.

HannesWell:
Great. Thanks, for reviewing.

Issue #252, opened 2021-06-28
Loop between "Building" and "Downloading Sources and Javadoc"


mickaelistria:
While I'm coding for Tycho, on snapshot branch, I see a lot of CPU used by Eclipse IDE and the Progress view seems to loop between "Building" and "Downloading Sources and Javadoc" pretty often.

mickaelistria:
It might just be a symptom of #123

laeubi:
Yep that seems to be a reincarnation of that bug. As mentioned there, it some time magically 'happens' but goes away without a clear indication why.

mickaelistria:
That typically happens when having some Tycho modules (incl tycho-core and most of content of tycho-bundles) open, but not all, and opening a new module.

reckart:
Happens for me a lot as well after updating to m2e 1.18.1 - different workspaces, some with some without bundles.

reckart:
Downgrading to m2e 0.17.2 seems to fix the issue.

mickaelistria:
I merged a patch that may improve things yesterday. Maybe it's also worth trying the latest snapshot of 1.18.2.

reckart:
Do you have the update site URL handy?

mickaelistria:
See https://github.com/eclipse-m2e/m2e-core/blob/master/CONTRIBUTING.md#-trying-latest-builds

reckart:
I have installed `m2e - Maven Integration for Eclipse (includes Incubating components) 1.18.2.20210721-064` - loops are still there. The funny thing is that "Download Artifact Sources" and "Download Artifact JavaDoc" are even both turned off in the preferences. "automatically update configuration" is also off. Still it loops on "Download sources and javadoc" <-> "Building" until I manually break the loop by killing the "Download sources and javadoc" task.

mickaelistria:
@reckart thanks for trying. Do you have a minimal project that can reproduce this issue in a predictable manner?

reckart:
Unfortunately not.

a1dutch:
`m2e - Maven Integration for Eclipse (includes Incubating components)	1.18.2.20210721-0646` fixed the endless building/downloading sources for my maven projects

ciros88:
I confirm build 1.18.2.20210721-0646 fixed that annoying problem.

HannesWell:
[...]

PR #303, opened 2021-08-21
Prevent endless download of artifact sources/javadoc (#252)

HannesWell:
The changes of this PR prevent endless download loops of sources/javadoc which can still occur under special circumstances as described in https://github.com/eclipse-m2e/m2e-core/issues/252#issuecomment-903173237.

The approach is to maintain a set of handled download requests, that guards against adding an equivalent request that is/was already handled in the current run of the DownloadSources-job.

Alternatively the queue could be cleaned in the end of each run, but this could suppress legitimate follow up download requests, or can the latter never happen?

### DIFF
--- a/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/DownloadSourcesJob.java
+++ b/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/DownloadSourcesJob.java
@@ -14,7 +14,6 @@
 package org.eclipse.m2e.jdt.internal;
 
 import java.io.File;
-import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -58,6 +57,7 @@ import org.eclipse.m2e.jdt.MavenJdtPlugin;
  *
  * @author igor
  */
+@SuppressWarnings("restriction")
 class DownloadSourcesJob extends Job implements IBackgroundProcessingQueue {
   private static Logger log = LoggerFactory.getLogger(DownloadSourcesJob.class);
 
@@ -137,6 +137,8 @@ class DownloadSourcesJob extends Job implements IBackgroundProcessingQueue {
 
   private final BlockingQueue<DownloadRequest> queue = new LinkedBlockingQueue<>();
 
+  private Set<DownloadRequest> requests = new HashSet<>();
+
   private final Set<IProject> toUpdateMavenProjects = new HashSet<>();
 
   private final Map<IPackageFragmentRoot, Attachments> toUpdateAttachments = new HashMap<>();
@@ -163,6 +165,7 @@ class DownloadSourcesJob extends Job implements IBackgroundProcessingQueue {
         if(!status.isOK()) {
           // or maybe just log and ignore?
           queue.clear();
+          requests.clear();
           toUpdateAttachments.clear();
           toUpdateMavenProjects.clear();
           return status;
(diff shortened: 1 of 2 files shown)


org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/DownloadSourcesJob.java
```diff
@@ -339,17 +335,14 @@ private File download(ArtifactKey artifact, List<ArtifactRepository> repositorie
...
-  }
-
-  public void addDownloadRequest(IProject project, IPackageFragmentRoot fragment, ArtifactKey artifact,
-      boolean downloadSources, boolean downloadJavadoc) {
     if(project == null || !project.isAccessible()) {
       return;
     }
-    queue.add(new DownloadRequest(project, fragment, artifact, downloadSources, downloadJavadoc));
+    DownloadRequest request = new DownloadRequest(project, fragment, artifact, downloadSources, downloadJavadoc);
+    if(requests.add(request)) { // guard against new requests that are/will be already downloaded in this run to prevent endless loops
```
mickaelistria, line 342: Could we start by something like `!queue.contains(request)` ? Wouldn't that already deliver a noticeable improvement without too much code complexity?

HannesWell: Unfortunately that would not help in the case described in https://github.com/eclipse-m2e/m2e-core/issues/252#issuecomment-903173237, because the new request is added when `DownloadSourcesJob.updateClasspath()` is called in `DownloadSourcesJob.run()`. But at that point the queue is already empty, so it would be added again.

That's the reason why the requests are cleared after `updateClasspath()` is called.

Alternatively we could either
- prevent `BuildPathManager.updateClasspath()` from creating new requests (but this would probably require more flags and also adds complexity).

or 
- simply clear the queue after `updateClasspath()` returned in `DownloadSourcesJob.run()`.
But, if any legitimate cases exists (do you know any?) where `updateClasspath()` creates follow-up downloads that downloads would be suppressed.

mickaelistria:
I admit I don't fully get the solution here; but since this has proven itself to be a fragile piece of code, it would be great if this bug could be covered by a test to prevent from future regression. Would that be doable?

HannesWell:
> 
> 
> I admit I don't fully get the solution here; but since this has proven itself to be a fragile piece of code, it would be great if this bug could be covered by a test to prevent from future regression. Would that be doable?

Fully agree. I'll try to set up some test-cases.

HannesWell:
> 
> 
> I admit I don't fully get the solution here; but since this has proven itself to be a fragile piece of code, it would be great if this bug could be covered by a test to prevent from future regression. Would that be doable?

I have created tests for this case and also other re-download cases and submitted them as a PR: https://github.com/tesla/m2e-core-tests/pull/130

Without this PR the test BuildPathManagerTest.testDownloadSources_009_redownloadSnapshotWithOlderSources fails.
If you debug the test and put a break-point into `DownloadSourcesJob` where the request is added to the queue:
`queue.add(request);`

.. just before in BuildPathManagerTest.testDownloadSources_009_redownloadSnapshotWithOlderSources the download of sources/javadoc is scheduled the second time:
```
    // download sources a second time time
    getBuildPathManager().scheduleDownload(project, true, false);
```
You will noticed that the break-point is hit indefinitely if the job is not cancelled (what the test case does after 10 round IIRC). When you look into the stack-trace you will find what I have described here: https://github.com/eclipse-m2e/m2e-core/issues/252#issuecomment-903173237

HannesWell:
This PR should not be merged up until #318 and https://github.com/tesla/m2e-core-tests/pull/130 are merged.

mickaelistria:
Great, thanks! Please rebase and merge when you're ready.

mickaelistria:
Rebased and merged as 7a56eef592cb3c389e348a4e3358eef2f5b7d05b ; thanks!

HannesWell:
Thanks for taking care of it.

PR #305, opened 2021-08-22
Add 'Update Project' to the Project menu

gzsombor:
I always bit confused, when I try to find the "Update Maven Project" item, and it's not in the 'Project' menu, and a bit annoyed, when I have to find the project or pom.xml in the huge package explorer tree. Maybe, I'm not alone with this.

### DIFF
--- a/org.eclipse.m2e.core.ui/plugin.xml
+++ b/org.eclipse.m2e.core.ui/plugin.xml
@@ -380,6 +380,18 @@
             properties="workspaceResulutionEnable,hasArtifactKey,hasProjectArtifactKey,isBuildDirectory"
             type="org.eclipse.core.runtime.IAdaptable"/>
       <propertyTester
+            id="org.eclipse.m2e.core.MavenPropertyTester"
+            class="org.eclipse.m2e.core.ui.internal.actions.MavenPropertyTester"
+            namespace="org.eclipse.m2e"
+            properties="hasMavenNature"
+            type="org.eclipse.core.resources.IResource"/>
+      <propertyTester
+            id="org.eclipse.m2e.core.MavenPropertyTester"
+            class="org.eclipse.m2e.core.ui.internal.actions.MavenPropertyTester"
+            namespace="org.eclipse.m2e"
+            properties="hasMavenNature"
+            type="org.eclipse.ui.IFileEditorInput"/>
+      <propertyTester
             id="org.eclipse.m2e.core.MavenPropertyTester2"
             class="org.eclipse.m2e.core.ui.internal.actions.MavenPropertyTester"
             namespace="org.eclipse.m2e"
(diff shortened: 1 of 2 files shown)


org.eclipse.m2e.core.ui/plugin.xml
```diff
@@ -674,4 +686,21 @@
...
+         point="org.eclipse.ui.menus">
+      <menuContribution
+            allPopups="false"
+            locationURI="menu:project">
+            <command
+                  commandId="org.eclipse.m2e.core.ui.command.updateProject"
+                  style="push">
+               <visibleWhen
+                     checkEnabled="false">
+                   <with variable="activeEditorInput">
```
mickaelistria, line 699: Actions in the project menu usually apply to the current selection, not particularly to the editor.

mickaelistria (changes requested):

org.eclipse.m2e.core.ui/plugin.xml
gzsombor: I've tried 
```
                       <with variable="selection">
                          <iterate>
                              <adapt
                                    type="org.eclipse.core.resources.IResource">
                                   <test property="org.eclipse.m2e.hasMavenNature" value="true"/>
                              </adapt>
                          </iterate>
                       </with>
```
but in that case, the menu item disappeared very easily. I mean, even if a java class is selected in the package explorer, I just had to edit that file, it goes away. Maybe, because that 'selection' turns into a text-selection, or something like that. I couldn't figure out the exact details, unfortunately.
What about having the 2 conditions, just or-ed together? It works better in my experience

org.eclipse.m2e.core.ui/plugin.xml
```diff
@@ -674,4 +686,31 @@
...
+      <menuContribution
+            allPopups="false"
+            locationURI="menu:project">
+            <command
+                  commandId="org.eclipse.m2e.core.ui.command.updateProject"
+                  style="push">
+               <visibleWhen
+                     checkEnabled="false">
+                   <or>
+                       <with variable="selection">
```
mickaelistria, line 700: This does not seems to work with the Project Explorer at least. Selecting a project in Project Explorer without editor open doesn't show the menu.

org.eclipse.m2e.core.ui/plugin.xml
```diff
@@ -674,4 +686,31 @@
...
       <remote  url="https://repo1.maven.org/maven2/archetype-catalog.xml" description="Maven Central"/>
    </extension>
+   <extension
+         point="org.eclipse.ui.menus">
+      <menuContribution
+            allPopups="false"
+            locationURI="menu:project">
+            <command
+                  commandId="org.eclipse.m2e.core.ui.command.updateProject"
+                  style="push">
```
mickaelistria, line 696: Another label should be used because the command label is "Update Project" (it's usually enough as we're in a Maven submenu). Here, without Maven context, it should be "Update Maven Project".
Or the other way round: the command should be relabeld to "Update Maven Project" (so the label also shows in Preference keys and Ctrl+3), and the label overriden only under "Maven" submenu.

mickaelistria (changes requested):

Issue #308, opened 2021-08-31
NPE during BuildPathManager refresh when working with remote projects


ivy-rew:
Our workspace contain remote projects, where the IResource.getLocation() will not return a local absolute path, but just null. This eclipse featureset provokes NPEs when Maven Dependency changes are propagated throughout the workspace. 

It looks as if we can easily workaround it, by making the BuildPathManager aware of this 'null' paths in remote projects. And rely upon resource URIs, which are always present in local and remote/cached projects. I'll setup a PR for this.

```java.lang.NullPointerException
	at org.eclipse.m2e.jdt.internal.BuildPathManager.resourceChanged(BuildPathManager.java:614)
	at org.eclipse.core.internal.events.NotificationManager$1.run(NotificationManager.java:305)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.core.internal.events.NotificationManager.notify(NotificationManager.java:295)
	at org.eclipse.core.internal.events.NotificationManager.handleEvent(NotificationManager.java:271)
	at org.eclipse.core.internal.resources.Workspace.broadcastEvent(Workspace.java:375)
	at org.eclipse.core.internal.resources.Resource.broadcastPreDeleteEvent(Resource.java:1851)
	at org.eclipse.core.internal.resources.Resource.delete(Resource.java:747)
	at org.eclipse.core.internal.resources.Project.delete(Project.java:320)
	at org.eclipse.ltk.core.refactoring.resource.DeleteResourceChange.perform(DeleteResourceChange.java:155)
	at ch.ivyteam.ivy.designer.refactor.restricted.change.DeleteProjectChange.perform(DeleteProjectChange.java:45)
	at org.eclipse.ltk.core.refactoring.CompositeChange.perform(CompositeChange.java:281)
	at org.eclipse.ltk.core.refactoring.CompositeChange.perform(CompositeChange.java:281)
	at org.eclipse.ltk.core.refactoring.CompositeChange.perform(CompositeChange.java:281)
	at org.eclipse.ltk.core.refactoring.PerformChangeOperation.lambda$0(PerformChangeOperation.java:258)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2292)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2317)
	at org.eclipse.ltk.core.refactoring.PerformChangeOperation.executeChange(PerformChangeOperation.java:295)
[...]

PR #309, opened 2021-08-31
#308 fix NPE when distributing classpath changes to remote projects

ivy-rew:
Work with 'locationURI' rather than with 'IResource.getLocation()', as the URI is defined in remote projects too, while the location may be null.

Change-Id: Idd6d26c794a1c6199694f6a66dafdb1d8f97899c

### DIFF
--- a/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/BuildPathManager.java
+++ b/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/BuildPathManager.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.MalformedURLException;
+import java.net.URI;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -139,7 +140,7 @@ public class BuildPathManager implements IMavenProjectChangedListener, IResource
 
   final File stateLocationDir;
 
-  final Map<String, InternalModuleInfo> moduleInfosMap = new ConcurrentHashMap<>();
+  final Map<URI, InternalModuleInfo> moduleInfosMap = new ConcurrentHashMap<>();
 
   private final DownloadSourcesJob downloadSourcesJob;
 
@@ -616,7 +617,7 @@ public class BuildPathManager implements IMavenProjectChangedListener, IResource
         log.error("Can't delete " + containerState.getAbsolutePath()); //$NON-NLS-1$
       }
 
-      moduleInfosMap.remove(project.getLocation().toString());
+      moduleInfosMap.remove(project.getLocationURI());
 
     } else if(IResourceChangeEvent.POST_CHANGE == type) {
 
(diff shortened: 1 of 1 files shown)


ivy-rew:
Doesn't look stable yet; I will investigate

HannesWell:
> 
> 
> Doesn't look stable yet; I will investigate

Don't worry. That is the usual failure that happens also on the master branch. And a fix is prepared but we depend on another project to perform a release (see for example PR #284)

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/BuildPathManager.java
```diff
@@ -616,7 +616,7 @@ public void resourceChanged(IResourceChangeEvent event) {
         log.error("Can't delete " + containerState.getAbsolutePath()); //$NON-NLS-1$
       }
 
-      moduleInfosMap.remove(project.getLocation().toString());
+      moduleInfosMap.remove(project.getLocationURI().toString());
```
mickaelistria, line 619: Couldn't we by the way increase typing and make this map use URIs as key?

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/BuildPathManager.java
```diff
@@ -666,7 +666,7 @@ private boolean requiresUpdate(IProject p, IProgressMonitor monitor) {
             if(moduleDescription == null) {
               return false;
             }
-            String location = p.getLocation().toString();
+            String location = p.getLocationURI().toString();
```
mickaelistria, line 669: Couldn't we store the URI directly?

ivy-rew: Yes, I agree that using URIs in the map would be more naturally. I'll update the PR towards that end.

laeubi: One remark here: URI defines a rather complex equals/hash code contract (what might be irrelevant here) just something to take in mind that two URIs are considered equal even if there String representation is NOT equal!

laeubi:
If we change here, I think `ConcurrentHashMap` is used wrong here... instead of manually fetch/update the map the `ConcurrentMap.compute(K, BiFunction<? super K, ? super V, ? extends V>)` should be used to update values in the map.

mickaelistria:
While I agree with improving how we use the map, it's something that would be better covered in another patch. The point about changing the type is related to that change by the fact that we change the type of keys we store and changing type makes it more expressive; optimizing the map is not really in the scope.
Note that since m2e is a pretty old code base; there are probably many occurrences of such maps where computeIfAbsent(...) should be used.

HannesWell:
> Note that since m2e is a pretty old code base; there are probably many occurrences of such maps where computeIfAbsent(...) should be used.

There are definitely many occurrences. I've the plan to clean up m2e to leverage java-11 collection methods (there are more 'new' collection methods) and more 'new' java features like try-with-resources. I already have a loose collection of incomplete clean-up commits on my computer but didn't have the time yet to finish them. But I'll try to do it in the near future. Nevertheless if somebody else is faster in providing PRs I'll be happy to review them.

ivy-rew:
Being a rare m2e contributor, I'd prefer to keep the map as is and just address the really required change to keep the infoMap compatible to remote projects.
So if it's good enough, I'd love to see it being merged as proposed here.

mickaelistria:
That's definitely good enough, thanks!

ivy-rew:
Thanks a lot @mickaelistria

Issue #310, opened 2021-08-31
NPE when creating problem markers on a remote project


ivy-rew:
Having a remote project with an unsolvable  'system' scoped dependency in my workspace, a problem marker is being created for it. However, that operation fails with an NPE as IResource.getLocation() is being used, which can return null for remote projects.

I'll propose a fix for it.

```
java.lang.NullPointerException
	at org.eclipse.m2e.core.internal.markers.SourceLocationHelper.findLocation(SourceLocationHelper.java:89)
	at org.eclipse.m2e.core.internal.markers.MavenMarkerManager.toMavenProblemInfos(MavenMarkerManager.java:211)
	at org.eclipse.m2e.core.internal.markers.MavenMarkerManager.addMarkers(MavenMarkerManager.java:65)
	at org.eclipse.m2e.core.internal.project.registry.ProjectRegistryManager.lambda$4(ProjectRegistryManager.java:750)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.executeBare(MavenExecutionContext.java:179)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.execute(MavenExecutionContext.java:153)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.execute(MavenExecutionContext.java:101)
	at org.eclipse.m2e.core.internal.project.registry.ProjectRegistryManager.execute(ProjectRegistryManager.java:1027)
	at org.eclipse.m2e.core.internal.project.registry.ProjectRegistryManager.readMavenProjectFacades(ProjectRegistryManager.java:735)
	at org.eclipse.m2e.core.internal.project.registry.ProjectRegistryManager.refresh(ProjectRegistryManager.java:431)
	at org.eclipse.m2e.core.internal.project.registry.ProjectRegistryManager.refresh(ProjectRegistryManager.java:364)
	at org.eclipse.m2e.core.internal.project.registry.ProjectRegistryManager.refresh(ProjectRegistryManager.java:315)
	at org.eclipse.m2e.core.internal.project.ProjectConfigurationManager.updateProjectConfiguration0(ProjectConfigurationManager.java:409)
	at org.eclipse.m2e.core.internal.project.ProjectConfigurationManager.lambda$1(ProjectConfigurationManager.java:358)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.executeBare(MavenExecutionContext.java:179)
[...]

PR #311, opened 2021-08-31
#310 fix: do not fail with an NPE on problems within remote projects

ivy-rew:
- handle IResource.getLocation(), which can be null for remote projects.

Change-Id: Ibd5db4e39814f681c93a77936b4fde4e84bf5fa5


mickaelistria:
Thanks!

PR #314, opened 2021-09-01
Use try-with-resource and implement Closeable in MutableProjectRegistry

HannesWell:
This PR aims to reduce the number of close operations in finally-blocks of try-constructs.
For this I use try-with-resources where applicable.
While in some cases this is only a syntactical difference, in some places this also has the effect that Exceptions that are thrown while closing the resource are not ignored any more. But I think this is OK, because first I think Exceptions while closing is rare and second I think that those Exceptions should be handled equivalently to for example cases where an exception is thrown while opening the resource.

At some locations I also simplified the code within the try block using Java's Files class.
Furthermore I let `MutableProjectRegistry` implement the AutoCloseable interface, because it already had a `close()` method that was only called in the finally-block of try statements. So I think this is suitable.

I started the Commit a while ago but didn't finish it yet. Alexanders recent PR(s) that also leverage try-with-resources remembered me to complete it, in order to avoid duplicated work. The currently pending PR of him also contains some try-with-resources applications that likely overlap with this PR, but the result should be the same.

### DIFF
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/actions/SelectionUtil.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/actions/SelectionUtil.java
@@ -265,7 +265,7 @@ public class SelectionUtil {
         File tempPomFile = null;
         try (InputStream is = storage.getContents()) {
           tempPomFile = File.createTempFile("maven-pom", ".pom"); //$NON-NLS-1$ //$NON-NLS-2$
-          Files.copy(storage.getContents(), tempPomFile.toPath());
+          Files.copy(is, tempPomFile.toPath());
           return readMavenProject(tempPomFile, monitor);
         } catch(IOException ex) {
           log.error("Can't close stream", ex); //$NON-NLS-1$
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/actions/SelectionUtil.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/actions/SelectionUtil.java
@@ -14,10 +14,9 @@
 package org.eclipse.m2e.core.ui.internal.actions;
 
 import java.io.File;
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.OutputStream;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -48,8 +47,6 @@ import org.eclipse.ui.IWorkbenchPage;
 import org.eclipse.ui.IWorkbenchWindow;
 import org.eclipse.ui.IWorkingSet;
 import org.eclipse.ui.PlatformUI;
-
-import org.codehaus.plexus.util.IOUtil;
 
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.execution.MavenExecutionRequest;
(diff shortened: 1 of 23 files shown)


akurtakov:
There seem to be some overlap with my yesterday fixes. Would you please fix the conflicts?

HannesWell:
I have rebased the commit, resolved the conflicts and forced pushed it.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryReader.java
```diff
@@ -65,53 +63,54 @@ public ProjectRegistryReader(File stateLocationDir) {
   public ProjectRegistry readWorkspaceState(final ProjectRegistryManager managerImpl) {
     if(stateFile.exists()) {
       final PackageAdmin packageAdmin = getPackageAdmin();
-      ObjectInputStream is = null;
-      try {
-        is = new ObjectInputStream(new BufferedInputStream(new FileInputStream(stateFile))) {
```
HannesWell, line 70: The creation of the ObjectInputStream was simply extracted into an own class to improve readability.
The same was done for the ObjectOutputStream below.
I think an anonymous class within a try-with-resources would not be very readable.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/actions/SelectionUtil.java
```diff
@@ -266,19 +263,13 @@ public static MavenProject getMavenProject(IEditorInput editorInput, IProgressMo
...
         File tempPomFile = null;
-        InputStream is = null;
-        OutputStream os = null;
-        try {
+        try (InputStream is = storage.getContents()) {
           tempPomFile = File.createTempFile("maven-pom", ".pom"); //$NON-NLS-1$ //$NON-NLS-2$
-          os = new FileOutputStream(tempPomFile);
-          is = storage.getContents();
-          IOUtil.copy(is, os);
+          Files.copy(storage.getContents(), tempPomFile.toPath());
```
mickaelistria, line 268: should this be
```suggestion
          Files.copy(is, tempPomFile.toPath());
```

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MutableProjectRegistry.java
```diff
@@ -35,7 +35,7 @@
  *
  * @author igor
  */
-public class MutableProjectRegistry extends BasicProjectRegistry implements IProjectRegistry {
+public class MutableProjectRegistry extends BasicProjectRegistry implements IProjectRegistry, AutoCloseable {
```
mickaelistria, line 38: Good idea.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryReader.java
mickaelistria: :+1:

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/actions/SelectionUtil.java
HannesWell: Yes you are right. Thanks for pointing out!

PR #319, opened 2021-09-04
Use try-with-resource and java.nio.file.Files

HannesWell:
Two more classes where try-with-resource can be used that somehow slipped through my previous PR.

Furthermore `java.nio.file.Files` is is used instead of `org.eclipse.jetty.util.IO`.

Additionally the Input- and OutputStreams of the request are now closed.


mickaelistria:
It looks like 1 extra test is failing with this change. Is this related?

PR #320, opened 2021-09-04
Use Comparator.comparing() and Comparator.thenComparing()

HannesWell:
Use Comparator.comparing() and Comparator.thenComparing() and the specializations for primitives, to simplify the construction of more complicated comparators.


mickaelistria:
Looks like the build failure wasn't related; let's merge

HannesWell:
Do you have an idea why the ConsoleTests suddenly were failing even when not related to the change?

Issue #324, opened 2021-09-06
Test failure with Maven 3.8.2


HannesWell:
Since Maven 3.8.2 is embedded (PR #323) `MavenImplTest.testReadRemoteParent` fails on the master branch:
```
testReadRemoteParent(org.eclipse.m2e.tests.embedder.MavenImplTest)  Time elapsed: 0.44 s  <<< FAILURE!
java.lang.AssertionError: remote-parent.artifact.file
	at org.eclipse.m2e.tests.embedder.MavenImplTest.testReadRemoteParent(MavenImplTest.java:430)
```
It looks like `parent.getArtifact().getFile()` returns null now.

Unless anybody else already knows how to address this, I will investigate it.

mickaelistria:
That's strange it didn't fail on the PR, did it?

HannesWell:
I have to admit that I don't know it. The Jenkins executors were very busy at the weekend and my PR's were waiting for free executors for a long time. When I looked for the result this morning the jobs were either still pending or were aborted after about 8h (I'm not sure if it was that PR or another one from the weekend).

But I can reproduce the error locally so I can investigate it.

mickaelistria:
Ok, I may have missed the failure before merging.
Maybe it's better if revert now and reopen a PR?

HannesWell:
> Ok, I may have missed the failure before merging.
> Maybe it's better if revert now and reopen a PR?

Yes that's probably better, I also thought about that already. I submit a PR.
Maven 3.8.2 seems to have more changes than I would expect from a minor version update.

mickaelistria:
Thanks. Can you please open a "revert revert" PR to check whether test fails there as well?

mickaelistria:
Let's close this ticket and track further discussion and work directly in PR #326

PR #323, opened 2021-09-05
[Runtime] Embed Maven 3.8.2

HannesWell:
Maven 3.8.2 was released:
https://maven.apache.org/docs/3.8.2/release-notes.html

I think I have to create a CQ as described in issue #159 for Maven 3.8.1?

PR #326, opened 2021-09-07
[Runtime] Embed Maven 3.8.3

HannesWell:
Second attempt to embed Maven 3.8.2 after PR #323.

As reported in issue #324 there were test-failure on the master branch, that are also expected now and have to be solved first.

### DIFF
--- a/m2e-maven-runtime/org.eclipse.m2e.maven.runtime/pom.xml
+++ b/m2e-maven-runtime/org.eclipse.m2e.maven.runtime/pom.xml
@@ -26,7 +26,7 @@
 
 	<properties>
 		<!-- maven core version -->
-		<maven-core.version>3.8.2</maven-core.version>
+		<maven-core.version>3.8.3</maven-core.version>
 		<!-- NOTE: When maven-core.version changes, this may impact the versions of the maven-resolver-*
 			jars that export the org.eclipse.aether.* packages in the org.eclipse.m2e.maven.runtime
 			bundle. So make sure the following variable has the value of the maven-resolver-* jars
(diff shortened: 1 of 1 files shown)


mickaelistria:
OK. so we see the expected test failure. Good! No let's just find a cause and a solution for it...

HannesWell:
Yep. I can investigate it.
But IMHO this should not be merged before https://github.com/eclipse/tycho/issues/261 is fixed, because otherwise you cannot build m2e with itself from the IDE and I suspect that it is the same for other Eclipse projects that use the P2Metadata goal.

mickaelistria:
A vote has started on the mailing-list and a Maven 3.8.3 release is going to happen soon.

HannesWell:
> 
> 
> A vote has started on the mailing-list and a Maven 3.8.3 release is going to happen soon.

Thanks for the hint! Do you mean this mail?:
https://lists.apache.org/thread.html/rec607477d7053e59e02825bafe2e13539bc8bd4484eee024cd9b6967%40%3Cdev.maven.apache.org%3E

I embedded the staged Maven 3.8.3 locally and the tests seem to pass now (except for the usual failure). Let's see if this also works on Jenkins.

HannesWell:
Build has just the usual build failure, so it looks like we are lucky and the problem (whatever it was) is solved with the new Maven version and we only have to wait for the official release.

mickaelistria:
Maven 3.8.3 was released today; you may update and un-draft this PR.

mickaelistria:
replaced with #343

HannesWell:
> 
> 
> replaced with #343

Thanks Mickael for taking care of it.

Issue #331, opened 2021-09-21
Download Jobs seems to run in infinite loop.


cdietrich:
On my workspace the DownloadJob runs all the time. 
it seems to schedule itself via

```
DownloadSourcesJob$DownloadRequest.<init>(IProject, IPackageFragmentRoot, ArtifactKey, boolean, boolean) line: 79	
DownloadSourcesJob.addDownloadRequest(IProject, IPackageFragmentRoot, ArtifactKey, boolean, boolean) line: 352	
DownloadSourcesJob.scheduleDownload(IProject, IPackageFragmentRoot, ArtifactKey, boolean, boolean) line: 342	
DownloadSourcesJob.scheduleDownload(IProject, ArtifactKey, boolean, boolean) line: 362	
BuildPathManager.scheduleDownload(IProject, MavenProject, ArtifactKey, boolean, boolean) line: 919	
BuildPathManager.configureAttachedSourcesAndJavadoc(IMavenProjectFacade, Properties, ClasspathDescriptor, IProgressMonitor) line: 343	
BuildPathManager.getClasspath(IMavenProjectFacade, int, Properties, boolean, IProgressMonitor) line: 270	
BuildPathManager.getClasspath(IProject, int, boolean, IProgressMonitor) line: 413	
BuildPathManager.getClasspath(IProject, int, IProgressMonitor) line: 393	
BuildPathManager.getClasspath(IProject, IProgressMonitor) line: 421	
BuildPathManager.updateClasspath(IProject, IProgressMonitor) line: 206	
DownloadSourcesJob.updateClasspath(BuildPathManager, Set<IProject>, Map<IPackageFragmentRoot,Attachments>, IProgressMonitor) line: 207	
DownloadSourcesJob.run(IProgressMonitor) line: 185	
```
the problematic place may be

```
  private void scheduleDownload(IProject project, MavenProject mavenProject, ArtifactKey artifact,
      boolean downloadSources, boolean downloadJavadoc) throws CoreException {
    ArtifactKey[] attached = getAttachedSourcesAndJavadoc(artifact, mavenProject.getRemoteArtifactRepositories(),
        downloadSources, downloadJavadoc);

    if(attached[0] != null || attached[1] != null) {
      downloadSourcesJob.scheduleDownload(project, artifact, downloadSources, downloadJavadoc);
    }
  }
```
the null checks dont also check the downloadSources and downloadJavadoc flags
[...]

PR #333, opened 2021-09-21
[DownloadSourcesJob] fix final clear of requests Set (#331)

HannesWell:
My last clean-up of the DownloadSourcesJob went wrong (maybe falsely resolved merge conflicts) and damaged the logic how the requests Set is cleared of `DownloadSourcesJob.run()`.

The `requests` set was fully cleared instead of retaining all elements in it that are also in the `queue` at that moment.

This is only harmful in case a new (valid) download-request was added during the classpath-update (or externally, but I'm not sure if this is possible). 
In this case the `requests` set does not contain that request in the queue (because it was cleared instead of only retaining all elements in the queue). So when the classpath is updated after that request is processed the `requests` set likely still does not contain the request, the request can be added again to the queue and a new run will be scheduled. After the next run the requests set would contain the request but it is again cleared after the classpath update is fully completed and the circle starts again.

The described scenario only occurred when for a SNAPSHOT version the sources.jar is last modified before the main jar in the local Maven repo and download sources is called via the right-click context menu of the Maven-Classpath container as described in https://github.com/eclipse-m2e/m2e-core/issues/331#issuecomment-924310353 and previous comments.

This PR also fixes issue #331.


HannesWell:
Build looks good. Only the usual failure.

PR #334, opened 2021-09-22
Build with Tycho 2.5.0

HannesWell:
This PR checks for regressions in the m2e build using the latest Tycho 2.5.0-SNAPSHOT.

It can be updated as soon as Tycho-2.5.0 is released or discarded.

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -3,6 +3,6 @@
   <extension>
     <groupId>org.eclipse.tycho.extras</groupId>
     <artifactId>tycho-pomless</artifactId>
-    <version>2.4.0</version>
+    <version>2.5.0</version>
   </extension>
 </extensions>
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -3,6 +3,6 @@
   <extension>
     <groupId>org.eclipse.tycho.extras</groupId>
     <artifactId>tycho-pomless</artifactId>
-    <version>2.4.0</version>
+    <version>2.5.0-SNAPSHOT</version>
   </extension>
 </extensions>
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -3,6 +3,6 @@
   <extension>
     <groupId>org.eclipse.tycho.extras</groupId>
     <artifactId>tycho-pomless</artifactId>
-    <version>2.5.0-SNAPSHOT</version>
+    <version>2.4.0</version>
   </extension>
 </extensions>
--- a/pom.xml
+++ b/pom.xml
@@ -37,7 +37,7 @@
 			This should usually be sync'd with the main feature version. -->
 		<m2e.version>1.18.2</m2e.version>
 
-		<tycho-version>2.4.0</tycho-version>
+		<tycho-version>2.5.0-SNAPSHOT</tycho-version>
 
 		<tycho.testArgLine>-Xmx800m</tycho.testArgLine>
 		<tycho.surefire.timeout>7200</tycho.surefire.timeout>
--- a/pom.xml
+++ b/pom.xml
@@ -37,7 +37,7 @@
 			This should usually be sync'd with the main feature version. -->
 		<m2e.version>1.18.2</m2e.version>
 
-		<tycho-version>2.5.0-SNAPSHOT</tycho-version>
+		<tycho-version>2.5.0</tycho-version>
 
 		<tycho.testArgLine>-Xmx800m</tycho.testArgLine>
 		<tycho.surefire.timeout>7200</tycho.surefire.timeout>


HannesWell:
The build fails while resolving the Target-Platform because of `org.apache.felix.resolver.reason.ReasonException: Uses constraint violations`.
I encountered the same errors locally on my computer.

laeubi:
Even if that won't make a difference here, but if you update the tycho version don't forget the pomless in .mvn/extensions.xml

HannesWell:
> 
> 
> Even if that won't make a difference here, but if you update the tycho version don't forget the pomless in .mvn/extensions.xml

Thanks for the hint, but I skipped the version update for `.mvn/extensions.xml` intentionally because the extension do not work with a SNAPSHOT version. I assume that in that early stage of the build the additional plug-in repositories are not considered.

laeubi:
>  because the extension do not work with a SNAPSHOT version

You need to define the snapshot repository in a settings.xml

HannesWell:
> > because the extension do not work with a SNAPSHOT version
> 
> You need to define the snapshot repository in a settings.xml

Good to know. But the EF-Jenkins servers seems not to have that set in their settings.xml:
https://ci.eclipse.org/m2e/job/m2e/job/PR-334/2/console

Anyway, usually this isn't a problem.

laeubi:
If you like maven support specify settings.xml on commandline as well, but we should really not give this too much effort as this is really not part of the problem here.

HannesWell:
> 
> 
> If you like maven support specify settings.xml on commandline as well, but we should really not give this too much effort as this is really not part of the problem here.

No no, I don't want to put effort on that topic. I'm totally fine with it as it is. I just wanted to indicate that I skipped the `extension.xml` intentionally. :)

laeubi:
We can try another run once the new snapshot build is available I have merged a fix for this right now see https://github.com/eclipse/tycho/pull/304

HannesWell:
Thank you Christoph for for fixing the problem in tycho.
But the build failed I with the same error. I assume the resolver parameters have to be tweaked? Can you propose a set of suspected working parameters?

laeubi:
Adding import package as described [here](https://github.com/eclipse/tycho/issues/303#issuecomment-925523026) to `org.eclipse.m2e.core.ui` should fix the issue. can you try to add this?

mickaelistria:
> Adding import package as described here to org.eclipse.m2e.core.ui should fix the issue. can you try to add this?

We shouldn't add import-package on bundle that do not effectively require the package. Such hacks shouldn't be merged.

laeubi:
the bundle actually depends on the `org.w3c.*` packages but do not import them directly but relies on its required-bundles to provide the dependency implicitly... and that's what causing this issue here I think even though PDE/equinox do not complain about this extended boot delegation it was always wrong to rely on it.

mickaelistria:
OK, it makes sense. Are the version range necessary for this to work though? It'd be nice if we can get rid of them so it would allow to use the packages from the Java API directly instead of having to install the extra org.w3c.dom.events bundle.

laeubi:
Version ranges are a crucial part when consuming an API in OSGi but for sure we can try if it suffice to not specify any range (=use largest version available).

laeubi:
I tried to reproduce this with a small test-project but didn't succeed yet so it seems not to be the combination of imports alone. Strange enough the m2e build succeeded for this bundle for me locally but I'll recheck...

laeubi:
I think the 2.5.0 SNAPSHOT is not updated when build from a branch... will check this

mickaelistria:
> I think the 2.5.0 SNAPSHOT is not updated when build from a branch... will check this

Right, automatic promotion is only happening from master ( https://github.com/eclipse/tycho/blob/master/Jenkinsfile#L27 ). We can either tweak the Jenkinsfile to allow deploying for any non-PR branch (but I don't know how to do that), or an alternative strategy is to run the `mvn deploy` step locally.

laeubi:
Okay that's the issue here :-\
https://ci.eclipse.org/tycho/job/tycho-github/job/tycho-2.5.x/ do not deploy snapshots
I have adjusted the jenkins file and added a not to https://github.com/eclipse/tycho/commit/3fe73b055fb2acc3a75a55de3b3c34a211500ed3

laeubi:
Just he usual test failure...

laeubi:
If I run with tycho 2.5.x build succeeds, if I specify `-Dtycho.equinox.resolver.batch.size=1` the build fails with the error. and also a more simpler test case fails in such a scenario.

HannesWell:
Sorry I lost track of the conversation in the Tycho-fixes.

> Adding import package as described [here](https://github.com/eclipse/tycho/issues/303#issuecomment-925523026) to `org.eclipse.m2e.core.ui` should fix the issue. can you try to add this?

Since the build succeeds now I assume this is not necessary any more? Should I do anything else?

laeubi:
> Since the build succeeds now I assume this is not necessary any more?

Nope no additional actions are required.

HannesWell:
> 
> 
> > Since the build succeeds now I assume this is not necessary any more?
> 
> Nope no additional actions are required.

OK, great. I will update this PR once Tycho 2.5.0 is released.

mickaelistria:
Tycho 2.5 was released, you can now update the PR and merge as soon as this works as expected.

PR #327, opened 2021-09-09
Clean up and simplify logback projects

HannesWell:
An attempt to clean up and simplify the `logback.appender` and `logback.configuration` plug-ins.

In order to direct the attention to the relevant parts of the logic utility code was extracted into own methods, but not re-used utility classes mere merged.

Furthermore added comments (with references to bug-reports) to locations that seem to be unnecessarily cumbersome at the first look, but turned out to be relevant in the end.
And some other minor improvements (e.g. avoid `File.toURL()`).

PR #339, opened 2021-09-29
Add missing and unify/simplify plug-in/feature Bundle-Vendor/Name data

HannesWell:
This PR improves and simplifies the plug-in/feature metadata in the following ways:
- adds Bundle-Vendor Manifest-headers where they were missing
- Unify the Bundle-Vendor of all plug-ins/features to be `Eclipse.org - m2e`
- Inline the values of `plugin.properties/bundle.properties` files into the Manifest.MF if they only contain `Bundle-Vendor` and `Bundle-Name`.
I think the chance that those values are internationalized is very low (and it would not work because not all plug-ins use plugin/bundle.properties) and the values are only used in the MANIFEST.MF. So IMHO this indirection only adds additional files and blows up the plug-ins without any benefits.
- Inline the value from the `org.eclipse.m2e.editor.xml/plugin.properties`. This properties file was not part of the build.properties `bin.includes`. Furthermore, because this plug-in already has a bundle.properties file, I fear that the keys are not properly replaced at runtime by their values. To avoid any further problems I simply inlined those values as well.

Besides those changes already done, the Bundle-Name headers could be unified as well, for example to all start with `M2E - ...` using a similar schema for all plug-ins. @mickaelistria what do you think, any objections?


mickaelistria:
Changing bundle name/vendor without other payload change will cause millions of users to re-download several dozens of MB of plugins for not added value when updating. That's not very geen-it  So I'm not in favor of pushing that network load without a compelling reason to do so.
But for bundles we already modified in more substantial way ans whise version was already bumped, itt seems like a decent change.

HannesWell:
> Changing bundle name/vendor without other payload change will cause millions of users to re-download several dozens of MB of plugins for not added value when updating. That's not very geen-it So I'm not in favor of pushing that network load without a compelling reason to do so. But for bundles we already modified in more substantial way ans whise version was already bumped, itt seems like a decent change.

Understand that, but the situation is not that bad. Summed up the five plug-ins that would require a version bump only from this have in total a size of about 100kB (without sources).
However I have reverted the changes for those plug-ins that didn't had a version bump in this release-cycle yet.
If PR #327 is merged before this one, the metadata updates can be applied for it as well.

Besides that I have unified the Bundle-Name headers of all plug-ins and features. They are now all start with `M2E` (which is similar most other Eclipse plug-ins, e.g. `EMF`) and the words in it are usually capitalized. Where it seemed appropriate I also adjusted the names themself.

Additionally that I adjusted the licence Header in the feature.properties of `org.eclipse.m2e.lemminx.feature ` and `org.eclipse.m2e.sse.ui.feature` because they said that Sonatype created the features in 2008 (`2008, 2010 Sonatype, Inc.`), but actually you created them in 2020 and as far as I know you are working for Red Hat. I hope this is correct.

mickaelistria:
See build failure. Apparently, some of the modifications here would still require a version bump.

HannesWell:
> 
> 
> See build failure. Apparently, some of the modifications here would still require a version bump.

I think the failure was because even tough the the `m2e.lifecyclemapping.defaults` plug-in effectivly did not change, because of the change that was subsequently reverted the latest git-timestamp had changed. 
I squashed and force pushed the commits to avoid a changed git-timestamp.

Besides that I changed the maven-runtime modules to use the Maven-project name as Bundle-Name and updated the project-name element accordingly. This ensure that both names are the same and avoids duplications.

HannesWell:
The build now completes but there is the following additional build failure in `org.eclipse.m2e.core.ui.tests`:
```

Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 5.751 s <<< FAILURE! - in org.eclipse.m2e.core.ui.tests.AllTests
testConsoleHasOutput(org.eclipse.m2e.core.ui.tests.ConsoleTest)  Time elapsed: 5.724 s  <<< FAILURE!
java.lang.AssertionError: missing output in console
	at org.eclipse.m2e.core.ui.tests.ConsoleTest.testConsoleHasOutput(ConsoleTest.java:61)


Results:

Failures: 
  ConsoleTest.testConsoleHasOutput:61 missing output in console
```

But I think this failure is unrelated.

mickaelistria:
> But I think this failure is unrelated.

This is not a failure we can ignore. It usually highlights an issue (version or classloading) with logback stuff.

mickaelistria:
A new build happen and the console test is green, so we're good here from test POV.

HannesWell:
> A new build happen and the console test is green, so we're good here from test POV.

Yes I triggered it, because I suspected that it does not fail consistently and I don't expect this change to influence the logback-configuration because no versions are changed.

The `ConsoleTest` seem to be a bit unstable respectively seem to have a race-condition. I have also see them failing locally without a reason (at least I couldn't explain it).

Do you have objections against the new names?

Once this PR is submitted I can create another (draf) PR with the pending metadata unifications of those plug-ins that did not yet have a version bump. As soon as there are other modifications in those plug-ins those changes can be applied (step by step) too.

mickaelistria:
No objection.

PR #340, opened 2021-10-02
[Maven-Runtime] clean-up and mark exported META-INF packages as internal

HannesWell:
This PR applies the following clean-up changes, mainly from PR #276 to reduce the number of changes there:
- mark exported META-INF packages as internal
- Remove unused build.properties files (the maven-runtime modules are pure Maven projects atm., therefore a build.properties file is not considered at all)
- update maven-bundle-plugin version
- add Automatic-Module-Name to generated MANIFEST.MF
- remove "Bundle" suffix from project-names (and thus also from the
Bundle-Names)
- add MavenDescriptor during build.
As soon as #276 is merged, the Maven pom.xml and pom.properties are embedded within `META-INF/maven` regardless of the archive-configuration of the maven-bundle-plugin because that plug-in does not build the archives then).

### DIFF
--- a/m2e-maven-runtime/org.eclipse.m2e.archetype.common/pom.xml
+++ b/m2e-maven-runtime/org.eclipse.m2e.archetype.common/pom.xml
@@ -22,7 +22,7 @@
 	<version>1.18.0-SNAPSHOT</version><!--Keep it in sync with version in target-platform.target -->
 	<packaging>bundle</packaging>
 
-	<name>M2E Maven Archetype Common Bundle</name>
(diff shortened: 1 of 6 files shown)


HannesWell:
Build looks good, just the usual failure.

Furthermore I extracted the jars of the runtime-modules I build locally and compared their content with the current master.
Except the expected changes (like the adjusted `Bundle-Name`, the x-internal directives and the files in `META-INF/maven`), the Export and Import-Package headers have a changed the order in their values, due to the update of the `maven-bundle-plugin` (I have double checked that it is just the order). So this change can be merged.

PR #343, opened 2021-10-04
Embed Maven 3.8.3

mickaelistria:
Also-By: Hannes Wellmann <wellmann.hannes1@gmx.net>


mickaelistria:
Reverted in #350 because it caused bugs #346 and #347. Will be reported to Maven.

Issue #347, opened 2021-10-06
NoClassDefFoundError when using some of projects dependencies


mcoufal:
I get NoClassDefFoundError when trying to use e.g. org.apache.log4j or org.junit.jupiter that is linked as maven dependency. My impression is that this will happen for plenty of maven dependencies.

How to reproduce:

1. Run Eclipse
2. Create a Maven project with this option enabled: "Create a simple project (skip archetype selection)"
3. Insert the following dependency into generated "pom.xml":
```
<dependencies>
	<dependency>
		<groupId>log4j</groupId>
		<artifactId>log4j</artifactId>
		<version>1.2.17</version>
	</dependency>
</dependencies>
```
4.  Right-click the project > Maven > Update Project, OK
5. Add new Java class file, paste following source code and save:
```
package testMvn;

import java.io.StringWriter;

import org.apache.log4j.Appender;
import org.apache.log4j.PatternLayout;
import org.apache.log4j.WriterAppender;

public class TestClass {

	public static void main(String[] args) {
		StringWriter errorOutputWriter = new StringWriter();
		Appender appender = new WriterAppender(new PatternLayout(PatternLayout.TTCC_CONVERSION_PATTERN),
				errorOutputWriter);

		org.apache.log4j.Logger.getRootLogger().addAppender(appender);

	}

}
```
6. Right-click the edited Java file > Run As > Java Application

The console view now contains an error:
`Error: Unable to initialize main class testMvn.TestClass
Caused by: java.lang.NoClassDefFoundError: org/apache/log4j/Layout`

My configuration:
Eclipse Platform I20211004-2030 (4.22)
Java version: 11.0.12
org.eclipse.m2e 1.18.2.20211005-0719

mickaelistria:
Is this a regression?

RalfNaujokat:
Hi.
I am using the latest snapshop,too, because of Java 17, 
and since the latest check-for-updates (and installing) in eclipse 
i get this NoClassDefFoundErrors when running tests (as mentioned in issue  #346)
for the second time.
This does also occur on 'Run as Java Application' on classes with a main method,
that have dependencies to other projects.
First run is Ok, second run fails with NoClassDefFoundError.

Looks like issue #346 is kind of associated to that.

mickaelistria:
[...]

PR #350, opened 2021-10-06
Issue #347 - Test reproducing issue

mickaelistria:


m2e-maven-runtime/org.eclipse.m2e.maven.runtime/pom.xml
```diff
@@ -26,7 +26,7 @@
 
 	<properties>
 		<!-- maven core version -->
-		<maven-core.version>3.8.3</maven-core.version>
+		<maven-core.version>3.8.1</maven-core.version>
```
HannesWell, line 29: Below `<maven-resolver.version>1.6.3</maven-resolver.version>` should also be reverted to 1.6.2 in case we really revert in the end. But I don't know if this affects this test.

Issue #348, opened 2021-10-06
M2E - PDE Integration v1.18.2: java.package; aQute.bnd.classfile [1.610.0,2.0.0)


minemeraj:
I am trying to install the M2E - PDE Integration plugin v1.18.2, but there following error occurs:

Your original request has been modified.
  "M2E - PDE Integration" is already installed, so an update will be performed instead.
Cannot complete the install because one or more required items could not be found.
  Software being installed: M2E - PDE Integration 1.18.2.20211004-1835 (org.eclipse.m2e.pde.feature.feature.group 1.18.2.20211004-1835)
  Missing requirement: biz.aQute.bndlib 6.0.0.202110041835 (biz.aQute.bndlib 6.0.0.202110041835) requires 'java.package; aQute.bnd.classfile [1.610.0,2.0.0)' but it could not be found
  Cannot satisfy dependency:
    From: M2E - PDE Integration 1.18.2.20211004-1835 (org.eclipse.m2e.pde.feature.feature.group 1.18.2.20211004-1835)
    To: org.eclipse.equinox.p2.iu; biz.aQute.bndlib [6.0.0.202110041835,6.0.0.202110041835]

Where I can find java.package; aQute.bnd.classfile [1.610.0,2.0.0)?

PR #356, opened 2021-10-08
[TargetPlatform] merge locations and specify versions

HannesWell:
At the moment m2e's target-file is composed of many different locations from different Eclipse project p2-repos.
But most of the IUs can be found in the `https://download.eclipse.org/releases/...` repo. Using that repo would reduce the number of locations and this simplify the target file.

With the reduced number of locations it is simpler to use stable repos (not like a latest repo that changes with each release) which gives us reproducible builds.

I have to admit that I never found and exact description of the difference between the `https://download.eclipse.org/releases/...` repos and the repos at `https://download.eclipse.org/eclipse/updates/...`, but my understanding is that the first contains all plug-ins and features used to build all of the pre-packaged Eclipse products while the latter only contains the plug-ins/features of the Eclipse platform/SDK? Is this correct?

With that understanding all IUs from the releases repo could be used for m2e's TP (because m2e only consumes them), except for `biz.aQute.bndlib`. `biz.aQute.bndlib` is included into the m2e-features and is contributed to the releases repo by m2e (at least I don't know any other Eclipse project using it, except Tycho), so if m2e would use that IU we would have boot-strapped us and could not update to a new version. This does not apply for the other IUs in the TP.

Only `org.eclipse.ui.tests.harness`, `org.eclipse.license.feature.group` and `org.eclipse.wildwebdeveloper.xml.feature.feature.group` cannot be consumed from the releases repo. The first two because they are not available and the latter because the releases repo only contains it in version 0.12.

For `biz.aQute.bndlib` I used a maven-target because we get a stable version with it and it is good if m2e uses its own features.

### DIFF
--- a/target-platform/target-platform.target
+++ b/target-platform/target-platform.target
@@ -4,26 +4,36 @@
 	<targetJRE path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11"/>
 	<locations>
 		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<repository location="https://download.eclipse.org/releases/2021-09/202109151000/"/>
+			<repository location="https://download.eclipse.org/eclipse/updates/4.21/"/>
 			<unit id="org.eclipse.sdk.feature.group" version="0.0.0"/>
 			<unit id="org.eclipse.equinox.p2.discovery.feature.feature.group" version="0.0.0"/>
+			<unit id="org.eclipse.ui.tests.harness" version="0.0.0"/>
+		</location>
+		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
+			<repository location="https://download.eclipse.org/egit/updates-5.13/"/>
 			<unit id="org.eclipse.jgit" version="0.0.0"/>
+		</location>
+		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
+			<repository location="https://download.eclipse.org/modeling/emf/emf/builds/release/2.27/"/>
 			<unit id="org.eclipse.emf.edit.ui.feature.group" version="0.0.0"/>
 			<unit id="org.eclipse.emf.ecore.edit.feature.group" version="0.0.0"/>
+		</location>
+		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
+			<repository location="https://download.eclipse.org/webtools/downloads/drops/R3.22.0/R-3.22.0-20210612170523/repository/"/>
 			<unit id="org.eclipse.wst.common.uriresolver" version="0.0.0"/>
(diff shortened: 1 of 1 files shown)


HannesWell:
> > For `biz.aQute.bndlib` I used a maven-target
> 
> That was previously not possible as without m2e/tycho support for maven targets it was not possible to consume it from maven, so the feature is now somewhat feeding itself :-)

That's right. And I think that's good, therefore I did this intentionally. :)

Furthermore I just noticed that with the latest snapshots `biz.aQute.bndlib` version `6.0.0` was already used, while in the latest release it was `5.3.0`. Should we stick with `6.0.0`? Then I can update this PR.

HannesWell:
> Haven't checked the release notes but if the code works wit 6.x I don't see a reason not to use the latest version.

OK, I have quickly read through it (https://github.com/bndtools/bnd/wiki/Changes-in-6.0.0) and didn't find anything that could affect us (but I'm also not 100% familiar with it). So let's go for 6.0.0

mickaelistria:
No objection, but we 1st need to query whether those bundles are fine to ship, using the dash-license tool. I'm on it.

mickaelistria (changes requested):
Actually, I have an objection about merging the TP locations. So maybe we should consider taking care of the BND stuff in a separate patch to move forward on that aspect.
My concern with merging the TP is that sometimes/often, m2e updates to one version of the dependencies that's not in SimRel yet. That can be newer Wild Web Developer, newer Platform... and in the end, there is also a cycle introduced by SimRel and m2e when relying on it.
In practice, maintaining a TP with different sites isn't much more complicated than maintaining a TP with less sites; and it gives a clearer definition of what projects/versions we decide to consider as main m2e "target". Since SimRel and m2e move at different pace, it's not really safe to assume that their requirement generally match. I'm basically afraid that such change would cause m2e to be less inclined to use newer deps that aren't in SimRel yet, while in practice, things happen the other way round: projects included in SimRel do lead what's in SimRel; SimRel isn't here to mandate what versions of X or Y included projects should use.

HannesWell:
Created PR #365 to separate the bnd changes.

mickaelistria:
> But because m2e does not publish those requirements, SimRel has to pick them from somewhere else anyway?

Yes, so for instance using newer m2e 1.18.2 will require SimRel to also use latest Wild Web Developer release. That's fine as it.

> What is important for me is, that we only use unmodified/stable repos for the TP (if possible), so nothing like a latest repo to avoid problems like in #348 that from one build to another a new (major) version of a dependency is used and we suddenly have problems. I really appreciate using the latest versions of dependencies, but I think the updates should happen in a controlled and aware manner.

That makes sense. Feel free to update the TP to avoid using "latest".

HannesWell:
> 
> 
> > But because m2e does not publish those requirements, SimRel has to pick them from somewhere else anyway?
> 
> Yes, so for instance using newer m2e 1.18.2 will require SimRel to also use latest Wild Web Developer release. That's fine as it.
> 
> > What is important for me is, that we only use unmodified/stable repos for the TP (if possible), so nothing like a latest repo to avoid problems like in #348 that from one build to another a new (major) version of a dependency is used and we suddenly have problems. I really appreciate using the latest versions of dependencies, but I think the updates should happen in a controlled and aware manner.
> 
> That makes sense. Feel free to update the TP to avoid using "latest".

Understand that. I have reverted the merge and replaced all "latest" locations with specific versions. I'll merge it if the build is successful.

PR #360, opened 2021-10-09
[Build] make build platform independent

HannesWell:
This PR makes the build platform independent by specifying the `release` option for the compiler as well as the supported environments. The latter is actually not required because m2e does not have platform specific fragments, but it silences the warnings emitted at the beginning of each build.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,6 +38,7 @@
 		<m2e.version>1.18.2</m2e.version>
 
 		<tycho-version>2.5.0</tycho-version>
+		<java.version>11</java.version>
 
 		<tycho.testArgLine>-Xmx800m</tycho.testArgLine>
 		<tycho.surefire.timeout>7200</tycho.surefire.timeout>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -285,6 +303,7 @@
 					<artifactId>tycho-compiler-plugin</artifactId>
 					<version>${tycho-version}</version>
 					<configuration>
+						<release>${java.version}</release>
```
mickaelistria, line 306: I don't think it's a good idea. The "release" value is supposed to be dynamically deduced from the Bundle-RequiredExecutionEnvironment or osgi.ee requirement. Hardcoding makes that having different BREEs in the project is not well supported anymore (imagine 1 bundle wants JavaSE-1.8 and another JavaSE-17), and makes migration to different/newer JRE harder because this pom would need to be modified.
I'm going to try removing it, and the following setting as well; since the workaround mentioned is supposed to be available in Tycho 2.5.

PR #362, opened 2021-10-09
[Indexer] depend on guava, slf4j, javax instead of embedding them

HannesWell:
This PR removes the following embedded dependencies from the `org.eclipse.m2e.maven.indexer` bundle and adds dependencies the corresponding bundles/packages instead:

- com.google.guava-guava
Replaced by `Require-Bundle: com.google.guava`
- org.slf4j
Replaced by Import-Package: `org.slf4j` just like for `m2e.maven.runtime` (the other sub-packages imported in m2e.maven.runtime do not occur in the sources of the embedded jars).
- javax.inject
Replaced with a Require-Bundle `javax.inject` just like the m2e.maven.runtime (actually both could also just import the package `javax.inject` and I don't know if it is a good idea to re-export `javax.inject` from m2e.maven.runtime but that's another issue)
- javax.annotation
Looks like it is actually never used. At least I found no occurrence of `javax.annotation` in the sources. Therefore I just removed it.

This change safes about 3MB of jars (mainly driven by the removal of guava). Before this changes the expanded jars folder (without sources) had about ~8.6MB, now it has ~5.6MB. So we over 30% in size.

### DIFF
--- a/m2e-maven-runtime/org.eclipse.m2e.maven.indexer/pom.xml
+++ b/m2e-maven-runtime/org.eclipse.m2e.maven.indexer/pom.xml
@@ -26,7 +26,6 @@
 
 	<properties>
 		<maven-indexer.version>6.0.0</maven-indexer.version>
-		<guava.version>30.1-jre</guava.version>
 	</properties>
 
 	<dependencies>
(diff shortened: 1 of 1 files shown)


HannesWell:
Actually the same could be done for the `m2e.maven.runtime` bundle and save a similar amount of jar-size.
As far as I can oversee it, this would just require to generalize `MavenEmbeddedRuntime.initClasspath(Bundle)`, but I think this should be done any ways.

mickaelistria:
Yet another awesome change! Since tests are passing, I think it's OK to merge it now.

PR #365, opened 2021-10-11
[TP] add biz.aQute.bndlib 5.3.0 via Maven-Target with fix version (#348)

HannesWell:
Don't use a latest version to avoid unintended/unaware automatic version
upgrades.

This solves issue #348 .

### DIFF
--- a/target-platform/target-platform.target
+++ b/target-platform/target-platform.target
@@ -47,7 +47,7 @@
 				<dependency>
 					<groupId>biz.aQute.bnd</groupId>
 					<artifactId>biz.aQute.bndlib</artifactId>
-					<version>5.3.0</version>
+					<version>5.1.2</version>
 					<type>jar</type>
 				</dependency>
 				<dependency>


target-platform/target-platform.target
```diff
@@ -47,10 +43,20 @@
...
 		<location includeSource="true" missingManifest="error" type="Maven">
-			<groupId>io.takari.m2e.workspace</groupId>
-			<artifactId>org.eclipse.m2e.workspace.cli</artifactId>
-			<version>0.3.1</version>
-			<type>jar</type>
+			<dependencies>
+				<dependency>
+					<groupId>biz.aQute.bnd</groupId>
+					<artifactId>biz.aQute.bndlib</artifactId>
+					<version>5.3.0</version>
```
mickaelistria, line 50: We need it to be 5.1.2 which was the last release approved ( https://dev.eclipse.org/ipzilla/show_bug.cgi?id=22694 )

mickaelistria:
@HannesWell I can take care of changing the version if it's easier for you (I'm almost fully available for m2e these days)

PR #367, opened 2021-10-11
[MavenEmbeddedRuntime] Generalize/simplify classpath computation

HannesWell:
This PR generalizes the classpath computation in `MavenEmbeddedRuntime` to add all directly required bundles of the `m2e.maven.runtime` bundle, that are specified via `Import-Package` or `Require-Bundle`, to the classpath of a launched Maven-build JVM.
At the moment the bundles to add are hard-coded into MavenEmbeddedRuntime, but are equivalent to the required-bundles/imported-packages of `m2e.maven.runtime`. But if we change the dependencies of `m2e.maven.runtime` the code in `MavenEmbeddedRuntime` would have to be adjusted as well, which brings the risk of forgetting it and different behaviour. Furthermore the new code is just simpler.

This PR also performs some minor code improvements and removes (now) unused methods.


mickaelistria:
That's a very nice change! Do you think it's safe enough for upcoming 1.18.2 or would you rather have it delayed to a future release?

Issue #280, opened 2021-07-19
Create a feature out of a maven-target location


laeubi:
There should be an option to create an autogenerated feature out of a maven target location that simply includes all (transitive) dependencies of that location.

That would help people to rapidly setup a maven dependency in case of feature based products. For finer grained control, user features can still be created/used.

laeubi:
This is currently blocked by https://bugs.eclipse.org/bugs/show_bug.cgi?id=576568

HannesWell:
Yes, I meant to create a feature out of a pom-module. That will be a useful feature!
I think m2e itself could use it to include the embedded jars in `m2e.archtype.common` and `m2e.maven.indexer` as bundles because those two bundles actually only serve as 'group' of jars, which would actually be simply a feature.

laeubi:
My change is accepted and will hopefully be merged soon, I have created a local POC that converts a pom dependency to a feature

![grafik](https://user-images.githubusercontent.com/1331477/137112663-09caeb63-47fc-46c0-812b-bfdc4f42e51b.png)

sadly PDE is not capable of displaying such a feature, so further changes are required to make this work more smoothly...

HannesWell:
That looks great! Really looking forward to this one.
I'm curious, how does one specify that and how the feature is created? Do I just have to flip a switch for a Maven-Target location and say create a feature for all bundles in this location? For example by specifying the feature-name as attribute of the target location? And will the feature then include all bundles of that target? Or will there be a dedicated "create-feature"-element in a Maven target where you can specify other data (like licence or URLs etc.) of the feature as well?

If you prefer, we can also discuss those things in a (draft) PR once your POC is successful and all PDE requirements are met.

laeubi:
The feature is currently in an experimental stage and the first goal would be to have *anything* generated that could be used as a feature, but there is also some time to decide an all of this as we have to wait for the next PDE release in December anyways, so first GA will be the 2022-03 eclipse release.

I really would like to have [Bug 576610](https://bugs.eclipse.org/bugs/show_bug.cgi?id=576610) merged as it simplifies inspecting the feature generated but will create a draft PR with a first version then and if you like any Alpha-Testing would be appreciated.

My current plans are as follows:

[...]

PR #376, opened 2021-10-14
Create a feature out of a maven-target location #280

laeubi:
First POC for transforming a pom into a feature

This transforms pom artifacts into features.

FYI @HannesWell 

This currently requires an ibuild of PDE to work!


laeubi:
@HannesWell @mickaelistria I have upgraded the target to use the latest 2021-12 release but now it seem I have to bump all versions (at least the build fails with that), is this intentional?

mickaelistria:
> it seem I have to bump all versions (at least the build fails with that), is this intentional?

Try rebasing your change on top of master first.

PR #377, opened 2021-10-14
Add support for resolving POMs via their relative path

josple:
Update the ProfileManager to always attempt to resolve POMs via the relative path before falling back to resolving via repos.


mickaelistria:
You need to sign the Eclipse Contributor Agreement to contribute to Eclipse project: https://www.eclipse.org/legal/ECA.php
Do you think you could add a test case for it?
Also, versions for modified bundles need to be increased (by 0.0.1 if it's only a fix) if they weren't modified since last release.

mickaelistria:
> EDIT: Actually maybe it is the right repo?

Yes it is!

josple:
> @mickaelistria: I don't think all the jars/repositories are necessary for this test
Yeah, I doubt it too.

But...I have moved all of the profile tests over into the m2e-core repo (*not* just added my new test) and it seems one of the pre-existing tests requires at least some of the jars - I tried removing all the jars last night (passed in my IDE as it caches the repo in the target directory) but it is failing (as you can see) due to missing the org.apache.maven.plugins:maven-resources-plugin:jar:2.6 jar....which is then going to pull in a whole bunch of other jars no doubt.

 I will have a quick look to see if I can modify the test to not require this...but if it can't be changed I wont have time trying to isolate which set of jars are required for this test.

mickaelistria:
And I believe you can write a test that doesn't require any dependency actually; with a very minimal pom.

josple:
> @mickaelistria: That's too much at once, please just consider adding your new profile test to this repo; don't bother with moving older stuff

I appreciate it would be nicer to do the move of existing tests first and then in a separate PR add the single additional test and src code fix...but tbf the change is not particular big:
* 1 test class (would need to be created anyway just without the pre-existing tests in it)
* 1 test helper (which would be needed anyway)
* a number of resource files (POMs) to support the pre-existing tests (would not have needed to be moved if I was just adding my test)
 
So given I have done the work and it isn't particularly sizable or invasive I think we can be pragmatic here and let it in. This would mean re-opening https://github.com/tesla/m2e-core-tests/pull/132 which is the other side of the change (i.e. the removal of the pre-existing tests from the legacy repo).

> And I believe you can write a test that doesn't require any dependency actually; with a very minimal pom.

Yes, my test only ever used one pom file from the local repo file system - so no problems there 👍 

What I do need help with is the current failure - I am not familiar with your versioning approach - I have bumped the version in the module I have made the fix to (i.e. the src code change) - I am not clear what it is now asking for - I have speculatively bumped the version in the feature.xml file...but I am just guessing at this point!

josple:
@mickaelistria , cool makes sense. I have removed all the pre-existing tests and their resources and removed the profile test helper to avoid a copy and paste IP issue over that as well.

Hopefully there wont be too many other versions that need random bumping...

mickaelistria:
That seems perfect, I'm merging it and this patch will be available in snapshots shortly!

HannesWell:
Thanks for your contribution @josple.

I have noticed that some files were committed using Windows-style line-endings (CRLF respectively "\r\n") usually it is recommended to set up git to commit using Linux-style line-endings (LF respectively "\n") otherwise one can get strange behaviour on windows, like changed files in your staging area that have no changes when you compare them. I have already fixed that in a follow up commit.

Additionally I have reset the `m2e-core-tests` submodule head to the previous commit since the PR in tesla/m2e-core-tests was abandoned and the referenced commit is not present in other developers git clone.

PR #378, opened 2021-10-16
Set m2e.version property for 1.18.3. development iteration

HannesWell:
To prepare for a the 1.1.8.3 development the `<m2e.version>` has to be bumped in the root `pom.xml`, otherwise the snapshots are published under the old URL.

The versions of the m2e.feature and m2e.sdk.feature were already bumped in the previous commit.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -35,7 +35,7 @@
 
 		<!-- Current target version of m2e, used to deduce some URL at verify and publish.
 			This should usually be sync'd with the main feature version. -->
-		<m2e.version>1.18.2</m2e.version>
+		<m2e.version>1.18.3</m2e.version>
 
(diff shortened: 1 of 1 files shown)


HannesWell:
Ideally the <m2e.version> property would not be necessary, because it can also be confused with the System-property of the same name that m2e sets to enable profiles that are only active in the IDE.
I think using the version of the m2e.feature instead would be handy. Then the property also don't have to be adjusted.

mickaelistria:
> I think using the version of the m2e.feature instead would be handy. Then the property also don't have to be adjusted.

It sounds like a good idea; especially since I don't see the m2e.version property clearly used in the pom files actually.
Maybe this property could be fully removed from pom file now; and the logic to retrieve the version (eg from feature) would be part of the Jenkinsfile.

PR #379, opened 2021-10-16
Normalize line-endings and restore m2e-core-tests submodule-head and m2e.core/.classpath

HannesWell:


HannesWell:
The m2e-core-tests submodule-head was set to the latest commit of the abandoned PR https://github.com/tesla/m2e-core-tests/pull/132. This PR reverts it.

Furthermore in `m2e.core/.classpath` the line `<classpathentry kind="output" path="bin"/>` was changed to `<classpathentry kind="output" path="target/classes"/>`. This sometimes happens automatically. I also observed this my workspace.
But I'm not sure when this happens exactly. I can only reliably reproduce it by right-clicking -> Maven -> Update project... but as a developer you should not have to do this. I also verified that the Oomph setup is not performing this change.

PR #396, opened 2021-11-11
Various cleanups

akurtakov:


laeubi:
Should we really bump versions for cleanup only actions?

mickaelistria:
> Should we really bump versions for cleanup only actions?

Only if this bundle was already released for the given version.

akurtakov:
It's different bytecode produce so yes. I try to enable warnings and cleanup actions for such things so in the future we have less and less such cleanup patches. I strongly encourage such route.

PR #403, opened 2021-11-15
[LemMinX] simplify code and use jars from m2e.maven.indexer

HannesWell:
This PR simplifies the code of the `org.eclipse.m2e.editor.lemminx` projects and
significantly reduces its size by using the jars from the `org.eclipse.m2e.maven.indexer` bundle instead of embedding them again. This reduces the size of its jar from 5.2M to 125kb and saves about 5MB, which is about 20% of the entire m2e updatesite.

I was not able to test this extensively but the embedded jars are the same and the tests passed.


mickaelistria (approved):
Looks great! Thanks!

Issue #406, opened 2021-11-17
1.19.0.20211116-0804 lost XML syntax coloring of pom.xml


dougbreaux:
After accepting: 

`M2E - Maven Integration for Eclipse (includes Incubating components)	1.19.0.20211116-0804`

I no longer have the normal XML coloring of my pom.xml files. I seem to have lost the sub-module:

` M2E - POM Editor (Legacy) using WebTools SourceEditing UI	1.18.2.20211002-1029`

I was also unable to revert to what I had before: 

`M2E - Maven Integration for Eclipse (includes Incubating components)	1.18.3.20211018-0804`

```
!ENTRY org.eclipse.equinox.p2.engine 4 4 2021-11-17 11:04:34.866
!MESSAGE An error occurred while collecting items to be installed
!SUBENTRY 1 org.eclipse.equinox.p2.engine 4 0 2021-11-17 11:04:34.866
!MESSAGE session context was:(profile=epp.package.jee, phase=org.eclipse.equinox.internal.p2.engine.phases.Collect, operand=, action=).
!SUBENTRY 1 org.eclipse.equinox.p2.artifact.repository 4 0 2021-11-17 11:04:34.866
!MESSAGE No repository found containing: org.eclipse.update.feature,org.eclipse.m2e.feature,1.18.3.20211018-0804
```

mickaelistria:
The WebTools editor is indeed removed by default. Is you file open using the "Generic Test Editor"? Do you have the `    M2E - POM Editor using LemMinX language server (includes Incubating components)` feature installed?

dougbreaux:
Have tried manually installing older version from Snapshots repo:
![image](https://user-images.githubusercontent.com/25640850/142250033-02f11f9a-3353-496e-b8d8-8db6a5d2591e.png)

I now have this hybrid-crazy that I'll probably come to regret, but that seems to be working so far:
![image](https://user-images.githubusercontent.com/25640850/142250212-3de00e1f-30d9-450e-9105-6a6129527ab3.png)

dougbreaux:
> 
> 
> The WebTools editor is indeed removed by default. Is you file open using the "Generic Test Editor"? Do you have the ` M2E - POM Editor using LemMinX language server (includes Incubating components)` feature installed?

Hmmm... I hadn't tried that LemMinX component. Guess I can try adding that.

HannesWell:
I also have the problem, that the installation of the m2e-LemMinX feature failed due to missing requirements.
[...]

PR #407, opened 2021-11-17
Add repository reference for LemMinX-Editor requirements (#406)

HannesWell:
As suggest in #406 this PR adds references pointing to the repositories providing the LemMinX requirements to the m2e-site repo. This simplifies installation of the `org.eclipse.m2e.lemminx.feature` (i.e. without need to specifiy more repo-URLs than the m2e one) for versions of m2e (release or snapshot) that were not yet assembled into an Eclipse Sim-Rel.

I added comments to m2e's target file to remind us to keep the repos-URLs in sync.
If one day any of those repos can be removed from the TP (possibly https://download.eclipse.org/lsp4e/releases if wildwebdeveloper regularly catches up), the corresponding reference should also be removed from the m2e-repo as well.

### DIFF
--- a/org.eclipse.m2e.site/category.xml
+++ b/org.eclipse.m2e.site/category.xml
@@ -22,4 +22,6 @@
       <category name="m2e"/>
    </feature>
    <category-def name="m2e" label="Maven Integration for Eclipse"/>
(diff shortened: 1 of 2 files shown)


mickaelistria:
Thanks!

PR #408, opened 2021-11-17
[SDK-Feature] Include all m2e features and add missing sources

HannesWell:
This PR suggests to include all m2e features into the m2e.sdk feature so it transitively contains all m2e plug-ins (plus bndlib and workspace.cli). Besides that this PR adds the missing source bundles.
Only the sources of the Maven-runtime components are not added because I assume that they will not work in the form they are produced now (they have to be adjusted in later PR, but I have to make myself familiar if and how nested sources jars are supported).

Having a sdk feature that contains all plug-ins of a project and their sources simplifies development of m2e-based plug-ins because one just has to add the m2e.sdk to the target-platform and gets all m2e plug-ins and their sources.

Furthermore this ensures that the version of the m2e.sdk is always the same as the version of the whole m2e project (because whenever m2e plug-in changes its version the m2e.sdk feature is forced by the build to have a corresponding version bump). This allows to use the m2e.sdk version to automatically derive the current snapshot repo version and makes the `m2e.version` property in the root pom.xml obsolete. I'm about to prepare a follow-up PR for this.

Additionally I (IMHO) enhanced the name and description of the SDK-feature.

### DIFF
--- a/org.eclipse.m2e.sdk.feature/feature.properties
+++ b/org.eclipse.m2e.sdk.feature/feature.properties
@@ -16,7 +16,7 @@
 # This file should be translated.
 
 # "featureName" property - name of the feature
-featureName=M2E - Extensions Development Support (optional)
+featureName=M2E - Complete Development Kit (optional)
 
 # "providerName" property - name of the company that provides the feature
 providerName=Eclipse.org - m2e
@@ -25,7 +25,7 @@ providerName=Eclipse.org - m2e
 #updateSiteName=The Eclipse Project Updates
 
 # "description" property - description of the feature
-description=M2Eclipse Extensions Development Support (Optional)
+description=The full M2Eclipse Software Development Kit that contains all m2e plug-ins and features and their sources. (Optional)
 
 # "copyright" property - text of the "Feature Update Copyright"
 copyright=\
--- a/org.eclipse.m2e.sdk.feature/feature.xml
+++ b/org.eclipse.m2e.sdk.feature/feature.xml
@@ -221,11 +221,4 @@
          version="0.0.0"
          unpack="false"/>
 
-   <plugin
-         id="org.eclipse.m2e.workspace.cli.source"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
-         unpack="false"/>
-
 </feature>
--- a/org.eclipse.m2e.sdk.feature/feature.xml
+++ b/org.eclipse.m2e.sdk.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.m2e.sdk.feature"
       label="%featureName"
-      version="1.18.3.qualifier"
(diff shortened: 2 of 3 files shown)


HannesWell:
Actually it would be very handy if Tycho had a (opt-in) feature to automatically add the source bundles to the feature. This would avoid the risk to forget some.

HannesWell:
Looks like the org.eclipse.m2e.workspace.cli.source is not properly signed.
Removed it in a fix-up commit that should be squashed together.

mickaelistria:
> Actually it would be very handy if Tycho had a (opt-in) feature to automatically add the source bundles to the feature.

Tycho can generate source features from "regular" features, so that would maybe be better for m2e to drop this feature and use source-feature generation instead.

HannesWell:
> 
> 
> > Actually it would be very handy if Tycho had a (opt-in) feature to automatically add the source bundles to the feature.
> 
> Tycho can generate source features from "regular" features, so that would maybe be better for m2e to drop this feature and use source-feature generation instead.

Thanks for that hint. Using this would be definitively make it simpler respectively harder to forget something.
Nevertheless I like the idea to have a skd.feature that contains all m2e plug-ins (+sources), it can be used in a target-platform for developers that develop plug-ins based on m2e. For example the Eclipse-Platform has it with `org.eclipse.sdk` or xText has it with `org.eclipse.xtext.sdk` and I personally find them very handy to create a target-platform with them.

On the other hand, having a source feature for each m2e feature adds five more features to the repo. If one wants to install them manually via Install New Software it could become a bit overwhelming. Having only one SDK feature instead would reduce the number of selectable feature. But the question is, is this relevant?

akurtakov:
You can have both. Let tycho generate source features for each feature with auto include for source bundles and etc. and keep the current sdk feature including the auto generated features. That way you get best of both worlds.

mickaelistria:
To get source in PDE, the .target only needs to have `includeSources=true` and the source bundles be in the repo. No need for the .target to reference a specific feature.

> You can have both. Let tycho generate source features for each feature with auto include for source bundles and etc. and keep the current sdk feature including the auto generated features. That way you get best of both worlds.

+1.

HannesWell:
> 
> 
> You can have both. Let tycho generate source features for each feature with auto include for source bundles and etc. and keep the current sdk feature including the auto generated features. That way you get best of both worlds.

That was also what I thought of. I'm sorry for not being clear.
I will propose a corresponding change in the next days. But I think the sources bundles for the runtime-component have to be adjusted before, because I don't expect them to work how they are build now.

> 
> 
> To get source in PDE, the .target only needs to have `includeSources=true` and the source bundles be in the repo. No need for the .target to reference a specific feature.

I know that, my point why I think sdk features are handy is that you can simply get everything from a project into the TP and can pick from it what you really need later while developing or building something else.

PR #409, opened 2021-11-17
[Build] Read m2e-version from m2e.sdk instead of m2e.version property

HannesWell:
This PR aims to remove the `m2e.version` property from the root `pom.xml` like it was suggested in #378.

The  `m2e.version` property in the root `pom.xml` is only used by the Jenkins file to deduce the name of the current snapshot repository.

The motivation to remove it is, that it might be confused with the System-property also named `m2e.version` which m2e creates when building Maven projects within the Eclipse-IDE to enable "m2e-profiles" that are intended to only run when the project is build by m2e within Eclipse-IDE.
Furthermore the m2e.version property in the root `pom.xml` has to be changed manually which is cumbersome and could be forgotten so the snapshot repo is not exactly named like the up-coming release version (like it is for example at the moment, 1.19.0 is planned but the repo is still named 1.18.3).

With PR #408 I suggested to include all m2e plug-ins (and their sources) into the m2e.sdk. This ensures that the m2e.sdk's version has to be changed with respect to semantic-versioning in correspondance to the greatest version change of all m2e plug-ins and features since the last release. So in the end the Tycho goal `tycho-p2-extras-plugin:compare-version-with-baselines` ensures that the m2e.sdk has the same version like the m2e-project as a whole. So its version can be used as replacement for the previous m2e.version property.

The only remaining task, this PR addresses, is to read the version from the m2e.sdk feature.
Since I'm not using the Linux CLI very often I'm bad in using corresponding Regular-Expressions and bash-commands to extract the version from the `feature.xml` directly (like it was done before). Therefore I adjusted the build to write into `org.eclipse.m2e.sdk.feature/target` a file named `m2e.version` that contains only the m2e.sdk version.
Reading this files content is simple in the Jenkins-File and IMHO less error-prone than having a complex RegEx to read the right version from a feature.xml. 

This is ready, as soon as #408 is merged.


mickaelistria:
I'm not sure the SDK feature is a good choice as it's likely to be removed in some future, replaced by features generated by tycho-source-feature-plugin.
Also, it's generally unsafe to assume that we can derive a "project version" (which is a kind of marketing version) from artifacts that try to follow semantic versioning. Overall, I'm wondering whether the proposed change makes things simpler than the current state where the version is explicitly defined somewhere.
However I agree removing m2e.version alltogether is a good idea. Maybe this m2e version can be hardcoded in Jenkinsfile instead of pom.xml; or maybe we can just stop using it and publish snapshots only to download.eclipse.org/technology/m2e/snapshots/latest so we don't need a version at all.

HannesWell:
> 
> 
> I'm not sure the SDK feature is a good choice as it's likely to be removed in some future, replaced by features generated by tycho-source-feature-plugin. Also, it's generally unsafe to assume that we can derive a "project version" (which is a kind of marketing version) from artifacts that try to follow semantic versioning.

There is indeed not a strong guarantee that the "project-version" has to be the same version as any artifact, but in case of m2e I expected to be the case. I wouldn't expect for example a m2e-version 2.0.0 to be released without any m2e-artifact making the same change. And since m2e.sdk is now a composite of all it reflects the change of the whole project. As long as the "project-version" also follows SemVer it should match. But you are definitely more experienced than I so you can judge it better if this also will be the case for the future.

>Overall, I'm wondering whether the proposed change makes things simpler than the current state where the version is explicitly defined somewhere. However I agree removing m2e.version alltogether is a good idea. Maybe this m2e version can be hardcoded in Jenkinsfile instead of pom.xml; or maybe we can just stop using it and publish snapshots only to download.eclipse.org/technology/m2e/snapshots/latest so we don't need a version at all.

If it should be hard-coded, it is definitely better done in the Jenkins file!

I also thought about dropping `download.eclipse.org/technology/m2e/snapshots/<m2e-version>`, but I think it is handy for people  that want to have the current snapshot because it fixes e.g. a bug, but don't want to get m2e-snapshots forever and actually just want the latest releases.
If they use `download.eclipse.org/technology/m2e/releases/latest` plus `download.eclipse.org/technology/m2e/snapshot/<curren m2 version>`, they get the current snapshot until the next release and then 'automatically' continue with the `releses/latest` repo.

What do you think about keeping this approach for now? Because a.t.m the m2e 'project-version' follows SemVer and if it changes one day, the  one day it changes M2E_VERSION can simply be hardcoded in the Jenkinsfile instead of being read from `org.eclipse.m2e.sdk.feature/target/m2e.version`.

mickaelistria:
Although I'm not personally so enthusiast, I think your suggestion isn't a bad one; so if you have a strong belief and some confidence that this is going to help, I support you in adopting it.
In worst case, we can change to hardcoding in Jenkinsfile later.

HannesWell:
> 
> 
> Although I'm not personally so enthusiast, I think your suggestion isn't a bad one; so if you have a strong belief and some confidence that this is going to help, I support you in adopting it. In worst case, we can change to hardcoding in Jenkinsfile later.

I think it is a trade-off between the convenience of a automatically derived version, that can also not be forgotten to be synchronized and potential 'arbitrary' version jumps for the projects. Since I don't see the later for m2e at the moment, I think for now it helps. And yes, if it does not work in the future we change to hardcoding it. 👍🏽

HannesWell:
It worked well. The snapshot repo at https://download.eclipse.org/technology/m2e/snapshots/1.19.0/ without further ado.
Unfortunately the old repo snapshots/1.18.3/ has to be deleted manually (which is now easily doable through the web-interface).

Alternatively the the `promote-snapshots-to-release` Jenkins job could be changed to purge all snapshot repos but snapshots/latest. A.t.m. it just deletes `snapshots/<release-version>/latest`.

Issue #431, opened 2021-12-08
Test failure at 	at org.eclipse.m2e.core.ui.tests.ConsoleTest.testConsoleHasOutput(ConsoleTest.java:61) in mvn clean verify build


vogella:
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 2.883 s <<< FAILURE! - in org.eclipse.m2e.core.ui.tests.AllTests
testConsoleHasOutput(org.eclipse.m2e.core.ui.tests.ConsoleTest)  Time elapsed: 2.857 s  <<< FAILURE!
java.lang.AssertionError: missing output in console
	at org.eclipse.m2e.core.ui.tests.ConsoleTest.testConsoleHasOutput(ConsoleTest.java:61)

HannesWell:
This is a duplicate of #341.

Issue #424, opened 2021-12-07
Downgrade the "Plugin Execution not covered by lifecycle configuration" error to warning / info or do not show it at all


vogella:
Currently helping a client and this client uses a lot of Maven. 

His and mine experience with Eclipse and importing a random Maven project are really bad as almost every Maven projects imports with errors. For example the "https://github.com/Atmosphere/atmosphere" project gives me the following errors:

Description	Resource	Path	Location	Type
Plugin execution not covered by lifecycle configuration: org.apache.maven.plugins:maven-checkstyle-plugin:3.1.1:checkstyle (execution: validate, phase: validate)	pom.xml	/annotations	line 3	Maven Project Build Lifecycle Mapping Problem
Plugin execution not covered by lifecycle configuration: org.apache.maven.plugins:maven-checkstyle-plugin:3.1.1:checkstyle (execution: validate, phase: validate)	pom.xml	/cpr	line 3	Maven Project Build Lifecycle Mapping Problem
Plugin execution not covered by lifecycle configuration: org.apache.maven.plugins:maven-checkstyle-plugin:3.1.1:checkstyle (execution: validate, phase: validate)	pom.xml	/jersey	line 3	Maven Project Build Lifecycle Mapping Problem
Plugin execution not covered by lifecycle configuration: org.apache.maven.plugins:maven-checkstyle-plugin:3.1.1:checkstyle (execution: validate, phase: validate)	pom.xml	/native	line 3	Maven Project Build Lifecycle Mapping Problem
Plugin execution not covered by lifecycle configuration: org.apache.maven.plugins:maven-pmd-plugin:3.13.0:check (execution: validate, phase: validate)	pom.xml	/annotations	line 3	Maven Project Build Lifecycle Mapping Problem
Plugin execution not covered by lifecycle configuration: org.apache.maven.plugins:maven-pmd-plugin:3.13.0:check (execution: validate, phase: validate)	pom.xml	/cpr	line 3	Maven Project Build Lifecycle Mapping Problem
Plugin execution not covered by lifecycle configuration: org.apache.maven.plugins:maven-pmd-plugin:3.13.0:check (execution: validate, phase: validate)	pom.xml	/jersey	line 3	Maven Project Build Lifecycle Mapping Problem
[...]

Issue #341, opened 2021-10-03
Test ConsoleTest.testConsoleHasOutput() is unstable


HannesWell:
The test case `ConsoleTest.testConsoleHasOutput()` seems to be unstable respectively to have a race-condition or comparable. Sometimes the test passes and sometimes not, even without changes. See for example https://ci.eclipse.org/m2e/job/m2e/job/master/89/ and https://ci.eclipse.org/m2e/job/m2e/job/master/90/ where the exact same commit was build. In Job 89 the test failed and in Job 90 it passed.

laeubi:
One remark for the test: I think the very first thing to change here is, that instead of

```
System.out.println("Console Text: " + ((TextConsole)mavenConsole).getDocument().get());
assertTrue("missing output in console", ((TextConsole)mavenConsole).getDocument().get().contains("BUILD SUCCESS"));
```
we should replace with:

1. extract the `((TextConsole)mavenConsole).getDocument().get()` into a local variable so there is no chance in having different output in the debug and the assert
2. Instead of query the console once, we could simply have a loop that fetches the text, check if it contains the desired string and sleep for a few milliseconds and then recheck. If this fails for more than 10 times we could fail the test.
3. It woudl be good to make the failing message more expressive e.g. `fail("Expecting BUILD SUCCESS in console output but got "+text)`;

laeubi:
Another aproach would be to add a [IPropertyChangeListener](https://javadoc.scijava.org/Eclipse/org/eclipse/ui/console/AbstractConsole.html#addPropertyChangeListener(org.eclipse.jface.util.IPropertyChangeListener)) and see if the text-console fires an event on text change, we can then use a countdonwn latch to signal the expected string was found + a timed wait on the latch.

HannesWell:
I fully agree with your remarks. I would even suggest to use Hamcrest matchers.
Using a IPropertyChangeListener seems to be a more stable approach to me. In case a `IConsoleListener` is added to the consoleManager, the sleep-loop could be fully omitted.

I can take care of this, unless you want to prepare a PR?

laeubi:
Just go a haed :-) I'm just not sure if the console emits an event on text change but we can try it out..

laeubi:
Duplicate of #431

vogella:
@HannesWell can you send the PR to stabilize the test?

HannesWell:
> @HannesWell can you send the PR to stabilize the test?

[...]

PR #432, opened 2021-12-08
Disables ConsoleTest test as it currently fails

vogella:
With the failing test, the build fails, so temporary disable the test.

For #431.


akurtakov:
What is the problem with the test? It's always better to fix test than to ignore it.

vogella:
@akurtakov it fails, see https://ci.eclipse.org/m2e/job/m2e/job/PR-425/2/testReport/junit/org.eclipse.m2e.core.ui.tests/ConsoleTest/testConsoleHasOutput/ or in a local build.

 Currently I'm trying to build m2e-core as I wanted to look into https://github.com/eclipse-m2e/m2e-core/issues/424 and this stops it. As a new guy to m2e-core I have no idea, why it fails.

HannesWell:
These failures are likely due to race-conditions and don't happen constantly. I have already started to work on a solution in context of issue #341 but didn't have the time to complete it yet, but I can try to priorities it.

As Alex suggested this should be the way to go.

vogella:
> but I can try to priorities it.

Would be nice.

PR #433, opened 2021-12-08
Adds Github build verification

vogella:
The build currently fails and complains about missing version increases, see https://github.com/vogellacompany/m2e-core/runs/4455549109?check_suite_focus=true#step:6:6755

@mickaelistria @laeubi what to do here? Send separate pull requests to increase the version? Or is the check incorrect, the Jenkins build seems not to care about this.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,43 @@
+# This workflow will build the m2e project with Maven
+
+name: Build m2e-core
+
+on:
+  push:
+    branches: 
+      - 'master'
+  pull_request:
+    branches: 
+     - 'master'
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 11
+      uses: actions/setup-java@v2
+      with:
+        java-version: '11'
+        distribution: 'adopt'
+    - name: Cache local Maven repository
+      uses: actions/cache@v2
+      with:
+        path: ~/.m2/repository
+        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
+          ${{ runner.os }}-maven-
+    - name: Generated Meta data
+      run: mvn generate-sources -f m2e-maven-runtime -Pgenerate-osgi-metadata -Dtycho.mode=maven
+    - name: Build m2e-core
+      run: mvn clean verify -Dmaven.test.failure.ignore=true --file pom.xml
+    - name: Upload Test Results
+      uses: actions/upload-artifact@v2
+      with:
+        name: test-results
+        if-no-files-found: error
+        path: |
+          ${{ github.workspace }}/*/target/surefire-reports/*.xml
+    


vogella:
Error:  Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:2.5.0:compare-version-with-baselines (check-no-version-regression) on project org.eclipse.m2e.maven.runtime.slf4j.simple: Only qualifier changed for (org.eclipse.m2e.maven.runtime.slf4j.simple/1.18.1.20211208-1002). Expected to have bigger x.y.z than what is available in baseline (1.18.1.20211008-0657) -> [Help 1]
Error:  Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:2.5.0:compare-version-with-baselines (check-no-version-regression) on project org.eclipse.m2e.model.edit: Only qualifier changed for (org.eclipse.m2e.model.edit/1.18.2.20211208-1002). Expected to have bigger x.y.z than what is available in baseline (1.18.2.20211002-1029) -> [Help 1]
Error:  Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:2.5.0:compare-version-with-baselines (check-no-version-regression) on project org.eclipse.m2e.logback.configuration: Only qualifier changed for (org.eclipse.m2e.logback.configuration/1.16.3.20211208-1002). Expected to have bigger x.y.z than what is available in baseline (1.16.3.20211002-1029) -> [Help 1]
Error:  
Error:  To see the full stack trace of the errors, re-run Maven with the -e switch.

.github/workflows/maven.yml
```diff
@@ -0,0 +1,37 @@
...
+      uses: actions/cache@v2
+      with:
+        path: ~/.m2/repository
+        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
+          ${{ runner.os }}-maven-
+    - name: Generated Meta data
+      run: mvn generate-sources -f m2e-maven-runtime -Pgenerate-osgi-metadata -Dtycho.mode=maven
+    - name: Build m2e-core
+      run: mvn clean verify --file pom.xml -fae
```
laeubi, line 35: You should use -Dmaven.test.failure.ignore=true instead and publish the test results as done in 
https://github.com/eclipse/tycho/blob/master/.github/workflows/maven.yml#L46
then another action can publish the test results see https://github.com/eclipse/tycho/blob/master/.github/workflows/check.yml

Or simply not use fae if you want to fail fast (what generally is better if we are not talking about integration tests)

laeubi (changes requested):

mickaelistria (changes requested):
We don't disable failing tests, we either fix them or live with the annoyance they cause.

mickaelistria:
>  Or is the check incorrect, the Jenkins build seems not to care about this.

The check is correct and Jenkins build cares about it. It's strange that bundles that were not modified are now treated as modified and generating new qualifiers. Isn't there an extra commit modifying them on your end?

vogella:
Thanks @laeubi and @mickaelistria I adjusted the commit and removed the removal of the failing test. AFAICS no other commits modified the bundles with the new qualifiers.

vogella:
Test failure is unrelated to this change.

mickaelistria:
thanks!

HannesWell:
Sorry for the question, but what is the advantage of having a Jenkins and GH-Actions build in parallel that do the same?

vogella:
Jenkins was not reliable in the past and AFAIK GitHub actions are faster

Hannes Wellmann ***@***.***> schrieb am Mi., 8. Dez. 2021,
20:59:

HannesWell:
> Jenkins was not reliable in the past and AFAIK GitHub actions are faster

Understand. Thanks for the explanation.

mickaelistria:
Currently, it's the opposite; GitHub Actions build is less reliable than Jenkins; it shows an unexpected failure.

vogella:
Which one? I checked multiple pull request and both Jenkins and build action fail. And #341 shows also in my local build.

laeubi:
See https://github.com/eclipse-m2e/m2e-core/runs/4457287427 for example fails with

> Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:2.5.0:compare-version-with-baselines (check-no-version-regression) on project org.eclipse.m2e.maven.runtime.slf4j.simple: Only qualifier changed for (org.eclipse.m2e.maven.runtime.slf4j.simple/1.18.1.20211208-1248). Expected to have bigger x.y.z than what is available in baseline (1.18.1.20211008-0657) -> [Help 1]

That is not the test-failure like on the Ci server.

laeubi:
The ci-build seems  to use xvnc to execute the core build, could this be an issue?

mickaelistria:
I don't think it's related to Xvnc. The qualifier is derived from Git history (last modification timestamp). I don't know what can cause it to be different on GH Actions.

HannesWell:
In my latest PR GH-Actions failed again (reporting the version regressions I cannot explain), while Jenkins succeeds.
Therefore I suggest to revert this PR.

mickaelistria:
> Therefore I suggest to revert this PR.

No objection against a revert. The Jenkins CI does the job, so if the Github Actions cause annoyance and no-one plans to get them working, then it's better to not have it.

PR #434, opened 2021-12-08
#280 rebased

laeubi:


laeubi:
@mickaelistria it still complains:

Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:2.5.0:compare-version-with-baselines (check-no-version-regression) on project org.eclipse.m2e.maven.runtime.slf4j.simple: Only qualifier changed for (org.eclipse.m2e.maven.runtime.slf4j.simple/1.18.1.20211208-1723). Expected to have bigger x.y.z than what is available in baseline (1.18.1.20211008-0657) -> [Help 1]

akurtakov:
I see that you're switching target platform. So there are at least two options:
* internal api is used which changed in a way that caused new bytecode to be generated
* lambdas order is different due to jdt.core being updated (see https://bugs.eclipse.org/bugs/show_bug.cgi?id=561135 )

mickaelistria:
The qualifier comes from Git history, binary comparison happens later.

laeubi:
Especially the slf4j simple API here does not contain any code and only depends on a maven deployed artifact so I'm confused. @HannesWell please correct me if I'm wrong.
Beside that should jdt updates not only be relevant if tycho is updated as well?

HannesWell:
> Especially the slf4j simple API here does not contain any code and only depends on a maven deployed artifact so I'm confused. @HannesWell please correct me if I'm wrong. Beside that should jdt updates not only be relevant if tycho is updated as well?

That's correct. The bundle `org.eclipse.m2e.maven.runtime.slf4j.simple` is just a 'clean' wrapper around the slf4j-simple bundle to have a bundle that is fully inactive/unusable in an OSGi Environment (like Eclipse) but can be actively used in a plain Java application like a Maven build is. This way the slf4j-loggers do not bind to slf4j-simple within Eclipse but do in a Maven-build JVM.
But as far as I can oversee it, the changes of this PR only affect the m2e.pde and m2e.pde.ui plug-in and the TP.
Is the change in the .gitignore intended? However my first thought is that it should not affect the build.

HannesWell:
But what's interesting is that the current master-brunch build of m2e also fails with the same error, but only the GitHub Action.

When I run the build (with this PR's commit), the build of the m2e.slf4j.simple module succeeds but m2e.pde has compile errors (but they also appear in the workspace).

laeubi:
Um... it seems I accidentally upgraded the Orbit from 2021-09 > 2021-12 instead of platform, still strange that this seem to force version bumps, maybe because the maven modules are build by tycho?

laeubi:
The CI is fine now and I'd like to merge this now, any concerns?

mickaelistria:
> any concerns?

No concern from me.

HannesWell:
> The CI is fine now and I'd like to merge this now, any concerns?

Sorry I didn't have to the time to test it yet, but I try to do it in the next days.
But if you prefere to merge now, I can test and report any findings afterwards.

If you want to merge, can you please use squash-and-merge?

laeubi:
Yep let me know if anything is missing and/or not working, as this includes some refactoring that blocks other features its better to start with it right now :-)

PR #436, opened 2021-12-09
Test upgrade the sdk feature

laeubi:


laeubi:
@HannesWell @mickaelistria this also happens when only upgrade the target. Should we investigate more here or should I try to bump the versions?

laeubi:
The "normal" ci build only fails with the usual test failure....

mickaelistria:
If the GitHub Actions build does not work because of an irrelevant error, we should disable it.

Issue #439, opened 2021-12-09
Adopt PDE configurations for maven-bundle-plugins


laeubi:
m2eclipse-tycho is no longer maintained but contains a very useful project configurator:

[PDEMavenBundlePluginConfigurator](https://github.com/tesla/m2eclipse-tycho/blob/master/org.sonatype.tycho.m2e/src/org/sonatype/tycho/m2e/internal/PDEMavenBundlePluginConfigurator.java)

this one instructs PDE to look at the right location for the `MANIFEST.MF` file. Although m2e should not host/contain specific code for all maven plugins, the maven-bundle-plugins (BND+Felix) are basically that "basic" that we should support them natively up to that point that the PDE project is configured accordingly and then the default lifecycle-mapping could be applied that is shipped with the plugin itself.

@mickaelistria what do you think? any concerns? I think a 1st-class support for plugin development in m2e is crucial for acceptance of more "maven-like" development.

mickaelistria:
I think it makes sense to include it in m2e, together in the same feature or bundle as the .target support.

laeubi:
> I think it makes sense to include it in m2e, together in the same feature or bundle as the .target support.

Yes that was also my intention as we already have a m2e.pde feature it would perfectly match to include it there.

laeubi:
@HannesWell due to mi limited testing this restores the behavior of adding PDE nature and setting the BUNDLE_ROOT for bnd-/felix-maven plugins when uninstall the no longer maintained tycho-connector. for felix I'll have then added the follwoing pom configuration:

```
<plugin>
	<groupId>org.apache.felix</groupId>
	<artifactId>maven-bundle-plugin</artifactId>
	<version>5.1.2</version>
	<extensions>true</extensions>
	<configuration>
		<instructions>
			<Export-Package>*</Export-Package>
			<Import-Package>!sun.*,*</Import-Package>
		</instructions>
		<supportIncrementalBuild>true</supportIncrementalBuild>
	</configuration>
	<executions>
		<execution>
			<goals>
				<goal>manifest</goal>
			</goals>
			<phase>process-classes</phase>
		</execution>
	</executions>
</plugin>
```

to get automatic bundle manifest generation on code changes and PDE recognizing the project as a bundle.

Just in case you wanted to try this out/help with testing :-)

laeubi:
[...]

PR #440, opened 2021-12-09
#439 - Adopt PDE configurations for maven-bundle-plugins

laeubi:
This add initial code contribution and project setup to m2e. This code is copied from (with slight adjustments e.g. package name and constant inlining):

- https://github.com/tesla/m2eclipse-tycho/blob/master/org.sonatype.tycho.m2e/src/org/sonatype/tycho/m2e/internal/PDEMavenBundlePluginConfigurator.java
- https://github.com/tesla/m2eclipse-tycho/blob/master/org.sonatype.tycho.m2e/src/org/sonatype/tycho/m2e/internal/PDEProjectHelper.java

both files are EPL 1.0 and the contribution is below 1000 LOC (including configuration files), so I would assume no IP review is required, but like to get a verification by @mickaelistria and @fbricon as project lead.

If the legal part is done I'd like to test and maybe enhance the code later on, e.g. currently I have only used it with felix plugin but the same should apply to the bnd-bundle plugin.

### DIFF
--- a/org.eclipse.m2e.pde.connector/build.properties
+++ b/org.eclipse.m2e.pde.connector/build.properties
@@ -1,0 +1,5 @@
+source.. = src/
+output.. = bin/
+bin.includes = META-INF/,\
+               .,\
+               plugin.xml
--- a/org.eclipse.m2e.pde.connector/build.properties
+++ b/org.eclipse.m2e.pde.connector/build.properties
@@ -2,4 +2,5 @@ source.. = src/
 output.. = bin/
 bin.includes = META-INF/,\
                .,\
-               plugin.xml
+               plugin.xml,\
+               lifecycle-mapping-metadata.xml
--- a/org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/Activator.java
+++ b/org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/Activator.java
@@ -1,0 +1,44 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Christoph Läubrich
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *      Christoph Läubrich - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.m2e.pde.connector;
+
+import java.util.Optional;
+
+import org.eclipse.pde.core.project.IBundleProjectService;
+import org.osgi.framework.BundleActivator;
+import org.osgi.framework.BundleContext;
+import org.osgi.util.tracker.ServiceTracker;
+
+public class Activator implements BundleActivator {
+
+	private static volatile ServiceTracker<IBundleProjectService, IBundleProjectService> bundleProjectServiceTracker;
+
+	@Override
+	public void start(BundleContext context) throws Exception {
+		bundleProjectServiceTracker = new ServiceTracker<>(context, IBundleProjectService.class, null);
+		bundleProjectServiceTracker.open();
+	}
+
+	@Override
+	public void stop(BundleContext context) throws Exception {
+		bundleProjectServiceTracker.close();
+	}
+
+	public static Optional<IBundleProjectService> getBundleProjectService() {
+		if (bundleProjectServiceTracker == null) {
+			return Optional.empty();
+		}
+		return Optional.ofNullable(bundleProjectServiceTracker.getService());
+	}
+
+}
--- a/org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/PDEMavenBundlePluginConfigurator.java
+++ b/org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/PDEMavenBundlePluginConfigurator.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008 Sonatype, Inc.
+ * Copyright (c) 2008, 2021 Sonatype, Inc.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -7,11 +7,12 @@
  *******************************************************************************/
 package org.eclipse.m2e.pde.connector;
 
-
 import java.io.File;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
+import org.apache.maven.model.Plugin;
 import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.project.MavenProject;
 import org.eclipse.core.resources.IMarker;
@@ -25,9 +26,12 @@ import org.eclipse.m2e.core.internal.IMavenConstants;
 import org.eclipse.m2e.core.internal.markers.MavenProblemInfo;
 import org.eclipse.m2e.core.internal.markers.SourceLocation;
 import org.eclipse.m2e.core.internal.markers.SourceLocationHelper;
+import org.eclipse.m2e.core.lifecyclemapping.model.IPluginExecutionMetadata;
 import org.eclipse.m2e.core.project.IMavenProjectFacade;
+import org.eclipse.m2e.core.project.configurator.AbstractBuildParticipant;
 import org.eclipse.m2e.core.project.configurator.AbstractProjectConfigurator;
 import org.eclipse.m2e.core.project.configurator.ILifecycleMappingConfiguration;
+import org.eclipse.m2e.core.project.configurator.MojoExecutionBuildParticipant;
 import org.eclipse.m2e.core.project.configurator.MojoExecutionKey;
 import org.eclipse.m2e.core.project.configurator.ProjectConfigurationRequest;
 import org.eclipse.m2e.jdt.IClasspathDescriptor;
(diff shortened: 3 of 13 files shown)


org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/PDEMavenBundlePluginConfigurator.java
```diff
@@ -0,0 +1,118 @@
+package org.eclipse.m2e.pde.connector;
```
mickaelistria, line 1: Missing copyright header.

mickaelistria (changes requested):
We'll need to check with IP team whether it's fine to just pick some other EPL code here.
I'm a bit worried about the file with missing copyright header being a blocker.

akurtakov:
Whenever new code that is not submitted by author and is not coming from project governed by EF a CQ has to be open. Please open new one of type Project Content (Content to be maintained by an Eclipse Foundation Project)  and refer to this PR.

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/PDEMavenBundlePluginConfigurator.java
laeubi: The project contains a feature file (that is the one published and distributed to the catalouge) that mention the EPL 1.0

https://github.com/tesla/m2eclipse-tycho/blob/master/org.sonatype.tycho.m2e.feature/license.html

laeubi: If git is right, the only contributor is @ifedorenko maybe he can claim here if it is EPL licensed?

```
git shortlog -n -s -- PDEMavenBundlePluginConfigurator.java
     5  Igor Fedorenko
```

laeubi:
I opened https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23882

laeubi:
The CQ require PMC approval: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23882

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/PDEMavenBundlePluginConfigurator.java
laeubi: my fault, i missed to copy the header while migration the file (no idea why it was lost...) I will force push it...

laeubi:
> The CQ require PMC approval

if @fbricon responsible for PMC for m2e? Can we ping someone else if not?

HannesWell:
Is it necessary to place the connector in a dedicated plug-in project?
Wouldn't it be possible to have it in the o.e.m2e.pde plug-in in a package named `org.eclipse.m2e.pde.connector`?
The TP related classes maybe would be better moved in their own package, `o.e.m2e.pde.target` (in a separate PR)?

laeubi:
> Is it necessary to place the connector in a dedicated plug-in project?

I think it is not strictly necessary but I don't see a reason why a dedicated plugin would harm here. Even though its unlikely if one likes to have the maven-bundle support but not the target support it could be installed separately.

HannesWell:
> > Is it necessary to place the connector in a dedicated plug-in project?
> 
> I think it is not strictly necessary but I don't see a reason why a dedicated plugin would harm here. Even though its unlikely if one likes to have the maven-bundle support but not the target support it could be installed separately.

I don't consider it harmful as well.
But having more plug-ins just fills the dev-workspace and adds management overhead (more versions to bump, more metadata files, etc.). Of course only very few big plug-ins is probably as bad too many very small plug-ins. So apart from the technical boundaries/requirements (like separate installability, dependencies) the decision for a new plug-in it should be weigh up sensibly. And IMHO in this case a new plug-in is not worth it. I think it would be better to introduce multiple packages in org.eclipse.m2e.pde.
But this is of course also a matter of preferences/habits, so there is not one _right_ solution.

laeubi:
I don't think any "count" is a valid measure of putting code into separate plugins. Especially when it comes to (semantic) versioning, smaller items are generally better than bigger ones. Just because both contribute to PDE don't make them related. The target part contributes to the target handling of pde and this contributes to the m2e connectors so this are two separate concerns that should be live in separate plugins.

HannesWell:
> I don't think any "count" is a valid measure of putting code into separate plugins. Especially when it comes to (semantic) versioning, smaller items are generally better than bigger ones. Just because both contribute to PDE don't make them related. The target part contributes to the target handling of pde and this contributes to the m2e connectors so this are two separate concerns that should be live in separate plugins.

That also makes sense.
But then the o.e.m2e.pde should be renamed to o.e.m2e.pde.target (or similar) as well as the corresponding ui plug-in to reflect the fact that it is dedicated to that concern. Otherwise the name is actually too general. But that's off topic of this PR.

laeubi:
> But then the o.e.m2e.pde should be renamed to o.e.m2e.pde.target

probably yes, but this would require using some kind of patch feature to make P2 aware of it (I have never used that) so for now I won't complicate it too much just for the sake of naming.

org.eclipse.m2e.pde.connector/build.properties
```diff
@@ -0,0 +1,5 @@
+source.. = src/
+output.. = bin/
+bin.includes = META-INF/,\
+               .,\
+               plugin.xml
```
HannesWell, line 5: The `lifecycle-mapping-metadata.xml` should be added to the `bin.includes`, shouldn't it?
Otherwise it will be missing in the build jars.

HannesWell (changes requested):
Just tested it with a little Maven-project that uses the maven-bundle-plugin:manifest goal and it worked very well.

There is just an error markers in the Dependencies-Tab of the Manifest-Editor, even tough the plug-in/package from the Maven-project was proposed by the Add... dialogues (for Require-Bundle and Import-Package). So I think this is an PDE issue.

I also like the warnings, in the POM regarding the missing configuration. This is very helpful.

Some parts of the code could be simplified using present-day Java-11 methods. But I think this should be done in a subsequent PR, so that the initial move to Eclipse has as few changes as possible and we get a more clear git-history.

HannesWell:
> > But then the o.e.m2e.pde should be renamed to o.e.m2e.pde.target
> 
> probably yes, but this would require using some kind of patch feature to make P2 aware of it (I have never used that) so for now I won't complicate it too much just for the sake of naming.

For the installation process an [uninstallBundle p2-touchpoint instruction](https://help.eclipse.org/latest/index.jsp?topic=/org.eclipse.platform.doc.isv/guide/p2_actions_touchpoints.html) could be used. But another issue would be down-stream users that use require the m2e.pde bundle or import a package from it. At the moment the content is de-facto public API because it is not internal.

Which gives me another question: Is the code in this new m2e.pde.connector plug-in meant to be public API or should it better be internal?

laeubi:
I consider none of the code in `m2e.pde.*` as public API beside that is very unlikely that anyone is already enhancing that stuff right now :-)

Don't know if one could mark a bundle as "internal" but I really don't like these eclipse extension as they are extremely confusing and against OSGi principles.

laeubi:
> There is just an error markers in the Dependencies-Tab of the Manifest-Editor, even tough the plug-in/package from the Maven-project was proposed by the Add... dialogues (for Require-Bundle and Import-Package). So I think this is an PDE issue.

Will you take a look at this?

I sometimes noted that PDE also marks bundles with unresolved requirements with such an error marker, e.g. if your maven-bundle import-package a dependency but this is not available at the current target platform (I opened [Bug 577605](https://bugs.eclipse.org/bugs/show_bug.cgi?id=577605) for such cases). If you double click the bundle it should open the manifest and show if there is any unresolved package.

org.eclipse.m2e.pde.connector/build.properties
laeubi: Good catch!

laeubi:
> Some parts of the code could be simplified using present-day Java-11 methods. But I think this should be done in a subsequent PR, so that the initial move to Eclipse has as few changes as possible and we get a more clear git-history.

Yep exactly main goal is to get the code in even though it seems we would have been faster If I have rewritten it from scratch as it seem no one of the PMC likes to approve the CQ :-(

HannesWell:
> I consider none of the code in `m2e.pde.*` as public API beside that is very unlikely that anyone is already enhancing that stuff right now :-)
> 
> Don't know if one could mark a bundle as "internal" but I really don't like these eclipse extension as they are extremely confusing and against OSGi principles.

Are you referring to the package name element *.internal.* or the Manifest header-arguments `;x-internal/friend` in the Manifest?
Personally I also don't like the internal-package names but at least it is a clear statement.

However I have noticed that this plug-in does not export its only package so it is not public (regardless of the name), so sorry for the disturbance in this regard. To keep the off-scope discussion about o.e.m2e.pde out of this PR I have created https://github.com/eclipse-m2e/m2e-core/pull/454, we should continue there if there is more to say.

HannesWell:
> > There is just an error markers in the Dependencies-Tab of the Manifest-Editor, even tough the plug-in/package from the Maven-project was proposed by the Add... dialogues (for Require-Bundle and Import-Package). So I think this is an PDE issue.
> 
> Will you take a look at this?
> 
I can, but at the moment I have too many other things on my list (in my mind), that I would like to do before. So if this flaw persists and nobody else takes care of it, I can do it later :)


> > Some parts of the code could be simplified using present-day Java-11 methods. But I think this should be done in a subsequent PR, so that the initial move to Eclipse has as few changes as possible and we get a more clear git-history.
> 
> Yep exactly main goal is to get the code in even though it seems we would have been faster If I have rewritten it from scratch as it seem no one of the PMC likes to approve the CQ :-(

Aren't @akurtakov and @vogella in the Eclipse PMC? Could you be so kind and have a look at this or refer to somebody else from PMC?

org.eclipse.m2e.pde.connector/build.properties
HannesWell: Thanks.
This is a problem I occasionally notice with my colleges (and sometimes even myself 😅). Maybe there should be some mechanism in Eclipse PDE/Equinox(probably from PDE to Equinox since build.properties are from PDE) that checks if a resource is also included in the bin.includes from the build.properties.

HannesWell (approved):

org.eclipse.m2e.pde.connector/build.properties
laeubi: The problem is that this is explicitly to choose what to include and what not, so a warning or something like that has to be provided by m2e scanning for "known" extension points then check if required resources (e.g. lifecycle-mapping-metadata.xml) are present and included. the OSGI-DS plugin does this already but its always a manual step to decide if a resource has to be included or not.

mickaelistria:
> Aren't @akurtakov and @vogella in the Eclipse PMC? Could you be so kind and have a look at this or refer to somebody else from PMC?

m2e is a project under technology. PMC members for Technology are listed at https://projects.eclipse.org/projects/technology/who .

HannesWell:
> m2e is a project under technology. PMC members for Technology are listed at https://projects.eclipse.org/projects/technology/who .

Thanks Mickael for that hint, I already wondered if there is more than one PMC, because I have the impression that the term Project is used redundantly. 

So could @guw, @caniszczyk or @waynebeaton be so kind and have a look at the associated CQ at
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23882

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/Activator.java
```diff
@@ -0,0 +1,44 @@
...
+package org.eclipse.m2e.pde.connector;
+
+import java.util.Optional;
+
+import org.eclipse.pde.core.project.IBundleProjectService;
+import org.osgi.framework.BundleActivator;
+import org.osgi.framework.BundleContext;
+import org.osgi.util.tracker.ServiceTracker;
+
+public class Activator implements BundleActivator {
```
vogella, line 22: I suggest to use the FrameworkUtil method to access OSGI services instead of an activator. Easier to read and avoids code.

laeubi: I don't see anything `org.osgi.framework.FrameworkUtil` can do here... If you mean `org.eclipse.core.runtime.ServiceCaller<Service>` this class introduces bad OSGi practice and I won't suggest using it anywhere at all...

vogella: FrameworkUtil.getBundle(this.getClass()).getBundleContext() can be used to access the bundle context and afterwards the service. via it.

vogella: In platform code we typically do not hold on to OSGi services but use util classes to access it dynanically to the listener is not needed. But having this activator is also not a really bad thing, if you prefer this pattern that is also fine of course. Activators is something which tend to grow over time and if large can slow down the startup of the application hence we try to avoid using them in platform code.

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/PDEMavenBundlePluginConfigurator.java
```diff
@@ -0,0 +1,174 @@
...
+		return false;
+	}
+
+	@Override
+	public AbstractBuildParticipant getBuildParticipant(IMavenProjectFacade projectFacade, MojoExecution execution,
+			IPluginExecutionMetadata executionMetadata) {
+		if ((isFelix(execution.getPlugin()) && isFelixBundleGoal(execution))
+				|| (isBND(execution.getPlugin()) && isBNDBundleGoal(execution))) {
+			return new MojoExecutionBuildParticipant(execution, true, true);
+		}
```
HannesWell, line 171:
```suggestion
		if ((isFelix(execution.getPlugin()) && isFelixManifestGoal(execution))) {
			return new MojoExecutionBuildParticipant(execution, true, true) {
				@Override
				public Set<IProject> build(int kind, IProgressMonitor monitor) throws Exception {
					SubMonitor subMonitor = SubMonitor.convert(monitor, 10);
					if (kind == FULL_BUILD || kind == CLEAN_BUILD) {
						deleteFelixManifestIncrementalData(subMonitor.split(1));
					} else {
						subMonitor.setWorkRemaining(9);
					}
					return super.build(kind, subMonitor.split(8));
					// The Manifest Mojo refreshes the IFile corresponding to the generated
					// MANIFEST.MF by using the AbstractEclipseBuildContext to write it.
					// -> No need to refresh the MANIFEST.MF-IFile 'manually'
				}

				@Override
				public void clean(IProgressMonitor monitor) throws CoreException {
					deleteFelixManifestIncrementalData(monitor);
				}

				private void deleteFelixManifestIncrementalData(IProgressMonitor monitor) throws CoreException {
					SubMonitor subMonitor = SubMonitor.convert(monitor, "Delete Felix-Manifest mojo data", 3);
					IMavenProjectFacade projectFacade = getMavenProjectFacade();
					File buildDir = new File(projectFacade.getMavenProject().getBuild().getDirectory());
					IResource targetFolder = MavenProjectUtils.getProjectResource(projectFacade.getProject(), buildDir);

					IFolder felixDataFolder = ((IFolder) targetFolder).getFolder("maven-bundle-plugin");
					if (!felixDataFolder.exists()) {
						felixDataFolder.refreshLocal(IResource.DEPTH_ZERO, subMonitor.split(1));
					}
					felixDataFolder.delete(true, false, subMonitor.split(2));
				}
			};
		} else if (isBND(execution.getPlugin()) && isBNDBundleGoal(execution)) {
			return new MojoExecutionBuildParticipant(execution, true, true);
		}
```

HannesWell: This uses the new method `MavenProjectUtils.getProjectResource()` introduce in PR #470.

HannesWell (changes requested):
I tested this PR with the pure Maven m2e's maven-runtime projects as proposed in PR #466 and found one problem with clean and full builds:

In the beginning of a clean/full build the `target/classes` folder is purged and with it the META-INF/MANIFEST.MF file located within it. At the same time `target/maven-bundle-plugin/org.apache.felix_maven-bundle-plugin_manifest_xx` is not touched. Consequently the `manifest` Mojo does not re-generate the MANIFEST.MF in the build when [supportIncrementalBuild](https://felix.apache.org/documentation/_attachments/components/bundle-plugin/manifest-mojo.html#supportIncrementalBuild) is enabled, because the mojo only checks `org.apache.felix_maven-bundle-plugin_manifest_xx` and not if the MANIFEST.MF is present.
To workaround this, `org.apache.felix_maven-bundle-plugin_manifest_xx` has to be  `target/maven-bundle-plugin` in each clean/full build before the `manifest` Mojo is executed.

laeubi:
> the `manifest` Mojo does not re-generate the MANIFEST.MF in the build when [supportIncrementalBuild](https://felix.apache.org/documentation/_attachments/components/bundle-plugin/manifest-mojo.html#supportIncrementalBuild) is enabled, because the mojo only checks `org.apache.felix_maven-bundle-plugin_manifest_xx` and not if the MANIFEST.MF is present.

I think this is a bug in the Felix Bundle plugin, if you agree, would you mind to open a ticket at felix (and maybe provide a PR to fix this)?

HannesWell:
> > the `manifest` Mojo does not re-generate the MANIFEST.MF in the build when [supportIncrementalBuild](https://felix.apache.org/documentation/_attachments/components/bundle-plugin/manifest-mojo.html#supportIncrementalBuild) is enabled, because the mojo only checks `org.apache.felix_maven-bundle-plugin_manifest_xx` and not if the MANIFEST.MF is present.
> 
> I think this is a bug in the Felix Bundle plugin, if you agree, would you mind to open a ticket at felix (and maybe provide a PR to fix this)?

Fully agree that this is a bug in the Felix `maven-bundle-plugin`. I can create an issue and provide a PR (I wanted to report/fix some other minor issues anyway, so now I have one more reason).
But if if this is fixed only future versions of the `maven-bundle-plugin` will work correctly.
It's probably discouraged to provide a workaround for old versions, but to make users aware of it we could emit a warning when a flawed version is used.

laeubi:
> But if if this is fixed only future versions of the `maven-bundle-plugin` will work correctly.

I won't mind to only support latest version given that this was flawed before and we have only limited dev-resources. I even don't see a reason to use old versions of the bundle-plugin... We should also not complicate the code with to much else/if/workaround/... and simply tell people when complain to use always the latest version of the plugin.

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/PDEMavenBundlePluginConfigurator.java
```diff
@@ -0,0 +1,174 @@
...
+										: Arrays.toString(BND_MANIFEST_GOALS))
+								+ ".",
+						IMarker.SEVERITY_WARNING, location);
+				problems.add(problem);
+				break;
+			}
+		}
+		if (problems.size() > 0) {
+			this.markerManager.addErrorMarkers(request.getPom(), IMavenConstants.MARKER_LIFECYCLEMAPPING_ID, problems);
+		}
```
HannesWell, line 102: This code does not consider that the resource of a problem could be the parent pom (or grand-parent etc.).
The following suggestion does consider this.
```suggestion
		if (!problems.isEmpty()) {
			IMavenProjectRegistry projectManager = MavenPlugin.getMavenProjectRegistry();
			for (MavenProblemInfo problem : problems) {
				String[] gav = problem.getLocation().getResourceId().split(":");
				IFile pom = projectManager.getMavenProject(gav[0], gav[1], gav[2]).getPom();
				markerManager.addErrorMarker(pom, IMavenConstants.MARKER_LIFECYCLEMAPPING_ID, problem);
			}
		}
```

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/PDEMavenBundlePluginConfigurator.java
```diff
@@ -0,0 +1,174 @@
...
+							FELIX_PARAM_SUPPORTINCREMENTALBUILD, Boolean.class, monitor);
+					if (supportIncremental == null || !supportIncremental.booleanValue()) {
+						SourceLocation location = SourceLocationHelper.findLocation(execution.getPlugin(),
+								"configuration");
+						MavenProblemInfo problem = new MavenProblemInfo(
+								"Incremental updates are currently disabled, set supportIncrementalBuild=true to support automatic manifest updates for this project.",
+								IMarker.SEVERITY_WARNING, location);
+						problems.add(problem);
+					}
+					hasManifestExecution = true;
```
HannesWell, line 79: The following code would give the user a warning to use the latest Maven-Bundle-Plugin (assuming the mentioned fixes will be integrated there). Maybe the message could be better.
```suggestion
					MavenVersion version = MavenVersion.parseMavenString(execution.getPlugin().getVersion());
					if (version != null && version.toReleaseVersion().compareTo(new MavenVersion("5.1.4")) < 0) {
						SourceLocation location = SourceLocationHelper.findLocation(execution.getPlugin(), "version");
						problems.add(new MavenProblemInfo(
								"The Maven-Bundle-Plugin only integrates well with Eclipse-M2E with version 5.4.0 or later",
								IMarker.SEVERITY_WARNING, location));
					}
					
					hasManifestExecution = true;
```

HannesWell (changes requested):
> > the `manifest` Mojo does not re-generate the MANIFEST.MF in the build when [supportIncrementalBuild](https://felix.apache.org/documentation/_attachments/components/bundle-plugin/manifest-mojo.html#supportIncrementalBuild) is enabled, because the mojo only checks `org.apache.felix_maven-bundle-plugin_manifest_xx` and not if the MANIFEST.MF is present.
> 
> I think this is a bug in the Felix Bundle plugin, if you agree, would you mind to open a ticket at felix (and maybe provide a PR to fix this)?

I opened issue [Apache FELIX-6493](https://issues.apache.org/jira/browse/FELIX-6493) and submitted PR https://github.com/apache/felix-dev/pull/124. Fell free to review it.

With those fixes the suggested workaround is not necessary any more and the manifest generation works well and even the error markers at the Maven-OSGi-Projects in the Manifest-Editor dependencies tab respectively the Target-Platform-State view magically vanished (I'm not sure why exactly and if this is permanently).

> > But if if this is fixed only future versions of the `maven-bundle-plugin` will work correctly.
> 
> I won't mind to only support latest version given that this was flawed before and we have only limited dev-resources. I even don't see a reason to use old versions of the bundle-plugin... We should also not complicate the code with to much else/if/workaround/... and simply tell people when complain to use always the latest version of the plugin.

I agree that we should not make it too complicated and a workaround is not necessary, but I think a warning would make the user experience much better. Especially because it could take some time until the new maven-bundle-plugin is available and is widely adopted. I think it would be a pity if this great feature would not work in the beginning and users first have to search the internet for a solution while we could simply give them a hint with a corresponding warning. :)

laeubi:
> I agree that we should not make it too complicated and a workaround is not necessary, but I think a warning would make the user experience much better. Especially because it could take some time until the new maven-bundle-plugin is available and is widely adopted. I think it would be a pity if this great feature would not work in the beginning and users first have to search the internet for a solution while we could simply give them a hint with a corresponding warning. :)

I think we can at least add a warning marker for older version and suggest an upgrade. But for such improvements I really want to wait until EMO approval of the code.

mickaelistria:
> I really want to wait until EMO approval of the code.

That's now good to merge from a legal POV according to recent comment on the CQ.

laeubi:
@HannesWell I'll merge this as-is now, so the code is in the repository can you create separate PR(s) for your enhancements?

HannesWell (approved):
> @HannesWell I'll merge this as-is now, so the code is in the repository can you create separate PR(s) for your enhancements?

Yes, please go ahead. I'll create a PR afterwards.

Issue #354, opened 2021-10-07
Multiple SLF4J bindings


kwin:
Eclipse comes with a default binding in Eclipse.app/Contents/Eclipse/configuration/org.eclipse.osgi/8/0/.cp/org/slf4j/impl/StaticLoggerBinder.class and m2e contains one in https://github.com/eclipse-m2e/m2e-core/blob/master/m2e-maven-runtime/org.eclipse.m2e.maven.runtime.slf4j.simple/pom.xml as well. So whenever you install "M2E - SLF4J over Logback Logging" you get this warning in the Maven Console.

```
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:...plugins/org.eclipse.m2e.maven.runtime.slf4j.simple_1.18.1.20211004-1243/jars/slf4j-simple-1.7.30.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [file:/.../Eclipse.app/Contents/Eclipse/configuration/org.eclipse.osgi/8/0/.cp/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.slf4j.impl.SimpleLoggerFactory]
```

That issue was already reported in https://bugs.eclipse.org/bugs/show_bug.cgi?id=506676.

kwin:
This affects some standard distributions e.g. "Eclipse IDE for Enterprise Java and Web Developers (2021-09)" (https://git.eclipse.org/c/oomph/org.eclipse.oomph.git/tree/setups/org.eclipse.products.setup#n1986).

mickaelistria:
@HannesWell Is this something you want in 1.18.2 ? Should we wait for it?

PR #446, opened 2021-12-11
[UITests] eliminate race-condition and rework ConsoleTest (#341)

HannesWell:
This PR aims to resolves issue https://github.com/eclipse-m2e/m2e-core/issues/341.

Previously test case did not ensure that the full console print-out written by the Maven-build process was transferred to the Maven-Processe's console-IDocument. The write to the console-IDocument happens in the UI-Thread while the test case is executed and reads from another (worker)-thread.
This lead to the suspected race-condition, because without synchronization it happened that, after the process had termianted, the test-thread read the console-document before the UI-thread wrote the process output to it.

Additionally this PR adds a check for a 'multiple SLF4J bindings' warning to guard against https://github.com/eclipse-m2e/m2e-core/issues/354.


HannesWell:
The failure in the Jenkins build is in `org.eclipse.m2e.editor.xml.WarnIfGroupSameThanParentTest.testMNGEclipse2552` and unrelated.

PR #448, opened 2021-12-11
[Tests] update 'm2e-core-tests' sub-module

HannesWell:

Issue #428, opened 2021-12-07
POM editor slow to recover from alleged errors during typing


2019-05-10:
Since update to 1.19 POM editor is very quick to "detect" and mark an error while typing, but exceedingly slow to recover from it once I've done typing.

Eclipse:
Version: 2021-09 (4.21.0)
Build id: 20210910-1417

M2E:
Eclipse.org - m2e	M2E Maven Integration for Eclipse Core	1.19.0.20211116-0804	org.eclipse.m2e.feature

PR #451, opened 2021-12-13
Fix 419 - Support Workspace resolution of artifacts

laeubi:
This adds basic support for referencing artifacts that are part of the workspace.

### DIFF
--- a/org.eclipse.m2e.pde/src/org/eclipse/m2e/pde/DependencyNodeGenerator.java
+++ b/org.eclipse.m2e.pde/src/org/eclipse/m2e/pde/DependencyNodeGenerator.java
@@ -1,0 +1,85 @@
+/*******************************************************************************
+ * Copyright (c) 2018, 2021 Christoph Läubrich
+ * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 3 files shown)


laeubi:
@HannesWell @mickaelistria 

There seems to be another unstable test now:

https://ci.eclipse.org/m2e/job/m2e/job/PR-451/2/testReport/org.eclipse.m2e.editor.xml/WarnIfGroupSameThanParentTest/testMNGEclipse2552/

could this be to the new default lemix editor being a bit slow? Maybe

- https://github.com/eclipse-m2e/m2e-core/issues/428

HannesWell:
I'm not sure why this happens, but I noticed this already in PR #446 and therefore already updated the m2e-core-tests submodule to the current master (#448). With they were at least successful when I run them locally in my Eclipse.
But on the master-branch build they did not appear (yet). So your suspicion could be right.

laeubi:
Okay I'll then merge this and we'll see later then. Maybe a similar fix as with the console would be appropriate to use a matcher/wait for some time until the expected message appears.

PR #453, opened 2021-12-14
[Build] remove forge-parent parent and clean up

HannesWell:
The currently specified `forge-parent` of m2e's root pom.xml is heavily outdated (10 years old), `forge-parent` is not maintained and its usage is generally not recommended any more (I can't find the reference at the moment). To make it short, we should remove it.

Additionally this PR contains some marginal clean-ups.

### DIFF
--- a/m2e-maven-runtime/pom.xml
+++ b/m2e-maven-runtime/pom.xml
@@ -26,7 +26,6 @@
 	<properties>
 		<dependency.jars.folder>jars</dependency.jars.folder>
 		<dependency.sources.folder>jars/sources</dependency.sources.folder>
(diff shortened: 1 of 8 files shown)


akurtakov:
Really welcome change. I've planned to do that some day when todo list gets under 100 items.

laeubi:
Finally https://www.eclipse.org/lists/tycho-dev/msg00269.html :-)

HannesWell:
> Really welcome change. I've planned to do that some day when todo list gets under 100 items.

Thanks. My experience from my few past years is that this will never happen, at least my list becomes longer and longe.

PR #454, opened 2021-12-14
[PDE] mark package org.eclipse.m2e.pde  as internal

HannesWell:
Because the only (exported) package of the `org.eclipse.m2e.pde `is not considered public API (as for example stated in https://github.com/eclipse-m2e/m2e-core/pull/440#issuecomment-993306343) we should mark the package as such to make users aware of that fact.
Otherwise downstream-adaptors could accidentally use the classes from that package like public API, unless they have searched in the depths of the m2e-github repo.

Since org.eclipse.m2e.pde has been released in this form already _actually_ it is already public API, but since this happened not so long ago and adaptors are unlikely (or if it happened there is a high chance that they are still active and can adapt accordingly or complain), I propose to make an exception and mark this package as internal afterwards.

@laeubi, @mickaelistria what do you think?

### DIFF
--- a/org.eclipse.m2e.pde/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.pde/META-INF/MANIFEST.MF
@@ -12,7 +12,7 @@ Require-Bundle: org.eclipse.core.runtime;bundle-version="3.19.0",
  org.eclipse.m2e.maven.runtime,
  org.eclipse.m2e.core,
  org.eclipse.core.resources
-Export-Package: org.eclipse.m2e.pde
+Export-Package: org.eclipse.m2e.pde;x-friends:="org.eclipse.m2e.pde.ui"
 Bundle-Activator: org.eclipse.m2e.pde.Activator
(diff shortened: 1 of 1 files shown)


laeubi:
If it makes the world better lets merge it :-)

> Otherwise downstream-adaptors could accidentally use the classes from that package like public API,

A matter of fact is that people will still use these if they think its useful and simply suppress the warning (as we do in m2e, tycho, ...)

laeubi (approved):

HannesWell:
> If it makes the world better lets merge it :-)
> 
> > Otherwise downstream-adaptors could accidentally use the classes from that package like public API,
> 
> A matter of fact is that people will still use these if they think its useful and simply suppress the warning (as we do in m2e, tycho, ...)

A little bit better, I think :-)
Of course it does not prevent external usage (I admit that I also sometimes use internal API if there is no public equivalent), but at least we tried our best and made down-stream users aware of it, so they (hopefully) know the implications. Unfortunately world is not perfect.

Issue #457, opened 2021-12-16
PDE integration requires the second page of the wizard (feature creation) even if the flag "Generate feature" is not selected.


vogella:
PDE integration requires the second page of the wizard (feature creation) even if the flag "Generate feature" is not selected.

![image](https://user-images.githubusercontent.com/139910/146330402-9bccfd35-90be-4e18-936f-d5c8af107bf7.png)

laeubi:
Can you check if it help to check/uncheck the box (might be an update issue?) and probably provide a patch?

vogella:
> Can you check if it help to check/uncheck the box (might be an update issue?) 

No it does not help.

laeubi:
The problem is only when adding a new item. Here PDE do not actually uses the wizard that is supplied by the extension but only the pages see `org.eclipse.pde.internal.ui.shared.target.AddBundleContainerSelectionPage.createTargetLocationProvisionerNode(...).new AbstractBundleContainerNode() {...}.createWizard()`

I'll have to take a look if I can fix this...

laeubi:
See

https://bugs.eclipse.org/bugs/show_bug.cgi?id=577861
https://git.eclipse.org/r/c/pde/eclipse.pde.ui/+/188942

vogella:
Thanks, Christoph.

PR #460, opened 2021-12-17
Fix #459 Distinguish between dependency resolution and scope

### LINKED ISSUE #459: Distinguish between dependency resolution and scope
### Discussed in https://github.com/eclipse-m2e/m2e-core/discussions/456

<div type='discussions-op-text'>

<sup>Originally posted by **laeubi** December 16, 2021</sup>
Currently the dependency scope is used in the target location as follows:

- if none (or empty) is give no dependencies are resolved
- if given, dependencies are resolved and filtered for the given scope (at least in the latest version)
- if a dependency is a pom dependency and no scope is given only direct dependencies are consider otherwise all  dependencies

While implementing that feature in tycho and trying to get both in sync I noticed that this is maybe not enough and it was at other places noticed by @HannesWell that it might not fit the expectation or confuse users.

I therefore like to propose the following adjustments that would better fit the maven model and the actual dependency resolution process:

- the scope is only used to filter the (possibly empty) list of dependencies, if it is empty or not given it is assumed to be compile scope (that's how maven actually handle it)
- add a new `includeDependencyDepth` that could take the value `NONE` (no dependencies included), `DIRECT` (only direct dependency child's are included), `INFINITE` (all transitive dependencies are included)
- for pom dependencies `NONE` is consider the same as `DIRECT`

The good thing is that we can simply assume that if `includeDependencyDepth` is missing and `includeDependencyScope` is specified we can assume `INFINITE` and thus would be able to be backward compatible to the existing format with auto-upgrading from the previous format.

Comments are welcome, if there are no concerns I'd like to get this feature into the upcoming 2.6.x release of tycho and prepare a patch for m2e as well.</div>

laeubi:
As discussed ...


org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/messages.properties
```diff
@@ -18,9 +18,11 @@ MavenTargetLocationWizard_12=Artifacts are resolved against the workspace <a hre
 MavenTargetLocationWizard_14=Label (optional)
 MavenTargetLocationWizard_9=Missing OSGi-Manifest
 MavenTargetLocationWizard_10=Dependencies scope
```
HannesWell, line 20: What do you think about "Include dependencies of scope"  instead?

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/messages.properties
```diff
@@ -18,9 +18,11 @@ MavenTargetLocationWizard_12=Artifacts are resolved against the workspace <a hre
...
+MavenTargetLocationWizard_17=Dependencies depth
```
HannesWell, line 21: What do you think about "Include dependencies for depth"  instead?
This proposal is IMHO not yet ideal but would be fit into the schema of my other proposal. Maybe you have a better one.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/MavenTargetLocationWizard.java
```diff
@@ -90,6 +92,9 @@ public MavenTargetLocationWizard(MavenTargetLocation targetLocation) {
 				targetLocation == null ? Messages.MavenTargetLocationWizard_1 : Messages.MavenTargetLocationWizard_2) {
 
 			private Link editInstructionsButton;
+			private Label scopeDescriptionLabel;
+
+			private Label includeLabel;
```
HannesWell, line 97: Shouldn't this be named `depthDescriptionLabel`?

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/MavenTargetLocationWizard.java
```diff
@@ -165,17 +179,41 @@ private Button createCheckBox(Composite composite, String text) {
...
 				layout.marginWidth = 0;
 				layout.marginHeight = 0;
 				composite.setLayout(layout);
-				scope = combo(new CCombo(composite, SWT.BORDER));
-				scopeLabel = new Label(composite, SWT.NONE);
-				scopeLabel.setText(Messages.MavenTargetLocationWizard_15);
-				scope.add(""); //$NON-NLS-1$
+				scope = combo(new CCombo(composite, SWT.READ_ONLY | SWT.BORDER | SWT.FLAT));
+				scopeDescriptionLabel = new Label(composite, SWT.NONE);
+				scopeDescriptionLabel.setText(Messages.MavenTargetLocationWizard_15);
```
HannesWell, line 216: This could be a tool-tip, too.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/MavenTargetLocationWizard.java
```diff
@@ -165,17 +179,41 @@ private Button createCheckBox(Composite composite, String text) {
...
+			private void createIncludeCombo(Composite parent) {
+				Composite composite = new Composite(parent, SWT.NONE);
+				GridLayout layout = new GridLayout(2, false);
+				layout.horizontalSpacing = 20;
+				layout.marginWidth = 0;
+				layout.marginHeight = 0;
+				composite.setLayout(layout);
+				include = new ComboViewer(combo(new CCombo(composite, SWT.READ_ONLY | SWT.BORDER | SWT.FLAT)));
+				includeLabel = new Label(composite, SWT.NONE);
+				includeLabel.setText(Messages.MavenTargetLocationWizard_16);
```
HannesWell, line 191: This could be a tool-tip instead.
At the moment the wizard appears a bit full.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/MavenTargetLocationWizard.java
```diff
@@ -261,21 +307,21 @@ private CCombo combo(CCombo combo) {
 	@Override
 	public boolean canFinish() {
 		if (isCreateFeature()) {
-			return page.isPageComplete();
+			return super.canFinish();
```
HannesWell, line 310: Is this and the changes in the following lines related to the inclusion-improvements or is it issue #457?

HannesWell (commented):
My first comments/ enhancement proposals regarding the UI.
I could not yet look into the enhanced Dependency-Inclusion logic yet and and I cannot say if I can do it before Monday, but I will do it latest on Monday.
I hope it is not urgent for you and sorry for the delay.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/messages.properties
laeubi: Maybe but the labels shouldn't be too long ...

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/messages.properties
laeubi: Same here ...

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/MavenTargetLocationWizard.java
laeubi: Maybe yes, anyways as it is never presented to the users its just up to us to choose the name, we still can refactor this later.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/MavenTargetLocationWizard.java
laeubi: For sure it would be cool if someone would take the time and effort to raft a super-cool styled wizard for this, but at least it fulfills the needs, I'd rather try to invest the time in adding the functional bits here, so sorry for no eye-candy :-D

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/editor/MavenTargetLocationWizard.java
laeubi: Actually the `isCreateFeature()` was inverse logic (returned false if a feature has to be created) as it has confused me I refactored it along with this change.

laeubi:
I'll just merge this right away as we need a corresponding change in Tycho as well, I think we can improve easily later on.

PR #467, opened 2021-12-28
[Build] improve snapshot deployment (remove <version>/latest and clear)

HannesWell:
With this PR I propose to
1. remove the 'latest' subfolder for the snapshot repo with a specific version to align better with snapshots/latest and releases.
The current `https://download.eclipse.org/technology/m2e/snapshots/1.19.1/latest/` would become `https://download.eclipse.org/technology/m2e/snapshots/1.19.1/`. Because there are no subversions or similar, the latest-subfolder has no advantage.

2. completely clear `download.e.o/tech/m2e/snapshots` before the snapshot-repo of a current master build is deployed.
 This way an old version repo is removed automatically once a new one is available.

3. inline only once used variables in the Jenkins file

@mickaelistria check you please check that the syntax is correct (I'm not a master of the linux shell).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -41,18 +41,16 @@ pipeline {
 						deployM2ERepository()
 						{
 							echo Deploy m2e repo to ${1}
-							ssh genie.m2e@projects-storage.eclipse.org "\
-								rm -rf  ${1}/* && \
-								mkdir -p ${1}"
+							ssh genie.m2e@projects-storage.eclipse.org "mkdir -p ${1}"
 							scp -r org.eclipse.m2e.site/target/repository/* genie.m2e@projects-storage.eclipse.org:${1}
 						}
 						M2E_VERSION=$(<"org.eclipse.m2e.sdk.feature/target/m2e.version")
-						SNAPSHOT_VERSIONED_AREA=/home/data/httpd/download.eclipse.org/technology/m2e/snapshots/${M2E_VERSION}/latest
-						SNAPSHOT_LATEST_AREA=/home/data/httpd/download.eclipse.org/technology/m2e/snapshots/latest
-
 						echo M2E_VERSION=$M2E_VERSION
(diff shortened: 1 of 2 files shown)


mickaelistria:
I have no objection here, however shouldn't the docs be updated as well? And also, what about even dropping the version part for snapshots?

HannesWell:
> I have no objection here, however shouldn't the docs be updated as well?

I could not find those versioned-snapshot repos mentioned anywhere but I agree they should be.

> And also, what about even dropping the version part for snapshots?

I thought about that as well, but I think it is handy for those that urgently want a fix for a problem or a new feature (or so) and therefore want to use the current snapshots but in general want to use the releases.
Those users (that likely have `m2e/releases/latest` repo as available updatesite) can add the version-specific snapshot repo and get the snapshots until the next release.
Since that version-specific snapshot repo is deleted with the release they "fallback" to the releases repo and from now on get only the regular release updates. They likely get an error message that that version-specific repo is not available anymore (and likely will remove it) but they wont stay at the snapshot repo accidentally forever.

HannesWell:
The [M2E's Jenkins Job `promote-snapshots-to-release `](https://ci.eclipse.org/m2e/job/promote-snapshots-to-release/) has to be updated accordingly since it promotes the version-specific repo.

HannesWell:
I'm going to merge this now and will update the `promote-snapshots-to-release` job accordingly.
If we found the version-specific snapshot repos to be unnecessary we can still delete them later.

Furthermore I'm going to delete [M2E's Jenkins project `download-cleanup`](https://ci.eclipse.org/m2e/job/download-cleanup/) since it is obsolete. The downloads are cleaned up now and can be cleaned up via the web-interface in the future by moving artifacts-to-delete to archive and deleting them from there.

HannesWell:
New deployment worked well. Old snapshot repos were deleted automatically and the current version specific one is at:
https://download.eclipse.org/technology/m2e/snapshots/1.19.2/

PR #466, opened 2021-12-24
[Maven-Runtime] convert projects into standard Maven-projects (#423)

HannesWell:
This PR has the goal to convert M2E's Maven runtime bundles (back) into full standard Maven projects that don't require special tooling/build as suggested in and to resolve #423.

To check for version regressions and to generate Eclipse-style build qualifiers the corresponding Tycho plug-ins are still used in the standard Maven build. This requires plug-in definitions like in the m2e-core/pom.xml, but those definitions should not be duplicated. At the same time `m2e-maven-runtime/pom.xml` cannot inherit from m2e's root pom.xml any more (because that one depends on the m2e-maven-runtime components) I introduced a m2e-parent Maven-project whose pom.xml contains the configuration shared by m2e-core/pom.xml and `m2e-maven-runtime/pom.xml`. Besides the build configuration also the legal elements. If you agree that this is a good idea, the m2e-parent could be introduced in a separate preceding PR?

As soon as #439 is merged the `.settings/org.eclipse.pde.core.prefs` and `.project` files can be completely removed.

With the current state of this PR the build is successful but in the IDE, the Maven-projects seem to not be considered by PDE as actually expected (and I already successfully tested with a small project). The m2e-maven-runtime bundles are always marked as missing in the Manifest's of other m2e plug-ins that require them.
@laeubi if you don't have an immediate idea why this is the case, I will investigate it in the next days.

PR #468, opened 2021-12-28
[Build] generate source-features by Tycho and include them into m2e-SDK

HannesWell:
This PR has the goal to generate source features for each m2e-feature using the `tycho-source:feature-source` Mojo and to include them in the m2e-sdk as suggested in PR #408.

This simplifies the aggregation of m2e-sources because they only have to be included per feature in the m2e-sdk feature (which is published in the m2e-repo) and the individually listed source plug-ins can be removed.
Listing the source-plug-ins can be forgotten easily (I did it when upgrading to `biz.aQute.bndlib`-6.0.0 which added `biz.aQute.bnd.util`, whose sources are missing a.t.m. and in the currently pending PRs that add new plug-ins the corresponding source entry is missing in the sdk too).

To make the `tycho-source-plugin:feature-source` mojo work, it was necessary to execute the `tycho-p2-plugin:p2-metadata` Mojo. It looks like that only the `p2-metadata` mojo adds the source-artifact to the Maven-Project during build. In order to not break local builds (with profile `eclipse-sign` disabled) the `p2-metadata` mojo has to run in every build and not only when `eclipse-sign` is activated.

Furthermore the source-plugins generated for the m2e-maven-runtime bundles like `org.eclipse.m2e.maven.runtime` did not work. This is probably because the embedded source jars are nested in the corresponding source plug-in and it looks like PDE cannot handle this at all. At least this is my impression when looking at `org.eclipse.pde.internal.core.ClasspathUtilCore.getSourceAnnotation()`. Therefore I completely excluded the corresponding source bundles for now.
@laeubi or @mickaelistria do you know a way how to assemble a source-plug-in for the nested jars of a plug-in?
In general I think down-stream users would find it very helpful to have sources for the Maven-classes etc.

### DIFF
--- a/m2e-maven-runtime/org.eclipse.m2e.archetype.common/build.properties
+++ b/m2e-maven-runtime/org.eclipse.m2e.archetype.common/build.properties
@@ -4,4 +4,3 @@ bin.includes = META-INF/,\
                .,\
                jars/
 bin.excludes = jars/sources/
-src.includes = jars/sources/
--- a/m2e-maven-runtime/org.eclipse.m2e.maven.indexer/build.properties
+++ b/m2e-maven-runtime/org.eclipse.m2e.maven.indexer/build.properties
@@ -5,4 +5,3 @@ bin.includes = META-INF/,\
                .,\
                jars/
 bin.excludes = jars/sources/
-src.includes = jars/sources/
--- a/m2e-maven-runtime/org.eclipse.m2e.maven.runtime.slf4j.simple/build.properties
+++ b/m2e-maven-runtime/org.eclipse.m2e.maven.runtime.slf4j.simple/build.properties
@@ -4,4 +4,3 @@ bin.includes = META-INF/,\
                .,\
                jars/
 bin.excludes = jars/sources/
-src.includes = jars/sources/
--- a/m2e-maven-runtime/org.eclipse.m2e.maven.runtime/build.properties
+++ b/m2e-maven-runtime/org.eclipse.m2e.maven.runtime/build.properties
@@ -4,4 +4,3 @@ bin.includes = META-INF/,\
                .,\
                jars/
 bin.excludes = jars/sources/
-src.includes = jars/sources/
--- a/m2e-maven-runtime/pom.xml
+++ b/m2e-maven-runtime/pom.xml
@@ -89,11 +89,57 @@
 							.classpath
 							.settings/
 							.polyglot.*
+							build.properties
 						</jgit.ignore>
+					</configuration>
+				</plugin>
+				<plugin>
+					<groupId>org.apache.maven.plugins</groupId>
+					<artifactId>maven-dependency-plugin</artifactId>
+					<version>3.2.0</version>
+					<configuration>
+						<overWriteReleases>true</overWriteReleases>
+						<overWriteSnapshots>true</overWriteSnapshots>
+						<excludeScope>provided</excludeScope>
+						<includeScope>runtime</includeScope> <!-- only include runtime and compile time dependencies, not test or provided! -->
+					</configuration>
+				</plugin>
+				<plugin>
+					<groupId>org.eclipse.tycho</groupId>
+					<artifactId>tycho-source-plugin</artifactId>
+					<version>${tycho-version}</version>
+					<configuration>
+						<additionalFileSets>
+							<fileSet>
+								<directory>${project.build.directory}/${dependency.sources.folder}</directory>
+								<includes>
+									<include>**/*.java</include>
+								</includes>
+							</fileSet>
+						</additionalFileSets>
 					</configuration>
 				</plugin>
 			</plugins>
 		</pluginManagement>
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-dependency-plugin</artifactId>
+				<executions>
+					<execution>
+						<id>fetch-flat-dependency-sources</id> <!-- These are used to build source-plugins -->
+						<goals>
+							<goal>unpack-dependencies</goal>
+						</goals>
+						<phase>process-resources</phase>
+						<configuration>
+							<outputDirectory>${project.build.directory}/${dependency.sources.folder}</outputDirectory>
+							<classifier>sources</classifier>
+						</configuration>
+					</execution>
+				</executions>
+			</plugin>
+		</plugins>
 	</build>
 
 	<profiles>
@@ -139,13 +185,6 @@
 					<plugin>
 						<groupId>org.apache.maven.plugins</groupId>
 						<artifactId>maven-dependency-plugin</artifactId>
-						<version>3.2.0</version>
-						<configuration>
-							<overWriteReleases>true</overWriteReleases>
-							<overWriteSnapshots>true</overWriteSnapshots>
-							<excludeScope>provided</excludeScope>
-							<includeScope>runtime</includeScope><!-- only include runtime and compile time dependencies, not test or provided! -->
-						</configuration>
 						<executions>
 							<execution>
 								<id>fetch-dependencies</id>
@@ -158,7 +197,7 @@
 								</configuration>
 							</execution>
 							<execution>
-								<id>fetch-dependency-sources</id>
+								<id>fetch-dependency-sources</id> <!-- These are used in the workspace -->
 								<goals>
 									<goal>copy-dependencies</goal>
 								</goals>
--- a/m2e-maven-runtime/pom.xml
+++ b/m2e-maven-runtime/pom.xml
@@ -113,7 +113,7 @@
 							<fileSet>
 								<directory>${project.build.directory}/${dependency.sources.folder}</directory>
 								<includes>
-									<include>**/*.java</include>
+									<include>**/*</include>
 								</includes>
 							</fileSet>
 						</additionalFileSets>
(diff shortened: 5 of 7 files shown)


laeubi:
> This is probably because the embedded source jars are nested in the corresponding source plug-in and it looks like PDE cannot handle this at all.

I think best would be if PDE is be able to handle this case. Another option might be that we add support to tycho to "flatten" source jars when it generates the source bundle, or we add some maven magic to do this.

HannesWell:
> > This is probably because the embedded source jars are nested in the corresponding source plug-in and it looks like PDE cannot handle this at all.
> 
> I think best would be if PDE is be able to handle this case. Another option might be that we add support to tycho to "flatten" source jars when it generates the source bundle, or we add some maven magic to do this.

Direct support would indeed be the best (I could have a look at it later). But I think some cooperation from Tycho is still required. The plugin-source-mojo already has the [distinctSourceRoots property](the https://www.eclipse.org/tycho/sitedocs/tycho-source-plugin/plugin-source-mojo.html#distinctSourceRoots) that aims to generate corresponding `roots` property values for the `Eclipse-SourceBundle` header. But from looking at the PDE code I found only a few places where `roots` is considered but they seem unconnected to source lookup or even unused. What also has to be considered is the shape of the Bundle respectively the source-bundle. The maven-runtime bundles use `Eclipse-BundleShape: dir` but the source plug-in does not copy that setting (maybe it should do it?). On the other hand it is possible to have jar-bundles with nested jars, in the case Equinox extracts the nested jars on first start into the configuration area of that bundle.
So it is definitely possible but some work has to be done to make it happen.

I also thought about working it around and `flatten` the sources into the source bundle by some special Maven treatment. The `Maven-Dependency-Plug-in` has the [unpack-dependencies goal](https://maven.apache.org/plugins/maven-dependency-plugin/unpack-dependencies-mojo.html) which could do the job in conjunction with [plugin-source-mojo-includes](https://www.eclipse.org/tycho/sitedocs/tycho-source-plugin/plugin-source-mojo.html#includes)

laeubi:
SOurce bundles are only relevant for PDE so I don't think Equinox itself has much to do with this or the Bundle-Shape setting. Beside that source-support in PDE very poor (e.g. requiring the special header in the first place is a bit strange).

HannesWell:
> SOurce bundles are only relevant for PDE so I don't think Equinox itself has much to do with this or the Bundle-Shape setting. Beside that source-support in PDE very poor (e.g. requiring the special header in the first place is a bit strange).

Of course this should only be handled in PDE (and likely Tycho has to support it too).
I just wanted to mention that both cases, jar or directory (but is likely a jar in most cases), of a source-bundle shape have to be considered.
It is very poor in that sense that it only supports flat sources (or nested <libName>src.zip files, but I have never seen those so I suspect this is legacy style?).

However for now I have updated the Maven-build to unpack all dependency jars into a `target/jars/sources` and include all java-files from that folder into the source-bundle. This way we have all source files merged together and as long as there are not two classes with the same name and package in the dependencies there should be no collision.
All other files from the source-jars, like Manifests or licence files, are discarded. One reason is that they could overwrite files from the m2e source-bundle itself (like its Manifest), another reason is that they could collide with another dependencies. For example two dependencies could have a licence.txt. Since the license files are already contained in the actual bundle I assume this is fine from a legal perspective, isn't it?
Otherwise there had to be distinct folders for each dependency which AFAIK is not supported by PDE.

HannesWell:
@mickaelistria if there are no legal concerns I would squash-and-merge?

mickaelistria:
> If there are no legal concerns I would squash-and-merge?

What specifically do you have in mind that would require some legal/IP attention here?

HannesWell:
> > If there are no legal concerns I would squash-and-merge?
> 
> What specifically do you have in mind that would require some legal/IP attention here?

That the current approach only extracts the `.java` files from the dependencies source jars but discards all licence, pom.xml, read-me, notes.txt or similar files that are in the original _source_-jars. The binary jars on the other hand are just nested into the `o.e.m2e.maven.runtime` bundles (and friends) as they are and therefore still contain such legal files (if they are present).

I would expect that the legal files in the source jars are the same like in the binary jars, but I'm not sure about this.

If you think we should include those files too, as a workaround all non-java files could be placed in dedicated folder within the `o.e.m2e.maven.runtime` (and others) source bundle.

mickaelistria:
OK, thanks for the clarification. I don't think we're allowed to ship source code in any way without the information required by the upstream project and license. So the legal files seem necessary.

HannesWell:
Alright.
Do you think a workaround like I have described is ok?
The problem is that at the moment (as far as I can tell) PDE does not supports nested source jars in source bundles and only finds source files if the package folder structure is placed in the root of the source bundle.

For the java files this is not a problem to extract them flat into the source bundle root because of different package names, but legal files will collide (two jars likely contain licence or notes.txt and definitely contain a pom.xml).
But I can try to place those legal files into dedicated subfolder, named like the jar.

laeubi:
Could it work to extract each jar in its own subfolder? Actually OSGI has `OSGI-OPT/src/` folder where it recommends to put sources into.

laeubi:
by the way I think we could still go with the embedded jars, if PDE does not support them we better add support there for it.

We could also go for the `OSGI-OPT/src/` way *inside the bundle itself* even if this increases the size of the binary a bit.

mickaelistria:
> Could it work to extract each jar in its own subfolder? Actually OSGI has OSGI-OPT/src/ folder where it recommends to put sources into.

That was also my first thought. If this can work for PDE, then I think it's the best way forward.

laeubi:
> > Could it work to extract each jar in its own subfolder? Actually OSGI has OSGI-OPT/src/ folder where it recommends to put sources into.
> 
> That was also my first thought. If this can work for PDE, then I think it's the best way forward.

If the sources are inside the bundle its mor JDT to handle that case not PDE (pde is only responsible for attaching additional source items to existing classpath items).

HannesWell:
> by the way I think we could still go with the embedded jars, if PDE does not support them we better add support there for it.
> 
Of course that will be better. I actually wanted to go this step as an intermediate/workaround solution because I expect adding support nested jars in PDE to be not too easy. But I can have a look at it.
We should also consider the switch to pure Maven projects with PR #466, so it should work then as well.

> We could also go for the `OSGI-OPT/src/` way _inside the bundle itself_ even if this increases the size of the binary a bit.

Storing the sources in the bundle not only increases the size a bit, the source-jars have often almost the same size of the binary-jar (some are much smaller, some only a bit). Therefore I don't think it is a good approach because it would almost double the size of the binary bundles even if one does not want the sources. Furthermore it does not prevent adjusting PDE to support it. Even tough JDT is responsible for resolving sources, I think PDE has to tell JDT where to search the sources and ASFAIK JDT does not know OSGi, does it?

laeubi:
> Therefore I don't think it is a good approach because it would almost double the size of the binary bundles even if one does not want the sources

Most OS licenses force to provide the source even if the user don't want it ;-)

And we are not talking about megabytes here, most m2e bundles are just a few kilobytes.

> Even tough JDT is responsible for resolving sources, I think PDE has to tell JDT where to search the sources and ASFAIK JDT does not know OSGi, does it?

Actually JDT should be able to discover sources inside a jar just fine, currently PDE does *attatch* sources to a jar because they are not located inside.

mickaelistria:
I'm definitely missing something: is the plan to ship source or source jars in the same bundle as binary ones? I was expecting it was more a matter of creating some extra source bundles.

laeubi:
> I'm definitely missing something: is the plan to ship source or source jars in the same bundle as binary ones? I was expecting it was more a matter of creating some extra source bundles.

The current intention is/was to ship an automatic generated **source bundle**. But as PDE does not fully understand what we generate, I suggested to maybe better ship the source inside the **binary bundle** as an alternative and hope JDT is handling that case better, beside that "source-bundle" is an _Eclipse-only-term_ and shipping the source inside the bundle is [recommended by OSGi specification](http://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html):

> A bundle is a JAR file that:
> 
>    [...]
> 
>  Can contain optional documentation in the OSGI-OPT directory of the JAR file or one of its sub-directories. Any information in this directory is optional. **For example, the OSGI-OPT directory is useful to store the source code of a bundle**. Management systems **may remove this information to save storage space in the OSGi framework**.

HannesWell:
> > Therefore I don't think it is a good approach because it would almost double the size of the binary bundles even if one does not want the sources
> 
> Most OS licenses force to provide the source even if the user don't want it ;-)
> 
> And we are not talking about megabytes here, most m2e bundles are just a few kilobytes.
>

Actually we _do_ talk about megabytes for the maven-runtime bundles:
- org.eclipse.m2e.archetype.common.source ~ 2.1MB
- org.eclipse.m2e.maven.indexer.source ~ 4.3MB
- org.eclipse.m2e.maven.runtime.source ~ 4.6MB
- org.eclipse.m2e.maven.runtime.slf4j.simple.source 18kB (that's indeed small)

These are the numbers Windows explorer shows me when the source-jars are nested into the the `m2e.maven.xx` source jars as they are. For most other m2e plug-ins you are right.

Just to make the issue more clear:
The problem are only the source bundles for maven-runtime components `org.eclipse.m2e.archetype.common.source`, `org.eclipse.m2e.maven.indexer.source`, `org.eclipse.m2e.maven.runtime.source` and `org.eclipse.m2e.maven.runtime.slf4j.simple.source` where the binary bundle-jars contain embed jars and the corresponding source bundles should be processable by PDE/JDT too.

For all other m2e plug-ins the source bundles generated by Tycho just work fine. So the entire discussion is only about the maven-runtime components.

> > Even tough JDT is responsible for resolving sources, I think PDE has to tell JDT where to search the sources and ASFAIK JDT does not know OSGi, does it?
> 
> Actually JDT should be able to discover sources inside a jar just fine, currently PDE does _attatch_ sources to a jar because they are not located inside.

Exactly. But AFAIK JDT only discovers sources inside a jar if the package-name directory structure containing the sources is located in the root of the jar (e.g. `jar:file://org.eclipse.m2e.maven.runtime.source!/ org/apache/maven/Maven.java`). I have to admit I didn't check it thoroughly but I would be surprised if JDT would discover sources in `OSGI-OPT/src` without any support (by PDE).

HannesWell:
The latest commit adds the non-java resources in dedicated sub-folders per jar to the source bundles, but leaves the the package-folder structure in the root of the source-bundle.

This is not nice and cumbersome in the build but at least the source bundles generated this way work already without further changes requirements on the consuming Eclipse.
If you are fine with this for now I would squash and merge it now. Support for embedded source-jars can then be added later to PDE. Otherwise I have to put this on hold because at the moment I have other things with higher priority on my TODO-List.

HannesWell:
I would be in favor to merge this as it is now (although it is not perfectly nice for the maven-runtime components).
We can still improve the way the sources of embedded jars are included later, but then we have at least working solution for now.

laeubi (approved):

HannesWell:
Thanks for the discussion!

PR #478, opened 2022-01-04
Build with Tycho 2.6.0

HannesWell:
- [x] update extensions after release too


HannesWell:
Only test `org.eclipse.m2e.editor.xml.WarnIfGroupSameThanParentTest.testMNGEclipse2552` which happens occasionally (looks like we have a race condition) and is not new.

PR #479, opened 2022-01-05
[PDE-Connector] support warning in parent, improve code and clean-up

HannesWell:
As discussed in PR #440 the code of the new PDEMavenBundlePluginConfigurator can be improved to handle warnings located in parent poms correctly and simplified/clean-up.

I noticed that most of the code of PDEProjectHelper is actually unused and removed it in a second commit (I plan to squash and merge them).

I think we should postpone the addition of the warning regarding 'incompatible' versions of maven-bundle-plugin until the fxies are actually merged in felix. I will prepare a corresponding follow-up draf PR once this is merged.

### DIFF
--- a/org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/PDEMavenBundlePluginConfigurator.java
+++ b/org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/PDEMavenBundlePluginConfigurator.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008, 2021 Sonatype, Inc.
+ * Copyright (c) 2008, 2022 Sonatype, Inc.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -8,15 +8,12 @@
 package org.eclipse.m2e.pde.connector;
 
 import java.io.File;
-import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 
 import org.apache.maven.model.Plugin;
 import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.project.MavenProject;
 import org.eclipse.core.resources.IMarker;
-import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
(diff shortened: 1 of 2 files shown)


laeubi:
> I noticed that most of the code of PDEProjectHelper is actually unused and removed it

I would like to keep the code for now untill we are sure it is not required anymore as these contain some usefull hints what PDE quirks we might need to be aware of.

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/PDEProjectHelper.java
```diff
@@ -1,260 +1,56 @@
 /*******************************************************************************
- * Copyright (c) 2011 Sonatype, Inc.
+ * Copyright (c) 2011, 2021 Sonatype, Inc.
```
laeubi, line 2: Happy new Year! :-)

HannesWell:
> > I noticed that most of the code of PDEProjectHelper is actually unused and removed it
> 
> I would like to keep the code for now untill we are sure it is not required anymore as these contain some usefull hints what PDE quirks we might need to be aware of.

Alright, makes sense. However I think we should format the file using the default m2e/Eclipse formatter.

HannesWell:
@laeubi do you have any further comments on this or are you fine with as it is?

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/PDEMavenBundlePluginConfigurator.java
```diff
@@ -53,93 +50,85 @@
...
-			}
-		}
-		return false;
+		return BND_MANIFEST_GOALS.contains(execution.getGoal());
 	}
 
 	@Override
 	public void configureClasspath(IMavenProjectFacade facade, IClasspathDescriptor classpath, IProgressMonitor monitor)
-			throws CoreException {
+			throws CoreException { // nothing to do
```
laeubi, line 116: PDEProjectHelper has some code that seems to handle class path changes, do you think we should call it here?

laeubi (approved):

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/PDEMavenBundlePluginConfigurator.java
HannesWell: I'm not yet sure about that, but I will check it as soon as I continue my work in this area!

HannesWell:
If the current build is successful I would like to merge this as it is to have it available in the upcoming release.

PR #505, opened 2022-01-20
Remove dep from lemminx editor on indexer

mickaelistria:
LemMinX-Maven doesn't require indexer any more


HannesWell:
I already wondered if this can be done with the latest update of leminx-maven.
Glad to see one more location less where the indexer is used.

PR #506, opened 2022-01-20
Update deps

akurtakov:
And reduce guava usage.


HannesWell:
Are you planning to continue to reduce the usage of guava?
Since the topic popped up in the PDE Gerrit this morning I remembered that I have some pending changes aiming to reduce the usage of guava and continued to complete this work.

At the moment I have some more changes, but I would like to avoid duplicated work.

akurtakov:
This was just long standing in my local git. I don't plan further work for now.

Issue #157, opened 2021-04-04
m2e Incremental Builder slowness


stas-bob:
I have a huge workspace with 150 maven projects.
On every incremental build m2e builder resolves dependencies. I looked into it with jvisualvm.
The progress view prints "Invoking Maven Builder on Project xy".
Maven is constantly building and building and building. It comes to an end finally but starts all over again on another workspace modification like saving a space into a java file.
I came across mavenProjectCache in ProjectRegistryManager. I noticed the cache size was 5 and on every incremental build the contents where diffrent. Cache items are replaced by others during one incremental build.
I then created a plugin which simply gets reference to the MavenPlugin and then goes through the fields with reflection and eventually replaced the google cache with another one with 500 elements.
From that point all my builder problems disappeard. The workspace is very fast now
I dont even notice any maven builder in progress view now.

Please make a public API/preference setting to modify cache size.

Or will i have side effects now? I worked with it for a while and didnt notice wrong behaviour.

mickaelistria:
150 modules is not huge, it's normal ;) FWIW m2e can scale with thousands of modules.
Can you please share your project, or just the pom files so this can be investigated? There may be some cases which are not handled optimally.
We cannot just increase the cache size too high: 1 Maven project in the cache costs ~1MB of RAM, we need to be careful of not growing it too big. Going to 500 is way too big, we could investigate something like 20 instead.
However, again the issue may just be a bug and just 1 necessary fix, so before blaming the cache, it should be investigated -thankd to a  reproducer- what are the alternative solutions.

stas-bob:
I can try to reconstruct the module structure without proprietary contents. Thats a lot of work but i want to try in a next response. 

Maybe you have an idea what i could already investigate myself? What could be the pitfalls?
I dont really know how the maven builder works but it iterates through numerous projects which are sometimes not even related to each other. Why is it necessary for the builder to invoke readProjectWithDependencies on every incremental build?
I attached a screenshot of the call tree during the "invoke maven builder" progress.
![1](https://user-images.githubusercontent.com/293153/113522055-066b5180-959e-11eb-94e4-7fe0c57da89b.png)

laeubi:
@mickaelistria my Dev-Machine has 64 GB if I can spend a few hundreds of them to speed up m2e why not :-)

I think it might not be the number of modules but if they are to related/chained here, maybe an intermediate solution would be to expose the cache size in the preferences and set it to 50 on default?

[...]

PR #507, opened 2022-01-20
Replace Guava with standard Java API

HannesWell:
This PR removes the dependency to Guava for all M2E plug-ins except for:
- org.eclipse.m2e.editor
- org.eclipse.m2e.maven.indexer
- org.eclipse.m2e.maven.runtime 

In some places the elements used from guava could be replaced quite easily with modern day java classes, in other places it took some more work and the resulting code was unfortunately a bit longer than before, especially in one use case of Multimaps.

@akurtakov what do you think? Do you know some better replacement strategies?

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/ProjectCachePlunger.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/ProjectCachePlunger.java
@@ -14,14 +14,14 @@
 package org.eclipse.m2e.core.internal.project;
 
 import java.io.File;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import com.google.common.collect.HashMultimap;
-import com.google.common.collect.Multimap;
 
 import org.eclipse.aether.RepositorySystemSession;
 import org.eclipse.aether.repository.RepositoryPolicy;
@@ -41,16 +41,16 @@ abstract class ProjectCachePlunger<Key> {
 
   private final Logger log = LoggerFactory.getLogger(getClass());
 
-  final Multimap<File, Key> projectKeys = HashMultimap.create();
+  final Map<File, Set<Key>> projectKeys = new HashMap<>();
 
-  final Multimap<Key, File> keyProjects = HashMultimap.create();
+  final Map<Key, Set<File>> keyProjects = new HashMap<>();
 
   public void register(MavenProject project, Key cacheKey) {
     // project.file is null for parent pom.xml resolved from repositories
     File file = project.getFile();
     if(file != null) {
-      projectKeys.put(file, cacheKey);
-      keyProjects.put(cacheKey, file);
+      projectKeys.computeIfAbsent(file, f -> new HashSet<>()).add(cacheKey);
+      keyProjects.computeIfAbsent(cacheKey, f -> new HashSet<>()).add(file);
     }
   }
 
@@ -62,14 +62,14 @@ abstract class ProjectCachePlunger<Key> {
     }
     final Set<File> affectedProjects = new HashSet<>();
 
-    for(Key cacheKey : projectKeys.removeAll(pom)) {
-      keyProjects.remove(cacheKey, pom);
+    for(Key cacheKey : removeAll(projectKeys, pom)) {
+      remove(keyProjects, cacheKey, pom);
       if(forceDependencyUpdate && RepositoryPolicy.UPDATE_POLICY_ALWAYS.equals(session.getUpdatePolicy())
           && session.getCache().get(session, cacheKey) == null) {
         session.getCache().put(session, cacheKey, Boolean.TRUE);
-        for(File affectedPom : keyProjects.removeAll(cacheKey)) {
+        for(File affectedPom : removeAll(keyProjects, cacheKey)) {
           affectedProjects.add(affectedPom);
-          projectKeys.remove(affectedPom, cacheKey);
+          remove(projectKeys, affectedPom, cacheKey);
         }
       }
       if(!keyProjects.containsKey(cacheKey)) {
@@ -77,8 +77,19 @@ abstract class ProjectCachePlunger<Key> {
         log.debug("Flushed cache entry for {}", cacheKey);
       }
     }
-
     return affectedProjects;
+  }
+
+  private static <K, V> Set<V> removeAll(Map<K, Set<V>> map, K key) {
+    Set<V> removed = map.remove(key);
+    return removed != null ? removed : Collections.emptySet();
+  }
+
+  private static <K, V> void remove(Map<K, Set<V>> map, K key, V value) {
+    map.computeIfPresent(key, (k, values) -> {
+      values.remove(value);
+      return values.isEmpty() ? null : values;
+    });
   }
 
   protected void disposeClassRealm(ClassRealm realm) {
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
@@ -33,19 +33,11 @@ import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
-import java.util.concurrent.ExecutionException;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import com.google.common.cache.Cache;
-import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.RemovalCause;
-import com.google.common.cache.RemovalListener;
-import com.google.common.collect.LinkedHashMultimap;
-import com.google.common.collect.Multimap;
 
 import org.eclipse.core.filesystem.EFS;
 import org.eclipse.core.resources.IFile;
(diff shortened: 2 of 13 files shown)


laeubi:
> especially in one use case of Multimaps

Reading though the javadoc Multimap seems to be quite easlie replaced by getOrDefault / computeIfAbsent or even Collectors.groupBy, could you explain what has made it complex here?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/M2EUtils.java
```diff
@@ -184,4 +189,38 @@ public static void copyProperties(Properties to, Properties from) {
       }
     }
   }
+
```
laeubi, line 192: For Hash operations I always found commons.DigestUtils quite handy...

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
```diff
@@ -1057,21 +1044,16 @@ MavenProject getMavenProject(final MavenProjectFacade facade, final IProgressMon
...
+    try {
+      return mavenProjects.computeIfAbsent(facade, fac -> mavenProjectCache.computeIfAbsent(fac,
+          f -> readProjectWithDependencies(f.getPom(), f.getResolverConfiguration(), monitor)));
+    } catch(IllegalStateException ex) { // thrown by method called in lambda to carry CoreException
+      Throwable cause = ex.getCause();
+      if(cause instanceof CoreException) {
+        throw (CoreException) cause;
       }
-      mavenProjects.put(facade, mavenProject);
+      throw new RuntimeException(cause); // this really should never happen
```
laeubi, line 1055: You should throw new AssertionError then ...

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
```diff
@@ -1057,21 +1044,16 @@ MavenProject getMavenProject(final MavenProjectFacade facade, final IProgressMon
...
-      } catch(ExecutionException ex) {
-        Throwable cause = ex.getCause();
-        if(cause instanceof CoreException) {
-          throw (CoreException) cause;
-        }
-        throw new RuntimeException(cause); // this really should never happen
+    try {
+      return mavenProjects.computeIfAbsent(facade, fac -> mavenProjectCache.computeIfAbsent(fac,
+          f -> readProjectWithDependencies(f.getPom(), f.getResolverConfiguration(), monitor)));
+    } catch(IllegalStateException ex) { // thrown by method called in lambda to carry CoreException
```
laeubi, line 1050: IllegalStateException is really bad choice I think... WHy not catch runtme exceptions at all and look if the cause if a CoreException, if not create a new one with "internal error" message?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
```diff
@@ -165,7 +156,7 @@
    */
   private final Map<MavenProjectFacade, MavenProject> legacyMavenProjects = new IdentityHashMap<>();
 
-  private final Cache<MavenProjectFacade, MavenProject> mavenProjectCache;
+  private final Map<MavenProjectFacade, MavenProject> mavenProjectCache;
```
laeubi, line 159: Should we use a ConcurrentMap here?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/M2EUtils.java
HannesWell: Good suggestion!
The library `apache.commons.codec` is much smaller and used already in M2E and other Eclipse plug-ins so I think it is a good trade. Thanks.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
HannesWell: That's right. Done.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
HannesWell: I would like to use that too, but I only know the approach via the `LinkedHashMap` that is applied below to implement a capacity restricted Map.
If you know a way to achieve this with a `ConcurrentHashMap` in atomic operations please let me know.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
HannesWell: Even with the latest change, so that it is something like?
```suggestion
     throw new AssertionError("Unexpected internal Error occured", ex); // this really should never happen
```

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/ProjectCachePlunger.java
```diff
@@ -62,25 +62,36 @@ public void register(MavenProject project, Key cacheKey) {
...
       if(!keyProjects.containsKey(cacheKey)) {
         flush(cacheKey);
         log.debug("Flushed cache entry for {}", cacheKey);
       }
     }
-
     return affectedProjects;
   }
 
+  private static <K, V> Set<V> removeAll(Map<K, Set<V>> map, K key) {
```
HannesWell, line 83: > > especially in one use case of Multimaps
> 
> Reading though the javadoc Multimap seems to be quite easlie replaced by getOrDefault / computeIfAbsent or even Collectors.groupBy, could you explain what has made it complex here?

In most cases you're right. When mentioning 'problems' I was mainly referring to this and the following method.
This is not really bad, but a bit longer than before.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
laeubi: Then access should probably be synchronized, actually this shows that guava provides more with the cache as is provided "by standard java" and we either need to implement a cache our self or keep guava for this use-case.

HannesWell: It is already synchronized, see the method `createProjectCache()`.
Most functionalities can be assembled with standard java, the only difference at the moment is that the concurrency-level is one while the Guava cache by default can handle up to 4 concurrent accesses, but this number is configurable.

HannesWell:
@laeubi do you or the others have any more remarks on this?

laeubi:
Nope

HannesWell:
> Nope

Then thanks for your review!

stas-bob:
The change of the cache to Map makes it really hard now to change the cache size. I managed to changed it with reflection in previous releases but now its really hard. Especially due to the size configured in the Constructor of LinkedHashMap.
I really require a higher cache size, otherwise the workspace becomes unusable. Please give the method createProjectCache an int parameter so that I can call it with my desired size.

HannesWell:
I have no objections to provide means to make this configurable (add a parameter and make it protected?).
But please not that this is an internal, non API class so it may change in the future at any time without any further notice.

Do you want to provide a PR that fits your needs?

I wonder if the cache size should become a preference? Other users could want to increase the cache size in the IDE. And you could of course then modify this via code too.

laeubi:
> I wonder if the cache size should become a preference?

Yes, making fields accessible for reflection is clearly not a project goal.

stas-bob:
To avoid confusion, of course I do not prefer any reflection logic. I prefer public API. I did reflection because there is no other alternative.
I already filed an issue about that

https://github.com/eclipse-m2e/m2e-core/issues/157

But I closed it, since I could not provide a reproducible example and I was able to fix it with reflection. But now the problem is back since the class changed. I dont blame anyone. I just ask for public API.

cpfeiffer:
FWIW, we also needed to increase the cache size (we cache 15 projects by default). We made it configurable with a system property `m2e.projectCacheSize`. Simple, only for "power users", no preferences UI needed.

See https://github.com/GEBIT/m2e-core/commit/3f9edce34f6332cac9d1d8c7b89462c554dcc9a9 for the patch.

Also see https://www.eclipse.org/lists/m2e-users/msg05094.html for a discussion on this topic. (While it discourages increasing the cache size, it really is a necessety for performance reasons.)

HannesWell:
> But I closed it, since I could not provide a reproducible example and I was able to fix it with reflection. But now the problem is back since the class changed. I dont blame anyone. I just ask for public API.

In the issue you mentioned Christoph already suggested to add a preference and Mickael expressed his consent: https://github.com/eclipse-m2e/m2e-core/issues/157#issuecomment-813287988
While this is indeed not ideal I agree that this is a suitable intermediate solution until there is a better one.

The fact that a cache exists is actually an implementation-detail and the class itself is not public API so I think it not would really make sense to add a public _API_ for that. We could add a public method for that, but creating a preference would also allow other users that do not want to build a patched m2e to increase it from the preference UI.
If you still want to build your own m2e you can set that preference easily from the code too.

Since the preference UI for M2E already exists, adding another int-item to it and reading it when creating the cache in `ProjectRegistryManager` should not be too much work.
I want to encourage one of you to submit a quality PR to add it :)
We committers will then review it and give feedback as necessary. If you have any questions, we will try to answer them, so please don't hesitate to ask.

HannesWell:
FYI: #850
But please keep in mind that this is internal and the System property may be removed at any time if it is not necessary any more.

PR #520, opened 2022-01-23
[Build] only sign plug-ins for m2e repo

HannesWell:
At the moment there are two profiles to sign the m2e-plugins, one in the root pom.xml and one in `org.eclipse.m2e.site/pom.xml`. The latter one is necessary because external bundles that are included into the m2e-repo are not jar-signed (with the Eclipse certificate). IIRC it is `org.eclipse.m2e.workspace.cli`.
Because the jar-signing has to be performed for the plug-ins in the m2e-repo anyways (already signed jars are not re-signed), I suggest to move the signing entirely to the `m2e-site` project and sign all included plug-ins there in one pass instead of signing each plugin-project individually. This simplifies the poms.

Additionally:
- merge definitions of tycho-p2-plugin
- simplify m2e.version file write (just write m2e.version file for each feature to avoid the need for ant-contrib)

@mickaelistria WDYT?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -216,6 +216,7 @@
 			<plugin>
 				<groupId>org.eclipse.tycho</groupId>
 				<artifactId>tycho-p2-plugin</artifactId>
+				<version>${tycho-version}</version>
 				<executions>
 					<execution>
 						<id>p2-metadata</id>
@@ -227,6 +228,15 @@
 				</executions>
 				<configuration>
 					<defaultP2Metadata>false</defaultP2Metadata>
+					<baselineReplace>all</baselineReplace>
+					<baselineRepositories>
+						<repository>
+							<url>https://download.eclipse.org/technology/m2e/releases/latest/</url>
+						</repository>
+						<repository>
+							<url>https://download.eclipse.org/technology/m2e/snapshots/latest/</url>
+						</repository>
+					</baselineRepositories>
 				</configuration>
 			</plugin>
 		</plugins>
@@ -281,22 +291,6 @@
 							<include>**/*Test.java</include>
 						</includes>
 						<forkedProcessTimeoutInSeconds>${tycho.surefire.timeout}</forkedProcessTimeoutInSeconds>
-					</configuration>
-				</plugin>
-				<plugin>
-					<groupId>org.eclipse.tycho</groupId>
-					<artifactId>tycho-p2-plugin</artifactId>
-					<version>${tycho-version}</version>
-					<configuration>
-						<baselineRepositories>
-							<repository>
-								<url>https://download.eclipse.org/technology/m2e/releases/latest/</url>
-							</repository>
-							<repository>
-								<url>https://download.eclipse.org/technology/m2e/snapshots/latest/</url>
-							</repository>
-						</baselineRepositories>
-						<baselineReplace>common</baselineReplace>
 					</configuration>
 				</plugin>
 				<plugin>
@@ -368,21 +362,14 @@
 
 	<profiles>
 		<profile>
-			<id>eclipse-sign</id>
+			<id>write-feature-version-file</id>
+			<activation>
+				<file>
+					<exists>feature.xml</exists>
+				</file>
+			</activation>
 			<build>
 				<plugins>
-					<plugin>
-						<groupId>org.eclipse.cbi.maven.plugins</groupId>
-						<artifactId>eclipse-jarsigner-plugin</artifactId>
-						<executions>
-							<execution>
-								<id>sign</id>
-								<goals>
-									<goal>sign</goal>
-								</goals>
-							</execution>
-						</executions>
-					</plugin>
 					<plugin>
 						<groupId>org.apache.maven.plugins</groupId>
 						<artifactId>maven-antrun-plugin</artifactId>
@@ -394,15 +381,10 @@
 									<goal>run</goal>
 								</goals>
 								<configuration>
-									<target> <!-- Write project.versions file read during build for the m2e-SDK feature -->
-										<taskdef resource="net/sf/antcontrib/antlib.xml" />
-										<if>
-											<equals arg1="${project.artifactId}" arg2="org.eclipse.m2e.sdk.feature" /> <!-- version-file file only required for m2e.sdk feature -->
-											<then>
-												<echo message="Write '${unqualifiedVersion}' to ${project.build.directory}/m2e.version" level="info" />
-												<echo message="${unqualifiedVersion}" file="${project.build.directory}/m2e.version" level="info" />
-											</then>
-										</if>
+									<target>
+										<!-- Write project.versions file read during build for the m2e-SDK feature -->
+										<echo message="Write '${unqualifiedVersion}' to ${project.build.directory}/m2e.version" level="info" />
+										<echo message="${unqualifiedVersion}" file="${project.build.directory}/m2e.version" level="info" />
 									</target>
 								</configuration>
 							</execution>


mickaelistria:
-1, usually, we want to sign the artifacts, not the site content, because artifacts may be consumed differently (imagine we publish them to some Maven repo in the future, they'd need to be jarsigned there as well ideally). So if one of both has to be removed, it's signing the p2 site, not signing the artifacts.

laeubi:
The signing of the P2 site is only required for external artifacts included (from other P2 site, Maven, ...) that are not signed at first place to fulfill sim-rel rules.

HannesWell:
> -1, usually, we want to sign the artifacts, not the site content, because artifacts may be consumed differently (imagine we publish them to some Maven repo in the future, they'd need to be jarsigned there as well ideally). So if one of both has to be removed, it's signing the p2 site, not signing the artifacts.

OK, that makes sense.

> The signing of the P2 site is only required for external artifacts included (from other P2 site, Maven, ...) that are not signed at first place to fulfill sim-rel rules.

Didn't the requirements change recently and PGP signing like it is required for Maven-Central is enough now, did it? Or did I understand that wrong? If PGP-signing is sufficient those signatures would have to be included instead, right?

laeubi:
The artifact in question (bnd-tool) is not PGP signed.

HannesWell:
Besides the bnd-lib/util bundles this affects also the `org.eclipse.m2e.workspace.cli`.

But do I get the PGP-signing story wrong?
It is a requirement for artifacts to be PGP/GPG signed: https://central.sonatype.org/publish/requirements/gpg/
And I thought GPG is an implementation of the PGP standard (and often used synonymous)?

Furthermore I thought there is currently some work done to allow those PGP/GPG signatures as an alternative to jar-signing in the SimRel to avoid the need for Eclipse-Orbit and to be able to consume artifacts from Maven-Central directly? The only thing that was to include those signatures as "trusted" signatures into the repo?
At least that is how I understood it, but I'm more a newbe in PGP/GPG signing than an expert.

laeubi:
> But do I get the PGP-signing story wrong?

Actually its a bit more complex:

- maven central requires you to sign the artifact
- currently m2e-pde/tycho does not download the signature and thus it is not available
- even if m2e/tycho would do that this won't work as expected because of how p2/platform currently implements the PGP story
- even if it would work, currently there is no plan to have pgp-signed artifacts in the sim-rel they are only used for artifacts not contributed to simrel
- even if we are there, the current implementation sign artifacts with an *own* PGP key (and not using the original one!)

Thus from my point of view it doesn't really makes a difference if we simply sign the jar here, there is nothing to gain ...

mickaelistria:
The PGP story in SImRel is still work in progress, but SimRel will probably mandate that the artifacts are signed with the Eclipse Project key (either Eclipse certificate for jarsigner, or the project PGP key for PGP). External signatures can be used as extra information, but the rule will be that the project must sign their artifacts, or maybe even SimRel will resign them.
We don't have to care about the upstream signature for external artifacts.

HannesWell:
Thank you both for the explanations. So the jar-signing should stay as it is.

I will close this PR and create a separate one for the additional clean-up changes that do not affect signing.

laeubi:
@HannesWell I think `org.eclipse.m2e.workspace.cli `should still be signed at first place and not at the repository level as it is produced at the m2e build.

> SimRel will probably mandate that the artifacts are signed with the Eclipse Project key (either Eclipse certificate for jarsigner, or the project PGP key for PGP)

I always found that strange... signing an external "thing" with my key feels plain wrong unless I really want to express that I have done a security-audit or alike but I think that's far beyond what we can offer for third-party dependencies. So I would strongly encourage to use the original key (and probably let eclipse/project key thrust to a certain degree common keys, eg. from apache or other Eclipse projects).

akurtakov:
> 
> > SimRel will probably mandate that the artifacts are signed with the Eclipse Project key (either Eclipse certificate for jarsigner, or the project PGP key for PGP)
> 
> I always found that strange... signing an external "thing" with my key feels plain wrong unless I really want to express that I have done a security-audit or alike but I think that's far beyond what we can offer for third-party dependencies. So I would strongly encourage to use the original key (and probably let eclipse/project key thrust to a certain degree common keys, eg. from apache or other Eclipse projects).

There are 2 sides of the issue here - technical and political.
* technically - we are just starting with PGP so having less keys and with chain of trust as all projects keys are signed with EF one makes it easier for everyone as there won't be random keys popping up to be approved as EF project keys will be "injected" in some way. 
* political - while I fully agree about not signing with project key third party content this is decision of EF/IDE WG and I'm not even sure it's mandatory for everything or only for participating in Simrel.

After all M2E is just a user/collaborator on the issue so technical decisions on PGP should be taken on Equinox p2 level and political on simrel/IDE WG level. So further discussions should be taken there as  here anything we say on the topic will have zero effect. 
Of course it's project's decision whether to participate in simrel if not wanting to play by these rules but M2E not participating in simrel will be counter productive for everyone IMHO.

laeubi:
@akurtakov sure, but I'm not a politician at all :-)

For sure m2e has to either accept or leave it (even though I don't like to choose between both), I just wanted to drop a note for those with more insight/influence on the EF/IDE/Simrel/... level and probably would consider this.

mickaelistria:
I personally find having consumer to sign their dependencies is a good thing, as it does indeed express that m2e trusts this particular external artifact and that people trusting m2e (and its PGP public key) imply that they do trust the external artifacts consumed by m2e in the same conditions.
Note that signing doesn't imply anything about security audit or even any form of audit of the content, it just implies that "project X has signed this artifact from project Y", the semantic of what this signature means is not pre-defined, it's project X to define what it means and users of project X to become aware of it and take a wider decision about their trust in project X.
Currently, PGP signature simply means "this was officially picked as a dependency of Eclipse m2e/SimRel so it has been through the license and other review steps defined in the Eclipse Development Process".

laeubi:
Maybe I'm wrong, but shouldn't then the original signature not be signed by the EF PGP key instead of creating a completely new signature?

mickaelistria:
That would be a possibility, but what would that change to end-users in the end? Seems a bit overkill.
Note that PGP signatures support in p2 does allow artifacts to have multiple signatures, so artifacts could have their "original" signature (meaning someone did initially trust this artifact), the project signature (m2e trusts this artifact) and the SimRel signature (SimRel trusts the artifact). Since we don't know what the original signature means, there is no particular reason to build on top of it, just resiging with a project signature that we know the semantics of the signature (conforms to EDP) seems much simpler and semantically stronger.

HannesWell:
> @HannesWell I think `org.eclipse.m2e.workspace.cli `should still be signed at first place and not at the repository level as it is produced at the m2e build.
> 
Regarding `org.eclipse.m2e.workspace.cli` I was wrong, it is already jar-signed.
For clarification that artifact is not build by `m2e-core` but, if I understand it correctly, it is from the sibling repository https://github.com/eclipse-m2e/org.eclipse.m2e.workspace. M2E-core includes it via the TP from Maven-Central.
Not sure why I got in my mind it would not be signed, but the m2e-master build logs stated that
`[INFO] Jar '/home/jenkins/agent/workspace/m2e_master/org.eclipse.m2e.site/target/repository/plugins/org.eclipse.m2e.workspace.cli_0.3.1.jar' is already signed and will *not* be resigned.`. And when looking at the mentioned repo it looks like the jar-signing is performed (I looked into the jars from Maven-Central and the certifacts are present).
So Christoph you are right, signing the m2e-repo is only done for the bnd-lib artifacts.

To the discussion if or if not dependencies should be (PGP) signed and how this should be done, I unfortunately cannot contribute anything useful.

PR #524, opened 2022-01-25
[Console] improve handling of linked Test-Reports

HannesWell:
This PR has to goal to improve the handling of links to test-reports displayed in the Console of a Maven build. This is the successor of PR #521.

Specifically, this involves the following points: 
- open JUnit-Test report if available (instead of plain-text summary file)
- Remove unused test-class search pattern

Since the question about the use-case of `Pattern TEST_CLASS_PATTERN = Pattern.compile("(?:  )test.+\\(([\\w\\.]+)\\)");` arose and Mickael suspected that it would be used in failed tests I tried it with a failing test:

grafik

The link in `[INFO] Running simpleProjectWithJUnit5Test.SimpleTest` is created by M2E's ConsoleLineTracker and when you click on it it (from now on) opens the JUnit view displaying the test results, almost as if you run it like JUnit test from the IDE.
The second link in the stack-trace `at simpleProjectWithJUnit5Test.SimpleTest.testSimple(SimpleTest.java:9)` is a `JavaStackTraceHyperlink` from `org.eclipse.jdt.internal.debug.ui.console`.

So the TEST_CLASS_PATTERN doesn't match here as well.
The pattern matches strings like the following (note that there have to be at least two white spaces in the beginning):
- `  test.(foo.bar.TestClass1)`
- `  test   (foo.bar.TestClass2)`
- `  test  some more text (foo.bar.TestClass)`
- `     testdkfjgdjklgdkflghdkfgdjgk(TestClass3)`

Regarding scanning the console for string `Running ` to find the name of executed test classes: As said in 
 https://github.com/eclipse-m2e/m2e-core/pull/521#issuecomment-1020562030 I think in general this is still the most performant way to add links and to find the corresponding test report files.
However it would probably be better to find those name using another pattern like Pattern.compile("Running ([\\w\\.]+)$") instead of just search for the String "Running ". With the current logic a link is created as soon as that string is written anywhere. The suggested pattern, although it still captures a lot (but for example no white-space) is at least a bit more specific.


mickaelistria:
Thanks for the details. I don't know much about this code, but I understand your comments and agree with it, so +1 from me to merge it as it (if all tests are still passing).

HannesWell:
I have updated the ConsoleLineTracker to use Regex Patterns instead of just marker Strings to be more precise in what they match and reduce the chance of false links. Since the Patterns start with the same strings as the markers did, applying those patterns on a not matching line should have a similar performance. Only a match will likely take a little longer but I think that's neglectable. 

> Thanks for the details. I don't know much about this code, but I understand your comments and agree with it, so +1 from me to merge it as it (if all tests are still passing).

Alright, thanks for having a look at this too. I will merge this later (given the tests pass).

PR #518, opened 2022-01-21
Remove SSE editor

mickaelistria:

PR #527, opened 2022-01-27
Remove WTP-based editor

mickaelistria:


mickaelistria:
Already tracked as #518

PR #535, opened 2022-01-30
[Console] Support content detection for ANSI-escaped console text

HannesWell:
With this PR the content detection of the `ConsoleLineTracker`, namely the detection of Running tests and the detection of forked VMs waiting for a debugger connection, also work for colored (i.e. ANSI escaped) print-outs.
Colored print-out is activated using the command line arguments `-Dstyle.color=always` or if one has the `m2e-chromatic` and `ANSI-Console` plug-ins installed from the marketplace.

I know that in general we should not support external plug-ins specifically. But the mentioned arguments is built into Maven and does not require any of the mentioned marketplace plug-ins (although the print out then becomes hardly readable). Therefore proper support can be tested independently and it requires not too much code. Therefore I think it's acceptable.

### DIFF
--- a/org.eclipse.m2e.launching/src/org/eclipse/m2e/internal/launch/MavenConsoleLineTracker.java
+++ b/org.eclipse.m2e.launching/src/org/eclipse/m2e/internal/launch/MavenConsoleLineTracker.java
@@ -87,6 +87,9 @@ public class MavenConsoleLineTracker implements IConsoleLineTracker {
 
   private static final int FAILED_ARTIFACT_ID = 1;
 
+  private static final Pattern ESCAPE_CHARACTERS = Pattern.compile("\\e\\[[\\d;]*?[^\\d;]");
+  // captures ANSI escape characters added when -Dstyle.color=always is set
+
   private boolean isMavenBuildProcess;
 
   private IConsole console;
@@ -96,6 +99,8 @@ public class MavenConsoleLineTracker implements IConsoleLineTracker {
   private IMavenProjectFacade mavenProject;
 
   private final Deque<IRegion> projectDefinitionLines = new ArrayDeque<>(2);
+
+  private final List<int[]> removedLineLocations = new ArrayList<>();
 
   @Override
   public void init(IConsole console) {
@@ -121,9 +126,12 @@ public class MavenConsoleLineTracker implements IConsoleLineTracker {
         if(runningTestMatcher.find()) {
           String testName = runningTestMatcher.group(TEST_CLASS_NAME);
           int start = runningTestMatcher.start(TEST_CLASS_NAME);
+          int end = start + testName.length();
+          start = getOriginalIndex(start, removedLineLocations);
+          end = getOriginalIndex(end, removedLineLocations);
 
           IHyperlink link = new MavenTestReportHyperLink(mavenProject, testName);
-          console.addLink(link, line.getOffset() + start, testName.length());
+          console.addLink(link, line.getOffset() + start, end - start);
           return;
         }
 
@@ -137,7 +145,7 @@ public class MavenConsoleLineTracker implements IConsoleLineTracker {
 
         Matcher failureMatcher = EXECUTION_FAILURE.matcher(text);
         if(failureMatcher.find()) {
-          addProjectLink(line, failureMatcher, FAILED_ARTIFACT_ID, FAILED_ARTIFACT_ID);
+          addProjectLink(line, failureMatcher, FAILED_ARTIFACT_ID, FAILED_ARTIFACT_ID, removedLineLocations);
         }
       } catch(BadLocationException ex) {
         // ignore
@@ -172,7 +180,8 @@ public class MavenConsoleLineTracker implements IConsoleLineTracker {
   private static final Pattern PACKAGING_TYPE_LINE = Pattern.compile("^\\[INFO\\] -+\\[ [\\w\\-\\. ]+ \\]-+$");
 
   private String getText(IRegion lineRegion) throws BadLocationException {
-    String line0 = getLineText(lineRegion);
+    removedLineLocations.clear();
+    String line0 = getLineText(lineRegion, removedLineLocations);
 
     if(projectDefinitionLines.size() < 2) {
       projectDefinitionLines.add(lineRegion);
@@ -187,14 +196,15 @@ public class MavenConsoleLineTracker implements IConsoleLineTracker {
       Iterator<IRegion> previousLines = projectDefinitionLines.descendingIterator();
 
       IRegion line1Region = previousLines.next();
-      String line1 = getLineText(line1Region);
+      String line1 = getLineText(line1Region, null);
 
       Matcher vMatcher = VERSION_LINE.matcher(line1);
       if(vMatcher.matches()) {
         String version = vMatcher.group(VERSION);
 
         IRegion line2Region = previousLines.next();
-        String line2 = getLineText(line2Region);
+        List<int[]> removedLine2Locations = new ArrayList<>();
+        String line2 = getLineText(line2Region, removedLine2Locations);
         Matcher gaMatcher = GROUP_ARTIFACT_LINE.matcher(line2);
         if(gaMatcher.matches()) {
           String groupId = gaMatcher.group(GROUP_ID);
@@ -203,7 +213,7 @@ public class MavenConsoleLineTracker implements IConsoleLineTracker {
           IMavenProjectRegistry projectManager = MavenPlugin.getMavenProjectRegistry();
           mavenProject = projectManager.getMavenProject(groupId, artifactId, version);
 
-          addProjectLink(line2Region, gaMatcher, GROUP_ID, ARTIFACT_ID);
+          addProjectLink(line2Region, gaMatcher, GROUP_ID, ARTIFACT_ID, removedLine2Locations);
         }
       }
     }
@@ -212,14 +222,27 @@ public class MavenConsoleLineTracker implements IConsoleLineTracker {
     return line0;
   }
 
-  private String getLineText(IRegion lineRegion) throws BadLocationException {
-    return console.getDocument().get(lineRegion.getOffset(), lineRegion.getLength()).strip();
+  private String getLineText(IRegion region, List<int[]> removedLocations) throws BadLocationException {
+    String rawText = console.getDocument().get(region.getOffset(), region.getLength()).strip();
+    Matcher m = ESCAPE_CHARACTERS.matcher(rawText);
+    return removedLocations == null ? m.replaceAll("") : m.replaceAll(mr -> {
+      removedLocations.add(new int[] {mr.start(), mr.end() - mr.start()});
+      return "";
+    });
   }
 
-  private void addProjectLink(IRegion line, Matcher matcher, int startGroup, int endGroup) {
+  private static int getOriginalIndex(int index, List<int[]> removedLocations) {
+    for(int i = 0; i < removedLocations.size() && removedLocations.get(i)[0] < index; i++ ) {
+      index += removedLocations.get(i)[1];
+    }
+    return index;
+  }
+
+  private void addProjectLink(IRegion line, Matcher matcher, int startGroup, int endGroup,
+      List<int[]> removedLocations) {
     IHyperlink link = new MavenProjectHyperLink(mavenProject);
-    int start = matcher.start(startGroup);
-    int end = matcher.end(endGroup);
+    int start = getOriginalIndex(matcher.start(startGroup), removedLocations);
+    int end = getOriginalIndex(matcher.end(endGroup), removedLocations);
     console.addLink(link, line.getOffset() + start, end - start);
   }
 
--- a/org.eclipse.m2e.core.ui.tests/src/org/eclipse/m2e/core/ui/tests/ConsoleTest.java
+++ b/org.eclipse.m2e.core.ui.tests/src/org/eclipse/m2e/core/ui/tests/ConsoleTest.java
@@ -17,6 +17,8 @@ package org.eclipse.m2e.core.ui.tests;
 import static java.util.Map.entry;
 import static java.util.function.Predicate.not;
 import static java.util.stream.Collectors.toList;
+import static org.eclipse.m2e.actions.MavenLaunchConstants.ATTR_COLOR_VALUE_ALWAYS;
+import static org.eclipse.m2e.actions.MavenLaunchConstants.ATTR_COLOR_VALUE_NEVER;
 import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.CoreMatchers.instanceOf;
 import static org.hamcrest.CoreMatchers.is;
@@ -103,7 +105,7 @@ public class ConsoleTest extends AbstractMavenProjectTestCase {
 		Path tempProjectFolder = copyTestProjectIntoWorkspace("simplePomOK");
 		File pomFile = tempProjectFolder.resolve(IMavenConstants.POM_FILE_NAME).toFile();
 
-		IDocument document = runMavenBuild(pomFile.getParent(), ILaunchManager.RUN_MODE);
+		IDocument document = runMavenBuild(pomFile.getParent(), ILaunchManager.RUN_MODE, ATTR_COLOR_VALUE_NEVER);
 
 		String consoleText = display.syncCall(document::get); // final document content could have changed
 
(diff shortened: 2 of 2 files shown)


mickaelistria:
I don't think it's a good idea to start adding such features in m2e. ANSI is a standard so those shoudl be directly added into Eclipse Console, as you could see in https://bugs.eclipse.org/bugs/show_bug.cgi?id=112948 . Adding it in standard Eclipse will benefit to a much wider world and attract more contributors who need this independently of Maven, and would also remove this duty from m2e, that can then keep focusing on Maven integration.

HannesWell:
> I don't think it's a good idea to start adding such features in m2e. ANSI is a standard so those shoudl be directly added into Eclipse Console, as you could see in https://bugs.eclipse.org/bugs/show_bug.cgi?id=112948 . Adding it in standard Eclipse will benefit to a much wider world and attract more contributors who need this independently of Maven, and would also remove this duty from m2e, that can then keep focusing on Maven integration.

I actually agree with you and would really appreciate if this would be part of standard Eclipse. It would then probably require a separate way to get the text without escape characters and to place a link using the unescpaed text's indices. But I think this should be feasible like I just did it here. Maybe just in more general form.

The reason I wanted to add this workaround to m2e is the time I expect support for ANSI escaped text will take until it will be available in standard Eclipse. I fear it will take some more time until the feature requested by the mentioned bug is available since it is dormant for three years now. I just tried to revive it yesterday and at least until now there was no reaction, but this of course still can change. At least the Maintainer of the ANSI-Console plug-ins seem to be active on GitHub so there might be a chance we can benefit from his knowledge.

I would also be fine if this becomes an intermediate "workaround", that is removed when there is support in the standard Eclipse console. You probably prefer to wait completely until this becomes available in Standard Eclipse, don't you?

laeubi:
For the [cucumber-eclipse](https://github.com/cucumber/cucumber-eclipse) project I use the ANSI-Console it works without any issues and having colors is quite cool: https://github.com/mihnita/ansi-econsole and the author seems to want contribute the code ...

mickaelistria:
> The reason I wanted to add this workaround to m2e is the time I expect support for ANSI escaped text will take until it will be available in standard Eclipse

That's up to you ;) Platform is open for contributions, and fast at reviewing/merging if everything is internal.
I personally don't expect the people who have written the ANSI Console plugin to contribute to Platform. It's been requested for a long time and they didn't seem interested. Many people prefer having stars on GitHub or Marketplace than just doing better software...

>  would also be fine if this becomes an intermediate "workaround"

Those workarounds are harmful for the wider community:
* They put technical value on the wrong layer, placing maintenance responsibility in the wrong hands
* They make the the value is not delivered to most of users who can benefit from it
* They reduce the chance of upstream projects including a real fix at some point
* They prevent upstream projects from growing their communities
* ...

The rule for best community OSS is to treat projects you depend on just like your projects and contribute to it as much as necessary for the success of your project.


For this particular change, it would be nice to extract this ANSI support as a specific IConsoleLineTracker first, and then to just put this IConsoleLineTracker into org.eclipse.debug.ui.

laeubi:
> I personally don't expect the people who have written the ANSI Console plugin to contribute to Platform.

Its not always easy especially as eclipse-platform is still not on github. Anyways if there is a need for it I could try to prepare a contribution....

HannesWell:
> > The reason I wanted to add this workaround to m2e is the time I expect support for ANSI escaped text will take until it will be available in standard Eclipse
> 
> That's up to you ;) Platform is open for contributions, and fast at reviewing/merging if everything is internal. I personally don't expect the people who have written the ANSI Console plugin to contribute to Platform. It's been requested for a long time and they didn't seem interested.

I know, its just that at the moment there are other things I would like to do first and it would take me some time to get familiar with the field of ANSI escaped text and the styling of Console print-out. Therefore I thought this relatively small change, to just not get harmed by colored print-outs, would be acceptable although I knew it was not ideal.

> > would also be fine if this becomes an intermediate "workaround"
> 
> Those workarounds are harmful for the wider community:
> 
I fully agree with all of your statements. Just as I said above.

OK, then lets abandon this change and hope (or better work) for a standard Eclipse console that supports colors. 

> 
> For this particular change, it would be nice to extract this ANSI support as a specific IConsoleLineTracker first, and then to just put this IConsoleLineTracker into org.eclipse.debug.ui.

You mean that there is something like an `AbstractANSIAwareConsoleLineTracker` that provides the unescaped text to subclasses and assists with placing links into the text just as I did here?

> For the [cucumber-eclipse](https://github.com/cucumber/cucumber-eclipse) project I use the ANSI-Console it works without any issues and having colors is quite cool: https://github.com/mihnita/ansi-econsole and the author seems to want contribute the code ...

It is indeed very cool. I really like to use the ANSI-Console plug-in in combination with the [m2e-chromatic plug-in](https://marketplace.eclipse.org/content/m2e-chromatic) which was just created because the Standard Eclipse Console does not support colors natively. See [Bug 544307](https://bugs.eclipse.org/bugs/show_bug.cgi?id=544307).
So once the Console understands ANSI m2e could ship this out of the box. Wouldn't this be some cool eye-candy for m2e 2.0, would it? 😎 

Yes  Mihai Nita expressed his interest in contributing his Plugin to Eclipse in [Bug 112948](https://bugs.eclipse.org/bugs/show_bug.cgi?id=112948) about three years ago but then nothing much happened.
Unfortunately, as Mickael said, there was no reaction on Mickaels comment two years and my comment two days ago. Could only guess the reason, maybe he had the impression there was not much assistance or so.

However I suggest to create and issue at https://github.com/mihnita/ansi-econsole to ask him if he wants to start/continue his work towards a contribution and to offer my (and Christoph if you want to participate also your) assistance for the 'Eclipse procedures'.

> > I personally don't expect the people who have written the ANSI Console plugin to contribute to Platform.
> 
> Its not always easy especially as eclipse-platform is still not on github. Anyways if there is a need for it I could try to prepare a contribution....

I would first try to convince Mihai Nita to work on this since he seems to have a lot of experience in the field and I'm not eager to work on this or do you really want to do it @laeubi?

laeubi:
> So once the Console understands ANSI m2e could ship this out of the box. Wouldn't this be some cool eye-candy for m2e 2.0, would it? sunglasses

m2e could simply include the ANSI console right now together with m2e-chromatic as it is apache licenced it would only require a CQ.

> Yes Mihai Nita expressed his interest in contributing his Plugin to Eclipse in [Bug 112948](https://bugs.eclipse.org/bugs/show_bug.cgi?id=112948) about three years ago but then nothing much happened.
> Unfortunately, as Mickael said, there was no reaction on Mickaels comment two years and my comment two days ago.

There are many bugs in eclipse that even have longer no reaction from eclipse committers so well its just a matter of of priorities maybe.

> I would first try to convince Mihai Nita to work on this since he seems to have a lot of experience in the field and I'm not eager to work on this or do you really want to do it @laeubi?

At laest @mihnita has had [fixed two bug upon my request](https://github.com/mihnita/ansi-econsole/issues?q=+is%3Aissue+author%3Alaeubi+) it seems fair to at least offering some help if desired. Its not a million LOC project anyways :-)

laeubi:
I opened https://github.com/mihnita/ansi-econsole/issues/70

HannesWell:
> > So once the Console understands ANSI m2e could ship this out of the box. Wouldn't this be some cool eye-candy for m2e 2.0, would it? sunglasses
> 
> m2e could simply include the ANSI console right now together with m2e-chromatic as it is apache licenced it would only require a CQ.

As a last resort we could think about that but I would prefer to try a transfer first, since it also allows better/deeper integration and the ANSI escaping can be considered in other changes and is not broken every second release.

> 
> > Yes Mihai Nita expressed his interest in contributing his Plugin to Eclipse in [Bug 112948](https://bugs.eclipse.org/bugs/show_bug.cgi?id=112948) about three years ago but then nothing much happened.
> > Unfortunately, as Mickael said, there was no reaction on Mickaels comment two years and my comment two days ago.
> 
> There are many bugs in eclipse that even have longer no reaction from eclipse committers so well its just a matter of of priorities maybe.
> 
It wouldn't surprise me if priorities are the case, I think everybody experienced it too and only sporadic reactions don't push priorities IMHO. Therefore my previous suggestion to express our interest first and see if he is interested again. :)

> > I would first try to convince Mihai Nita to work on this since he seems to have a lot of experience in the field and I'm not eager to work on this or do you really want to do it @laeubi?
> 
> At laest @mihnita has had [fixed two bug upon my request](https://github.com/mihnita/ansi-econsole/issues?q=+is%3Aissue+author%3Alaeubi+) it seems fair to at least offering some help if desired. Its not a million LOC project anyways :-)

That's right :)

> I opened [mihnita/ansi-econsole#70](https://github.com/mihnita/ansi-econsole/issues/70)

Thanks for creating the issue!

mihnita:
Hi all,

First, I must apologize for being unresponsive.

The last few months have been "a bit busy", and in fact I think that will still be the case until end of August or so.

I've answered to issue https://github.com/mihnita/ansi-econsole/issues/70

But TLDR: I will check it if is possible to dual license Eclipse & Apache,
and with contact all contributors if adding an Eclipse license would be OK.

Mihai

HannesWell:
Now that we have built in support for Console colors in Eclipse-Debug (https://github.com/eclipse-platform/eclipse.platform.debug/pull/50) and support of that in M2E (https://github.com/eclipse-m2e/m2e-core/pull/873) I think we can and even should consider this again. Otherwise the link-creation and debugger attachment don't work with the colored console.

org.eclipse.m2e.launching/src/org/eclipse/m2e/internal/launch/MavenConsoleLineTracker.java
```diff
@@ -212,14 +222,27 @@ private String getText(IRegion lineRegion) throws BadLocationException {
     return line0;
   }
 
-  private String getLineText(IRegion lineRegion) throws BadLocationException {
-    return console.getDocument().get(lineRegion.getOffset(), lineRegion.getLength()).strip();
+  private String getLineText(IRegion region, List<int[]> removedLocations) throws BadLocationException {
```
HannesWell, line 225: @mihnita does the console provide a better (built in) way to obtain the unescaped text?

mihnita: > @mihnita does the console provide a better (built in) way to obtain the unescaped text?
> Or should we add that to the console, maybe in a more efficient way?

Nothing public.
The only place where that is needed is to put text in clipboard without codes.
And that is a simple regexp replace in [`AnsiConsoleUtils`](https://github.com/mihnita/eclipse.platform.debug/blob/master/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ansi/AnsiConsoleUtils.java)
```
AnsiConsoleUtils.ESCAPE_SEQUENCE_REGEX_TXT.matcher(plainText).replaceAll("");
```

mihnita: BTW, I know why the external maven uses colors on the level and the embedded one does not.

`org.eclipse.m2e.maven.runtime` requires `org.eclipse.m2e.maven.runtime.slf4j.simple`, which contains (in `jars` the simple slf4j logger (`slf4j-simple-*.jar`)

That logger does not output colors.

The external maven has a `maven-slf4j-provider-*.jar` in the `libs` folder that implements what's needed to behave like a slf4j provider.
And that provider produces color output for the levels.

What we see colored right now with the embedded maven are messages logged.
Because if I do `logger.warning("Hello \033[91m RED \033[m world!");` the message itself has colors.

---
Some options:
1. leave it as is
2. remove `slf4j-simple`, take `maven-slf4j-provider`, and we will behave 100% like the external maven
3. replace `slf4j-simple` with logback, which allows for colors and a lot of customization.
4. replace `slf4j-simple` with our own slf4j-provider, where we have full control and we know that nobody will change it behind our back

I played with option 3, and worked without problems.
And I don't expect trouble with option 2 either.
Option 4 would be more work.

mihnita: I've also checked option 2, using `maven-slf4j-provider`
Works, and the output looks like the one from external maven.

HannesWell: > > @mihnita does the console provide a better (built in) way to obtain the unescaped text?
> > Or should we add that to the console, maybe in a more efficient way?
> 
> Nothing public. The only place where that is needed is to put text in clipboard without codes. And that is a simple regexp replace in [`AnsiConsoleUtils`](https://github.com/mihnita/eclipse.platform.debug/blob/master/org.eclipse.ui.console/src/org/eclipse/ui/internal/console/ansi/AnsiConsoleUtils.java)
> 
> ```
> AnsiConsoleUtils.ESCAPE_SEQUENCE_REGEX_TXT.matcher(plainText).replaceAll("");
> ```

Thanks for that hint. Unfortunately that class is not accessible because the package is not exported.

In general and as it has already been said by Mickael earlier in this PR, unescaping of the console document text should not be done in M2E because others are probably also need this capability (getting the unescaped text, adding a link based on the unescaped offsets and lengths, etc.). Therefore I think it is better handled in the `o.e.ui.console` or `o.e.debug.ui`.
I'm thinking about having something like un-escaped views of an `IRegion` that provides the text without escape characters. Maybe it is even better to add an `getUnescapedDocument()` to the `IConsole`.
Or simply just build the entire document model by default on the unescaped text. So that if one really wants the raw text, it has to be obtained via `getRawDocument()` or so.
But that is something we should discuss in a dedicated issue in eclipse.platform.debug.

I also noticed that if the output is colored links are not displayed. But that is also an issue for platform.debug.


For now to support colored print-out in M2E I think this is the way we have to go until platform provides a more general one.

HannesWell: > BTW, I know why the external maven uses colors on the level and the embedded one does not.

What you describes makes totally sense to me, although I have not yet fully tested it but you are right with the jars in the external maven installation.

I'm in favor of option 2. The logging is a bit difficult here, because we want to have the standard logging when we run a Maven-build launched from the IDE, but at the same time want to use the logback implementation when maven runs in the IDE and that only be achieved the way it is now (at least nobody had a better idea how to make that `generally` better).

But option 2 looks like a very good way to go that is simple to achieve. Do you want to provide a PR for that?
You just have to replace the slf4j-simple by `org.apache.maven:maven-slf4j-provider` in the `org.eclipse.m2e.maven.runtime.slf4j.simple` and adjust the version to match the embedded maven version. For that you probably want to move the `maven-core.version` property up to the `m2e-maven-runtime/pom.xml`.

Actually we should then remove the `simple` suffix from `org.eclipse.m2e.maven.runtime.slf4j.simple` and set it's version also to the maven version. In the require-bundle in the `org.eclipse.m2e.maven.runtime/pom.xml` has to be adjusted to the names as well.


Option 3 and 4 seem to be a bit overkill respectively overprotective. We already rely on what Maven is giving us and therefore we can rely on the logger as well.

HannesWell: Looks like the renaming is a bit more work. It is probably simpler to split this, if you want you can just replace the logger impl and I take care of all other adjustments in the maven-components. Alternatively I can do all at once and mention you as coauthor of the commit.

mihnita: What do you think about adding `org.apache.maven:maven-slf4j-provider` directly to `org.eclipse.m2e.maven.runtime` and removing `org.eclipse.m2e.maven.runtime.slf4j.simple` completely?
It really is part of maven, covered by the same license, same version, etc.

M.

HannesWell: > What do you think about adding `org.apache.maven:maven-slf4j-provider` directly to `org.eclipse.m2e.maven.runtime` and removing `org.eclipse.m2e.maven.runtime.slf4j.simple` completely?

I already thought about that for a while. But until know there where technical difficulties that make it harder to merge the sfl4j-simple jar into m2e.maven.runtime than keeping it separate.
But I had a new idea that should make it simple. I'm currently trying it out and will let you know if that has worked.
If it works as I intend it, then it should be a simple change to use `maven-slf4j-provider`.

HannesWell: https://github.com/eclipse-m2e/m2e-core/pull/875

HannesWell:
Just created https://github.com/eclipse-platform/eclipse.platform.debug/issues/76 to handle the issue of making ConsoleLineTrackers work with ANSI-escape sequences in general.
Until this is completed in Eclipse-Platform-Debug I think we should merge this now to keep the link creation and automatic attachment of debuggers working.

HannesWell:
Until https://github.com/eclipse-platform/eclipse.platform.debug/issues/76 is resolved we should have this change to keep the console-line tracking working with colored print-out.

Issue #118, opened 2021-03-30
Bug 486737 - Support for Maven Extensions (LifeCycleParticipants)


mickaelistria:
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=486737

> It would be nice if m2e supported Maven's lifecycle participants by instantiating them and calling their AbstractMavenLifecycleParticipant.afterProjectsRead() method.
> 
> The other lifecycle methods do not make sense in m2e context.
> 
> To make things as safe as possible, one might explicitly whitelist those participants that shall be called during an m2e build.

laeubi:
I think m2e already supports extensions but not extensions defined in the `.mvn` folder, maybe we should rephrase the ticket?

cpfeiffer:
AFAIK only the patch from this ticket supports lifecycle participant extensions inside m2e. The extensions in the `.mvn` folder are Maven Core Extensions (see #116).

laeubi:
Okay It seems I confused this with
- https://github.com/eclipse-m2e/m2e-core/issues/116

as far as I understood the discussion at the Bugzilla there where concerns regarding the set of "reactor projects" are there any progress regarding that discussion?

laeubi:
I just discovered there is a `org.eclipse.m2e.core.mavenComponentContributors` extension point maybe it would be more suitable for plugins that want to contribute to m2e to add such an extension?

laeubi:
Requires
- https://github.com/eclipse-m2e/m2e-core/issues/546

cpfeiffer:
@laeubi hi from eclipsecon :smile: This is the ticket we were talking about, yesterday. All the details are in the bugzilla issue at https://bugs.eclipse.org/bugs/show_bug.cgi?id=486737

laeubi:
@cpfeiffer Tanks for bringing attention to this long outstanding issue.
It seems we have most things at hand to implement this, to bring this to life I think we should probably do th follwoing two steps first:

1. Enhance the m2e lif-cycle mapping format in a way a plugin can choose if it want to participate or not, we currently already have `pluginExecutions` > `pluginExecution` > `action` what the can take a `configurator` and we should enhance this to additional can contain `lifeCycleParticipants` > `lifeCycleParticipant` mentioning the class that should be enabled in the lif-cycle of project reads and sessions. One might think to also explicticly state the supported methods, but thats probably to much here
[...]

PR #538, opened 2022-01-31
Add packagingTypes extension point

kjsmita6:
Our application has a few custom Maven packaging types that do not show up in the drop-down and must be typed in manually in the pom.xml. This small change adds an extension point that allows the developers of the product/plugins to define packaging types that will be added to the pom editor.

The extension point is defined by the following:
extensionpoint
which will add a packaging type `test-packaging-type`. This type does not actually exist and is only here for testing, which is why the next screenshot shows an error. The plugin developers should make sure that the packaging type actually will exist in the launched product.

Then, in the launched product, see the type in the pom editor drop-down.
ccombo

This is my first contribution to any Eclipse project, so please let me know if I did something incorrectly.

### DIFF
--- a/org.eclipse.m2e.core/plugin.properties
+++ b/org.eclipse.m2e.core/plugin.properties
@@ -44,3 +44,4 @@ extension-point.changed.name = mavenProjectChangedListeners
 extension-point.lifecycleMappingMetadataSource.name = Lifecycle Mapping Metadata Source
 extension-point.projectConversionParticipants.name= Project conversion participants
 extension-point.workspaceClassifierResolvers.name= Workspace classifier resolvers
+extension-point.artifactPackagingTypes.name= Artifact Packaging Types
--- a/org.eclipse.m2e.core/plugin.xml
+++ b/org.eclipse.m2e.core/plugin.xml
@@ -22,6 +22,7 @@
    <extension-point id="conversionEnabler" name="%extension-point.conversionEnabler.name" schema="schema/conversionEnabler.exsd"/>
    <extension-point id="incrementalBuildFrameworks" name="%extension-point.incrementalBuildFrameworks.name" schema="schema/incrementalBuildFrameworks.exsd"/>
    <extension-point id="workspaceClassifierResolvers" name="%extension-point.workspaceClassifierResolvers.name" schema="schema/workspaceClassifierResolvers.exsd"/>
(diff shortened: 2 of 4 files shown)


mickaelistria:
Thanks, we'll try to look at it shortly. We first need a positive review from CI to validate this change doesn't introduce regressions.

laeubi:
I think this is another reincarnation of missing extension support in m2e as packaging types are normally declared via a plugin in the pom already.

mickaelistria:
> I think this is another reincarnation of missing extension support in m2e as packaging types are normally declared via a plugin in the pom already.

That's right!
So I don't think we want a new extension point for that, if this information is supposed to be accessible dynamically from the Maven project.

mickaelistria:
Sorry @kjsmita6 but as @laeubi identified, this proposal is not really correct nor sufficient for m2e. We can achieve much better results by looking up build extensions (that's not trivial though, as #118 tracks). So I think we're going to close this PR.
However you may want to create a new PR adding a comment in the Overview page to remind that with extension support, it would be possible to dynamically list all packaging types that make sense for the current project.

laeubi:
> (that's not trivial though, as #118 tracks

That said, we already have some workarounds for packaging types in place in m2e, maybe this could simply be reused her, take a look at:

https://github.com/eclipse-m2e/m2e-core/blob/840fb2e973d5edc1286d2c27dd24aff9a707f875/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/DefaultMavenDependencyResolver.java#L98-L105

one could the parse these plugins for `components.xml` descriptors declaring new packaging types (thats how maven seems to handle them).

mickaelistria:
That would be very good and it may open some first breaches towards tackling #118.
@kjsmita6 are you interested in implementing the approach suggest by @laeubi , introspecting the project plugins for declared packaging types?

kjsmita6:
@mickaelistria I could be interested in working on this, as it's something we need in our product. I will investigate some things, try and implement them, and create another PR when necessary.

mickaelistria:
@kjsmita6 Thanks! And feel free to ping on issues or on the m2e-dev mailing-list if you need any assistance.

laeubi:
> And feel free to ping on issues or on -the m2e-dev mailing-list- if you need any assistance.

Better use the https://github.com/eclipse-m2e/m2e-core/discussions

PR #541, opened 2022-01-31
[Core] simplify SourceLocationHelper

HannesWell:


mickaelistria:
Looks like this is causing regressions. Can you please investigate?

HannesWell:
> Looks like this is causing regressions. Can you please investigate?

The reason was that the `SourceLocationHelper` now creates `SourceLocation` that have now have a non null resource path and the `SourceLocationHelperTest` had to be adapted accordingly. I have done this in https://github.com/tesla/m2e-core-tests/pull/142

In general I think more resource paths are better because it allows a more precise placement of the error marks in the pom files, for example in the constructor of `MavenProblemInfo` or `MavenProblemInfo.processMarker()`.

mickaelistria:
> In general I think more resource paths are better 

+1
I'll merge the tests and retrigger.

HannesWell:
> > In general I think more resource paths are better
> 
> +1 I'll merge the tests and retrigger.

Thank you Mickael.

PR #548, opened 2022-02-01
[Core|Launching] Simplifications and removal of unused elements

HannesWell:
This PR simplifies code and removes unused elements in m2e.core and m2e.launching.
And additionally strengthens the reference from `m2e.launching` to `org.eclipse.m2e.workspace.cli`.


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenPluginActivator.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2018 Sonatype, Inc. and others.
+ * Copyright (c) 2010, 2021 Sonatype, Inc. and others.
```
akurtakov, line 2: It's 2022 :)

HannesWell: You git me. 😅 I did most of this last year already. Thanks for the hint, I have corrected it.

HannesWell:
I just noticed that this change broke some plugins from the `com.ifedorenko.m2e.mavendev.feature`.
The reason is that it uses the just removed internal method from the internal class `Bundles`:
https://github.com/ifedorenko/com.ifedorenko.m2e.mavendev/blob/3ca383541f66d2859136bfbd40c7c5a36ce56075/com.ifedorenko.m2e.mavendev.launch.ui/src/com/ifedorenko/m2e/mavendev/launch/ui/internal/BuildProgressActivator.java#L293

The plug-ins seems not be developed anymore since the repo has been archived. So I think we have to live with that now.
IIRC I have them installed because they are part of the Tycho Oomph setup. So I think they should be removed from there:
https://github.com/eclipse/tycho/blob/1f15e7e73cbd1425bd957aa2df8e003b26d93491/setup/Tycho.setup#L92

Nevertheless some of the features might be worth to be contributed to m2e directly.

PR #554, opened 2022-02-03
Reference Eclipse Platform repo

mickaelistria:
To resolve JDT in case it's missing

### DIFF
--- a/org.eclipse.m2e.site/category.xml
+++ b/org.eclipse.m2e.site/category.xml
@@ -23,5 +23,6 @@
    </feature>
    <category-def name="m2e" label="Maven Integration for Eclipse"/>
    <repository-reference location="https://download.eclipse.org/wildwebdeveloper/releases/0.13.2/" enabled="true" />
-   <repository-reference location="https://download.eclipse.org/lsp4e/releases/0.20.1/" enabled="true" />
+   <repository-reference location="https://download.eclipse.org/lsp4e/releases/0.20.2/" enabled="true" />
+   <repository-reference location="https://download.eclipse.org/eclipse/updates/4.22/" enabled="true" />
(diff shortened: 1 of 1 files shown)


HannesWell:
When the latest repository is referenced, future updates of old releases may not work because they only refer to the then latest repo and not to the repo current at the time when the repo was created.

mickaelistria:
I'm torn here, as it's usually a good thing to move towards newer release of dependencies...

HannesWell:
> I'm torn here, as it's usually a good thing to move towards newer release of dependencies...

Fully agree and I also try to use the latest dependencies if possible. But I think it should happen in a controlled manner.
And has the potential to create some hard to predict/understand behavior that might de-facto break repositories that just worked, without changing them.

laeubi:
I wonder if we could add some magic to tycho that would simply add [the locations from the target](https://github.com/eclipse-m2e/m2e-core/blob/master/target-platform/target-platform.target)... that way we don't need to manage this at two locations.

Or[ if PDE would support variables](https://bugs.eclipse.org/bugs/show_bug.cgi?id=578430) we could define them in the pom.xml properties and have only placeholders.

HannesWell:
> I wonder if we could add some magic to tycho that would simply add [the locations from the target](https://github.com/eclipse-m2e/m2e-core/blob/master/target-platform/target-platform.target)... that way we don't need to manage this at two locations.
> 
> Or[ if PDE would support variables](https://bugs.eclipse.org/bugs/show_bug.cgi?id=578430) we could define them in the pom.xml properties and have only placeholders.

These are two ideas I also had already and I expect to be very helpful in scenarios like this one!
Is there already an issue for Tycho where it can be discussed?

PR #564, opened 2022-02-05
Fix #563  Make the life-cycle-mapping available as a view in eclipse

### LINKED ISSUE #563: Make the life-ycle-mapping aviable as a view in eclipse
https://github.com/eclipse-m2e/m2e-core/issues/539#issuecomment-1026778207 shows a screenshot of the life-cycle mapping result when `Project Properties > Maven > Lifecycle Mapping`

This is very useful when investigating problems with life-cycle mappings but a bit hard to find/use. e.g. one has to change something, the open the dialog again and so on.

I think it would be useful if this could simply be a view that always shows the data of the current selected project.

laeubi:

### DIFF
--- a/org.eclipse.m2e.core.ui/plugin.properties
+++ b/org.eclipse.m2e.core.ui/plugin.properties
@@ -98,6 +98,7 @@ action.add.plugin.tooltip = Add Maven Plugin
 action.assignWorkingSet.label=Assign Working Sets...
 category.maven.name = Maven
 view.repos.name = Maven Repositories
+view.mappings.name = Maven Lifecycle Mappings
 view.build.name = Maven Workspace Build
 page.maven.name = Maven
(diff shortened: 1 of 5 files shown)


laeubi:
@mickaelistria any concerns?

mickaelistria:
To be honest, I don't think a view for this is worth it, I wouldn't have invested personally in it, nor supported any of my team-mate to invest in it. A project property page as it is right now is fundamentally more correct as people don't need it constantly shown on the IDE, and the lifecycle mappins are usually project information (user can customize that per project).
But we have to value diversity of opinion and preferences in the project, so if you think it's worth the addition and commit to maintenance that goes with it, I don't think any committer is entitled to veto such harmless additions.
So it's only your call here, you don't need my review ;)

Issue #549, opened 2022-02-02
Improve conflict handling of lifecycle mappings


kwin:
In case two extensions with overlapping lifecycle mapping (like PDE Connector and Sling IDE Tooling  in https://github.com/eclipse-m2e/m2e-core/blob/50ed92a39fa68f2749d2ee7cec0bc29af51a1d32/org.eclipse.m2e.pde.connector/lifecycle-mapping-metadata.xml#L39 and  https://github.com/apache/sling-ide-tooling/blob/93bcb69fe7cfad6b68d5caf838a14d3bdf887b92/eclipse/eclipse-m2e-ui/lifecycle-mapping-metadata.xml#L32) only the errors 

```
Conflicting lifecycle mapping (plugin execution "biz.aQute.bnd:bnd-maven-plugin:6.1.0:bnd-process (execution: generate-osgi-metadata, phase: process-classes)"). To enable full functionality, remove the conflicting mapping and run Maven->Update Project Configuration.
Plugin execution not covered by lifecycle configuration: biz.aQute.bnd:bnd-maven-plugin:6.1.0:bnd-process (execution: generate-osgi-metadata, phase: process-classes)
```

are emitted and incremental support for bnd-process is completely disabled.

Instead in case of conflicts it should be possible for a user to decide on a per-project level which lifecycle mapping should be used.

laeubi:
@kwin thanks for your suggestion, do you plan to give it a try and provide a PR for this?

laeubi:
@mickaelistria @HannesWell could you help here? I think this can become important once we release 1.20 as there might also be collisions with the old takari mapping for tycho.

mickaelistria:
I don't plan to work on this issue. If the solution is to remove the old takari mapping of Tycho, then we can imagine have the m2e-PDE feature adding a negative requirement towards the takari stuff so we at least ensure both cannot be installed together.

laeubi:
Okay I'll try to take a look at this I think this is crucial for this release as we will otherwise get complains from users and the current behavior is simply not helpful.

HannesWell:
> Okay I'll try to take a look at this I think this is crucial for this release as we will otherwise get complains from users and the current behavior is simply not helpful.

Thank you Christoph! I agree that this is important but I actually didn't have the plan to work on this, so I'm thankful that you take over. If you need somebody to review, please let me know.

laeubi:
This is how it will look like for a execution phase with duplicate mappings:

[...]

PR #565, opened 2022-02-05
Improve conflict handling of lifecycle mappings #549

laeubi:
This is a first step improving the following things

- record what is causing the conflict
- add the exception to the maven problem
- show problem markers at the correct location in the pom
- store the conflicting bundles at the marker

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/lifecyclemapping/DuplicateLifecycleMappingMetadataException.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/lifecyclemapping/DuplicateLifecycleMappingMetadataException.java
@@ -1,0 +1,44 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Christoph Läubrich and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *      Christoph Läubrich - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.m2e.core.internal.lifecyclemapping;
+
+import java.util.Arrays;
+
+import org.eclipse.m2e.core.internal.lifecyclemapping.model.LifecycleMappingMetadata;
+import org.eclipse.m2e.core.internal.lifecyclemapping.model.LifecycleMappingMetadataSource;
+
+
+/**
+ * DuplicateLifecycleMappingMetadataException
+ *
+ */
+public class DuplicateLifecycleMappingMetadataException extends DuplicateMappingException {
+
+  private static final long serialVersionUID = 1L;
+  private LifecycleMappingMetadata[] lifecyclemappings;
+
+  public DuplicateLifecycleMappingMetadataException(LifecycleMappingMetadata... lifecyclemappings) {
+    super(Arrays.stream(lifecyclemappings).map(LifecycleMappingMetadata::getSource)
+        .toArray(LifecycleMappingMetadataSource[]::new));
+    this.lifecyclemappings = lifecyclemappings;
+  }
+
+  /**
+   * @return Returns the lifecyclemappings.
+   */
+  public LifecycleMappingMetadata[] getConflictingMappings() {
+    return this.lifecyclemappings;
+  }
+
+}
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/lifecyclemapping/DuplicateMappingSourceProblem.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/lifecyclemapping/DuplicateMappingSourceProblem.java
@@ -1,0 +1,82 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Christoph Läubrich and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *      Christoph Läubrich - initial API and implementation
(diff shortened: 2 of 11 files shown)


laeubi:
As this already has visual and useful enhancement for the user I'd like to merge this as a first step.

The next step then should be to have a quick-fix to actually ignore a certain conflicting mapping.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/lifecyclemapping/DuplicatePluginExecutionMetadataException.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.eclipse.m2e.core.internal.lifecyclemapping.model.LifecycleMappingMetadataSource;
+import org.eclipse.m2e.core.internal.lifecyclemapping.model.PluginExecutionMetadata;
+
+
+/**
+ * DuplicatePluginExecutionMetadataException
+ *
+ */
+public class DuplicatePluginExecutionMetadataException extends DuplicateMappingException{
```
HannesWell, line 25: Here Eclipse complains about a missing "static final serialVersionUID field of type long" as well.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/lifecyclemapping/DuplicateMappingSourceProblem.java
```diff
@@ -0,0 +1,82 @@
...
+    Bundle[] bundles = Arrays.stream(conflictingSources).map(LifecycleMappingMetadataSource::getSource)
+        .filter(Bundle.class::isInstance).map(Bundle.class::cast).toArray(Bundle[]::new);
+    marker.setAttribute(IMavenConstants.MARKER_DUPLICATEMAPPING_SOURCES, bundles.length);
+    for(int i = 0; i < bundles.length; i++ ) {
+      Bundle bundle = bundles[i];
+      String prefix = IMavenConstants.MARKER_DUPLICATEMAPPING_SOURCES + "." + i + ".";
+      Version version = bundle.getVersion();
+      String v = version.getMajor() + "." + version.getMinor() + "." + version.getMicro();
+      marker.setAttribute(prefix + "bundleName",
+          Objects.requireNonNullElse(bundle.getHeaders().get(Constants.BUNDLE_NAME), bundle.getSymbolicName()));
```
HannesWell, line 76: Why do you check the headers first? Can the bundle.getSymbolicName() be `null`?

HannesWell (commented):
This is indeed already an improvement and the code looks good in general, I just added few minor remarks.
Do you have a simple example that can be used as a test-case?

Some places like `FailedMappingMetadataSource` could be polished a bit, IMHO the (non-) javadoc is not very useful. Also the constructors of MavenProblemInfo could be further streamlined. I can add corresponding suggestions here or add a follow up PR. As you prefer.

laeubi:
> I can add corresponding suggestions here or add a follow up PR.

Follow Up is a bit better, you might also want to take a look at the eclipse settings here we ship they generate javadoc with copyright info of sonatype in 2010 whenever one creates a new file ...

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/lifecyclemapping/DuplicateMappingSourceProblem.java
laeubi: I'd just like to get the "name" of the bundle as a human readable name, so we later can use this in a quickfix and falling back to the symbolic name if non is given. The symbolic name should always be there.

laeubi:
> Do you have a simple example that can be used as a test-case?

It is really a bit tricky to actually test this, I have done the following:

1. Import the https://github.com/tesla/m2eclipse-tycho into my workspace
2. Add a stripped down version of [the sling tooling ](https://github.com/apache/sling-ide-tooling/tree/93bcb69fe7cfad6b68d5caf838a14d3bdf887b92/eclipse/eclipse-m2e-ui) to the workspace where I deleted all but `BndMavenPluginProjectConfigurator` and `MavenBundlePluginProjectConfigurator` and replaced them with a dummy implementation doing simply nothing.
3. Started the dev-eclipse and created a bnd and felix bundle project [test-projects.zip](https://github.com/eclipse-m2e/m2e-core/files/8012721/test-projects.zip)
.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/lifecyclemapping/DuplicateLifecycleMappingMetadataException.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.eclipse.m2e.core.internal.lifecyclemapping.model.LifecycleMappingMetadata;
+import org.eclipse.m2e.core.internal.lifecyclemapping.model.LifecycleMappingMetadataSource;
+
+
+/**
+ * DuplicateLifecycleMappingMetadataException
+ *
+ */
+public class DuplicateLifecycleMappingMetadataException extends DuplicateMappingException {
```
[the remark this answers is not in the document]
laeubi: Done.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/lifecyclemapping/DuplicatePluginExecutionMetadataException.java
laeubi: Done

laeubi:
@HannesWell rebased and updated, if the build is fine and there are no further comments I'd like to merge this to start with the quick-fixes.

HannesWell:
> > I can add corresponding suggestions here or add a follow up PR.
> 
> Follow Up is a bit better, you might also want to take a look at the eclipse settings here we ship they generate javadoc with copyright info of sonatype in 2010 whenever one creates a new file ...

Alright 👍🏽 
That should definitely be fixed. I already thought multiple times that we should unify the settings (especially regarding formatting) in the entire m2e-project.

HannesWell (approved):
Thanks for sharing the steps to reproduce. Looks indeed a bit cumbersome, so I will skip this part for now.

> @HannesWell rebased and updated, if the build is fine and there are no further comments I'd like to merge this to start with the quick-fixes.

No further comments from my side. Thank you!

PR #568, opened 2022-02-07
Add a quickfix for conflicting life-cycle mappings to be ignored on the workspace

laeubi:
This adds a quickfix that if applied filters unwanted life-cycle mapping sources based on the goal or packaging type.

### DIFF
--- a/org.eclipse.m2e.core/mdo/lifecycle-mapping-metadata-model.xml
+++ b/org.eclipse.m2e.core/mdo/lifecycle-mapping-metadata-model.xml
@@ -43,6 +43,13 @@
             <multiplicity>*</multiplicity>
           </association>
         </field>
+        <field>
+          <name>lifecycleMappingFilters</name>
+          <association>
+            <type>LifecycleMappingFilter</type>
+            <multiplicity>*</multiplicity>
+          </association>
+        </field>
       </fields>
       <codeSegments>
         <codeSegment>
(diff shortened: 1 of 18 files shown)


kwin:
How to edit/remove the filter once it is set for a project? Is there any UI hooks planned apart from the quickfix for adding?

kwin:
A project-specific filter might be more useful than a workspace-specific one...

laeubi:
> How to edit/remove the filter once it is set for a project?

grafik

>  Is there any UI hooks planned apart from the quickfix for adding?

I don't think so, even though it would be possible to e.g. add a dedicated editor for such kind of things.

> A project-specific filter might be more useful than a workspace-specific one...

Technically it would already work to add this to the pom as well, but this would require another quickfix see:

- [LifecycleMappingResolution.java](https://github.com/eclipse-m2e/m2e-core/blob/master/org.eclipse.m2e.editor/src/org/eclipse/m2e/editor/internal/lifecycle/LifecycleMappingResolution.java)
- [LifecycleMappingOperation](https://github.com/eclipse-m2e/m2e-core/blob/master/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/editing/LifecycleMappingOperation.java)

but this PR would be for Workspace only right now.

laeubi:
> A project-specific filter might be more useful than a workspace-specific one...

By the way: I'm not sure if a per-project setting it actually useful. These connectors are installed in the Workspace and I can't think about a use-case where I want one connector for one project and another for another project.

Its more that (in the case of `m2e` + `sling`) I can express a preference what of both connector I prefer to do the job for me if there are conflicts (that's why one need to choose individual goals/phases because that might differ e.g. one connector might provide superior features for a certain phase or packaging).

If I add something on the project level (and then check in the pom) it might be that a users has only installed `m2e` or `sling` and if the one the user has is excluded might be confused.

laeubi:
@HannesWell @mickaelistria  this is fine from my side and ready to be merged.

target-platform/target-platform.target
```diff
@@ -45,6 +45,7 @@
 			<repository location="https://download.eclipse.org/cbi/updates/license/"/>
 			<unit id="org.eclipse.license.feature.group" version="0.0.0"/>
 		</location>
+		<location path="${eclipse_home}" type="Profile"/>
```
mickaelistria, line 48: Is this needed as part of this change?

laeubi: No this should not be part of the change...

mickaelistria:
No objection to merge.

HannesWell:
No objections from me either.
Like in the preceding PR due to a lack of simple example I cannot test this quickly but looking over the code and from your description there is no reason to further delay this.

Issue #546, opened 2022-02-01
Provide an IMaven per project


laeubi:
Currently IMaven is accessed via a static singleton.

We should change this in the following way:

1. There should be an adapter that allows adapting an IProject -> IMaven when the project is a m2e project we return a project scoped IMaven (could as a first step always return the static global one)
2. Register the global IMaven as a service
3. remove and/or make the accessor for IMaven internal

HannesWell:
For clarification: What is the advantage?

> 
>     1. There should be an adapter that allows adapting an IProject -> IMaven when the project is a m2e project we return a project scoped IMaven (could as a first step always return the static global one)
> 
This could be done similar like in JDT where one can get a `IJavaProject` using `JavaCore.create(IProject)`.

In general I like it when there are only few, ideally just one, entry-points into a 'topic' from where I can get everywhere through a few intuitive steps. It makes it much easier to orientate oneself compared to when everything is distributed over multiple places that I all have to know.

laeubi:
Adapters are much more flexible and do allow a loose coupling of things without exposing API (and you know how hard it is to change API ...)

And I don't see much value of `MavenPlugen.getInstance().getMaven(...)` over `Adapters.adapt(project, IMaven.class)`

HannesWell:
> Adapters are much more flexible and do allow a loose coupling of things without exposing API (and you know how hard it is to change API ...)
> 
It's indeed hard. But what I always wonder about Adapters is that by providing a way as you suggested, you de-facto make it _API_ that that call will always return a corresponding `IMaven` object. You cannot simply change that an return for example an `IMaven2` object instead or just nothing. API is not only about class names and method signatures it also about behavior and behavior is IMHO much harder to enforce because it require extensive tests.

> And I don't see much value of `MavenPlugen.getInstance().getMaven(...)` over `Adapters.adapt(project, IMaven.class)`

I would suggest to make `getMaven()` a static method of `MavenPlugen`. Then `MavenPlugin` could one of the few entry-points into the Maven-world mentioned above. If there are more methods one can then enter `MavenPlugen` and see what the ContentAssist suggests me respectively what's available. With the Adapter approach it would have to be documented somewhere else (but where?) or I would have to guess it. And documentation tends to be outdated.
But this is a general question I also have for the Service approach and I could not yet get an answer for.

laeubi:
[...]

Issue #543, opened 2022-02-01
Additional "runtime" scoped dependencies support?


laeubi:
### Discussed in https://github.com/eclipse-m2e/m2e-core/discussions/461

<div type='discussions-op-text'>

<sup>Originally posted by **laeubi** December 18, 2021</sup>
Now that we correctly filter dependencies scopes it shows that it might cause confusion for[ runtime scoped dependencies](https://github.com/eclipse/tycho/issues/447).

The `runtime` scope is a bit special as maven [defines it as:](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html)

> runtime
> This scope indicates that the dependency is not required for compilation, but is for execution. Maven includes a dependency with this scope in the runtime and test classpaths, but not the compile classpath.

That mean that if we want it currently one has to select `test` scope.

I see the following options:

1. We allow to select `runtime` scope and handle it as something between `compile` and `test` (include everything `compile` would include and include `runtime` dependencies but not `test` scoped ones.
2. We add a boolean attribute `includeRuntime` that additionally include runtime scoped dependencies 
3. We do nothing an require people to explicitly include those if the only want compile scope

@akurtakov @mickaelistria @HannesWell any preferences?</div>

@akurtakov @mickaelistria @HannesWell I think it would be good t have this in before the next release see also
 - https://github.com/eclipse/tycho/issues/447

mickaelistria:
Solution 1 seems to be the most correct from Maven and PDE POV. The "runtime" scope should also be the default scope in the UI as per expectation of target platform definition.

laeubi:
I'll try to fix this this week, as I think we should include thin the the 1.20 release.

laeubi:
Target used for testing [Issue_543.target.zip](https://github.com/eclipse-m2e/m2e-core/files/8000846/Issue_543.target.zip)

laeubi:
@vogella @akurtakov @HannesWell this should shortly be available in the snapshot please test this if it works for all your cases (but keep in mind that tycho do *not* support this at the moment!) if no issues arise I'll adjust tycho support accordingly.

laeubi:
Also fixed in tycho now: https://github.com/eclipse/tycho/issues/447

PR #569, opened 2022-02-07
[Core][Projects] Get packaging types from build plugins

kjsmita6:
This is an improved version of #538. Instead of creating an extension point, the project will now look through all of its build/extension plugins, open the jar files, and find/parse components.xml (as suggested by @laeubi ). 

Two things I am unsure of:
1. I'm not super knowledgeable in Maven/m2e internals. I know enough about it, but not as much as might be required. I tested this code with my own product's custom packaging types and it was able to load all of them into the pom editor from components.xml. I'm not sure if everyone has it set up the same way. If not, this probably won't work.
2. I don't think m2e can use any plexus XML parsing code/if there is a way to parse components.xml already. If there is I can remove my own parsing code in favor of an existing one.

Again, please let me know if anything can be changed/improved. I am happy to have feedback and fix whatever is necessary. This is something we really need in our product so I would really like to have it included in m2e.

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
@@ -24,16 +24,6 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.jar.JarEntry;
-import java.util.jar.JarFile;
-
-import javax.xml.parsers.DocumentBuilder;
-import javax.xml.parsers.DocumentBuilderFactory;
-
-import org.w3c.dom.Document;
-import org.w3c.dom.Element;
-import org.w3c.dom.Node;
-import org.w3c.dom.NodeList;
 
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
@@ -41,12 +31,8 @@ import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
 
-import org.apache.maven.artifact.Artifact;
-import org.apache.maven.artifact.DefaultArtifact;
-import org.apache.maven.artifact.handler.DefaultArtifactHandler;
 import org.apache.maven.artifact.repository.ArtifactRepository;
 import org.apache.maven.execution.MavenSession;
-import org.apache.maven.model.Plugin;
 import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.project.MavenProject;
 
@@ -54,7 +40,6 @@ import org.eclipse.m2e.core.embedder.ArtifactKey;
 import org.eclipse.m2e.core.embedder.ArtifactRef;
 import org.eclipse.m2e.core.embedder.ArtifactRepositoryRef;
 import org.eclipse.m2e.core.embedder.IMaven;
-import org.eclipse.m2e.core.internal.MavenPluginActivator;
 import org.eclipse.m2e.core.lifecyclemapping.model.IPluginExecutionMetadata;
 import org.eclipse.m2e.core.project.IMavenProjectFacade;
 import org.eclipse.m2e.core.project.MavenProjectUtils;
@@ -73,12 +58,6 @@ public class MavenProjectFacade implements IMavenProjectFacade, Serializable {
   public static final String PROP_LIFECYCLE_MAPPING = MavenProjectFacade.class.getName() + "/lifecycleMapping";
 
   public static final String PROP_CONFIGURATORS = MavenProjectFacade.class.getName() + "/configurators";
-
-  private static final String COMPONENTS_PATH = "META-INF/plexus/components.xml";
-
-  private static final String ARTIFACT_HANDLER_ROLE = "org.apache.maven.artifact.handler.ArtifactHandler";
-
-  private final String[] PACKAGING_TYPES_IGNORING = new String[] {"jar", "war", "ear", "ejb", "pom", "maven-plugin"};
 
   private final ProjectRegistryManager manager;
 
@@ -114,8 +93,6 @@ public class MavenProjectFacade implements IMavenProjectFacade, Serializable {
   private final Set<ArtifactRepositoryRef> artifactRepositories;
 
   private final Set<ArtifactRepositoryRef> pluginArtifactRepositories;
-
-  private Set<String> availablePackagingTypes;
 
   // TODO make final
   private Set<ArtifactRef> artifacts; // dependencies are resolved after facade instance is created
@@ -168,8 +145,6 @@ public class MavenProjectFacade implements IMavenProjectFacade, Serializable {
       this.pluginArtifactRepositories.add(new ArtifactRepositoryRef(repository));
     }
 
-    this.availablePackagingTypes = new LinkedHashSet<>();
-
     timestamp = new long[ProjectRegistryManager.METADATA_PATH.size() + 1];
     IProject project = getProject();
     int i = 0;
@@ -205,8 +180,6 @@ public class MavenProjectFacade implements IMavenProjectFacade, Serializable {
     this.artifactRepositories = new LinkedHashSet<>(other.artifactRepositories);
 
     this.pluginArtifactRepositories = new LinkedHashSet<>(other.pluginArtifactRepositories);
-
-    this.availablePackagingTypes = new LinkedHashSet<>(other.availablePackagingTypes);
 
     this.timestamp = Arrays.copyOf(other.timestamp, other.timestamp.length);
   }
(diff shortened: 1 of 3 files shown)


laeubi:
> 1\. I'm not sure if everyone has it set up the same way. If not, this probably won't work.

I think it don't has to work for everyone, we can improve it later if there are other means of defining this. Also with #546 we might be able to leverage other techniques if full extension support is available.

> 2\. I don't think m2e can use any plexus XML parsing code/if there is a way to parse components.xml already.

I think standard java xml parsing is fine and might even be more flexible

org.eclipse.m2e.editor/src/org/eclipse/m2e/editor/pom/OverviewPage.java
```diff
@@ -349,6 +354,17 @@ private void createArtifactSection(FormToolkit toolkit, Composite composite, Wid
     artifactPackagingCombo.add("ear"); //$NON-NLS-1$
     artifactPackagingCombo.add("pom"); //$NON-NLS-1$
     artifactPackagingCombo.add("maven-plugin"); //$NON-NLS-1$
+
```
laeubi, line 357: Maybe it would be worth here to use a TreeSet<String> to first collect all packing types (initialized with the hard-coded defaults) so there is no need for filtering and no possible duplicates.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
```diff
@@ -59,6 +74,12 @@
 
   public static final String PROP_CONFIGURATORS = MavenProjectFacade.class.getName() + "/configurators";
 
+  private static final String COMPONENTS_PATH = "META-INF/plexus/components.xml";
+
+  private static final String ARTIFACT_HANDLER_ROLE = "org.apache.maven.artifact.handler.ArtifactHandler";
+
+  private final String[] PACKAGING_TYPES_IGNORING = new String[] {"jar", "war", "ear", "ejb", "pom", "maven-plugin"};
```
laeubi, line 81: I think this is not necessary if we make the UI collecting everything in a Set first (see comment below).

laeubi (commented):
I think this is fine, just one minor adjustment.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
```diff
@@ -59,6 +74,12 @@
 
   public static final String PROP_CONFIGURATORS = MavenProjectFacade.class.getName() + "/configurators";
 
+  private static final String COMPONENTS_PATH = "META-INF/plexus/components.xml";
+
+  private static final String ARTIFACT_HANDLER_ROLE = "org.apache.maven.artifact.handler.ArtifactHandler";
```
mickaelistria, line 79: Let's drop it and directly use `ArtifactHandler.ROLE` from Maven API

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
```diff
@@ -59,6 +74,12 @@
 
   public static final String PROP_CONFIGURATORS = MavenProjectFacade.class.getName() + "/configurators";
 
+  private static final String COMPONENTS_PATH = "META-INF/plexus/components.xml";
```
mickaelistria, line 77: I'm annoyed I didn't find constant in Plexus or Maven for that...

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
mickaelistria, line 81: Ideally, the solution shouldn't require such exception and would be generic enough to work without them. (It's not a blocker for a merge though)

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/IMavenProjectFacade.java
```diff
@@ -156,4 +156,18 @@ MojoExecution getMojoExecution(MojoExecutionKey mojoExecutionKey, IProgressMonit
...
+  void updateAvailablePackagingTypes();
+
+  /**
+   * Gets any custom packaging types found using updateAvailablePackagingTypes. If this method is called while
+   * availablePackagingTypes is empty, updateAvailablePackagingTypes will be called
+   * 
+   * @see IMavenProjectFacade#updateAvailablePackagingTypes()
+   * @return A set of all custom packaging types available to this project, or an empty set if none are found
+   */
+  Set<String> getAvailablePackagingTypes();
```
mickaelistria, line 172: Since this is only useful for the UI (OverviewPage), I don't think we should make it an API here, and instead we should probably just move the code to OverviewPage directly.

mickaelistria (changes requested):
It's great progress.
However, all that should be moved to Overview page as it's not really useful as part of the IMavenProjectFacade in general and we prefer keeping such important API minimal and more focused.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/IMavenProjectFacade.java
laeubi: I think `getAvailablePackagingTypes()` is good here but  `updateAvailablePackagingTypes()` should not be part of the API.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
laeubi: I even don't found the place where it is parsed ...

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
laeubi: See my suggestions, I think the overview page should simply use a Set<String> and we should not bother here....

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/IMavenProjectFacade.java
laeubi: Why I think its useful in the API is that I would expect in the future that this information is read via the maven-extension mechanism and then we don't want to remember the Overviewpage uses some kind of own solution.
Also we probably should check elsewhere in the code if we have hard-coded defaults that better call this method instead.

mickaelistria: > Why I think its useful in the API is that I would expect in the future that this information is read via the maven-extension mechanism and then we don't want to remember the Overviewpage uses some kind of own solution.

That's not how API are supposed to be growing. Some code is promoted as an API if there is a clear need from it, from multiple consumers and if it makes sense to host the logic for those clients into the object because of some strong coupling. It's not the case here. IMavenProjectFacade represents one project and its info. The available types is most often useless for a project and not really related to the project but more to the classrealm, the set is moveable so it's unclear from API what to expect as value... It could be some utility class somewhere, but definitely not a getter in IMavenProjectFacade.
At this stage, there is no good reason to make it API, so we should avoid future pain and not make it API, at least not that way.
The more API != the better.

laeubi:
@mickaelistria I think we should include this in 1.20, there is not much to be left, it is just a bit moving things here or there ... and it is local enough to have low risk for regression.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/IMavenProjectFacade.java
laeubi: Then lets move it to OverviewPage but include it in 1.20 ...

laeubi:
@kjsmita6 do you think you can move the code to the OverviewPage and update your PR as of today?

mickaelistria:
It's not ready and I don't want to wait further with the release. I see a bad trend of the release scope growing everyday. At some point, if we want 2.0 to happen, we need to make 1.20 highest priority.
If we wait for this one to be complete, then another interesting change will come up, and we'll then wait for this other interesting change and so on. It's already been so for the 3 last weeks. We don't have 3 more weeks to waste if we want 1.20 in next SimRel.

laeubi:
> If we wait for this one to be complete, then another interesting change will come up, and we'll then wait for this other interesting change and so on. It's already been so for the 3 last weeks. We don't have 3 more weeks to waste if we want 1.20 in next SimRel.

Didn't we say we don't want to be coupled to SimRel time-schedule? :-)
I don't say we should include other poping up but this one.

akurtakov:
We don't want to be coupled as there are many other things to release at end of cycle thus Mickael tried to release it earlier but it ended up with endless feature growth. If 1.20 don't make it 2022-03 there is no point in even releasing it as 2022-06 will have 2.0 and many(most?) users will not get 1.20 at all.

laeubi:
> as 2022-06 will have 2.0

Sorry of being pessimistic but I don't see any guarantee that we will have 2.0 in 2022-06 as its completely unclear how far we got so maybe 2022-06 will still ship 1.20, that's why I have completed some important features (especially for supporting pom-dependencies) and those where people otherwise would complain (conflicting mappings).

And I don't see any 'endless' growth' there is a clear decrease in PRs targeting for 1.20 and waiting for this where a user has actually invest time for a completely different solution to came up with a important improvement for his usecase sound fair and waiting for at least another 3 Months until it bubbles up in the IDE seems not very profitable.

akurtakov:
2022-06 should ship 2.0 - it's a matter of getting rid of all old stuff that's planned and get 2.0 out. Doable even in April IMHO. Any feature addition that don't make it should be for 2.1,... It's important to deliver faster rather than prolong releases.

laeubi:
> It's important to deliver faster rather than prolong releases.

But that's impossible with current eclipse-rules ... anyways I don't get why we repel first-time contributors due to delay their contribution because of a unclear "feeling" delaying the 1.20 at one or two days won't harm if we simply agree not including anything else.

If someone is eager to start right now, one could simply do it on a branch and we merge it in after the release, this change is local enough to be not affected by any API changes.

laeubi:
> Mickael asked for 1.20 release January 21st - a release should have been done in the next few days.

Yeah bust asking for a release is not enough, I asked for a planed date and it was said we should not plan a date, its a bit late to complain now that there is a date (now) but that was never communicated ... Also one should check tickets and assign milestones, the code was simply not ready at this point of time to be released,

> someone else could have picked the releng had and do 1.20.1 or 1.21 in two weeks and repeated even one more time if needed.

someone sometimes aka never ... I just can strongly encourage to include this as it is clear and small and valuable and unlikely we will ever do a 1.20.1 once we started with the 2.0 stuff. Anyways of course no one is oblique to follow this advice, its just makes me sad that we often have this 'If I'm not affected I don't care' habit. For me https://github.com/eclipse-m2e/m2e-core/issues/549 was a clear blocker as we would otherwise would get instant complains from m2e users once this is released to a wider audience. The second one was https://github.com/eclipse-m2e/m2e-core/issues/543 what will be crucial if we want to support maven locations in the upcoming 2022-06 and I can't really find anything else that was done for the 1.20 in the meantime so its really a bit unfair to claim we are adding "endless new features" here...

laeubi:
> 2022-06 should ship 2.0

Could you then please add the latest date we should have the 2.0 available[ here as a due-date](https://github.com/eclipse-m2e/m2e-core/milestones/5/edit)?

akurtakov:
> > 2022-06 should ship 2.0
> 
> Could you then please add the latest date we should have the 2.0 available[ here as a due-date](https://github.com/eclipse-m2e/m2e-core/milestones/5/edit)?

Done. May 24th is the latest date at which release should be public in order to make it to M3 as per https://wiki.eclipse.org/Category:SimRel-2022-06 . Wed is last contribution day in these weeks. Thu is aggregation. Fri is going public.

laeubi:
> > > 2022-06 should ship 2.0
> > 
> > 
> > Could you then please add the latest date we should have the 2.0 available[ here as a due-date](https://github.com/eclipse-m2e/m2e-core/milestones/5/edit)?
> 
> Done. May 24th is the latest date at which release should be public in order to make it to M3 as per https://wiki.eclipse.org/Category:SimRel-2022-06 . Wed is last contribution day in these weeks. Thu is aggregation. Fri is going public.

Thanks, I have for now set Friday for 1.20 please adjust if that seems unrealistic.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/IMavenProjectFacade.java
kjsmita6: Moved to overview page.

org.eclipse.m2e.editor/src/org/eclipse/m2e/editor/pom/OverviewPage.java
```diff
@@ -342,13 +365,17 @@ private void createArtifactSection(FormToolkit toolkit, Composite composite, Wid
...
     artifactPackagingCombo = new CCombo(artifactComposite, SWT.FLAT);
+    packagingTypes = new LinkedHashSet<>();
+
+    packagingTypes.add("jar"); //$NON-NLS-1$
+    packagingTypes.add("war"); //$NON-NLS-1$
+    packagingTypes.add("ejb"); //MNGECLIPSE-688 : add EAR & EJB Support //$NON-NLS-1$
+    packagingTypes.add("ear"); //$NON-NLS-1$
+    packagingTypes.add("pom"); //$NON-NLS-1$
+    packagingTypes.add("maven-plugin"); //$NON-NLS-1$
+    updateAvailablePackagingTypes(); // dynamically load available packging types from build plugins
```
laeubi, line 376: I think you could simply place the code from updateAvailablePackagingTypes() right here and safe the field `packagingTypes` then and use a local variable instead.

laeubi (changes requested):
Just a minor adjustment.

org.eclipse.m2e.editor/src/org/eclipse/m2e/editor/pom/OverviewPage.java
kjsmita6: Have this done but just waiting for final review from @mickaelistria so I can just push all changes at once.

laeubi: Just squash your commits and push them here, so we can see the current status for this

mickaelistria: +1 with @laeubi here, please push your progress without waiting for anyone. FWIW, I'm fine with this new patch, and would also like to see the changes suggested by @laeubi implemented. After that, I think it's find to merge as it's an isolated change bringing a great feature.

kjsmita6: Alright, pushed my change. Everything should be here now.

mickaelistria (approved):
That's a great improvement to me! Thanks a lot.
@laeubi OK to merge?

laeubi (approved):

PR #571, opened 2022-02-08
Make Eclipse-M2E launch configuration feature based

HannesWell:
This PR changes the Eclipse-M2E launch configuration to be based on features.

Compared to before where all workspace plug-ins were part of the launched eclipse this has the following advantages 
- test bundles are not part of the launched Eclipse
- the logback configuration plugin is activated automatically
- only the LemMinX editor feature is included and the old editor is ignored.
- other plug-ins in the workspace that are unrelated to M2E are not part of the launched Eclipse
- One recognizes immediately if a new plug-in is not added to a corresponding feature.

This improved launch-configuration relies on recent changes in PDE that were published with Eclipse 2022-03 Milestone 2.
Since  2022-03 M2 is already part of the Oomph Product Catalog I think it is acceptable to require M2E contributors to use that?

@laeubi this is a little taste of how simple you can create an M2E product/launch configuration in the future (the product part is not yet fully completed).

Site note: the bundles `org.apache.xerces` and `org.apache.xml.resolver` are only included because PDE currently fails to resolve it successfully which is why they are not added automatically.

### DIFF
--- a/setup/Eclipse-M2E-IDE.launch
+++ b/setup/Eclipse-M2E-IDE.launch
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <launchConfiguration type="org.eclipse.pde.ui.RuntimeWorkbench">
+    <setAttribute key="additional_plugins">
+        <setEntry value="org.eclipse.m2e.logback.configuration:1.16.3.qualifier:default:true:default:true"/>
+    </setAttribute>
     <booleanAttribute key="append.args" value="true"/>
     <booleanAttribute key="askclear" value="true"/>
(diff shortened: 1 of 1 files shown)


laeubi:
> This improved launch-configuration relies on recent changes in PDE that were published with Eclipse 2022-03 Milestone 2.
> Since 2022-03 M2 is already part of the Oomph Product Catalog I think it is acceptable to require M2E contributors to use that?

I think we should only use released versions for m2e development.

HannesWell:
Since Eclipse 2022-03 has been released today, we can merge this now.
The good news is that `org.apache.xerces`/`org.apache.xml.resolver` are now resolved successfully and therefore don't have to be added explicitly anymore.

PR #572, opened 2022-02-08
[Console] support eclipse-plugin type projects build with Tycho

HannesWell:
With this change the `MavenConsoleLineTracker` is enhanced to support `Eclipse Plug-in` projects build with Tycho.
As a consequence all recently added/enhanced features like the test-report and project links as well as automatic debugger attachment now wok for `Eclipse Plug-in` projects too.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,19 @@
 # Eclipse m2e - Release notes
 
+## 2.1.0
+
+* 📅 Release Date: _expected_ December 2022
+* All changes: https://github.com/eclipse-m2e/m2e-core/compare/2.0.2...2.1.0
+
+### Console support for polyglot Maven projects and projects without Maven nature
+
+The M2E Maven-Console now also supports tracking of so called _polyglot_ Maven-projects. Those are projects that don't have a standard `pom.xml` and whose Maven-model is instead created from another source. One prominent example in the Eclipse world are Eclipse-PDE projects that are build with Tycho(-pomless).
+
+Due to this new support, polyglot Maven projects now also benefit from the 
+[Improved links to JUnit test-reports and project file in the Console](https://github.com/eclipse-m2e/m2e-core/blob/master/RELEASE_NOTES.md#improved-links-to-junit-test-reports-and-project-pomxml-in-the-console-of-a-maven-build)
+as well as the capability to
+[Automatically launch and attach Remote-Application-Debugger when Maven plug-in starts a forked JVM that waits for a debugger](https://github.com/eclipse-m2e/m2e-core/blob/master/RELEASE_NOTES.md#automatically-launch-and-attach-remote-application-debugger-when-maven-plug-in-starts-a-forked-jvm-that-waits-for-a-debugger) introduced in previous releases.
+
 ## 2.0.2
 
 * 📅 Release Date: August 30th 2022
@@ -83,7 +97,6 @@ Further information, how to activate the debug mode of forked JVMs for the plug-
 - [Tycho-Surefire](https://www.eclipse.org/tycho/sitedocs/tycho-surefire-plugin/test-mojo.html#debugPort)
 - [Tycho-Eclipserun](https://www.eclipse.org/tycho/sitedocs/tycho-extras/tycho-eclipserun-plugin/eclipse-run-mojo.html#jvmArgs)
 
-
 #### Improved links to JUnit test-reports and project pom.xml in the Console of a Maven build
 
 Clicking on the link placed at the name of a running test-class now opens the `JUnit` view that displays the test-reports of the executed tests:
@@ -97,7 +110,6 @@ For each project build a link is now added to the project's headline, which open
 In case of a build failure another link, that opens the `pom.xml` of the failed project, is added to the line that that describes the failure at the very end of the Maven build print-out:
 
 ![grafik](https://user-images.githubusercontent.com/44067969/152247987-01ee209a-ad6c-454e-92c1-e2aa75388931.png)
-
 
 #### the m2e-pde editor now supports generation of a feature from a location:
 
--- a/m2e-maven-runtime/org.eclipse.m2e.maven.runtime/pom.xml
+++ b/m2e-maven-runtime/org.eclipse.m2e.maven.runtime/pom.xml
@@ -162,6 +162,7 @@
 							-exportcontents: \
 								META-INF.plexus;-noimport:=true;x-internal:=true,\
 								META-INF.sisu;-noimport:=true;x-internal:=true,\
+								org.eclipse.m2e.internal.maven.listener;x-friends:="org.eclipse.m2e.launching",\
 								org.apache.maven.*;provider=m2e;mandatory:=provider,\
 								org.codehaus.plexus.*;provider=m2e;mandatory:=provider,\
 								org.sonatype.plexus.*;provider=m2e;mandatory:=provider,\
@@ -186,5 +187,22 @@
 				</plugin>
 			</plugins>
 		</pluginManagement>
+		<plugins>
+			<plugin>
+				<groupId>org.eclipse.sisu</groupId>
+				<artifactId>sisu-maven-plugin</artifactId>
+				<version>0.3.5</version>
+				<executions>
+					<execution>
+						<?m2e execute onIncremental?>
+						<id>index-plexus</id>
+						<phase>process-classes</phase>
+						<goals>
+							<goal>main-index</goal>
+						</goals>
+					</execution>
+				</executions>
+			</plugin>
+		</plugins>
 	</build>
 </project>
--- a/m2e-maven-runtime/org.eclipse.m2e.maven.runtime/src/main/java/org/eclipse/m2e/internal/maven/listener/M2EMavenBuildDataBridge.java
+++ b/m2e-maven-runtime/org.eclipse.m2e.maven.runtime/src/main/java/org/eclipse/m2e/internal/maven/listener/M2EMavenBuildDataBridge.java
@@ -1,0 +1,232 @@
+/********************************************************************************
+ * Copyright (c) 2022, 2022 Hannes Wellmann and others
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Hannes Wellmann - initial API and implementation
+ ********************************************************************************/
+
+package org.eclipse.m2e.internal.maven.listener;
+
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
+import java.nio.ByteBuffer;
+import java.nio.channels.ServerSocketChannel;
+import java.nio.channels.SocketChannel;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.StringJoiner;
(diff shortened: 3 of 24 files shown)


laeubi:
Just one note about the wording: Tycho projects are also Maven projects, there is no such thing as "pure" Maven or "eclipse plugin" maven.
The only difference might be that the _Tycho-Surefire plugin_ emits other messages as the _Maven Surefire Plugin_.

laeubi:
@HannesWell is this planned for 1.20 release? If yes please assign a milestone.

mickaelistria:
> is this planned for 1.20 release? If yes please assign a milestone.

I suggest we just stop planning for 1.20 or we will never deliver it. If it's good to merge, we merge right now, if not we delay to next release; but I don't think such improvement needs to be seen as a blocker for the release.

akurtakov:
M3 is in 2 weeks and projects delivering new features are supposed to be done by that time. Thus any new features for 1.20 should be a total no-go from now on and effort being to release by end of next week.

HannesWell:
> Just one note about the wording: Tycho projects are also Maven projects, there is no such thing as "pure" Maven or "eclipse plugin" maven. The only difference might be that the _Tycho_-Surefire plugin emits other messages as the _Maven_ Surefire Plugin. As both are using surefire under the hood, this might be more a matter of surefire versions anyways.

For this case the relevant difference between Eclipse Plug-in project build using Eclipse's Tycho Plug-ins for Maven and an ordinary Maven project (is ordinary better?) is that the latter is available in the `IMavenProjectRegistry` while the former is not. It is not present either because it has the Maven-nature not attached (which seems to happen frequently for Eclipse Plug-in Projects that only have a pom.xml to enhance the build) or it does not even have a pom.xml file at all (and uses a polyglot file during build).

I changed `pure Maven project` to `ordinary Maven project`. Or do you have a better phrase that is better?

RELEASE_NOTES.md
```diff
@@ -16,6 +16,7 @@ Further information, how to activate the debug mode of forked JVMs for the plug-
 - [Tycho-Surefire](https://www.eclipse.org/tycho/sitedocs/tycho-surefire-plugin/test-mojo.html#debugPort)
 - [Tycho-Eclipserun](https://www.eclipse.org/tycho/sitedocs/tycho-extras/tycho-eclipserun-plugin/eclipse-run-mojo.html#jvmArgs)
 
+This works for ordinary Maven projects as well as for Eclipse Plug-in projects build with Eclipse Tycho.
```
laeubi, line 19: This does not sound correct. There are no 'ordinary' projects compared to tycho projects because they are both "maven" builds.

I would suggest to either remove this or explicitly name the mojos of what build output we support.

From the other changes it seems more that you try to support pomless builds here and those without a maven nature. In this case it has nothing to do with building with tycho I think but more with finding sources in non-m2e projects.
A better aproach might be to use **JDT** to dearch fir the type instead of searching for a certain project making assumptions about the artifact-id.

RELEASE_NOTES.md
```diff
@@ -31,6 +32,7 @@ In case of a build failure another link, that opens the `pom.xml` of the failed
 
 ![grafik](https://user-images.githubusercontent.com/44067969/152247987-01ee209a-ad6c-454e-92c1-e2aa75388931.png)
 
+This works for ordinary Maven projects as well as for Eclipse Plug-in projects build with Eclipse Tycho.
```
laeubi, line 35: See above comment

HannesWell:
> @HannesWell is this planned for 1.20 release? If yes please assign a milestone.

Yes

> I suggest we just stop planning for 1.20 or we will never deliver it. If it's good to merge, we merge right now, if not we delay to next release; but I don't think such improvement needs to be seen as a blocker for the release.

The should be ready now (if the build succeeds).

laeubi (dismissed):

org.eclipse.m2e.launching/META-INF/MANIFEST.MF
```diff
@@ -18,7 +18,8 @@ Require-Bundle: org.eclipse.core.runtime,
  org.eclipse.m2e.maven.runtime;bundle-version="[1.16.0,2.0.0)",
  org.eclipse.m2e.core;bundle-version="[1.16.0,2.0.0)",
  org.eclipse.m2e.core.ui;bundle-version="[1.16.0,2.0.0)",
- org.eclipse.m2e.workspace.cli;bundle-version="0.1.0"
+ org.eclipse.m2e.workspace.cli;bundle-version="0.1.0",
+ org.eclipse.pde.core
```
fbricon, line 22: Please don't couple m2e launching with PDE. If you want to add PDE specific behavior contributed to m2e launching, please introduce an extension point.

laeubi: Good catch! This would require to install PDE in all cases!

laeubi:
> M3 is in 2 weeks and projects delivering new features are supposed to be done by that time. Thus any new features for 1.20 should be a total no-go from now on and effort being to release by end of next week.

Good point, lets finish[ these open PRs](https://github.com/eclipse-m2e/m2e-core/pulls?q=is%3Aopen+is%3Apr+milestone%3A1.20.0) and then enter "stabilization period" until end of next week. If issues arise, we can decide whether its a critical bug that needs be fixed or delay it to 2.0 ...

mickaelistria:
> Good point, lets finish[ these open PRs](https://github.com/eclipse-m2e/m2e-core/pulls?q=is%3Aopen+is%3Apr+milestone%3A1.20.0) and then enter "stabilization period" until end of next week

I removed the PRs from the milestone. The plan now is to release 1.20.0 today, start 2.0 tomorrow and maybe branch some 1.20.1 is some regressions have happened in 1.20 and we don't feel comfortable shipping the fix in 2.0 only.

laeubi:
> > Good point, lets finish[ these open PRs](https://github.com/eclipse-m2e/m2e-core/pulls?q=is%3Aopen+is%3Apr+milestone%3A1.20.0) and then enter "stabilization period" until end of next week
> 
> I removed the PRs from the milestone. The plan now is to release 1.20.0 today, start 2.0 tomorrow and maybe branch some 1.20.1 is some regressions have happened in 1.20 and we don't feel comfortable shipping the fix in 2.0 only.

I think the packaging type fix is valuable and quite advanced, given that the author has came up with a completely new solution on our request it seems fair to take it into account give he has a strong use-case for this.

laeubi:
@HannesWell do you mind opening a dedicated issue for this one? We can then better track to solve this for 2.0 I think the MavenProjectregistry might be enhanced to at laest return a (temporary) project for projects that have a pom.xml but not a maven nature.

RELEASE_NOTES.md
HannesWell: > This does not sound correct. There are no 'ordinary' projects compared to tycho projects because they are both "maven" builds.
> 
> I would suggest to either remove this or explicitly name the mojos of what build output we support.

This list was actually only intended as a list of examples respectively a reference for the (IMHO) most common use cases. All kind of plug-ins that launch other VMs are supported as long as one knows the right properties to command the plug-in to launch the VM the with corresponding debugger flags. The text `Listening for transport dt_socket at address:` is emitted by the JVM (you can see it also wenn launching JVM with corresponding flags manually) so I think this is quite generic.

But I think I will just remove the remark. As long as it works people are usually happy and I assume most wont except a difference in the first place any ways.

> 
> From the other changes it seems more that you try to support pomless builds here and those without a maven nature. In this case it has nothing to do with building with tycho I think but more with finding sources in non-m2e projects. A better aproach might be to use **JDT** to dearch fir the type instead of searching for a certain project making assumptions about the artifact-id.

The root problem is to find out which project from the workspace the print-out refers to. And since Maven runs in a separate process the information is not available in the Eclipse VM.
The cleanest way would probably to add an Event-Spy via a Maven-Extension to the Maven-VM that connects to the Eclipse VM and reports the information need here to the Eclipse-VM, namely groupId, artifactId and version as well as the project location. The latter can be used to search an `IProject` by its URI using the ResourcePlug-ins workspace. The ConsoleLineTracker then just has to place the links at the right text.

Igor Fedorenko has done something similar for the MavenDev view:
https://github.com/ifedorenko/com.ifedorenko.m2e.mavendev/blob/master/com.ifedorenko.m2e.mavendev.launch.ui.eventspy/src/main/java/com/ifedorenko/m2e/mavendev/launch/ui/eventspy/internal/ExecutionEventListener.java

But this would take a bit more time and I won't be able to complete this within the few next days.

org.eclipse.m2e.launching/META-INF/MANIFEST.MF
HannesWell: With the approach suggested above coupling to PDE should not be necessary anymore.

HannesWell:
> I removed the PRs from the milestone. The plan now is to release 1.20.0 today, start 2.0 tomorrow and maybe branch some 1.20.1 is some regressions have happened in 1.20 and we don't feel comfortable shipping the fix in 2.0 only.

OK, then lets postpone this PR.

> @HannesWell do you mind opening a dedicated issue for this one? We can then better track to solve this for 2.0 I think the MavenProjectregistry might be enhanced to at laest return a (temporary) project for projects that have a pom.xml but not a maven nature.

I can open another issue. The approach written above, will likely be a better solution because it avoids further workarounds.

HannesWell:
I just implemented the solution described in https://github.com/eclipse-m2e/m2e-core/pull/572#discussion_r801392629, were a `EventSpy` is added that participates in the Maven build and sends the relevant Maven-project data via a UNIX domain socket to M2E running in the Eclipse-JVM (and yes that works on Windows as well. In fact I have only tested it on Windows yet).


With the possibility to consider Maven-extensions and therefore loading polyglot-projects Tycho projects are already recognize by the `MavenConsoleLineTracker`. Nevertheless the now presented approach, to obtain the 'real' data from the Maven-model in the Maven build process, the has several advantages that IMHO justify the increased complexity:
- Recognizes all projects in the workspace, even if a project has no Maven-Nature (which is the case for many PDE-projects).
- Finds the right project even if two projects with the same GAV (but different project-name) exist in the workspace.
- Avoids the need to load the Maven-model when looking up the project for a given GAV
  - Improved performance due to less file-operations while tracking the console-text
  - No need resource-locks/workspace-rules while tracking the console-text

HannesWell:
Does anybody have any remarks or objections about the new approach?
If there are no general objections, I will continue and polish this to make it ready for submission.
Of course suggestions for improvements are welcome :)

HannesWell:
Actually the `M2EMavenBuildProjectSpy` could be moved to the m2e.maven.runtime bundle.
This would avoid to create an extra Plugin for it, would make it more clear what dependencies are available (i.e. only what's in the Maven-runtime) and thus would prevent adding more dependencies accidentally. Furthermore it would allow to place the code for de- and encoding of the transferred data in one class and use it on the Maven and IDE site without the need to use `Eclipse-ExtensibleAPI`.

HannesWell:
> Actually the `M2EMavenBuildProjectSpy` could be moved to the m2e.maven.runtime bundle.

This causes some difficulties during build due to the up-front manifest generation. Therefore that part should be postponed until #466 is completed. But since this is all internal it can be done at any time.

HannesWell:
After thinking about it again, I noticed that when using Java-17 features respectively having classes compiled for Java-17 in an extension for the maven-runtime that is then always added by m2e, this effectively mandates to launch the embedded Maven-runtime with Java-17, which is artificial since Maven currently only requires Java-7 (IIRC). Therefore I have to change this back to a traditional lookback-socket connection.

HannesWell:
For me this is ready for submission.

@fbricon, @laeubi do you want to review it again?

laeubi:
If you are fine with it I'm also fine :-)

HannesWell:
> If you are fine with it I'm also fine :-)

Great. Then lets see if this works out in the wild.

PR #592, opened 2022-02-26
[TP|Build] Update to wildwebdeveloper 0.13.3 and lsp4e 0.20.2

HannesWell:


HannesWell:
The reference to the lsp4e 0.20.2 repository in the category.xml was already updated before.

Issue #591, opened 2022-02-24
m2e contributes an embedded jar file that contains unsigned content and therefore cannot be notarized on macOS


jonahgraham:
The 2022-03 M3 build currently cannot be fully notarizared due to unsigned content that comes from (AFAICT) m2e. Although it is an upstream project (jansi) that contains the actual native parts.

The error message from the notarization is:
```json
{
    "logFormatVersion": 1,
    "jobId": "5d49fcdd-e490-479a-8139-a66499aa33f7",
    "status": "Invalid",
    "statusSummary": "Archive contains critical validation errors",
    "statusCode": 4000,
    "archiveFilename": "eclipse-java-2022-03-M3-macosx-cocoa-x86_64-1445086157277215824.dmg",
    "uploadDate": "2022-02-24T15:01:21Z",
    "sha256": "02f07c7ecc3aa198325f76809a52bd7adc9d4c18c242730969ad6652cda9bb0d",
    "ticketContents": null,
    "issues": [
      {
        "severity": "error",
        "code": null,
        "path": "eclipse-java-2022-03-M3-macosx-cocoa-x86_64-1445086157277215824.dmg/Eclipse.app/Contents/Eclipse/plugins/org.eclipse.m2e.maven.runtime_1.18.2.20211123-1648/jars/jansi-2.4.0.jar/org/fusesource/jansi/internal/native/Mac/x86/libjansi.jnilib",
        "message": "The binary is not signed.",
        "docUrl": null,
        "architecture": "i386"
      },
      {
        "severity": "error",
        "code": null,
        "path": "eclipse-java-2022-03-M3-macosx-cocoa-x86_64-1445086157277215824.dmg/Eclipse.app/Contents/Eclipse/plugins/org.eclipse.m2e.maven.runtime_1.18.2.20211123-1648/jars/jansi-2.4.0.jar/org/fusesource/jansi/internal/native/Mac/x86/libjansi.jnilib",
        "message": "The signature does not include a secure timestamp.",
        "docUrl": null,
        "architecture": "i386"
      },
      {
        "severity": "error",
        "code": null,
[...]

PR #593, opened 2022-02-26
Fix 591: sign jnilib files

HannesWell:
This PR has the goal to resolve issue #591 by implementing for the m2e build the steps that are performed in Orbit to sign Mac jnilibs for the m2e build, according to https://github.com/eclipse-m2e/m2e-core/issues/591#issuecomment-1050419195.

@jonahgraham, @laeubi lets see if this works.

### DIFF
--- a/m2e-maven-runtime/pom.xml
+++ b/m2e-maven-runtime/pom.xml
@@ -312,5 +312,78 @@
 				</plugins>
 			</build>
 		</profile>
+		<profile>
+			<id>eclipse-sign-jnilibs</id>
+			<activation>
+				<file>
+					<exists>eclipse-sign-jnilibs.properties</exists>
+				</file>
+			</activation>
+			<!-- 
+			To activate jnilib signing for a bundle, create a file in the project called 'eclipse-sign-jnilibs.properties' and
+			define as value of the key 'jars.directory' the directory that contains the jars whose *.jnilib files have to be signed.
+			The following ant-script then extracts all .jnilib files, sign them and repacks them into the jar file. 
+			-->
+			<build>
+				<plugins>
+					<plugin>
+						<groupId>org.apache.maven.plugins</groupId>
+						<artifactId>maven-antrun-plugin</artifactId>
+						<executions>
+							<execution>
+								<id>sign-jnilibs-files</id>
+								<goals>
+									<goal>run</goal>
+								</goals>
+								<phase>generate-resources</phase> <!-- Do this before extracting sources-->
+								<configuration>
+									<target>
+										<!-- See last answer of https://stackoverflow.com/questions/4368243/maven-antrun-with-sequential-ant-contrib-fails-to-run/45958355 -->
+										<!-- and http://ant-contrib.sourceforge.net/tasks/tasks/index.html -->
+										<taskdef resource="net/sf/antcontrib/antlib.xml" />
+										<loadproperties srcFile="${project.basedir}/eclipse-sign-jnilibs.properties" prefix="signProperties." />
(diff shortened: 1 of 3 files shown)


jonahgraham:
Looks promising. If you have a prerelease that you can put in simrel I can run the full notarization process to confirm no issues. Just ping me.

HannesWell:
> Looks promising. If you have a prerelease that you can put in simrel I can run the full notarization process to confirm no issues. Just ping me.

Yes, finally the result looks fine (in the second Patch-set the exec-task was not configured to fail the build and therefore it passed although signing failed). The print-out of the Manve-build looks like the jnilib files of the main artifact and the sources are signed successfully and the jansi-jar embedded into the m2e.maven.runtime bundle contains a jnilib file that is different to the one in the 1.20.0 release (but the jansi version did not change). So I think the signing and re-packing of the jansi-jar and jansi-source jar works successfully now.

@jonahgraham can you verify from the archived artifacts of the build job that is works? I have no Mac to check that.
This is the link:
https://ci.eclipse.org/m2e/job/m2e/job/PR-593/6/artifact/org.eclipse.m2e.site/target/repository/plugins/org.eclipse.m2e.maven.runtime.source_1.18.3.20220226-1734.jar

@mickaelistria, @akurtakov, @laeubi and @jonahgraham I suspect we have to create a m2e `1.20.1` release from this for the Sim-Rel? If yes, I can prepare a dedicated branch for this and back-port this change. Would we also require a dedicated Jenkins job for that branch or could we replay the build of the master branch with a adjusted configuration to build from that branch?

laeubi:
@HannesWell is really only the source jar that changes here? Then for me it seems a bit strange to require any signing / releasing here as source jars are obviously do not contain any code that is ever executed ...

m2e-maven-runtime/pom.xml
```diff
@@ -312,5 +312,78 @@
 				</plugins>
 			</build>
 		</profile>
+		<profile>
+			<id>eclipse-sign-jnilibs</id>
```
mickaelistria, line 316: Shouldn't it be part of the regular "sign" profile that's already defined in parent?

HannesWell: But this would activate the subsequent mac-signing for all projects and not only for those that have the `eclipse-sign-jnilibs.properties` in their root folder.

mickaelistria: ok, makes sense. It's a pity Maven doesn't allow condition executions.

laeubi: > ok, makes sense. It's a pity Maven doesn't allow condition executions.

You can open an enhancement request to maven :-)

mickaelistria: Yep, just opened https://issues.apache.org/jira/browse/MNG-7426

HannesWell: > ok, makes sense. It's a pity Maven doesn't allow condition executions.

That would indeed makes thinks simpler. I also added the follow up #596 to not fail the build locally.
If conditional executions would be possible it could be handled in one profile.
When thinking about it again it would probably be simpler to make the execution of the ant-task conditional repetitively skip the signing procedure if `eclipse-sign-jnilibs.properties`.
But actually I hope that this is fixed in jansi and this work-around can be fixed soon.


> > ok, makes sense. It's a pity Maven doesn't allow condition executions.
> 
> You can open an enhancement request to maven :-)

I would support that. But I expect that such a feature was already requested many times (but I might be wrong).

PR #597, opened 2022-02-27
Consider maven-enforcer-plugin, requireJavaVersion for deriving JRE

### LINKED ISSUE #590: Improve heuristics for setting JRE classpath container in AbstractJavaProjectConfigurator
Currently the JRE container set in https://github.com/eclipse-m2e/m2e-core/blob/94d02ac79993cb3c3ddb6f45d522f66c0fe3d466/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java#L218 depends on the given `environmentId`.

The `enviromentId` is determined from the target compiler compliance level in https://github.com/eclipse-m2e/m2e-core/blob/94d02ac79993cb3c3ddb6f45d522f66c0fe3d466/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java#L193, which is derived from the maven-compiler-plugin's release or target plugin parameter.

This heuristic fails for all projects which require building on a newer version but for an older java version (cross compilation).
I propose two changes:

1. evaluate "requireJavaVersion" rule from Maven enforcer preferably (https://maven.apache.org/enforcer/enforcer-rules/requireJavaVersion.html)
2. add a workspace preference to optionally always use the workspace default JRE

kwin:
Allow to optionally always use default JRE
This closes #590

### DIFF
--- a/org.eclipse.m2e.feature/feature.xml
+++ b/org.eclipse.m2e.feature/feature.xml
@@ -127,7 +127,6 @@
          download-size="0"
          install-size="0"
          version="0.0.0"
-         fragment="true"
          unpack="false"/>
 
    <plugin
--- a/org.eclipse.m2e.jdt.ui/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.jdt.ui/META-INF/MANIFEST.MF
@@ -2,13 +2,13 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.m2e.jdt.ui;singleton:=true
-Bundle-Version: 1.17.2.qualifier
-Bundle-Localization: fragment
+Bundle-Version: 1.18.0.qualifier
+Bundle-Localization: plugin
 Export-Package: org.eclipse.m2e.jdt.ui.internal;x-internal:=true,
  org.eclipse.m2e.jdt.ui.internal.actions;x-internal:=true,
  org.eclipse.m2e.jdt.ui.internal.filter;x-internal:=true
-Fragment-Host: org.eclipse.m2e.jdt;bundle-version="[1.16.0,2.0.0)"
-Require-Bundle: org.eclipse.core.runtime,
+Require-Bundle: org.eclipse.m2e.jdt;bundle-version="[1.18.4,2.0.0)",
+ org.eclipse.core.runtime,
  org.eclipse.jface,
  org.eclipse.debug.core,
  org.eclipse.jdt.core,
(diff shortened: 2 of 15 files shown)


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/preferences/MavenPreferenceConstants.java
```diff
@@ -166,6 +166,13 @@
    **/
   String P_AUTO_UPDATE_CONFIGURATION = PREFIX + "autoUpdateProjects"; //$NON-NLS-1$
 
+  /**
+   * boolean
+   *
+   * @since 2.0
+   **/
+  String P_ALWAYS_CONFIGURE_DEFAULT_JRE = PREFIX + "alwaysConfigureDefaultJRE"; //$NON-NLS-1$
+
```
mickaelistria, line 175: Those preferences that purely belong to JDT integration should be part of the org.eclipse.m2e.jdt bundle

mickaelistria (changes requested):
This change should be fully contained in the org.eclipse.m2e.jdt bundle.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/preferences/MavenPreferenceConstants.java
kwin: So a new preferences page just for this boolean flag?

mickaelistria: Yes. And maybe other will follow.
Also, please define in the dialog what you mean by the "default JRE". Default is not an explicit enough terms as Maven and JDT have their defaults. If this is JDT's default JRE, the preference page could link to JDT's one.

kwin:
> This change should be fully contained in the org.eclipse.m2e.jdt bundle.

I guess you mean both org.eclipse.m2e.jdt and org.eclipse.m2e.jdt.ui, right?

mickaelistria:
> I guess you mean both org.eclipse.m2e.jdt and org.eclipse.m2e.jdt.ui, right?

Right. Sorry for being unclear about that.

kwin:
@mickaelistria Can I convert module `org.eclipse.m2e.jdt.ui` from fragment to real plugin then, to leverage Bundle-Activator for a new `AbstractUIPlugin`?

mickaelistria:
> Can I convert module org.eclipse.m2e.jdt.ui from fragment to real plugin then, to leverage Bundle-Activator for a new AbstractUIPlugin?

Sure. And if there are packages you need to export from org.eclipse.m2e.jdt, please export them with `x-friends:="org.eclipse.m2e.jdt.ui".

kwin:
I am not sure why the build fails, as I actually did revert all changes done in m2e.core. @mickaelistria Do I need to squash first?

mickaelistria:
Such build failure can sometimes happen if the patch is not based on top of master? Can you please try rebasing on top of master and then push force to your branch?

kwin:
@mickaelistria After rebase and squash still the same issue.

mickaelistria:
It's not the same issue. The new one says that you need to bump the version of the modified bundles that were not yet modified since last release. So org.eclipse.m2e.jdt.ui version should have +0.0.1.

mickaelistria:
Thanks!

PR #610, opened 2022-03-14
[RelEng] Set up automatic license check

HannesWell:
This PR adds a GH workflow to automatically check the license status of all m2e (transitive) dependencies.
It is heavily inspired from [Tycho's workflows](https://github.com/eclipse/tycho/tree/master/.github/workflows) and the one of the eclipse.releng.aggregator.

What I would like to do additionally with this workflow is to add the possibility to re-trigger the license-check and to request reviews just from posting a corresponding comment (a.t.m it is `/check-licenses` respectively `/check-licenses-request-review`.
I don't want to file a review-request on every PR with new dependencies to let a committer decide if it is wanted (otherwise it could be abused by externals).

The only think left is a Token for the EFN Gitlab that is placed in this GH repositories secret store.
@mickaelistria, @laeubi, @akurtakov Do you think that would be possible?

### DIFF
--- a/.github/workflows/licensecheck.yml
+++ b/.github/workflows/licensecheck.yml
@@ -1,0 +1,31 @@
+# This workflow will check for Maven projects if the licenses of all (transitive) dependencies are vetted.
+
+name: License vetting status check
+
+on:
+  push:
+    branches: 
+      - 'master'
+      - 'm2e-*'
+  pull_request:
+    branches: 
+     - 'master'
+     - 'm2e-*'
+
+jobs:
+  check-licenses:
+    runs-on: ubuntu-latest
+
+    steps:
+
+    - uses: actions/checkout@v3
+    - uses: actions/setup-java@v3
+      with:
+        java-version: '11'
+        distribution: 'temurin'
+        cache: 'maven'
+
+    - run: mvn clean generate-sources -f m2e-maven-runtime/pom.xml -B -Dtycho.mode=maven -Pgenerate-osgi-metadata
+
+    - name: Check license vetting status
+      run: mvn -U -B -ntp org.eclipse.dash:license-tool-plugin:license-check -Ddash.fail=true
--- a/pom.xml
+++ b/pom.xml
@@ -355,6 +355,10 @@
 			<id>cbi-releases</id>
 			<url>https://repo.eclipse.org/content/repositories/cbi-releases/</url>
 		</pluginRepository>
+		<pluginRepository>
+			<id>dash-licenses-snapshots</id>
+			<url>https://repo.eclipse.org/content/repositories/dash-licenses-snapshots/</url>
+		</pluginRepository>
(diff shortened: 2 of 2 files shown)


akurtakov:
@HannesWell what is the issue here? It would be nice to get this one.

HannesWell:
> @HannesWell what is the issue here? It would be nice to get this one.

Nothing in particular, just got distracted by other things. But I will prioritize this.

HannesWell:
The license-check itself now runs on each commit to a PR (initial and subsequent).

This workflow also has the ability to create a license-vetting request for the IP-team by adding a comment that starts with `/request-license-review`.

The intention is to issue those requests directly from GitHub.

But this would require a personal authentication access token for Eclipse Foundations GitLab as described here:
https://github.com/eclipse/dash-licenses#automatic-ip-team-review-requests
This would have to be stored in the GH secrets-store of this repo (or organization). The big question is, do you think we get one that can be used by all committers? If you think there is a chance I can create an Help-Desk issue.

It then should be ensured that only the comment of committers can create a request (to prevent abuse by others).
I'm not sure if this happens automatically because only the comments created by users with write-access get access granted to the secrets or if it has to be checked explicitly. The latter is possible in general, although I did not yet find an elegant solution for it.

mickaelistria:
> The intention is to issue those requests directly from GitHub.

I don't think it's worth it. Many deps automatically fix themselves on ClealyDefined over time; automating here doesn't seem necessary and turns the workflow very complex when a minimal Maven execution could fit. Could we start by a basic Maven execution that only votes according to dash license tool results?

HannesWell:
> > The intention is to issue those requests directly from GitHub.
> 
> I don't think it's worth it. Many deps automatically fix themselves on ClealyDefined over time; automating here doesn't seem necessary and turns the workflow very complex when a minimal Maven execution could fit. Could we start by a basic Maven execution that only votes according to dash license tool results?

I have prepared and tested (with my forks) a reusable workflow that I would like to contribute to the dash-licenses repo.
However I agree that we should now start for m2e with a basic workflow that only checks the status.

As part of my testing mentioned above I created a vetting request for `maven/mavencentral/io.takari.m2e.workspace/org.eclipse.m2e.workspace.cli/0.3.1`, that although hosted in this organization, seems to be not be in the database:
https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/2322

Should we wait until that request has been finished so that we will only have green-checkmarks?

HannesWell:
> As part of my testing mentioned above I created a vetting request for `maven/mavencentral/io.takari.m2e.workspace/org.eclipse.m2e.workspace.cli/0.3.1`, that although hosted in this organization, seems to be not be in the database: https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/2322

The request was resolved successfully so this can now be merged with all check-marks green.

PR #626, opened 2022-04-01
Build with Tycho 2.7.1-SNAPSHOT

HannesWell:


mickaelistria:
Why not 3.0.0-SNAPSHOT? Unless this job is only for testing 2.7.1 before it gets released?

HannesWell:
It is for testing 2.7.1 :)

HannesWell:
To simplify future testing of Tycho-Extensions SNAPSHOTs I think we should add the specialized Maven-settings file permanently to the repo.

When a future Snapshot version is then tested one can replay the Jenkins build and only has to modify the Maven command by adding (locally or on Jenkins): `--settings setup/tychoSnapshotSettings.xml`

This also allows to remove the Tycho-Snapshot repo from the pom.xml.

PR #631, opened 2022-04-10
Remove unused Maven-Wrapper

HannesWell:
The Maven-wrapper is not used, is it @mickaelistria?


mickaelistria:
I can't tell for everyone, but "official" build doesn't use it, and I don't either. @laeubi do you?

laeubi:
Nope never used that.

PR #636, opened 2022-04-13
Support contribution of Eclipse features via Maven repositories.

pzi-dsa:
Convert Maven artifacts containing a feature.xml into an Eclipse feature, rather than an OSGi bundle, similar to pom-derived features. 

Supplement to eclipse/tycho#882

Note:
For some reason, I'm getting a NPE when selecting the feature in the "Open Plug-in Artifact" menu. I don't think it's directly related to this addition, as the same problem occurs for pom-derived features.
I think the problem is that the menu tries to read the location of the feature.xml file via IFeatureModel#getInstallLocation. Given that null values are allowed, I don't think it's m2e related but rather a JFace bug. Does anyone have further insights into this?
[Stacktrace.txt](https://github.com/eclipse-m2e/m2e-core/files/8481469/Stacktrace.txt)

### DIFF
--- a/org.eclipse.m2e.pde/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.pde/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: M2E PDE Integration
 Bundle-SymbolicName: org.eclipse.m2e.pde;singleton:=true
-Bundle-Version: 1.20.0.qualifier
+Bundle-Version: 2.0.0.qualifier
 Automatic-Module-Name: org.eclipse.m2e.pde
 Bundle-RequiredExecutionEnvironment: JavaSE-11
 Require-Bundle: org.eclipse.core.runtime;bundle-version="3.19.0",
@@ -17,5 +17,6 @@ Bundle-Activator: org.eclipse.m2e.pde.Activator
 Bundle-ActivationPolicy: lazy
 Import-Package: org.apache.commons.codec.digest;version="1.14.0",
  org.apache.commons.io;version="2.6.0",
- org.apache.commons.io.output;version="2.8.0"
+ org.apache.commons.io.output;version="2.8.0",
+ org.slf4j
 Bundle-Vendor: Eclipse.org - m2e
--- a/org.eclipse.m2e.pde/src/org/eclipse/m2e/pde/MavenTargetLocation.java
+++ b/org.eclipse.m2e.pde/src/org/eclipse/m2e/pde/MavenTargetLocation.java
@@ -9,10 +9,14 @@
  *
  * Contributors:
  *      Christoph Läubrich - initial API and implementation
+ *      Patrick Ziegler - Support contribution of Eclipse features via Maven repositories
  *******************************************************************************/
 package org.eclipse.m2e.pde;
 
+import java.io.File;
+import java.io.IOException;
 import java.io.PrintWriter;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -27,8 +31,12 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Predicate;
+import java.util.jar.JarFile;
 import java.util.stream.Collectors;
+import java.util.zip.ZipEntry;
 
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.io.output.StringBuilderWriter;
 import org.apache.maven.RepositoryUtils;
 import org.apache.maven.artifact.repository.ArtifactRepository;
@@ -62,14 +70,19 @@ import org.eclipse.pde.core.IModel;
 import org.eclipse.pde.core.target.ITargetDefinition;
 import org.eclipse.pde.core.target.TargetBundle;
 import org.eclipse.pde.core.target.TargetFeature;
+import org.eclipse.pde.internal.core.ICoreConstants;
 import org.eclipse.pde.internal.core.ifeature.IFeature;
 import org.eclipse.pde.internal.core.ifeature.IFeaturePlugin;
 import org.eclipse.pde.internal.core.target.AbstractBundleContainer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @SuppressWarnings("restriction")
 public class MavenTargetLocation extends AbstractBundleContainer {
 
+	private static final Logger LOGGER = LoggerFactory.getLogger(MavenTargetLocation.class);
 	private static final String SOURCE_SUFFIX = ".source";
+	private static final String NOT_A_FEATURE = "not_a_feature";
 	public static final String ELEMENT_CLASSIFIER = "classifier";
 	public static final String ELEMENT_TYPE = "type";
 	public static final String ELEMENT_VERSION = "version";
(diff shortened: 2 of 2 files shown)


org.eclipse.m2e.pde/src/org/eclipse/m2e/pde/MavenTargetLocation.java
```diff
@@ -273,6 +274,10 @@ private boolean isPomType(Artifact artifact) {
 		return POM_PACKAGE_TYPE.equals(artifact.getExtension());
 	}
 
+	private boolean isFeature(Artifact artifact) {
+		return new FeatureParser().parse(artifact.getFile()) != null;
```
laeubi, line 278: I think parsing the feature to check is a bit heavy here, if there is not other way at laest the parse feature infos should be returned and used to prevent additional parsing afterwards.

laeubi (changes requested):

laeubi:
> but rather a JFace bug

I think PDE is to blame here please open a bug report there as well.

org.eclipse.m2e.pde/src/org/eclipse/m2e/pde/MavenTargetLocation.java
pzi-dsa: Done. I now simply check, whether the artifact contains a "feature.xml" in the root directory without looking at its content.
The file is only parsed once and only if the artifact is an actual feature.

HannesWell:
> > but rather a JFace bug
> 
> I think PDE is to blame here please open a bug report there as well.

@laeubi Didn't you add support for "in-memory" Features without a `feature.xml` file recently to PDE? I'm not sure if this is exactly this issue, but do you use the latest version of Eclipse?

org.eclipse.m2e.pde/src/org/eclipse/m2e/pde/MavenJarFeatureModel.java
```diff
@@ -0,0 +1,119 @@
...
+import org.eclipse.pde.internal.core.feature.Feature;
+import org.eclipse.pde.internal.core.feature.FeatureChild;
+import org.eclipse.pde.internal.core.feature.FeaturePlugin;
+import org.eclipse.pde.internal.core.ifeature.IFeature;
+import org.eclipse.pde.internal.core.ifeature.IFeatureChild;
+import org.eclipse.pde.internal.core.ifeature.IFeatureInfo;
+import org.eclipse.pde.internal.core.ifeature.IFeaturePlugin;
+
+@SuppressWarnings("restriction")
+public class MavenJarFeatureModel extends AbstractFeatureModel {
```
laeubi, line 21: I wonder if we probably can reuse the default feature model of PDE here?

pzi-dsa: You're right. With the feature.xml, we can delegate the whole task to PDE.
I've updated the code to reflect that. Though PDE doesn't seem to be able to access the file if contained within a jar, so right it's necessary to extract it first.

laeubi (changes requested):
Sorry I lost a bit track of this, but lets try to finish this!

org.eclipse.m2e.pde/src/org/eclipse/m2e/pde/MavenTargetLocation.java
```diff
@@ -273,8 +284,59 @@ private boolean isPomType(Artifact artifact) {
...
+				if(featureFile.exists()) {
+					return featureFile;
+				}
+				
+				try (JarFile jar = new JarFile(artifact.getFile())) {
+					ZipEntry entry = jar.getEntry(ICoreConstants.FEATURE_FILENAME_DESCRIPTOR);
+					
+					// feature.xml is missing -> not an Eclipse feature
+					if(entry == null) {
+						return null;
```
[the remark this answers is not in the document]
pzi-dsa: Good point. Checking the artifact for a feature.xml more than once is excessive.
As you suggested, a marker file is now created when no feature.xml is found.

org.eclipse.m2e.pde/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: M2E PDE Integration
 Bundle-SymbolicName: org.eclipse.m2e.pde;singleton:=true
-Bundle-Version: 1.20.0.qualifier
+Bundle-Version: 1.21.0.qualifier
```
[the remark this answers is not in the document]
pzi-dsa: Done

org.eclipse.m2e.pde/src/org/eclipse/m2e/pde/MavenTargetLocation.java
```diff
@@ -273,8 +285,67 @@ private boolean isPomType(Artifact artifact) {
...
+				
+				// May have already been unpacked -> reuse
+				if(featureFile.exists()) {
+					return featureFile;
+				}
+				
+				File markerFile = new File(file.getParentFile(), NOT_A_FEATURE);
+				
+				// Artifact has already been checked during an earlier cycle
+				if(markerFile.exists()) {
```
laeubi, line 318: Maybe you want to check that the lastChange() of the marker file is >= than the last change of the file :-)

pzi-dsa: Yeah... I was only thinking about releases, where the artifact file wouldn't be updated. But for snapshots, it has to be done.
Updated.

laeubi (approved):
Looks good!

laeubi:
@HannesWell do you like to review this as well?

org.eclipse.m2e.pde/src/org/eclipse/m2e/pde/MavenTargetLocation.java
```diff
@@ -273,8 +285,67 @@ private boolean isPomType(Artifact artifact) {
...
+		File baseFile = artifact.getFile();
+		
+		if(baseFile == null) {
+			return null;
+		} else if (baseFile.isDirectory()) {
+			File featureFile = new File(baseFile, ICoreConstants.FEATURE_FILENAME_DESCRIPTOR);
+			return featureFile.exists() ? featureFile : null;
+		} else if (DEFAULT_PACKAGE_TYPE.equals(FilenameUtils.getExtension(baseFile.getName()))) {
+			File featureFile = unpackFeatureFile(artifact, cacheManager);
+			return featureFile;
```
HannesWell, line 298: The variable can be in-lined.
```suggestion
			return unpackFeatureFile(artifact, cacheManager);
```

org.eclipse.m2e.pde/src/org/eclipse/m2e/pde/MavenTargetLocation.java
```diff
@@ -273,8 +285,67 @@ private boolean isPomType(Artifact artifact) {
...
+				try (JarFile jar = new JarFile(artifact.getFile())) {
+					ZipEntry entry = jar.getEntry(ICoreConstants.FEATURE_FILENAME_DESCRIPTOR);
+					
+					// feature.xml is missing -> not an Eclipse feature
+					if(entry == null) {
+						FileUtils.touch(markerFile);
+						return null;
+					}
+					
+					FileUtils.copyInputStreamToFile(jar.getInputStream(entry), featureFile);
```
HannesWell, line 331: Instead of using the apache-commons-io you could use `Files` from Java's standard library.
```suggestion
					Files.copy(jar.getInputStream(entry), featureFile.toPath());
```

org.eclipse.m2e.pde/src/org/eclipse/m2e/pde/MavenTargetLocation.java
```diff
@@ -273,8 +285,67 @@ private boolean isPomType(Artifact artifact) {
...
+			File featureFile = unpackFeatureFile(artifact, cacheManager);
+			return featureFile;
+		}
+		
+		return null;
+	}
+	
+	private File unpackFeatureFile(Artifact artifact, CacheManager cacheManager) {
+		try {
+			return cacheManager.accessArtifactFile(artifact, (file) -> {
```
HannesWell, line 306: It is not necessary to put the only argument of a lambda into paranthesis:
```suggestion
			return cacheManager.accessArtifactFile(artifact, file -> {
```

HannesWell (approved):
Thank you for your contribution.
The change looks good to me, but I didn't have the time to to test it.

I only have a few minor code-style recommendations, but thy are not necessary to merge this.

pzi-dsa:
> I only have a few minor code-style recommendations, but thy are not necessary to merge this.

It probably makes sense to fix them now. Otherwise I doubt it'll ever get changed :-)

pzi-dsa:
I've integrated the suggested codestyle changes. Unless there is an issue during the build job, that should be all?

laeubi:
> And by the way we should really add some unit tests for Maven-targets. :)

Contributions are welcome :-D

HannesWell:
@laeubi do you want to merge this?
We both agreed it is ready, but since you took the most care about it I don't want to take away the final action from you.

laeubi:
I restarted the build now, no idea why it was failing, beside that I think its not relevant who press the button ;-)

HannesWell:
> I restarted the build now, no idea why it was failing

This is likely again a problem due to the p2-metadata plug-ins baseline replacement.
Due to more changes in the last days there is a newer `org.eclipse.m2e.sdk.feature` that is now in the baseline and replaces the one from this PR:
```
[INFO] --- tycho-p2-plugin:2.7.2:p2-metadata (p2-metadata) @ org.eclipse.m2e.sdk.feature ---
[WARNING] MavenProject: org.eclipse.m2e:org.eclipse.m2e.sdk.feature:2.0.0-SNAPSHOT @ /home/jenkins/agent/workspace/m2e_PR-636/org.eclipse.m2e.sdk.feature/.tycho-consumer-pom.xml: baseline and build artifacts have same version but different contents
   no-classifier: different
      META-INF/ECLIPSE_.RSA: present in baseline only
      META-INF/ECLIPSE_.SF: present in baseline only
      feature.xml: different
   classifier-sources-feature: not present in baseline

[INFO] MavenProject: org.eclipse.m2e:org.eclipse.m2e.sdk.feature:2.0.0-SNAPSHOT @ /home/jenkins/agent/workspace/m2e_PR-636/org.eclipse.m2e.sdk.feature/.tycho-consumer-pom.xml
    The main artifact has been replaced with the baseline version.
```

But that newer m2e.sdk feature references and older version of the m2e.pde feature, which causes this failure:

```
[INFO] --- tycho-p2-repository-plugin:2.7.2:assemble-repository (default-assemble-repository) @ org.eclipse.m2e.repository ---
[INFO] {osgi.os=win32, osgi.ws=win32, org.eclipse.update.install.features=true, osgi.arch=x86_64}
[ERROR] Cannot resolve project dependencies:
[ERROR]   Software being installed: org.eclipse.m2e.repository raw:1.16.0.'SNAPSHOT'/format(n[.n=0;[.n=0;[-S]]]):1.16.0-SNAPSHOT
[ERROR]   Missing requirement: org.eclipse.m2e.sdk.feature.feature.group 2.0.0.20220509-1843 requires 'org.eclipse.equinox.p2.iu; org.eclipse.m2e.pde.feature.feature.group [1.20.100.20220222-1059,1.20.100.20220222-1059]' but it could not be found
[ERROR]   Cannot satisfy dependency: org.eclipse.m2e.repository raw:1.16.0.'SNAPSHOT'/format(n[.n=0;[.n=0;[-S]]]):1.16.0-SNAPSHOT depends on: org.eclipse.equinox.p2.iu; org.eclipse.m2e.sdk.feature.feature.group 0.0.0
```

This should vanish if a rebase is done.

> beside that I think its not relevant who press the button ;-)

Good to know, I just assume that for some people it is important.

So I will jump in and submit this. I strongly predict that this will work once rebased on the master and @pzi-dsa already had to wait long enough.

HannesWell:
@pzi-dsa, thanks again for your contribution!

PR #637, opened 2022-04-13
Update deprecated group id & artifact id of Tycho pomless extension

pzi-dsa:


mickaelistria:
@laeubi the test failure seems to tell that there are cases when new pomless extensions fails?

laeubi:
m2e currently uses functionality that is incompatible with the new tycho-build extension.

pzi-dsa:
> Therefore I suggest to abandon this PR

Sure thing. My thanks to all of you for looking into this and explaining the issue.

PR #639, opened 2022-04-15
Make the Adapter factory more lazy

laeubi:
See also https://github.com/eclipse-platform/eclipse.platform.ui/issues/11


mickaelistria:
Thanks, I wasn't aware of that property and how it can prevent from starting the plugin.

PR #641, opened 2022-04-15
[Build] Read M2E_VERSION directly from m2e.sdk/feature.xml

HannesWell:
This avoids the need to create an m2e.version file using the `maven-antrun-plugin`.

The regex pattern is intentionally designed to be fragile, because I would rather fail the deployment than pushing the repo to some unintended path.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,8 +46,17 @@ pipeline {
 								mkdir -p ${1}"
 							scp -r org.eclipse.m2e.site/target/repository/* genie.m2e@projects-storage.eclipse.org:${1}
 						}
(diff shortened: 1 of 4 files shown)


HannesWell:
Jenkins says the syntax is erroneous I will investigate it.

HannesWell:
Since I don't have the Replay permission anymore and my other testing (ab) using one of by PDE Jenkins PR branches constantly failed with syntax errors I'm uncertain how to properly fix this. At least from the mentioned testing it is not only about escaping the back-slashes.

Therefore I'll revert this one for now and start a whole new second attempt.

HannesWell:
> Jenkins says the syntax is erroneous I will investigate it.

Fixed with #648.

PR #642, opened 2022-04-15
Register the IResourceChangeListener as a service

laeubi:


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenPluginActivator.java
```diff
@@ -216,14 +219,13 @@ public void start(final BundleContext context) throws Exception {
...
-    workspace.addResourceChangeListener(mavenBackgroundJob,
-        IResourceChangeEvent.POST_CHANGE | IResourceChangeEvent.PRE_CLOSE | IResourceChangeEvent.PRE_DELETE);
+    context.registerService(IResourceChangeListener.class, mavenBackgroundJob, getMaskProperties(
+        IResourceChangeEvent.POST_CHANGE | IResourceChangeEvent.PRE_CLOSE | IResourceChangeEvent.PRE_DELETE));
 
     this.projectManager = new MavenProjectManager(managerImpl, mavenBackgroundJob, stateLocationDir);
     this.projectManager.addMavenProjectChangedListener(new WorkspaceStateWriter(projectManager));
     if(updateProjectsOnStartup || managerImpl.getProjects().length == 0) {
-      this.projectManager.refresh(new MavenUpdateRequest(workspace.getRoot().getProjects(), //
+      this.projectManager.refresh(new MavenUpdateRequest(ResourcesPlugin.getWorkspace().getRoot().getProjects(), //
```
mickaelistria, line 228: Can't we get it from context then?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenPluginActivator.java
```diff
@@ -234,7 +236,8 @@ public void start(final BundleContext context) throws Exception {
     this.configurationManager = new ProjectConfigurationManager(maven, managerImpl, modelManager, mavenMarkerManager,
         mavenConfiguration);
     this.projectManager.addMavenProjectChangedListener(this.configurationManager);
-    workspace.addResourceChangeListener(configurationManager, IResourceChangeEvent.PRE_DELETE);
+    context.registerService(IResourceChangeListener.class, configurationManager,
```
mickaelistria, line 239: What about the removeResourceListener then? Shouldn't we unregister the service instead?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenPluginActivator.java
merks, line 228: Is this an example of removing it?   It's just moving the call, isn't it?

laeubi: No it is just here to change the easy ones first and reducing the parts in the code where it is called :-)

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenPluginActivator.java
laeubi: All services are automatically disposed when the bundle is stopped.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenPluginActivator.java
laeubi: Yeah, I just aimed to fix the obvious ones first.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenPluginActivator.java
mickaelistria: So the removeResourceListener can be removed?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenPluginActivator.java
```diff
@@ -263,6 +266,12 @@ public void start(final BundleContext context) throws Exception {
     bundleContext.registerService(IMavenProjectRegistry.class, projectManager, null);
   }
 
+  private static Dictionary<String, ?> getMaskProperties(int mask) {
+    Dictionary<String, Object> properties = new Hashtable<>();
+    properties.put("event.mask", mask);
+    return properties;
```
HannesWell, line 272: In case you don't want to create a Hashtable you could use
```suggestion
    return FrameworkUtil.asDictionary(Map.of("event.mask", mask));
```

laeubi: I don't know if wrapping things more really makes things more easy to understand but nice to know in case one has already a Map ;-)

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenPluginActivator.java
laeubi: Of course yes, I have updated the PR

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenPluginActivator.java
laeubi: I now replaced this with a service tracker. instead.

mickaelistria:
I'm fine merging it, but it's important to acknowledge that this code here is becoming more complicated than using `ResourcesPlugin.getWorkspace()`. I hope this actually demonstrates how people who are willing to adopt alternatives to this API may actually need guidance as replacing it is far from being trivial.

laeubi:
Most of the complexity is already there and cause by we can't really use DS here to make this much more clean and consistent (e.g. by not having an  activator at all). I also need do duplicate some functionality from the resource plugin but hopefully that will be available in the next release.
Beside that, currently one can't use the SafeParticipant in a declarative way. I'll try to open a PR for that as well...

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenPluginActivator.java
HannesWell: That's right. IIRC a method for a singleton Dictonary exists somehwere but I did not find it quickly.
Any way I just wanted to mention it.

laeubi: We hoepfully could remove that method on the next eclipse release.

laeubi:
I have opened https://github.com/eclipse-platform/eclipse.platform.resources/issues/55 to track the issue.

PR #643, opened 2022-04-16
Remove Depprecated MavenRuntimeManager and deprecated methods

laeubi:


laeubi:
The problem is that there are test referencing this, I'm currently struggling a bit with that :-)

mickaelistria:
If those tests are not relevant to m2e's functionaliy, then you can just remove them.

HannesWell:
Thanks for starting the removal of deprecated elements.

If I searched right, there are many more deprecated elements to remove.
Should we create a dedicated issue to track the progress of the removal (IIRC there was only a discussion)? And do you plan to remove the remaining elements too?

mickaelistria:
Was merged as https://github.com/eclipse-m2e/m2e-core/commit/fee4e7aea7e47a6152cf8c385cb829ec98af6c53

PR #646, opened 2022-04-16
[Build] Fix escaping of M2E_VERSION regex pattern in Jenkinsfile

HannesWell:
Attempt to fix failure on master after #641.


HannesWell:
Close this one. In case the problem can be fixed easily, it is done by #648 where Jenkins (hopefully) picks the Jenkinsfile from the PR.

PR #647, opened 2022-04-16
Revert "[Build] Read M2E_VERSION directly from m2e.sdk/feature.xml"

HannesWell:
Reverts eclipse-m2e/m2e-core#641


HannesWell:
No need to revert the change anymore.
The problem was fixed with #648.

PR #649, opened 2022-04-17
[Maven-Runtime] migrate to biz.aQute.bnd:bnd-maven-plugin

HannesWell:
This PR migrates the generation of the MANIFEST.MF for the Maven-runtime components from Apache Felix' `maven-bundle-plugin `to the `biz.aQute.bnd:bnd-maven-plugin` as suggested by Tom Watson a while ago (about a month).

The `bnd-maven-plugin` is seen as a replacement for the `maven-bundle-plugin` and seems to be more actively maintained.
At Felix no committer reacted to any of my PRs (https://github.com/apache/felix-dev/pull/123 and https://github.com/apache/felix-dev/pull/124) yet that were already created by the end of last year even tough @laeubi asked them via mail to look at them a month ago and one committer said he would do that.

With that being said and since the maven-bundle-plugin has some flaws that make it problematic to use with M2E in the IDE (which I attempted to fix with the mentioned PR), I think we should move on. M2E supports both plug-ins in the IDE and the way dependencies are embedded with the `bnd-maven-plugin` could simplify the solution for the problem that we have in PR #466 when it comes to the question how the content of embedded jars is exported  to other Plug-ins on JDT's classpath to allow successful compilation in the IDE.

### DIFF
--- a/m2e-maven-runtime/org.eclipse.m2e.archetype.common/pom.xml
+++ b/m2e-maven-runtime/org.eclipse.m2e.archetype.common/pom.xml
@@ -19,7 +19,7 @@
 	</parent>
 
 	<artifactId>org.eclipse.m2e.archetype.common</artifactId>
-	<version>1.18.1-SNAPSHOT</version>
+	<version>1.18.2-SNAPSHOT</version>
 	<packaging>eclipse-plugin</packaging>
 
 	<name>M2E Maven Archetype Common</name>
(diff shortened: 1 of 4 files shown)


mickaelistria:
I really like this change! Moving to BND seems the way to go, both technically and strategically as I see more and more opportunities of collaboration between Eclipse projects (Tycho, PDE...) and BND.

laeubi:
> Moving to BND seems the way to go, both technically and strategically

Just as a matter of fact, maven-bundle-plugin uses BND under the hood (as Tycho uses BND as well to perform some tasks)...

HannesWell:
I agree. Collaboration and joint efforts for the basics of fully open projects is IMO beneficial and better for all compared to everybody reinventing the wheel. But I think this is already happening many contributors of BND, Felix and Eclipse appear in multiple of those projects. 

But as Christoph mentions the `maven-bundle-plugin` already uses BND, which also makes the transition relatively easy. For most headers/instructions only the syntax had to be adjusted (remove the angle-brackets and closing tags and make underscores a dash). Also the most basic headers could be removed because BND already uses a good default for them.
The greatest work was to find/understand a suitable replacement for the Embed-Dependency capability which is the `includeresource` instruction and adapt to it:
https://bnd.bndtools.org/instructions/includeresource.html


> Beside it would really be good if the felix people would officially discontinue their plugin if the don't like to maintain it anymore...

Agree. @tjwatson already expressed that intend a while ago in the mail mentioned initially. At least M2E is now ready for this. :)

PR #650, opened 2022-04-17
[Build] Simplify maven.yml and update action versions

HannesWell:


HannesWell:
Since this action is deactivated there is no need to await build completion.

PR #651, opened 2022-04-18
Rename m2e.site to m2e.repository

HannesWell:
The term `site` is outdated. Since this project itself is not published we don't have to worry about any API associated problems.


HannesWell:
Renaming went well. The master build succeed and the snapshot repo was also deployed successfully after the renaming.

PR #656, opened 2022-04-23
Revert .gitmodules change and update m2e-core-tests

HannesWell:


HannesWell:
@laeubi you can delete your `Issue_644_adjust` branch now.

PR #662, opened 2022-04-29
Use DS in profile manager

laeubi:


org.eclipse.m2e.profiles.ui/src/org/eclipse/m2e/profiles/ui/internal/actions/ProfileSelectionHandler.java
```diff
@@ -75,9 +77,20 @@ public class ProfileSelectionHandler extends AbstractHandler {
...
   private static final Logger log = LoggerFactory.getLogger(ProfileSelectionHandler.class);
 
+  @Inject
+  private IProfileManager profileManager;
+
+  /**
+   * 
+   */
+  public ProfileSelectionHandler() {
+    ContextInjectionFactory.inject(this, MavenProfilesUIActivator.getServiceContext());
```
HannesWell, line 87: The following variant approach would avoid changes in `MavenProfilesUIActivator`.
```suggestion
    BundleContext bundleContext = FrameworkUtil.getBundle(ProfileSelectionHandler.class).getBundleContext();
    IEclipseContext serviceContext = EclipseContextFactory.getServiceContext(bundleContext);
    ContextInjectionFactory.inject(this, serviceContext);
```

AFAICT `MavenProfilesUIActivator` until now is effectively only used to hold the plug-in ID string constant, which is only used to create I status objects. If those stati are created by the static factories the entire activator could be removed.

Or instead of going the indirection via the E4-Context respectively the `ContextInjectionFactory`, why not do the following and make `profileManager` final?
```suggestion
    BundleContext bundleContext = FrameworkUtil.getBundle(ProfileSelectionHandler.class).getBundleContext();
    profileManager = bundleContext.getService(bundleContext.getServiceReference(IProfileManager.class));
```
As you said before the service objects acquired by a bundle are unget() as soon as the bundle stops and because a handler object probably also lives until that point in time ungetting it would not free the service-object much earlier.

org.eclipse.m2e.profiles.core/src/org/eclipse/m2e/profiles/core/internal/MavenProfilesCoreActivator.java
```diff
@@ -10,83 +10,59 @@
...
+   */
+  public static MavenProfilesCoreActivator getDefault() {
+    return instance;
+  }
 
-	public IProfileManager getProfileManager() {
-		return profileManager;
-	}
+  @Deprecated(forRemoval = true) //only used for test code at the moment...
+  public synchronized IProfileManager getProfileManager() {
```
HannesWell, line 62: Are you planning to update the tests accordingly to use the service from the registry? Then we could entirely remove this activator.
You could do it like in your last change (and Mickael waits until the .gitmodules change is reverted before this is merged :) 😄).

HannesWell (commented):
Looks good in general and I really like OSGi (D)S more and more. :)
I only have to comments.

org.eclipse.m2e.profiles.ui/src/org/eclipse/m2e/profiles/ui/internal/actions/ProfileSelectionHandler.java
HannesWell: > AFAICT `MavenProfilesUIActivator` until now is effectively only used to hold the plug-in ID string constant, which is only used to create I status objects. If those stati are created by the static factories the entire activator could be removed.
> 
A suggestion for that is https://github.com/eclipse-m2e/m2e-core/pull/664

org.eclipse.m2e.profiles.core/OSGI-INF/org.eclipse.m2e.profiles.core.internal.management.ProfileManager.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
HannesWell, line 1: When this change is submitted after you released Tycho 2.7.2 (which I assume since you used the `classpath=true` setting), then this file can be ignored

org.eclipse.m2e.profiles.core/src/org/eclipse/m2e/profiles/core/internal/MavenProfilesCoreActivator.java
laeubi: My idea ways that I first merge this, then adjust the test, then remove the activator, its a bit more steps but I think better than all theses hassle with .gitmodules changes ...

org.eclipse.m2e.profiles.ui/src/org/eclipse/m2e/profiles/ui/internal/actions/ProfileSelectionHandler.java
laeubi: `BundleContext bundleContext = FrameworkUtil.getBundle(ProfileSelectionHandler.class).getBundleContext();` misses a lot sanity checks, I already prepared a change in E4 to make this more convenient in the next dev cycle:

- https://github.com/eclipse-platform/eclipse.platform.runtime/issues/43

laeubi: `profileManager = bundleContext.getService(bundleContext.getServiceReference(IProfileManager.class));`

This also misses some sanity checks, I think we should not use such low-level API when we have some DI in place (e.g. E4, DS, ...)

HannesWell: > `BundleContext bundleContext = FrameworkUtil.getBundle(ProfileSelectionHandler.class).getBundleContext();` misses a lot sanity checks, I already prepared a change in E4 to make this more convenient in the next dev cycle:
> 
>     * [Add a method EclipseContextFactory.getServiceContext(Class<?>) method eclipse-platform/eclipse.platform.runtime#43](https://github.com/eclipse-platform/eclipse.platform.runtime/issues/43)

That change is handy and the sanity checks are indeed missing. But I think it cannot happen that this class is loaded in a non OSGi environment and we can ensure the bundle is started by ensuring `ActivationPolicy: lazy` is set in this plug-ins manifest. Therefore I doubt that they will ever fail here.

> `profileManager = bundleContext.getService(bundleContext.getServiceReference(IProfileManager.class));`
> 
> This also misses some sanity checks, I think we should not use such low-level API when we have some DI in place (e.g. E4, DS, ...)

Are you referring to the null-check on the `ServiceReference` returned by `getServiceReference()`? Or is there more I oversee?
If not, we can ensure respectively be sure that a service is provided since we do that, can't we?

org.eclipse.m2e.profiles.core/src/org/eclipse/m2e/profiles/core/internal/MavenProfilesCoreActivator.java
HannesWell: Sure, that's also possible. As you prefer. 👍🏽

org.eclipse.m2e.profiles.core/OSGI-INF/org.eclipse.m2e.profiles.core.internal.management.ProfileManager.xml
HannesWell: Whichever happens first^^

org.eclipse.m2e.profiles.ui/src/org/eclipse/m2e/profiles/ui/internal/actions/ProfileSelectionHandler.java
laeubi: I thing you know Murphys law: the [principle](https://dictionary.cambridge.org/dictionary/english/principle) that if it is [possible](https://dictionary.cambridge.org/dictionary/english/possible) for something to go [wrong](https://dictionary.cambridge.org/dictionary/english/wrong), it will go [wrong](https://dictionary.cambridge.org/dictionary/english/wrong) ... and even in the worst possible way ;-)

HannesWell: You are not wrong , but I would say if we set `ActivationPolicy: lazy` and the `BundleContext` is still null here, there is some bigger problem. :)
And mainly I would not like to keep the activator (and abandon https://github.com/eclipse-m2e/m2e-core/pull/664) just to have a saver way to get the bundle-context (if you want to be very safe, you also would have to check that the BundleContext passed to the start method is not null).

mickaelistria:
I see I'm marked as a reviewer. I overall support the move to DS, but don't have the knowledge (nor much time right now) to go into details. So I happily trust that if @laeubi and @HannesWell are happy with this change, it can be merged.

org.eclipse.m2e.profiles.core/OSGI-INF/org.eclipse.m2e.profiles.core.internal.management.ProfileManager.xml
HannesWell: This file can now be ignored.

HannesWell:
@laeubi can you please update the PR and adapt to the removal of the `o.e.m2e.profiles.ui` activator?
I can prepare a PR to update the tests so we can already test and submit this in one go (using the temporary modified .gitmodules). Then the activator `MavenProfilesCoreActivator` can probably removed completely.

Btw. what do you think about something like a [`useService(Class<T>)` method](https://github.com/HannesWell/equinox.framework/blob/abeac0ca04667d981e1da0743db7cfae3b60f289/bundles/org.eclipse.osgi/osgi/src/org/osgi/framework/FrameworkUtil.java#L1071)? I think this would be a very convenient and save way for single use of a service when a ServiceTracker is not wanted.
If you just want any implementation of a service, you could then simply use the following code (once OSGi requires Java>9):
```
try (var myService = FrameworkUtil.useService(MyService.class)){
  myService.get().someMethod();
}
```
This is just an idea I had recently that is not worked out, but if some people find it useful I think about proposing it for OSGi. I think this would lower the barrier to use OSGi Services outside of DS.

org.eclipse.m2e.profiles.core/src/org/eclipse/m2e/profiles/core/internal/management/ProfileManager.java
```diff
@@ -64,15 +66,23 @@
...
 public class ProfileManager implements IProfileManager {
 
+  @Reference
+  IProjectConfigurationManager configurationManager;
+
+  @Reference
+  IMaven maven;
+
+  @Reference
+  ILog log;
```
HannesWell, line 79: This does not work because the platform does not register an ILog service implementation.

HannesWell (commented):
I have created a draft PR to adjust the tests already: https://github.com/tesla/m2e-core-tests/pull/153

Besides that I rebased this PR in my m2e-fork and made the necessary adjustments to the changes that happened in the meantime: https://github.com/HannesWell/m2e-core/tree/pr-662
Feel free to use it to update this PR. :)
It also includes a temporary update of the `.gitmodules` file.

org.eclipse.m2e.profiles.core/src/org/eclipse/m2e/profiles/core/internal/management/ProfileManager.java
laeubi: It will on the next release... so one need an ibuild currently.

laeubi:
@HannesWell feel free to simply merge your PR and closing this :-)

org.eclipse.m2e.profiles.core/src/org/eclipse/m2e/profiles/core/internal/management/ProfileManager.java
HannesWell: Understand, great.
For curiosity, how do you ensure that a component gets the Log associated with its bundle injected?
Can you please link the PR?

HannesWell:
> @HannesWell feel free to simply merge your PR and closing this :-)

Done. :)
Superseded by https://github.com/eclipse-m2e/m2e-core/pull/693

org.eclipse.m2e.profiles.core/src/org/eclipse/m2e/profiles/core/internal/management/ProfileManager.java
laeubi: See 
- https://github.com/eclipse-platform/eclipse.platform.runtime/pull/34

HannesWell: Thanks. Learned one more thing.

Issue #123, opened 2021-03-30
M2Eclipse gets stuck in endless update loop


laeubi:
This has been reported in the following Bugs but its unclear how to reproduce this yet:

- https://bugs.eclipse.org/bugs/show_bug.cgi?id=558317
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=567959
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=553432
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=558689
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=563140

DieterMai:
I closed ticket Bug 563140 [1] since no longer reproducible

[1] https://bugs.eclipse.org/bugs/show_bug.cgi?id=563140

HannesWell:
Is there any progress on this issue? I run into it in my Tycho-Dev workspace (provisioned using the Tycho Oomph-Setup), using Eclipse-for-Commiter latest (currently at 2021-03) and m2e 1.17.2.

I sampled my Dev-Eclipse using VisualVM and from this it looks like the loop is in the LifecycleMappingFactory.

![grafik](https://user-images.githubusercontent.com/44067969/113278366-253dbf80-92e2-11eb-85bb-ee184590ddcf.png)

laeubi:
@HannesWell Good idea. The problem is that this not occurs always, sometimes it helps to update one or two projects. then it works for some time and then comes back. I'm currently also again caught in an endless circle :-1:

HannesWell:
Yes this issue just appeared recently to me. I could make it work once by updating one or two projects as you said. But now this also did not work. It is just terribly annoying, I can hardly change a line of code in my Tycho-Dev workspace.

Anyway. I just attached a debugger and my first impression is that the `DependencyResolutionContext` passed to the `ProjectRegistryManager.refresh` method gets always new elements added when `refreshPhase2` is called. But I will investigate further.

laeubi:
ProjectRegistryManager line 504 a new facade is created but I can't tell from the code if this is intentional.

laeubi:
@mickaelistria can you take a look at this? even worse now my IDE constantly get stuck and I need to kill the process :-(

laeubi:
One think that might help here:
If I understand the code right, it triggers some kind of "project changed", maybe it would be better to just note all projects that require such a change event in a list, complete the cycle and then trigger one big "project(s) change" event afterwards?

[...]

PR #665, opened 2022-05-03
M2Eclipse gets stuck in endless update loop

laeubi:
Fix https://github.com/eclipse-m2e/m2e-core/issues/123

Lets see if simple things just work...


laeubi:
@mickaelistria this has solved my endless cycle but two tests complain, can you take a look what these tests are missing or what feature they depend on so I can try to enhance the solution?

mickaelistria:
Those tests are important. I think you can reproduce them manually to get a sense of what are the bad effects of your proposal. Basically, in some cases, some projects really need to be processed multiple times and your code skips that. The condition to skip a refresh needs to be narrower.

laeubi:
> what are the bad effects

Stuck in an endless cycle is (for me) much more bad :-)

> The condition to skip a refresh needs to be narrower.

I'll try a more sophisticated check ... :-\

mickaelistria (approved):
This seems quite good!

laeubi:
If you think that it don't harm I'd like to merge this, maybe it will need improvements but I can only check for sure next time the build-loop occurs again ...

PR #669, opened 2022-05-05
Remove deprecated APIs in org.eclipse.m2e.jdt

mickaelistria:

PR #672, opened 2022-05-05
Remove deprecated unused APIs in org.eclipse.m2e.core

HannesWell:
This PRs applies the simple removals of API in o.e.m2e.core, those API that is not used.


HannesWell:
This requires PR https://github.com/eclipse-m2e/m2e-core/pull/669 to be merged first because the last class using one of the removed methods is removed there.

PR #673, opened 2022-05-05
Major-version bump of m2e.core.ui to 2.0.0 after deprecated API removal

HannesWell:


org.eclipse.m2e.editor/META-INF/MANIFEST.MF
```diff
@@ -16,7 +16,7 @@ Require-Bundle: org.eclipse.core.runtime,
  org.eclipse.ui.workbench,
  org.eclipse.m2e.core;bundle-version="1.16.0",
  org.eclipse.m2e.maven.runtime;bundle-version="1.16.0",
- org.eclipse.m2e.core.ui;bundle-version="[1.17.2,2.0.0)",
+ org.eclipse.m2e.core.ui;bundle-version="[1.17.2,3.0.0)",
```
HannesWell, line 19: Should we raise the upper-bound to 2.0.0?

mickaelistria: I don't think it's necessary, as the bundle doesn't use any new API nor uses any removed API. It's still compatible with 1.17.2.

HannesWell: At the moment yes. But this might change in the future. We should keep that then in mind (some tooling support would of course be better).

laeubi: > I don't think it's necessary, as the bundle doesn't use any new API nor uses any removed API

The recommendation from OSGi is to use the version that the bundle was compiled against, unless we plan to compile (and test!) using 2.x bundles against older m2e releases we should raise this, I also don't see any advantage in haveing lower bounds for our own API ...

HannesWell: That's right.
And the chances that we will forget to raise the lower bound if we one day use new API are quite high I think.
And yes the question is: What do we gain with the lower bound?

akurtakov: Just bump it. It's our own API so mix and match between m2e versions shouldn't be done.

PR #674, opened 2022-05-05
[TP] add Mockito, extends Test-utils and update m2e-core-tests submodule

HannesWell:
Update the TP and `AbstractMavenProjectTestCase` to prepare for the changes in:
- https://github.com/tesla/m2e-core-tests/pull/147

This performs also the verification build of that PR.
The .gitmodules update has to be reverted before this is merged. But this way a verification build of the test changes is possible in advance.


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "m2e-core-tests"]
 	path = m2e-core-tests
-	url = https://github.com/tesla/m2e-core-tests.git
-	branch = master
+	url = https://github.com/hannesWell/m2e-core-tests.git
```
akurtakov, line 3: This part is here by mistake, right?

target-platform/target-platform.target
```diff
@@ -31,6 +31,7 @@
 			<unit id="ch.qos.logback.core" version="0.0.0"/>
 			<unit id="ch.qos.logback.classic" version="0.0.0"/>
 			<unit id="ch.qos.logback.slf4j" version="0.0.0"/>
+			<unit id="org.mockito" version="0.0.0"/>
```
akurtakov, line 34: Mockito in Orbit has been maintained mainly by Platform needs but we moved to using the one from maven central directly. So this one is most likely stuck at this version and even worse uses different symbolic name compared to upstream.  It would be best to use approach similar to https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target#L322

.gitmodules
HannesWell: It is here to test my PR in `tesla/m2e-core-tests` before it is merged there.
I stated in my initial comment that this part has to be reverted before this can be merged and that's why I made this a draft.
But since there are test-failures I have to fix them in `tesla/m2e-core-tests`  first it can be merged there.

target-platform/target-platform.target
HannesWell: Thanks for that hint! I will change it.

mickaelistria: I think it would be better to just consume it from Platform, eg on line 11, than to re-consume it from Maven. This will give some consistency with Platform version.

HannesWell: > I think it would be better to just consume it from Platform, eg on line 11, than to re-consume it from Maven. This will give some consistency with Platform version.

That's reasonable. I changed it accordingly.

HannesWell:
The changes in m2e-core-tests are ready for submission now. As soon as they have been submitted I can update and submit this PR too.

PR #675, opened 2022-05-05
Clean-up and simplify MavenImpl

HannesWell:
Before we remove the deprecated methods, `MavenImpl` can be cleaned-up.

I can then take care of removing the deprecated methods.

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
@@ -122,7 +122,6 @@ import org.apache.maven.plugin.Mojo;
 import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.plugin.MojoNotFoundException;
 import org.apache.maven.plugin.PluginConfigurationException;
-import org.apache.maven.plugin.PluginContainerException;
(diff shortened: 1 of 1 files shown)


laeubi:
As we are talking about cleanup:

I noticed there are some (inconsistent) formatting and code templates enabled on the project level.
I wonder if we can agree to remove all thos project level source config and simply use "Eclipse" Formatting profile in all cases?

HannesWell:
> I noticed there are some (inconsistent) formatting and code templates enabled on the project level. I wonder if we can agree to remove all thos project level source config and simply use "Eclipse" Formatting profile in all cases?

Yes, I fully support that! I noticed that as well in the past but did not yet have the time to bring this up.

akurtakov:
> > I noticed there are some (inconsistent) formatting and code templates enabled on the project level. I wonder if we can agree to remove all thos project level source config and simply use "Eclipse" Formatting profile in all cases?
> 
> Yes, I fully support that! I noticed that as well in the past but did not yet have the time to bring this up.

Fully agree on using "Eclipse" formatting profile. But let's keep the project level source config files saying that, otherwise contributors that have changed their workspace formatting profile to smth else will submit different formatting.

laeubi:
> . But let's keep the project level source config files saying that

I think we can better configure that at the SDK setup I think, these "project settings" tend to diverge as new items are added or defaults are changed.

PR #680, opened 2022-05-07
Remove deprecated IMaven API

HannesWell:
Remove the deprecated API of `org.eclipse.m2e.core.embedder.IMaven`.

This requires adjustments in the m2e-core tests which should be tested first:
https://github.com/tesla/m2e-core-tests/pull/149

### DIFF
--- a/org.eclipse.m2e.apt.core/src/org/eclipse/m2e/apt/internal/AbstractAptConfiguratorDelegate.java
+++ b/org.eclipse.m2e.apt.core/src/org/eclipse/m2e/apt/internal/AbstractAptConfiguratorDelegate.java
@@ -363,8 +363,9 @@ public abstract class AbstractAptConfiguratorDelegate implements AptConfigurator
       MojoExecution mojoExecution) throws CoreException {
     PluginExecution execution = new PluginExecution();
     execution.setConfiguration(mojoExecution.getConfiguration());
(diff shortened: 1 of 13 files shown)


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "m2e-core-tests"]
 	path = m2e-core-tests
-	url = https://github.com/tesla/m2e-core-tests.git
```
HannesWell, line 3: This is only for testing the m2e-core test changes and has to be reverted before submission.

HannesWell:
As already written in the associated update of the m2e-core-tests:
After some reconsideration I think the the method `MavenImpl.resolveParentProject()` has to stay because it's functionality to reliably resolve the IFile of the parent MavenProject is still needed and I don't know a suitable better way.
Nevertheless it can be removed from the `IMaven` API and just be used internally in m2e. In case it is needed by down-stream users we can either re-add it or think of a better way, but at the moment I don't know any but I'm also not deeply familiar with this part of m2e.

mickaelistria:
@HannesWell chanes in tests are now merged.

HannesWell:
> @HannesWell chanes in tests are now merged.

Thanks Mickael!

Since you did not objected yet, I assume you agree with this change?

mickaelistria:
> Since you did not objected yet, I assume you agree with this change?

Of course! But a necessary import has been removed apparently.

.gitmodules

PR #681, opened 2022-05-08
Remove deprecated MavenRuntime and IMavenLauncherConfiguration API

HannesWell:
The APIs `MavenRuntime` and `IMavenLauncherConfiguration` are deprecated but their functionality is still required internally to prepare a Maven-build launch.

So I think the suitable way is to remove the interface `MavenRuntime` and to make `IMavenLauncherConfiguration` internal by moving it into the internal class `AbstractMavenRuntime`. Alternatively `IMavenLauncherConfiguration` could stay a top-level interface in `org.eclipse.m2e.core.internal.launch`. I have no strong opinion which one is better.

This requires the adjustment of the corresponding package-import in the test, but since this breaks the compilation of the test if until this is submitted, I would like to have PR #680 submitted first.


HannesWell:
Created https://github.com/tesla/m2e-core-tests/pull/150 to adapt the tests to this change.
@mickaelistria could please merge that as well, once we have a successful build here. Thank you!

.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "m2e-core-tests"]
 	path = m2e-core-tests
-	url = https://github.com/tesla/m2e-core-tests.git
-	branch = master
```
HannesWell, line 4: This has to be reverted before this change is submitted and after the test-changes are submitted.

HannesWell:
Since no objections were raised, I assume everybody is fine with this and I will submit this as soon as the build succeeds.
Because without this change the tests are broken we should not leave this open longer than necessary.

HannesWell:
I'm puzzled why the build failed the second time with:
```
Running org.eclipse.m2e.tests.discovery.LifecycleMappingDiscoveryTest
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGBUS (0x7) at pc=0x00007f34ad2e509d, pid=3143, tid=3201
#
# JRE version: OpenJDK Runtime Environment AdoptOpenJDK-11.0.11+9 (11.0.11+9) (build 11.0.11+9)
# Java VM: OpenJDK 64-Bit Server VM AdoptOpenJDK-11.0.11+9 (11.0.11+9, mixed mode, tiered, compressed oops, g1 gc, linux-amd64)
# Problematic frame:
# C  [libc.so.6+0x15a09d]  __strcasecmp_l_avx+0x49
#
# Core dump will be written. Default location: Core dumps may be processed with "/usr/lib/systemd/systemd-coredump %P %u %g %s %t %c %h" (or dumping to /home/jenkins/agent/workspace/m2e_PR-681/m2e-core-tests/org.eclipse.m2e.tests/core.3143)
#
# An error report file with more information is saved as:
# /home/jenkins/agent/workspace/m2e_PR-681/m2e-core-tests/org.eclipse.m2e.tests/hs_err_pid3143.log
#
# If you would like to submit a bug report, please visit:
#   https://github.com/AdoptOpenJDK/openjdk-support/issues
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
```

mickaelistria:
That could be just bad luck (some bug in the JDK build); let's re-try the build.

HannesWell:
> That could be just bad luck (some bug in the JDK build); let's re-try the build.

Sure. Force pushed an amended commit, maybe some freshness helps too.
 
Some quick internet-search gave me this: https://bugs.openjdk.java.net/browse/JDK-8254139
This somehow sounds similar, but I'm not very familiar with these JDK internals so it could be completely unrelated.
But that happens also on Jenkins.

mickaelistria:
If this happens again, you may want to try to use `temurin-jdk11-latest` instead of adoptopenjdk in the Jenkinsfile, as it's a newer build of OpenJDK (11.0.13), or we may even consider to use `openjdk-jdk17-latest` but it means we're not testing against Java 11 by default; which would be totally OK with me.

HannesWell:
> If this happens again, you may want to try to use `temurin-jdk11-latest` instead of adoptopenjdk in the Jenkinsfile, as it's a newer build of OpenJDK (11.0.13), or we may even consider to use `openjdk-jdk17-latest` but it means we're not testing against Java 11 by default; which would be totally OK with me.

Thanks for that hint. I already noticed that therere are more recent JDK-11 builds and wondered if they are available. Thanks for the hint. This build succeeded 🎉, nevertheless I'm about to prepare a PR to use `temurin-jdk11-latest`.
As long as we are not using Java-17 I think we should build/test with Java-11. With Jonah's latest post on the mailing-list I also thought about if we should consider to move to Java-17 but I don't see an urgent reason at the moment for m2e (but I could be wrong) and therefore would suggest to move to Java-17 when Platform is doing the move?

akurtakov:
> As long as we are not using Java-17 I think we should build/test with Java-11. With Jonah's latest post on the mailing-list I also thought about if we should consider to move to Java-17 but I don't see an urgent reason at the moment for m2e (but I could be wrong) and therefore would suggest to move to Java-17 when Platform is doing the move?

As tm4e is going to be Java 17, wildwebdeveloper and m2e will inherit some Java 17 dependencies so it may make sense to move the whole project to Java 17 prior to Platform.

mickaelistria:
IMO, upcoming 2.0 should be Java 11; and next 2.1 should be Java 17.

PR #683, opened 2022-05-09
Replace uses of deprecated methods in MavenJRETab and WorkingSetGroup

HannesWell:
... and perform minor clean ups and formatting.

Not setting `IJavaLaunchConfigurationConstants.ATTR_VM_INSTALL_NAME` and `IJavaLaunchConfigurationConstants.ATTR_VM_INSTALL_TYPE` to the config and simply letting the `JREsComboBlock` of the `MavenJRETab` handle the JREs also fixes the quirk that when one creates a new maven launch-config `Alternate JRE` is selected by default instead of Workspace default:

grafik

### DIFF
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/components/WorkingSetGroup.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/components/WorkingSetGroup.java
@@ -101,8 +101,8 @@ public class WorkingSetGroup {
 
     workingsetComboViewer = new ComboViewer(workingsetCombo);
     workingsetComboViewer.setContentProvider(new IStructuredContentProvider() {
-        @Override
-        public Object[] getElements(Object input) {
+      @Override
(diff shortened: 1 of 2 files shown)


mickaelistria:
It seems https://github.com/tesla/m2e-core-tests/blob/7a254858e3569548140aaeca9b6c678711bfbbba/org.eclipse.m2e.tests/src/org/eclipse/m2e/tests/embedder/MavenRuntimeManagerTest.java still requires cleanup/removal.

HannesWell:
> It seems https://github.com/tesla/m2e-core-tests/blob/7a254858e3569548140aaeca9b6c678711bfbbba/org.eclipse.m2e.tests/src/org/eclipse/m2e/tests/embedder/MavenRuntimeManagerTest.java still requires cleanup/removal.

I think the Problem is more that the tests are already adapted for PR #681 but that PR is not yet submitted.
I oversaw that Aleks already had submitted the change.

HannesWell:
Temporary included to commit to remove the deprecated API that let the tests fail. Should succeed now.

PR #685, opened 2022-05-10
Remove remaining deprecated APIs from org.eclipse.m2e.core

HannesWell:
This PR replaces the use and removes the remaining deprecated API from m2e.core.

The only problem that is not yet resolved is how to properly replace `ProjectImportConfiguration.getProjectName()` respectivly `ProjectImportConfiguration.getProject()`.


HannesWell:
I moved the implementations of the deprecated methods of `ProjectImportConfiguration` that are still used (and I think required) internally to `ProjectConfigurationManager` respectively `MavenProjectWizard`.
With that this PR is ready and all deprecated API should be removed from M2E.

PR #686, opened 2022-05-13
Add a dev target that includes the current eclipse isntallation

laeubi:
I sometimes use the current eclipse installation so I don't need to import all bundles (especially the runtime ones).
With the new target reference in PDE one could easily create such a target without modify the original one.

### DIFF
--- a/target-platform/dev.target
+++ b/target-platform/dev.target
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<?pde version="3.8"?>
+<target name="dev">
+	<locations>
+		<location path="${eclipse_home}" type="Profile"/>
+		<location type="Target" uri="file:${project_loc:/target-platform}/target-platform.target"/>
+	</locations>
+</target>


target-platform/dev.target
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<?pde version="3.8"?>
+<target name="dev">
```
HannesWell, line 3: Maybe a more expressive name like `m2e-with-running-eclipse` would be better.
Some devs have multiple projects in their workspace.

mickaelistria:
Do we want to advertise contributors that they should use this target? If it's not meant to be "standard" as part of some workflow documented/advertised/maintained by the project, it doesn't make sense to have it part of m2e and would better remain a developer decision IMO.
So if you think it's best for contributors, please also adapt the CONTRIBUTING.md file and other relevant setup files.

laeubi:
I'll leave the decision a bit to @HannesWell as he has done most of the good work about setting up m2e IDE. Given one uses this target it is not strictly required to have e.g. the runtime projects imported unless one plans to change things here given that we use a current release in the setup...

HannesWell:
Personally I'm not a big fan of using the running-platform as Target-Platform because what I use for development is something different than what I'm developing. And not using the running-platform also makes sure that there are no (or at least less) unwanted interferences that might cause hard to find problems for contributors. Therefore I'm in favour to provide an as similar as possible set-up for all contributors. If somebody decides to use something else its his/her decision.

The runtime components are now-days relatively easy to set up using the Oomph setup and will be just regular projects without any special care once https://github.com/eclipse-m2e/m2e-core/pull/466 is completed. So for me that is not a relevant argument any more. But if you want to add the Maven-Runtime to the tp, you could also add the m2e snapshot p2-repo.

The only useful reason I see for using the running-platform is if one wants to add other plug-ins to a launched Eclipse-App to test something. But then those other plug-ins/features could also be locally added to the TP (either directly or the same way as you did).

I'm in favour to keep the current TP the default. Actually it would not harm to add this TP to the repo, but Mickael is right that this will then mainly just lay around and look unused and one day one will come and remove it because there is no direct use-case.

And since you showed that is really easy to set up such a TP, I think it is fine to not add it to the repo and if one needs it, one can create it locally. If you think it is important to mention we could update the doc accordingly and give that hint to fellow contributors.

PR #693, opened 2022-05-14
Use DS in profile manager

HannesWell:
Follow-up of PR https://github.com/eclipse-m2e/m2e-core/pull/662 that is rebased to the current master and adapted to the changes that happened in the meantime.


org.eclipse.m2e.profiles.core.tests/src/org/eclipse/m2e/profiles/core/internal/management/MavenProfileManagerTest.java
```diff
@@ -16,30 +16,43 @@
...
 
 
 public class MavenProfileManagerTest extends AbstractMavenProjectTestCase {
 
-  private final IProfileManager profileManager = MavenProfilesCoreActivator.getDefault().getProfileManager();
+	private static ServiceTracker<IProfileManager, IProfileManager> profileManagerTracker;
+
+	@BeforeClass
+	public static void setUpProfileManagerTracker() throws BundleException {
+		BundleContext context = FrameworkUtil.getBundle(MavenProfileManagerTest.class).getBundleContext();
```
HannesWell, line 47: This plug-ins manifest specifies `Bundle-ActivationPolicy: lazy` so the `BundleContext` should never be null here.
Furthermore I think this code will never run outside of OSGi so I also expect that `getBundle()` never returns null unless the entire framework is in a weird state.

HannesWell (commented):
The build succeeded. 
@laeubi if you are fine with this change as well, I'm going to merge it.

org.eclipse.m2e.profiles.core.tests/src/org/eclipse/m2e/profiles/core/internal/management/MavenProfileManagerTest.java
laeubi: For testing purpose this should be fine.

org.eclipse.m2e.profiles.ui/src/org/eclipse/m2e/profiles/ui/internal/actions/ProfileSelectionHandler.java
```diff
@@ -72,9 +78,19 @@ public class ProfileSelectionHandler extends AbstractHandler {
 
   private static final Logger log = LoggerFactory.getLogger(ProfileSelectionHandler.class);
 
+  @Inject
+  private IProfileManager profileManager;
+
+  public ProfileSelectionHandler() {
+    BundleContext context = FrameworkUtil.getBundle(ProfileSelectionHandler.class).getBundleContext();
+    IEclipseContext serviceContext = EclipseContextFactory.getServiceContext(context);
```
HannesWell, line 86: This plug-ins manifest specifies `Bundle-ActivationPolicy: lazy` so the `BundleContext` should never be null here.
Furthermore I think this code will never run outside of OSGi so I also expect that `getBundle()` never returns null unless the entire framework is in a weird state.

org.eclipse.m2e.profiles.ui/src/org/eclipse/m2e/profiles/ui/internal/actions/ProfileSelectionHandler.java
```diff
@@ -72,9 +78,19 @@ public class ProfileSelectionHandler extends AbstractHandler {
 
   private static final Logger log = LoggerFactory.getLogger(ProfileSelectionHandler.class);
 
+  @Inject
+  private IProfileManager profileManager;
+
+  public ProfileSelectionHandler() {
+    BundleContext context = FrameworkUtil.getBundle(ProfileSelectionHandler.class).getBundleContext();
```
HannesWell, line 85: This plug-ins manifest specifies `Bundle-ActivationPolicy: lazy` so the `BundleContext` should never be null here.
Furthermore I think this code will never run outside of OSGi so I also expect that `getBundle()` never returns null unless the entire framework is in a weird state.

org.eclipse.m2e.profiles.core.tests/src/org/eclipse/m2e/profiles/core/internal/management/MavenProfileManagerTest.java
HannesWell: Ups, I actually intended to add this comment in the class below.

org.eclipse.m2e.profiles.ui/src/org/eclipse/m2e/profiles/ui/internal/actions/ProfileSelectionHandler.java
laeubi: This still do not handle dynamics, we should not use the bare OSGi API when there is no need for it and we can use a DI technology instead.

HannesWell: What's then the approach you suggest?

laeubi: Sorry for confusion ... it seems it get messed up with the recent push, as long as Eclipse offers not the `EclipseContextFactory.getServiceContext(<class>);` method think we can go with this...

HannesWell: > Sorry for confusion ... it seems it get messed up with the recent push, as long as Eclipse offers not the `EclipseContextFactory.getServiceContext(<class>);` method think we can go with this...

No problem. Makes sense to use that ones available.

PR #694, opened 2022-05-15
Unify/simplify Plug-in labels for m2e.jdt.ui and m2e.lifecylcemapping.defaults

HannesWell:
Improve metadata as in #339 for remaining plug-ins that had other changes for this release in the meantime.

This applies all currently possible/wanted changes of https://github.com/eclipse-m2e/m2e-core/pull/402.


HannesWell:
A second look revealed that https://github.com/eclipse-m2e/m2e-core/pull/402 can be submitted as a whole. So there is no need for this partial submission.

Issue #702, opened 2022-05-16
Remove org.eclipse.m2e.core.internal.MavenPluginActivator.getWagonManager()


laeubi:
It seems this is not used anywhere and also is just a convince method.

PR #703, opened 2022-05-16
Some cleanup

laeubi:
Fix https://github.com/eclipse-m2e/m2e-core/issues/702


HannesWell:
I wonder why a second lookup() method was even added? Seems totally duplicated.
Looks good that it is gone.

PR #704, opened 2022-05-16
Disable concurrent builds, as building older commits do not help much.

laeubi:
Currently a PR is build multiple times if before the build finished there is a new commit in the meantime.


laeubi:
This could be merged as soon as the Jenkin jobs successfully starts, not code changes are included here, we just need to make sure the syntax is correct.

Issue #666, opened 2022-05-04
maven.multiModuleProjectDirectory in <repositories> not resolved


GaZaTu:
yo,

i have a root pom.xml with an entry like this:
```xml
<repositories>
  <repository>
    <id>local-maven-repo</id>
    <url>file:///${maven.multiModuleProjectDirectory}/.local-maven-repo</url>
  </repository>
</repositories>
```
to provide all child maven modules with access to that in-project maven repo.

This works with IntelliJ and mvn-cli and used to work with Eclipse 2021-09 and redhat.java v1.1.0 but since updating to 2022-03 it results in an exception like this:
```
Failed to read artifact descriptor for GROUP_ID:ARTIFACT_ID:jar:0.0.1-SNAPSHOT

org.eclipse.aether.resolution.ArtifactDescriptorException: Failed to read artifact descriptor for GROUP_ID:ARTIFACT_ID:jar:0.0.1-SNAPSHOT
	at org.apache.maven.repository.internal.DefaultArtifactDescriptorReader.loadPom(DefaultArtifactDescriptorReader.java:255)
	at org.apache.maven.repository.internal.DefaultArtifactDescriptorReader.readArtifactDescriptor(DefaultArtifactDescriptorReader.java:171)
	at org.eclipse.aether.internal.impl.collect.DefaultDependencyCollector.resolveCachedArtifactDescriptor(DefaultDependencyCollector.java:538)
	at org.eclipse.aether.internal.impl.collect.DefaultDependencyCollector.getArtifactDescriptorResult(DefaultDependencyCollector.java:523)
	at org.eclipse.aether.internal.impl.collect.DefaultDependencyCollector.processDependency(DefaultDependencyCollector.java:410)
	at org.eclipse.aether.internal.impl.collect.DefaultDependencyCollector.processDependency(DefaultDependencyCollector.java:362)
	at org.eclipse.aether.internal.impl.collect.DefaultDependencyCollector.process(DefaultDependencyCollector.java:349)
	at org.eclipse.aether.internal.impl.collect.DefaultDependencyCollector.collectDependencies(DefaultDependencyCollector.java:254)
	at org.eclipse.aether.internal.impl.DefaultRepositorySystem.collectDependencies(DefaultRepositorySystem.java:284)
	at org.apache.maven.project.DefaultProjectDependenciesResolver.resolve(DefaultProjectDependenciesResolver.java:169)
[...]

PR #705, opened 2022-05-16
Basedir multi module project directory

mickaelistria:


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
```diff
@@ -174,7 +174,7 @@
 
 
 @Component(service = {IMaven.class, IMavenConfigurationChangeListener.class})
-public class MavenImpl implements IMaven, IMavenConfigurationChangeListener {
+public class MavenImpl implements IMaven, IMavenConfigurationChangeListener, Cloneable {
```
laeubi, line 177: I don't think that `Cloneable` is a good choice here and a (private) copy constructor would be much more suitable.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
```diff
@@ -1289,4 +1295,36 @@ public MavenExecutionResult execute(MavenExecutionRequest request) {
...
+    while(current != null) {
+      if(new File(current, ".mvn").isDirectory()) {
+        return current;
+      }
+      current = current.getParentFile();
+    }
+    return basedir;
+  }
+
+  public MavenImpl cloneForBasedir(File basedir) {
```
laeubi, line 1317: Why not passing an (optional) basedir to the createExecutionContext()? If we really need project specific maven impls then this hsould be handled in the IMavenFacade where you already added a `getMaven()` method

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
```diff
@@ -532,7 +535,10 @@ public List<MojoExecution> getExecutionPlan(String lifecycle, IProgressMonitor m
 
   @Override
   public IMaven getMaven() {
-    // TODO: if project has Maven customization (such as a .mvn, MAVEN_OPTS...)
-    return MavenPlugin.getMaven();
+    if(maven == null) {
+      File basedir = pomFile != null ? (pomFile.isDirectory() ? pomFile : pomFile.getParentFile()) : null;
+      maven = ((MavenImpl) MavenPlugin.getMaven()).cloneForBasedir(basedir);
```
laeubi, line 540: Actually the search for the `.mvn` folder should happen here and we should then get a MavenImpl per `.mvn` folder (that could be shared by multiple projects).

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
```diff
@@ -917,6 +918,9 @@ static class Context {
 
     request.setLocalRepository(getMaven().getLocalRepository());
     request.setWorkspaceReader(getWorkspaceReader(state, pom, resolverConfiguration));
+    if(pom != null && pom.getLocation() != null) {
+      request.setMultiModuleProjectDirectory(MavenImpl.computeMultiModuleProjectDirectory(pom.getLocation().toFile()));
```
laeubi, line 922: If we know the pom why have all the basedir hassle above? It seems this is only required here, because if the request was not aquired by a project then the `MultiModuleProjectDirectory` won'T have a real meaning

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
mickaelistria: What's the particular issue with Cloneable?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
mickaelistria: Because we basically want several MavenImpl in the end, 1 per project; and not 1 MavenImpl with multiple contexts. Stateful is beautiful!
But yes, we can move it to MavenProjectFacade, I think the only reason why I left it here instead is because I was too lazy to add a setter for the new basedir.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
laeubi: I found this rather descriptive: https://www.vojtechruzicka.com/java-cloning-problems/ so cloneable is hardly useful in most cases, if one needs multiple instances of a component you might want to use a factory component instead or the component context.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
laeubi: > Stateful is beautiful!

Yeah but the basedir is not really a state, and I think the idea is that we can share things across projects, an `IMaven` per projects would probably also blow up the memory consumption.

mickaelistria: > an IMaven per projects would probably also blow up the memory consumption.

Not for the moment as it's a shallow copy and things like ClassRealm (which are actually consuming RAM) are still shared.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
laeubi: Actually you can use the IProject location, also pomless *has* poms if we would use more maven like execution and if we would support extensions, what I try to in cooperate e.g. with #688 ... so from maven POV there is always a pom but not necessarily named `pom.xml` so one could safely assume that such a `MultiModuleProjectDirectory` would require some kind of "entry-point".

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
laeubi: > reusing as much as possible and only overriding the basedir to customize the executionRequest

But that the property of the `ExecutionRequest`. If we are talking about extensions and alike we need a *deep* copy. So for the sake of MultiModuleProject dir I think this is not something we should "clone" for and for the other case we should use a separate instance (as it would require new container).

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
laeubi: But that's the point here, if you just use IMaven to pass around a path, why not pass the path itself?
Especially as if an `IMaven` is bound to a project, where you could always get the basedir of the corresponding project instead...

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
mickaelistria: This patch is not about extensions at the moment. Let's stick to what it's supposed to fix now and we can improve later to grow to other cases.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
laeubi: > We don't need an IProject to get an IMavenProjectFacade, and it's important that things remain as such.

Okay then maybe `IMavenProjectFacade.getProject()` gives a wrong impression here, anyway I assume that each `IMavenProjectFacade` has a directory (aka basedir?) assigned, so this should be the entry point (a pom.xml is not needed). So then `IMavenExecutionContext.getExecutionRequest()` should contain a variant where one can pass a `IMavenProjectFacade` then which is then used to construct the appropiate Execution context without the need to store the basedir in the IMaven.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
laeubi: Yep that's why I think it should not be needed to clone/copy/... anything in `IMaven` at the moment but purely handle this in the `ExecutionContext`.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
mickaelistria: > But that's the point here, if you just use IMaven to pass around a path, why not pass the path itself?

Eventually, we'll pass more than the path. But the current issue is that it's much much more difficult to pass the path along the whole call chain as it requires to go through legacy APIs that do not allow propogation (don't accept extra args). I started trying that, had to create many more methods to IMaven, deprecate some more, it was basically not a good looking solution.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
mickaelistria: > IMavenProjectFacade.getProject()

I think so, and it's logged in some other issues that we may want to get rid of it to provide more clarity and flexibility (basically removing some apparent constraints).

> IMavenExecutionContext.getExecutionRequest() should contain a variant where one can pass a IMavenProjectFacade then which is then used to construct the appropiate Execution context without the need to store the basedir in the IMaven.

Again, that's what I started with, and it becomes a maze.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
laeubi: I don't think we must pass much around, but in 
`ProjectRegistryManager.configureExecutionRequest(MavenExecutionRequest, IProjectRegistry, IFile, ResolverConfiguration)`

if there is a pom set the multimodule dir. In all other cases there won't be an executable project anyways?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
laeubi: We might then simply set the property here as a first start this should cover all relevant use-cases (see comment above).

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
mickaelistria: Again, I tried, and it's not as simple as you think. The ProjectRegistryManager is just the very most basic place where the execution request is created, and it's an easy one to tweak; but in many places, the chain is IMaven.createExecutionContext().getExecutionRequest(), and both those methods are API and one of them would need to be deprecated/replaced; which is something we are happy to avoid by having the "basedir" (or other project specific settings) included in the IMaven directly.

laeubi: > but in many places, the chain is IMaven.createExecutionContext().getExecutionRequest(), and both those methods are API and one of them would need to be deprecated/replaced; which is something we are happy to avoid by having the "basedir" (or other project specific settings) included in the IMaven directly.

But how is then than the basedir ever injected? if not through the project facade? Can you give one example of such call?

mickaelistria: In the end, we do need some specific IMaven (per project, or per group of project sharing the same .mvn, whatever), so what it the issue with using MavenProjectFacade.getMaven() to decide of the configuration of the IMaven? I really don't get what is the problem here, and what you're suggesting recommends to keep using singletons and tweaks 11 places where we can just tweak 3 and allow objects to carry more context and more customization, as it is functionally needed.

mickaelistria: > The issue is that you don't really have a specific IMaven you clone a static singleton just to pass a File object (for each call)

Still, it's a different Maven, that will allow futher configurations. How it's constructed, what's clones and what's not doesn't really matter in term of coupling from consumer end.

> constructing other objects later on this just feels wrong

I really don't get the same impression. On the other end, tweaking several methods to always receive the same arguments, or at least arguments that are coupled together is a sign that a dedicated instance can be worth it. I would suggest you refer to "Clean Code" and "Design Patterns" books that spend time on finding the smell of when singletons are ready to become instances. And this kind of "propagate 1 value in many places" is the main smell.

>  And given that this only works if I call method X because someone before has called Y in the same thread makes me thing that this is just waiting for bad things to happen (even though I can't proof it).

That threading stuff is indeed a hotspot for issues, but as far as I looked up, it doesn't cause any issue compared to currnet state.

> either there is already a proper set up context from the project manager anyways

Unfortunately, the request in project manager is not coupled to further invocations, so what happens in it has no impact on next calls to IMaven.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
mickaelistria: It's not "simply" and what you suggest is adding 1 more API when we can do without it.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
mickaelistria: We already have 1 new API to customize how project execute stuff. We need to be able to leverage it. What you suggest is adding an extra API that does cover only a small part of what the current API -which already exists- can enable. So the proposal is weaker than the current state while adds complexity/API. Multiplying entry points that can achieve the same thing is not really helpful.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
laeubi: I don't know what new API you are referring here...  I'm refring to setting this in `ProjectRegistryManager.configureExecutionRequest(MavenExecutionRequest, IProjectRegistry, IFile, ResolverConfiguration)` this won't require new API and should suffice here.

mickaelistria: That's only 1 place where this API is used, and not sufficient. This method is only used in ProjectRegistryManager, some other places do require to have this (and more) set specifically for a given project, and that's for this kind of cases that we've created `MavenProjectFacade.getMaven()`

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
laeubi: > We already have 1 new API to customize how project execute stuff.

**Setup** and **Execution** are just two different parts in maven, so if you want to configure the execution it should be local to that execution and not global to the setup phase. and I don't propose any new API here, but cloning is the wrong technique here, it will completely disconnect you from any required OSGi life-cycle. If you still think than my non-new-api and local-injection of that property is not sufficient I would suggest we first solve the problem of multiple (independent) IMaven instances before proceed any further here. It will be required anyways for supporting extensions.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
laeubi: > That's only 1 place where this API is used, and not sufficient.

I still don't see why it is not sufficient, can you please give a concrete code reference where it wont work?

> This method is only used in ProjectRegistryManager,

Yes and this are the places where current code starts the call, and all other code just "connects" there, because if there is no project context there is no MultiModule directory and that is the place where the project facade is managed! So I don't see where (and why) a project is actually executed if not through the ProjectRegistryManager?

laeubi: Also if you take a look at the stack trace of https://github.com/eclipse-m2e/m2e-core/issues/666 you will clearly see that this is processed through the maven builder that exactly setup the execution context in that way in the stack frame here:

at org.eclipse.m2e.core.internal.project.registry.ProjectRegistryManager.refresh(ProjectRegistryManager.java:506)

and that's where the issue arises not elsewhere in the code, because there are only a build of the project from that point and all other code simply attach to that execution afterwards!

mickaelistria: If we fix only the ProjectRegistryManager, then further errors of that sort will happen eg if AbstractLifecycleMapping is used because the remote repo wouldn't be resolved and some dependencies/plugins won't be properly resolved for the project at that stage (although they were resolved earlier).

PR #706, opened 2022-05-16
[Logback] reduce rolling log file limit

HannesWell:
In order to reduce the disk-space occupation by m2e I want to propose the reduce the max-size of each log-file from 100MB to 10MB. Given that the rolling file-appender creates up to 10 files, this reduces the total log-file size from 1GB to 100MB per workspace!
I think this is more than sufficient. Who want's to know what exactly happens in its workspace 3months ago?
Personally I never look into the logs and given that the MavenConsole shows the log content in the IDE, from my standpoint, the rolling file appender can be removed completely.


HannesWell:
@laeubi, @mickaelistria do you have any opinion on this one?

mickaelistria:
> do you have any opinion on this one?

I don't

laeubi:
Never noticed that and can't remember any bug report where this was useful, so from my side lets reduce this.

PR #707, opened 2022-05-16
Prevent NPE on m2.pde.ui bundle stop

HannesWell:
It can happen that not all classes referenced in the m2e.pde.ui bundle's activator are initialized when the bundle is stopped. I think this happens when a user has not opened all tabs of the target-editor.

If m2e.pde.ui is stopped, e.g. when the application shuts down those classes are initialized during Bundle-stop. The initialization requires a Display instance that is null then this causing an NPE and finally a InitializationException.
This seems not to be harmful but can be disturbing for users as this shows up on the Error-log at the next start.


laeubi (approved):

HannesWell:
Thanks Christoph for reviewing.

Issue #709, opened 2022-05-17
remove addSettingsChangeListener/removeSettingsChangeListener from IMaven


laeubi:
The methods are not used anywhere and it seems we can reduce the footprint of the IMaven interface here.

mickaelistria:
No objection.

PR #714, opened 2022-05-17
remove add/removeSettingslistener from the IMaven interface

laeubi:
Fix https://github.com/eclipse-m2e/m2e-core/issues/709

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
@@ -37,10 +37,13 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Deactivate;
 import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.component.annotations.ReferenceCardinality;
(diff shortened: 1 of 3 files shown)


laeubi:
FYI @HannesWell

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
```diff
@@ -188,7 +191,8 @@ public class MavenImpl implements IMaven, IMavenConfigurationChangeListener, Clo
 
   private final ConverterLookup converterLookup = new DefaultConverterLookup();
 
-  private final List<ISettingsChangeListener> settingsListeners = new ArrayList<>();
+  @Reference(cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC)
```
HannesWell, line 194: `cardinality = ReferenceCardinality.MULTIPLE` is actually not necessary isn't it? (But I'm fine with making this explicit.
IIRC with the dynamic-policy new/removed service implementations are added/removed from the list instead of creating a new list, aren't they?

HannesWell (approved):
Looks good. Thank you!

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
laeubi: > `cardinality = ReferenceCardinality.MULTIPLE` is actually not necessary isn't it?

maybe I just always add it to make it clear :-)

PR #716, opened 2022-05-17
Move the basedir entirely into the IMavenExecutionContext

laeubi:
This scopes the basedir inside a unique IMavenExecutionContext, instead
of trying to provide a project specific IMaven, the IMavenProjectFacade
now allows of creation of a project scoped IMavenExecutionContext to be
used, see https://github.com/eclipse-m2e/m2e-core/discussions/710 for
further context.

@mickaelistria WDYT?

### DIFF
--- a/org.eclipse.m2e.apt.core/src/org/eclipse/m2e/apt/internal/AbstractAptConfiguratorDelegate.java
+++ b/org.eclipse.m2e.apt.core/src/org/eclipse/m2e/apt/internal/AbstractAptConfiguratorDelegate.java
@@ -52,6 +52,7 @@ import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.project.MavenProject;
 
 import org.eclipse.m2e.apt.internal.utils.ProjectUtils;
+import org.eclipse.m2e.core.MavenPlugin;
 import org.eclipse.m2e.core.project.IMavenProjectFacade;
 import org.eclipse.m2e.core.project.configurator.AbstractBuildParticipant;
 import org.eclipse.m2e.core.project.configurator.AbstractProjectConfigurator;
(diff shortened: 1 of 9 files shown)


mickaelistria:
I'm a deeply torn here. On one hand, I like the idea of hiding IMaven as much as possible as this is ultimately an interface that is not nice to use; but on the other hand, having access to IMaven instead of "just" IMavenExecutionContext can enable more project-specific configuration and to expose them fully from the facade.
I need to spend some more time to audit the IMaven class and clearly make my mind on whether there are still some methods that we are likely to be useful in a project specific context or not.

laeubi:
Sure take the time just some notes here

- I think `IMaven` is just to "fat" to be usefully defined on a per project basis and I found that we weather can add such project specific thing as a direct method to `IMavenProjectFacade` itself if it is execution independent. If it is bound to the execution, we should add it to the `IMavenExecutionContext` or the `IMaven` method could be context sensitive (e.g. fetching the current Thread Context and then use that information to fetch additional data, delegate to the MavenExecutionContext or whatever)
- We could just start right now with `IMavenExecutionContext` (as it would become handy anyways) and if we later still found  `IMaven` could still have project specific ones

mickaelistria:
FWIW, I think it's fundamentally correct: the Maven instance/installation doesn't seem to be fundamentally configured by the `${basedir}`, `${basedir}` only belongs to 1 Maven execution execution and not to the Maven configuration, it a N-1 mapping, not a 1-1. So passing the basedir when requesting a context is more correct.
However, I think we should still keep `getMaven()` in IMavenProjectFacade (and make it return a MavenImpl in MavenProjectFacade implementation, the singleton MavenImpl we have for now). but also offer a `createExecutionContext` as you suggest, which would consume `MavenProjectFacade.getMaven()` instead of the singleton one.

laeubi:
> However, I think we should still keep `getMaven()` in IMavenProjectFacade

done.

mickaelistria:
Thanks, I appreciate that you make this effort of a patch that can please both of us ;)

PR #719, opened 2022-05-18
Move execution part out of IMaven

laeubi:
This is archived by two ways:

1) IMavenProjectFacade now has a method to create an execution from it
2) IMavenExecutionContext now has a method join/canJoin

That way we don't need IMaven to perform what
IMavenExecutionContext#join now offers and we can execute without
reference IMaven as it actually only delegates to the
IMavenExecutionContext.

This is based on 
 - https://github.com/eclipse-m2e/m2e-core/pull/716
 
to show further usages and already give an out-sight of the further refactoring suggested in 
- https://github.com/eclipse-m2e/m2e-core/discussions/710


laeubi:
FYI @mickaelistria to see only changes between Commits you can choose "Files changed" > "Changes From ..." > the commit this is based on.

Just in case the other PR is merges, this needs rebase + adjustment of the test du to changed method signatures, thats why I marked this as a DRAFT for now.

mickaelistria:
I'm pretty sure those methods have a lot of consumers, that would then be broken with next release. The impact here is IMO too big (unlike other removals). We should `@Deprecate(forRemoval = true)` and document that `createExecutionContext().execute(...) should be used instead and adapt m2e-core so the new methods are used.
Such a patch could be merged without much concerns.

Then we can discuss other things like what the IMavenProjectFacade should expose, but I don't think both are so highly correlated so it's better to keep them in separate PRs so we make more efficient progress.

laeubi:
> I'm pretty sure those methods have a lot of consumers, that would then be broken with next release.

Don't we made 2.0 release to break things? :-)

> We should `@Deprecate(forRemoval = true)` and document that `createExecutionContext().execute(...) should be used instead

Sure that should be possible, I'll do that once we are confident with the change and make sure code in m2e itself do not longer use these methods.

> Then we can discuss other things like what the IMavenProjectFacade should expose, but I don't think both are so highly correlated so it's better to keep them in separate PRs so we make more efficient progress.

Do you refer to this PR or further parts? Not sure what exactly should be a separate PR.

mickaelistria:
> Don't we made 2.0 release to break things? :-)

Not really to break. Initial scope was to remove @Deprecated stuff (so we keep compatibility with up-to-date consumer who're supposed to have abandoned those APIs), then we've grown the scope to remove all APIs that seem unused (by "validating" unused as no relevant search result on GitHub) to also keep some compatibility.
But here, it's removing one of the most important method of that class, that is widely used. So it basically implies breaking all connectors and extenders. This is not what we want to achieve.

> Do you refer to this PR or further parts? Not sure what exactly should be a separate PR.

I refer to addition of IMavenProjectFacade.createExecutionContext(), which is not so related to this PR and already discussed in https://github.com/eclipse-m2e/m2e-core/pull/716

laeubi:
> I refer to addition of IMavenProjectFacade.createExecutionContext(), which is not so related to this PR

Alright, yes as described in the PR description we should first finish https://github.com/eclipse-m2e/m2e-core/pull/716 and I'll rebase this here then.

laeubi:
Replaced by: https://github.com/eclipse-m2e/m2e-core/pull/722

PR #722, opened 2022-05-18
Move execution part out of IMaven

laeubi:
This is archived by two ways:

1) IMavenProjectFacade now has a method to create an execution from it
2) IMavenExecutionContext now has a method join/canJoin

That way we don't need IMaven to perform what
IMavenExecutionContext#join now offers and we can execute without
reference IMaven as it actually only delegates to the
IMavenExecutionContext.

Existing methods are now marked for removal as they are probably widely
used.

See also https://github.com/eclipse-m2e/m2e-core/discussions/710

this replaces https://github.com/eclipse-m2e/m2e-core/pull/719

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMavenExecutionContext.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMavenExecutionContext.java
@@ -16,13 +16,18 @@ package org.eclipse.m2e.core.embedder;
 import org.eclipse.aether.RepositorySystemSession;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.Status;
 
 import org.apache.maven.artifact.repository.ArtifactRepository;
 import org.apache.maven.execution.MavenExecutionRequest;
+import org.apache.maven.execution.MavenExecutionResult;
 import org.apache.maven.execution.MavenSession;
+import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.project.ProjectBuildingRequest;
 
+import org.eclipse.m2e.core.internal.embedder.MavenExecutionContext;
+import org.eclipse.m2e.core.project.IMavenProjectFacade;
 import org.eclipse.m2e.core.project.IMavenProjectRegistry;
 
 
@@ -84,6 +89,20 @@ public interface IMavenExecutionContext {
   <V> V execute(MavenProject project, ICallable<V> callable, IProgressMonitor monitor) throws CoreException;
 
   /**
+   * @since 1.4
+   */
+  void execute(MavenProject project, MojoExecution execution, IProgressMonitor monitor) throws CoreException;
+
+  /**
+   * Execute the given {@link MavenExecutionRequest} in the context of m2eclipse and return the result, this could be
+   * seen as an embedded run of the maven cli, at least it tries to replicate as much as possible from that.
+   * 
+   * @param request a {@link MavenExecutionRequest}
+   * @return the result of the execution
+   */
+  MavenExecutionResult execute(MavenExecutionRequest request);
+
+  /**
    * @throws IllegalStateException if called outside of {@link #execute(MavenProject,ICallable, IProgressMonitor)}
    * @since 1.4
    */
@@ -105,5 +124,25 @@ public interface IMavenExecutionContext {
    * @since 1.4
    */
   ProjectBuildingRequest newProjectBuildingRequest();
+
+  /**
+   * Either join the currents thread {@link IMavenExecutionContext} or creates a fresh one from the supplied factory if
+   * this thread is currently not executing anything. Be aware that because of this might return an already executing
+   * context, this context should not be configured in any way, use {@link IMaven#createExecutionContext()} or
+   * {@link IMavenProjectFacade#createExecutionContext()} to create a specific context that could be configured.
+   * 
+   * @return a {@link IMavenExecutionContext} that could be used for execution
+   * @throws CoreException
+   */
+  static IMavenExecutionContext join(IMavenExecutionContextFactory factory) throws CoreException {
+    IMavenExecutionContext context = MavenExecutionContext.getThreadContext();
+    if(context == null) {
+      if(factory == null) {
+        throw new CoreException(Status.error("Nothing to join and no factory given!"));
+      }
+      return factory.createExecutionContext();
+    }
+    return context;
+  }
 
 }
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
@@ -30,10 +30,8 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
@@ -63,7 +61,6 @@ import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.MultiStatus;
-import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.osgi.util.NLS;
 
@@ -94,7 +91,6 @@ import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.InvalidRepositoryException;
 import org.apache.maven.artifact.repository.ArtifactRepository;
 import org.apache.maven.cli.configuration.SettingsXmlConfigurationProcessor;
-import org.apache.maven.eventspy.internal.EventSpyDispatcher;
 import org.apache.maven.execution.DefaultMavenExecutionRequest;
 import org.apache.maven.execution.DefaultMavenExecutionResult;
 import org.apache.maven.execution.MavenExecutionRequest;
@@ -106,9 +102,7 @@ import org.apache.maven.extension.internal.CoreExports;
 import org.apache.maven.extension.internal.CoreExtensionEntry;
 import org.apache.maven.lifecycle.LifecycleExecutor;
 import org.apache.maven.lifecycle.MavenExecutionPlan;
-import org.apache.maven.lifecycle.internal.DependencyContext;
 import org.apache.maven.lifecycle.internal.LifecycleExecutionPlanCalculator;
-import org.apache.maven.lifecycle.internal.MojoExecutor;
 import org.apache.maven.model.ConfigurationContainer;
 import org.apache.maven.model.Model;
 import org.apache.maven.model.Plugin;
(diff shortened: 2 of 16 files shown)


org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMavenExecutionContext.java
```diff
@@ -106,4 +125,31 @@ public interface IMavenExecutionContext {
...
+   * configured in any way, use {@link IMaven#createExecutionContext()} or
+   * {@link IMavenProjectFacade#createExecutionContext()} to create a specific context that could be configured.
+   * 
+   * @return a {@link IMavenExecutionContext} that could be used for execution
+   * @throws CoreException
+   */
+  static IMavenExecutionContext join() throws CoreException {
+    IMavenExecutionContext context = MavenExecutionContext.getThreadContext();
+    if(context == null) {
+      return MavenPlugin.getMaven().createExecutionContext();
```
mickaelistria, line 140: I don't think it's a good thing to have in IMavenExecutionContext, that is supposed to be deduced from one particular IMaven instance, a reference to the common/singleton IMaven. IMO, it's still something that makes more sense to keep on the consumer end or in IMaven.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMavenExecutionContext.java
```diff
@@ -106,4 +125,31 @@ public interface IMavenExecutionContext {
...
+  /**
+   * This method allows to check if the current thread can join a running execution
+   * 
+   * @return <code>true</code> if the current thread has an attached {@link IMavenExecutionContext} or
+   *         <code>false</code> otherwise.
+   */
+  static boolean canJoin() {
+    return MavenExecutionContext.getThreadContext() != null;
+  }
+
```
mickaelistria, line 154: Seems not used

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
```diff
@@ -632,14 +602,6 @@ public Artifact resolve(String groupId, String artifactId, String version, Strin
     return resolve(artifact, remoteRepositories, monitor);
   }
 
-  private IMavenExecutionContext context() {
```
mickaelistria, line 635: IMO, it's better to make it an API here than to have a static accessor API in IMavenExecutionContext; because how to choose or create the executionContext depends on the IMaven instance (it uses `createExecutionContext()`).

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMavenExecutionContext.java
laeubi: This is (as of now) the usual way for any code that do not want to be specific, to use the global execution as a default.
If the code cares one needs to create a special `IMavenExecutionContext` and not use join.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMavenExecutionContext.java
laeubi: Yes this is for completeness here, so if code want to do the following:

```
if (canJoin()) {
  //if something is running we want to attach
  join().execute(...)
} else {
 //we don't want the global context but can't join, do something special here
}
```

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
laeubi: See comments above, if one wants to distinguish the cases one needs to use `canJoin()/join()` instead, because code using this currently explicitly is "thread context aware" and code that don't want to creates own ones in the first place.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMavenExecutionContext.java
mickaelistria: > the usual way for any code that do not want to be specific

Yes, but we already have 1 entry point for the global execution (`MavenPlugin.getMaven()`), I don't think we should create new ones; ideally `MavenExecutionContext` shouldn't reference any `MavenCore.getMaven()` and should only use its own IMaven reference. Having consumers invoking `MavenPlugin.getMaven().getExecutionContext()` (was `context()` would IMO be cleaner) as it forces user to care about choosing IMaven they want to use to execute.

laeubi:
Here is an example from the last PR where one can see how you enter a specialized context for a contex-aware function:

```
 MavenProject mavenProject = mavenFacade.getMavenProject();
  return mavenFacade.createExecutionContext().execute(mavenProject, (context, monitor) -> {
    return mavenProject.getMaven().getMojoParameterValue(mavenProject, parameter, asType, mojoExecution.getPlugin(),
        execution, mojoExecution.getGoal(), null);
  }, null);
```

`IMaven#getMojoParameterValue` uses join() to join the already running (project scoped) execution here. Just in case `IMaven#getMojoParameterValue` is called *outside* it will use the global context as a fallback (as there is nothing that could be done anyways as using the global context).

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMavenExecutionContext.java
laeubi: > Yes, but we already have 1 entry point for the global execution (`MavenPlugin.getMaven()`)

That's exactly what I want to get rid of.

> ideally `MavenExecutionContext` shouldn't reference any `MavenCore.getMaven()`

sure, actually its not the `MavenExecutionContext` (object) but a static method I just feel its best placed in the interface as in a separate class/util method.

mickaelistria: > That's exactly what I want to get rid of.

We won't be able to remove `MavenPlugin.getMaven()` for a long time, and in any case, we'll probably still need it forever for compatibility and because we do need an access to default IMaven.
Replacing 1 static access by another doesn't change much IMO, and this hides to users some important coupling without removing this coupling. I don't get the benefit of not having it a (non-static) method in IMaven.

laeubi: > We won't be able to remove `MavenPlugin.getMaven()` for a long time, and in any case, we'll probably still need it forever for compatibility and because we do need an access to default IMaven.

I don't want to get rid of the `MavenPlugin.getMaven()`, but about IMaven#execute(....) methods! That's why they are marked for removal now, the context where we Execute is `IMavenExecutionContext`

> I don't get the benefit of not having it a (non-static) method in IMaven.

because it has nothing to do with `IMaven`, it is just a fallback for the case there is no active executing method, and we need this for two cases:

1. I want to execute something, I don't know (or care) about `IMaven` but want to join if the current thread already has a "transaction"
2. for backward compatibility reasons and we don't want the user to check for null all the time.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMavenExecutionContext.java
laeubi: This is now used.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMavenExecutionContext.java
laeubi: I have now adjusted to not create an execution implicitly but forcing any code that uses join to run inside a execute code block.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/MavenExecutionContextFactory.java
```diff
@@ -0,0 +1,28 @@
...
+package org.eclipse.m2e.core.embedder;
+
+import org.eclipse.core.runtime.CoreException;
+
+
+/**
+ * a {@link MavenExecutionContextFactory} is a supplier for new {@link IMavenExecutionContext}s
+ *
+ */
+public interface MavenExecutionContextFactory {
```
mickaelistria, line 20: We should probably keep the `I` prefix for consistency.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMavenExecutionContext.java
```diff
@@ -106,4 +125,46 @@ public interface IMavenExecutionContext {
...
+   * @return a {@link IMavenExecutionContext} that could be used for execution
+   * @throws CoreException
+   */
+  static IMavenExecutionContext joinOrCreate(MavenExecutionContextFactory factory) throws CoreException {
+    if(canJoin()) {
+      return join();
+    }
+    return factory.createExecutionContext();
+  }
+
```
mickaelistria, line 159: I'm not sure all that should be static: at the moment, it is assumed that if there is a thread context, it's good to use it; but is it always true? ie if the current thread context doesn't use the same IMaven, is it the one that should be used?
I still think `static` is harmful here. Could those be non-static methods in the IMavenExecutionContext factory; or just a new non-static method in IMaven? I know you'd like to avoid it, but I think most ways around it are actually bringing more complexity and less flexiiblity. What is particularly wrong with `IMaven.context()` being an API that returns a suitable context (may reuse one from the thread or create one)?

laeubi: `MavenExecutionContext` is an internal class and as you can see exactly that method is used here :-)

mickaelistria: The method could be moved to `IMavenExecutionContext` if it's generally useful. And could be static as it really doesn't depend on any state beyond the static current thread.

laeubi: Well that's what this wrapper method (join()) do + some checks as interface can't contain (private) fields

mickaelistria: Yes, but then why an extra new method with more behavior if we can just expose an existing method? Seems to me that this method -as it takes a factory as input- does require extra interface and so on; while just exposing `getThreadContext()` wouldn't require so much.
Then, clients can do `Optional.of(IMavenExecutionContext.getThreadContext()).orElse(projectFacade::createExecutionContext)` which gives them more knowledge and responsibility over the context they want ot use. They may often realize that `projectFacade.createExecutionContext()` might be a better choice while this method would hide such possible improvements.

laeubi: > Yes, but then why an extra new method with more behavior if we can just expose an existing method? Seems to me that this method -as it takes a factory as input- does require extra interface and so on;

this is because you have had concerns using the `MavenPlugin().getMaven()` here is bad, so I added the opportunity to pass arbitrary "factory" method to construct a default.

> Then, clients can do

There is just one problem `createExecutionContext` throws an CoreException and thus you can't simply do as you suggested here (thats also why I needed a new interface in contrast to just using `Supplier<..>`.

We could change this, but this would be a (compile time) API break.

mickaelistria:
> any chance we can come to a conclusion here?

I still don't see the value of this `join` and `canJoin` methods over introducing `getThreadContext()` as API (I think it just returns null without failure when there is no context, so it can do both).

>  think this do not make things worse (but maybe a bit more clear) compared to current state

Assing APIs is worse than not adding APIs ;)

laeubi:
> I still don't see the value of this `join` and `canJoin` methods over introducing `getThreadContext()` as API (I think it just returns null without failure when there is no context, so it can do both).

That's why I suggest we merge this and then you can try to get rid of it rigtht after that, I think this is better checked by actual code changes. If we don't have this helper methods code at the caller-side will get much more complicated, or we introduce the same helper method on other sites...

laeubi:
> I still don't see the value of this `join` and `canJoin` methods over introducing `getThreadContext()`

I have removed the `canjoin` now and squashed the both join variants into one... home this is satisfactory now...

mickaelistria:
> That's why I suggest we merge this and then you can try to get rid of it rigtht after that, I think this is better checked by actual code changes.

OK, let's merge and I'll remove.

> If we don't have this helper methods code at the caller-side will get much more complicated,

I think it can actually be a good thing that the caller have to care about the context when they need to wrap a Maven execution.

PR #726, opened 2022-05-19
Add ServiceUsage utility and OSGiServiceInjector test-rule

HannesWell:
This PR is is intended to incubate/try out two utilities to simplify the correct usage of OSGi services:

- **ServiceUsage** which is effectively an AutoClosable Wrapper around a service usage that can be used inside a try-with-resources block and ensures that the service object is `unget` after it has been used. It is currently located in `M2EUtils` and can be obtained via the `useService(Class)` respectively `useService(Class, Function)`.
- **OSGiServiceInjector** JUnit-4 rule which injects OSGi service implementation into correspondingly annotated Test-class fields.
The implementation currently based on the E4's `ContextInjectionFactory` but it likely has to be implemented 'manually' by using a ServiceTracker per field that is closed after the evaluation because the `EclipseContextOSGi` ungets the service-reference immediately. Using custom annotations like `@InjectService` would also make it more clear what is happening and would allow parameters if necessary (to specify the element type of a collection?).
The same functionality could be implemented for JUnit-5 using a corresponding `Extension`.

I already applied them to a few locations within the m2e-core code-base and in https://github.com/tesla/m2e-core-tests/pull/155 but there are more possible.

I think similar functionality could also be useful for many Eclipse-Platform projects because Services are there often not used correctly.
Finally it could maybe even be considered for inclusion into OSGi itself respectively [osgi-test](https://github.com/osgi/osgi-test/) (although `useService()` requires Java-9+).

@laeubi what do you think?


org.eclipse.m2e.tests.common/src/org/eclipse/m2e/tests/common/OSGiServiceInjector.java
```diff
@@ -0,0 +1,49 @@
...
+  public Statement apply(Statement base, FrameworkMethod method, Object target) {
+    return new Statement() {
+      @Override
+      public void evaluate() throws Throwable {
+        Bundle bundle = FrameworkUtil.getBundle(target.getClass());
+        BundleContext context = bundle.getBundleContext();
+        if(context == null) {
+          throw new IllegalStateException("Test bundle <" + bundle.getSymbolicName() + "> not started");
+        }
+        // TODO: BundleContextOSGi ungets services immediately!
```
laeubi, line 38: This is something that should be addressed at E4 and not at m2e

laeubi:
If you want to only access a service once eclipse has "ServiceCaller" class already, but one should be aware that getting/ungetting a service might create noise, so I think this should really only be used in very rare cases.

HannesWell:
> If you want to only access a service once eclipse has "ServiceCaller" class already, but one should be aware that getting/ungetting a service might create noise, so I think this should really only be used in very rare cases.

Thank you for the suggestion.
I just drafted a PR to add a corresponding method to it:
https://github.com/eclipse-equinox/equinox/pull/146

I plan to create a corresponding PR to add something like the suggested Junit-Rule to inject OSGi-services for tests to [org.eclipse.test](https://github.com/eclipse-platform/eclipse.platform.releng/tree/master/bundles/org.eclipse.test) and will then abandon this PR.

laeubi:
> Finally it could maybe even be considered for inclusion into OSGi itself respectively [osgi-test](https://github.com/osgi/osgi-test/) (although `useService()` requires Java-9+).

OSGi Test already has ways to fetch services by annotations:

https://github.com/osgi/osgi-test/blob/main/org.osgi.test.junit5/README.md#testing-with-osgi-services

I really like to see this getting used but currently it is a bit horrible to use inside PDE see:
- https://github.com/osgi/osgi-test/pull/503

and not very intuitive to include it in Tycho see:
- https://github.com/eclipse-tycho/tycho/issues/1556

so probably it would be much more better to improve the support here at PDE/Tycho than adding more "helpers" ...

PR #729, opened 2022-05-22
Add support to lookup of extensions

laeubi:
This add first support of lookup extension components, currently only
project scoped extensions are supported but could be enhanced later to
support core extensions as well.

### DIFF
--- a/org.eclipse.m2e.core.tests/src/org/eclipse/m2e/core/ExtensionsTest.java
+++ b/org.eclipse.m2e.core.tests/src/org/eclipse/m2e/core/ExtensionsTest.java
@@ -1,0 +1,39 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Christoph Läubrich and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.m2e.core;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+import java.util.Collection;
+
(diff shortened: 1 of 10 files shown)


laeubi:
@kjsmita6 this should now be the "dynamic" variant of your change, do you think you can test this with your custom plugin as well? With tycho extension it looks like this:

grafik

laeubi:
@mickaelistria any concerns?

kjsmita6:
@laeubi Looks good with my testing (checked out your fork). In my pom I have defined:
```xml
  <build>
        <plugins>
            <plugin>
                <groupId>com.tibco.ep</groupId>
                <artifactId>ep-maven-plugin</artifactId>
                <extensions>true</extensions>
            </plugin>
        </plugins>
    </build>
```

And shows up in the pom editor as:
Screenshot (57)

org.eclipse.m2e.core.tests/src/org/eclipse/m2e/core/ExtensionsTest.java
```diff
@@ -0,0 +1,30 @@
+package org.eclipse.m2e.core;
```
mickaelistria, line 1: Please include legal header.

mickaelistria (changes requested):

mickaelistria:
@laeubi I'd like to merge it ASAP (only concern is the missing copyright header on new file) as this improves an important part of m2e and really emphasizes the progress that are possible with recent changes.

org.eclipse.m2e.core.tests/src/org/eclipse/m2e/core/ExtensionsTest.java
laeubi: Done.

laeubi:
I have added the header now.

mickaelistria:
Thanks, that's a great patch!

laeubi:
> Thanks, that's a great patch!

Its just a first small step so don't be to enthusiastic :-)

PR #732, opened 2022-05-23
Cut dependency to maven archetype runtime from m2e.core

### LINKED ISSUE #731: Cut dependency to maven archetype runtime from m2e.core
Related Eclipse bug - https://bugs.eclipse.org/bugs/show_bug.cgi?id=517822

> When embedding m2e-core to jdt.ls, we bring in the maven archetype runtime as dependency, which could be avoided if m2e-core were refactored to not depend on that runtime directly.
> We should extract all archetype related operations into a separate plugin.

snjeza:
Fixes #731 

Signed-off-by: Snjezana Peco <snjezana.peco@redhat.com>

### DIFF
--- a/m2e-maven-runtime/org.eclipse.m2e.archetype.common/pom.xml
+++ b/m2e-maven-runtime/org.eclipse.m2e.archetype.common/pom.xml
@@ -97,8 +97,8 @@
 						<![CDATA[
 							-exportcontents: \
 								META-INF.plexus;-noimport:=true;x-internal:=true,\
-								org.apache.maven.archetype.*;provider=m2e;mandatory:=provider;x-internal:=true,\
-								org.codehaus.plexus.velocity;provider=m2e;mandatory:=provider;x-internal:=true
+								org.apache.maven.archetype.*,\
+								org.codehaus.plexus.velocity
 							Require-Bundle: org.eclipse.m2e.maven.runtime;bundle-version="[1.18.0,1.19.0)"
 						]]>
 						</bnd>
--- a/org.eclipse.m2e.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.core/META-INF/MANIFEST.MF
@@ -8,7 +8,6 @@ Bundle-Vendor: %Bundle-Vendor
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.osgi;bundle-version="3.10.0",
  org.eclipse.m2e.maven.runtime;bundle-version="1.16.0",
- org.eclipse.m2e.archetype.common;bundle-version="1.16.0",
  org.eclipse.m2e.workspace.cli;bundle-version="0.1.0",
  org.eclipse.core.runtime;bundle-version="3.12.0",
  org.eclipse.core.resources;bundle-version="3.9.0",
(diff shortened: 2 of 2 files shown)


mickaelistria (changes requested):
I don't think just marking "optional" is a good thing, nor that it is technically correct.
One alternative would be to replace "Require-Bundle" by "Import-Package" and then in JDT-LS, you can create a fake bundle exporting the package and the Archetype class as a mock.

snjeza:
> I don't think just marking "optional" is a good thing,...

@mickaelistria @HannesWell I have updated the PR.

fbricon:
Why make it simple when you can make it complicated?
Not only do we need to add bloat to the m2e manifest, but we also impose the responsibility of providing mocks to jdt.ls.

laeubi (changes requested):
> Why make it simple when you can make it complicated?

Because the "simple" solution is plainly wrong and will break m2e ... The org.eclipse.m2e.archetype.common is not optional for m2e and doing so would require a lot of "bloat" to handle that correctly.

I'd just want to quote the original request here (emphasis by me):

> When embedding m2e-core to jdt.ls, we bring in the maven archetype runtime as dependency, which could be avoided if m2e-core were refactored to not depend on that runtime directly.
> 
> **We should extract all archetype related operations into a separate plugin**.

Given that this PR does not solve the issue in any way I think.

> Not only do we need to add bloat to the m2e manifest, but we also impose the responsibility of providing mocks to jdt.ls.

Import-package is not bloat but good practice in OSGi... and yes, if we do not refactor m2e-core in a way that it does not depend on the archetype then jdt.ls has to provide some kind of _polyfill_.

On the other hand one should ask: Whats bad of embedding the archetype stuff (beside the 2mb of additional jar size)?

mickaelistria:
Thanks @snjeza. I've opened https://github.com/eclipse-m2e/m2e-core/pull/735 which just import fewer packages. If build succeeds and JDT-LS can provide mocks for those packages to not require all archetype stuff. However, as @laeubi noticed, a longer term solution would be to extract, or even remove, parts that are related to archetypes from m2e.core; particularly as it seems like they're mostly used in m2e.core.ui.

fbricon:
For sure, the best solution is to refactor the archetype stuff out of core. @snjeza's solution was an attempt at getting a cheap solution for our issue, turns out it failed. Had that worked, that would have proven m2e-core could be used without the archetype stuff, making it effectively optional. Now that that point is moot, I'm definitely not found of using required packages, when it requires listing a ton of packages, but I'll take it as an interim solution, short of proper refactoring. 

Cutting archetype out of core has been discussed informally for a long time, before jdt.ls even existed, as newer versions bring even more baggage (eg. groovy), that'll make the whole thing even heavier. Today, simply making jdt.ls (~5%) leaner is our immediate goal.

laeubi:
@fbricon just remove the bundle from the manifest and you will see all the places where a compile error occur, all those classes won't be loadable once we make it optional and the bundle is not there, I would suspect that then even most parts of m2e fail with classnotfound exceptions.

fbricon:
@laeubi I'm well aware. It all depends on the how m2e classes are loaded. Anyways we've already established this doesn't work in practice, so we can move on

snjeza:
> @fbricon just remove the bundle from the manifest and you will see all the places where a compile error occur, all those classes won't be loadable once we make it optional and the bundle is not there, I would suspect that then even most parts of m2e fail with classnotfound exceptions.

There are 85 compiler errors

HannesWell:
Since the dependency from m2e.core to archetype was removed thanks to @laeubi via the PRs #758, #762 and #763 I think this now obsolete and can be closed.

PR #737, opened 2022-05-24
Remove IMavenExecutionContextFactory

mickaelistria:
This interface is never used directly and it doesn't seem profitable to
introduce a new API and more complex type hierarchy for that. This will
allow easier changes in the future.


laeubi:
The idea is/was that one can write generic methods that can be used by either of both, of course now it is not used as you replaced all (m2e internal) calls with repetitive code blocks.

But for example the "getExecutionContext" method that is currently private to MavenImpl could be a (default) implemented method in this interface, so everyone extending that interface gets a `getExecutionContext` for free...

mickaelistria:
> The idea is/was that one can write generic methods that can be used by either of both, of course now it is not used as you replaced all (m2e internal) calls with repetitive code blocks.

We can refactor later if there is a clear need for a generic method. But with the current code and current needs, there is nothing that makes the interface profitable, so let's minimize the API changes.

> so everyone extending that interface gets a getExecutionContext for free..

I'm not sure we really want that. Actually, I think we'd probably do better moving away from using the ThreadLocal context later in the API and make it internal to the execution context implementation.
Anyway, that part is a bit out of the scope.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
```diff
@@ -274,4 +273,18 @@ default <V> V execute(ICallable<V> callable, IProgressMonitor monitor) throws Co
...
+   * <p>
+   * <b>IMPORTANT:</b> When in the context of a particular project, it's usually better to use
+   * {@link IMavenProjectFacade#createExecutionContext()} which will include more project-specific configuration and
+   * will lead to more accurate and consistent results compared to Maven CLI commands.
+   * </p>
+   *
+   * @since 1.4
+   * @see IMavenProjectFacade#createExecutionContext()
+   */
+  IMavenExecutionContext createExecutionContext();
```
laeubi, line 288: And now instead of implement / define that twice you could add this (and default implement it) in IMavenExecutionContextFactory ;-)

PR #741, opened 2022-06-01
[Runtime] Embed Maven 3.8.6

HannesWell:
Maven 3.8.6 development has been completed: https://www.mail-archive.com/dev@maven.apache.org/msg127028.html
Lets try it. Once it has been released this can be completed.

### DIFF
--- a/m2e-maven-runtime/org.eclipse.m2e.maven.runtime/pom.xml
+++ b/m2e-maven-runtime/org.eclipse.m2e.maven.runtime/pom.xml
@@ -19,14 +19,14 @@
 	</parent>
 
 	<artifactId>org.eclipse.m2e.maven.runtime</artifactId>
-	<version>1.18.4-SNAPSHOT</version>
+	<version>1.18.5-SNAPSHOT</version>
 	<packaging>eclipse-plugin</packaging>
 
 	<name>M2E Embedded Maven Runtime (includes Incubating components)</name>
 
 	<properties>
 		<!-- maven core version -->
-		<maven-core.version>3.8.5</maven-core.version>
+		<maven-core.version>3.8.6</maven-core.version>
 		<!-- NOTE: When maven-core.version changes, this may impact the versions of the maven-resolver-*
 			jars that export the org.eclipse.aether.* packages in the org.eclipse.m2e.maven.runtime
 			bundle. So make sure the following variable has the value of the maven-resolver-* jars
@@ -138,7 +138,7 @@
 					<artifactId>bnd-maven-plugin</artifactId>
 					<configuration>
 						<bnd>
-						<![CDATA[
+							<![CDATA[
 							-exportcontents: \
 								META-INF.plexus;-noimport:=true;x-internal:=true,\
 								META-INF.sisu;-noimport:=true;x-internal:=true,\
@@ -157,7 +157,7 @@
 								org.eclipse.m2e.maven.runtime.slf4j.simple;bundle-version="[1.18.0,1.19.0)",\
 								com.google.guava
 						]]>
-						<!-- All direct dependencies specified as Require-Bundle or Import-package are added to the classpath of a launched 
+							<!-- All direct dependencies specified as Require-Bundle or Import-package are added to the classpath of a launched 
 							Maven-Build-JVM. See MavenEmbeddedRuntime for details. -->
 						</bnd>
 					</configuration>


HannesWell:
Btw. what do you think about aligning the version of `org.eclipse.m2e.maven.runtime` with the embedded maven version?
Although `org.eclipse.m2e.maven.runtime` is not exactly maven, I think Maven is the most relevant part and I think it would be handy to be able to derive the embedded version from the plug-in version.

mickaelistria:
> Btw. what do you think about aligning the version of org.eclipse.m2e.maven.runtime with the embedded maven version?

Sounds OK to me.

laeubi:
Why not, but actually I wonder if we should not ask the maven people to add OSGi metadata in the first place?
For the SPI stuff we could add appropriate http://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.loader.html headers or ask to use the [GlassFish ServiceLoader](https://github.com/eclipse-ee4j/glassfish-hk2-extra/blob/master/osgi-resource-locator/src/main/java/org/glassfish/hk2/osgiresourcelocator/ServiceLoader.java).

mickaelistria:
> Why not, but actually I wonder if we should not ask the maven people to add OSGi metadata in the first place?

It's a much more complex thing to implement. Also, based on my impressions with lemminx (which can be run in OSGi way but...) it's not so much adding the metadata that helps but also having real supports (eg tests) for OSGi environments. Without them, it's relatively easy to ship good jars that are just broken bundles, particularly when playing with extensibility and classloading. If Maven adds metadata and doesn't test and m2e is the main or only client, then it means that m2e is basically the project supporting the OSGi execution of Maven, that's a lot of additional duties over just embedding the distribution.

laeubi:
> that's a lot of additional duties over just embedding the distribution.

I think it doesen't matter much if we rebundle the stuff or if we do the same at the maven project, testing new maven releases is required anyway and we should participate there e.g. by testing snapshots. If we would include everything as regular maven artifacts, it would be much more easier to do so and even people would be able to upgrade maven if they like independent from m2e, also the recent discussion about the "size" of m2e could be solved as optional or transtive deps might just be excluded, while currently we ship all in some big bundles.

mickaelistria:
> I think it doesen't matter much if we rebundle the stuff or if we do the same at the maven project

I've had multiple ppposite experiences with that since I started to do Eclipse development 14 years ago qp until now. So I don't share your optimism. But m2e has the advantage of having a very good test suite with good coverage, so it gives a good harness to try it if you're interested.

> people would be able to upgrade maven if they like independent from m2e

I don't think m2e users do care about that and they will always take the version shipped by m2e anyway.

> recent discussion about the "size" of m2e could be solved as optional or transtive deps might just be excluded

If we use Import-Package in m2e, then we already have modularity and those people who have particular issues with the size can bundle maven however they want (eg create a com.acme.m2e.custom.maven.runtime to use in place of org.eclipse.m2e.maven.runtime).

HannesWell:
> Why not, but actually I wonder if we should not ask the maven people to add OSGi metadata in the first place? 

At the moment I would also prefer to continue embedding the jars for the following reasons:
- m2e.maven.runtime contains many different jars from different vendors that IIRC are mostly non OSGi compliant. So it likely would require some work to provide suitable manifests for all jars and to update the dependencies so that maven in the end uses the new ones.
- A regular Maven installation is a specific set of jars and in order to ensure exactly the same behavior like a regular installation I think we should not use the dynamics of OSGi here.
- Since the m2e.maven.runtime is not an ideal candidate to be OSGi-ifyed I had the intention to make it a "show-case"/"dog-feeding" case to build an Maven-based OSGi-bundle that embeds Maven-dependencies and works in the workspace and build without any further actions. Basically what https://github.com/eclipse-m2e/m2e-core/pull/466 should be in the end.

laeubi:
> * m2e.maven.runtime contains many different jars from different vendors that IIRC are mostly non OSGi compliant.

m2e target locations can generate compliant jars for them ... beside that we could also ask those to add OSGi headers, one just has to start somewhere...

> * A regular Maven installation is a specific set of jars and in order to ensure exactly the same behavior like a regular installation I think we should not use the dynamics of OSGi here.

Having one big bundle is not more dynamic than having many ones (just more flexible)

HannesWell:
YES! It finally works 😄 🎉

mickaelistria:
@HannesWell great work on the IP stuff!

HannesWell:
> @HannesWell great work on the IP stuff!

Thanks :)

HannesWell:
Maven 3.8.6 has been released.
All tests are green and we have IP clearance. So this is ready for submission.

I will perform the suggested version alignment as soon as we have updated to archtype 3.x since the version change requires changes in the embedded archetype and I don't want to change the archetype version just for this cosmetic change. But I think the update will happen soon.

HannesWell:
> Btw. what do you think about aligning the version of `org.eclipse.m2e.maven.runtime` with the embedded maven version? 

Done with https://github.com/eclipse-m2e/m2e-core/pull/779.

PR #750, opened 2022-06-05
Add BREE Header for Maven-Runtime components

HannesWell:
To fix warnings about missing Bundle-RequiredExecutionEnvironment add a
corresponding bnd-instruction to set it.

Although Maven 3 actually only requires Java-1.7, BND-tools generates a
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"

In order to not cause conflicts Java 1.8 is used as BREE as well.

Additionally align the JRE-container in the .classpath file to not cause
another warning.

### DIFF
--- a/m2e-maven-runtime/pom.xml
+++ b/m2e-maven-runtime/pom.xml
@@ -27,6 +27,7 @@
 		<dependency.jars.folder>jars</dependency.jars.folder>
 		<dependency.sources.folder>jars/sources</dependency.sources.folder>
 		<failIfMacSigningFailed>false</failIfMacSigningFailed>
+		<mavenJavaBREE>JavaSE-1.8</mavenJavaBREE>
 	</properties>
 
 	<modules>
@@ -71,6 +72,7 @@
 							Import-Package: !*
 							Automatic-Module-Name: ${bsn}
 							Eclipse-BundleShape: dir
+							Bundle-RequiredExecutionEnvironment: ${mavenJavaBREE}
 						]]>
 						</bnd>
 					</configuration>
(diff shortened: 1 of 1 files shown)


HannesWell:
Alternatively the BREE could be set to Java-17 to align with the rest of the project.

laeubi:
> @mickaelistria, @laeubi what do you think?

I think PDE should be fixed to understand `Require-Capability`, BREE is deprecated by the OSGi spec:

> The Bundle-RequiredExecutionEnvironment manifest header provides the same function as the [osgi.ee Namespace](http://docs.osgi.org/specification/osgi.core/7.0.0/framework.namespaces.html#framework.namespaces.osgi.ee). It allows a bundle to depend on the execution environment. **This header is deprecated** but must be fully supported by a compliant framework. Bundles should not mix these headers but use either an osgi.ee requirement or this header.

http://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.module.bree

laeubi:
> Although Maven 3 actually only requires Java-1.7, BND-tools generates

As BND uses class info to derive this data, there might be a (dependency?) class that actually uses Java 1.8

HannesWell:
> I think PDE should be fixed to understand `Require-Capability`, BREE is deprecated by the OSGi spec:

Good point. I created https://github.com/eclipse-pde/eclipse.pde/issues/140 to track that for PDE.

Since osgi.ee and BREE should not be mixed I'm closing this one and we have to wait for a fix in PDE.


> > Although Maven 3 actually only requires Java-1.7, BND-tools generates
> 
> As BND uses class info to derive this data, there might be a (dependency?) class that actually uses Java 1.8

Assumed the same and a quick search revealed that at least `io.takari.aether:aether-connector-okhttp` has a property `takari.javaSourceVersion=1.8`. So that's probably the cause.

PR #753, opened 2022-06-05
Update to Maven Archetype 3.2.1

HannesWell:
This draft provides an intermediate state of the migration to Archetype 3.2.1.
The dependencies are updated and the code compiles.
I also tested remote archtype catalogs and they worked well. The defaultLocal catalog failed in my test and arbitrary local catalog is not yet tested.

I pushed this branch into the m2e-core repo directly. So @laeubi if you want to use this and it is suitable, you can directly work on this branch.

This also includes some parts of https://github.com/eclipse-m2e/m2e-core/pull/735 and fixes restriction warnings in m2e.core and m2e.core.ui.

### DIFF
--- a/m2e-maven-runtime/org.eclipse.m2e.archetype.common/pom.xml
+++ b/m2e-maven-runtime/org.eclipse.m2e.archetype.common/pom.xml
@@ -19,15 +19,13 @@
 	</parent>
 
 	<artifactId>org.eclipse.m2e.archetype.common</artifactId>
-	<version>1.18.2-SNAPSHOT</version>
+	<version>3.2.1-SNAPSHOT</version>
 	<packaging>eclipse-plugin</packaging>
 
 	<name>M2E Maven Archetype Common</name>
 
 	<properties>
-		<archetype-common.version>2.4</archetype-common.version>
-		<dom4j.version>2.1.3</dom4j.version>
-		<commons-collections.version>3.2.2</commons-collections.version>
+		<archetype-common.version>3.2.1</archetype-common.version>
 	</properties>
 
 	<dependencies>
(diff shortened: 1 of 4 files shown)


laeubi:
> So @laeubi if you want to use this and it is suitable, you can directly work on this branch.

@HannesWell I think if all compiles and all test are passing we are ready to merge this, but it seems there are test failures, do you plan to take a look into those or should I take over?

laeubi:
> pushed this branch into the m2e-core repo directly.

By the way this is not strictly necessary, you can enable the check mark "allow maintainer to edit this branch" or one could simply import your PR.

HannesWell:
> > So @laeubi if you want to use this and it is suitable, you can directly work on this branch.
> 
> @HannesWell I think if all compiles and all test are passing we are ready to merge this, but it seems there are test failures, do you plan to take a look into those or should I take over?

I can have a short look this evening. I suspect the problem is in wrong/missing configuration in the `getArchetypeCatalog()` implementations in `ArchetypeCatalogFactory`.
But I don't want to spent much more time on this, since you plan a greater rework and I think its not worth to invest much in something that is removed soon anyway. I will let you know if I have given up. :)


> > pushed this branch into the m2e-core repo directly.
> 
> By the way this is not strictly necessary, you can enable the check mark "allow maintainer to edit this branch" or one could simply import your PR.

Indeed. I have seen "allow maintainer to edit this branch"  before but never used it respectively have seen that it is used, so I don't know how (well) this works. Yes it can be imported as well, I just wanted to avoid a later closing+re-creating PRs.

laeubi:
@HannesWell I think you are right that it is not worth fixing the tests ... I'll try to apply some smaller changes and then try to apply your migration to Archetype 3!

HannesWell:
> @HannesWell I think you are right that it is not worth fixing the tests ... I'll try to apply some smaller changes and then try to apply your migration to Archetype 3!

Alright, thanks for taking over. Let me know if there is something I can help with/review.

akurtakov:
VERY NICE ! 
@waynebeaton , looks like license review is moving to a whole new level with this feature!

laeubi:
This is now included as part of other changes.

PR #754, opened 2022-06-06
Add support for maven-rooted PlexusContainer

laeubi:
This adds support for PlexuContainers rooted as the "mvn" folder and
lookup of components in three different context:

- IMaven lookup in the default maven context not containing any
extensions
- IMavenProjectFacade lookup in the projects maven context containing
core extensions
- IMavenExecutionContext lookup in the executions maven context
containing core and project scoped extensions

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IComponentLookup.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IComponentLookup.java
@@ -1,0 +1,44 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Christoph Läubrich
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *      Christoph Läubrich - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.m2e.core.embedder;
+
+import java.util.Collection;
+
+import org.eclipse.core.runtime.CoreException;
+
+
+/**
+ * A {@link IComponentLookup} looks up a component in a given context
+ * 
+ * @noimplement This interface is not intended to be implemented by clients.
+ * @since 2.0
+ */
+public interface IComponentLookup {
+
+  /**
+   * Lookup a single component from this context.
+   * 
+   * @param clazz the requested role
+   * @return The component instance requested.
+   * @throws CoreException if the requested component is not available
+   */
+  <C> C lookup(Class<C> type) throws CoreException;
+
+  /**
+   * Look up a list of components in a given context
+   * 
+   * @param <C>
+   * @param extensionType
+   * @return a (possibly empty) list of the requested components
+   * @throws CoreException
+   */
+  <C> Collection<C> lookupCollection(Class<C> type) throws CoreException;
+}
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
@@ -56,7 +56,7 @@ import org.eclipse.m2e.core.project.IMavenProjectFacade;
  * @author igor
  * @noimplement This interface is not intended to be implemented by clients.
  */
-public interface IMaven {
+public interface IMaven extends IComponentLookup {
 
   // POM Model read/write operations
 
@@ -229,8 +229,6 @@ public interface IMaven {
    */
   ClassLoader getProjectRealm(MavenProject project);
 
-  // execution context
-
   /**
    * This is convenience method fully equivalent to
    *
@@ -241,8 +239,8 @@ public interface IMaven {
    * return context.execute(callable, monitor);
    * </pre>
    *
-   * @deprecated should be replaced with the fully equivalent code mentioned in this javadoc, e.g. inside a private
-   *             util method
+   * @deprecated should be replaced with the fully equivalent code mentioned in this javadoc, e.g. inside a private util
+   *             method
    * @since 1.4
    */
   @Deprecated(forRemoval = true)
@@ -262,17 +260,6 @@ public interface IMaven {
   default <V> V execute(ICallable<V> callable, IProgressMonitor monitor) throws CoreException {
     return IMavenExecutionContext.getThreadContext().orElseGet(this::createExecutionContext).execute(callable, monitor);
   }
-
-  /**
-   * Lookup a component from the embedded PlexusContainer.
-   * @param clazz the requested role
-   * @return The component instance requested.
-   * @throws CoreException if the requested component is not available
-   *
-   * @since 1.10
-   */
-  <T> T lookup(Class<T> clazz) throws CoreException;
-
 
   /**
    * Creates and returns new, possibly nested, maven execution context for this Maven embedder.
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
@@ -22,7 +22,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Deque;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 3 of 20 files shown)


laeubi:
This is another step in extension support and for example allows to import a pomless-bundle as a maven project into m2eclipse:

grafik

HannesWell (commented):
Thank you for your work on this one.
I only have a few minor remarks.

org.eclipse.m2e.core/.classpath
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-17">
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-16">
```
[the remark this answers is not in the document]
laeubi: No it seems eclipse (or m2e?) has changed this automatically

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
```diff
@@ -56,7 +56,11 @@
  * @author igor
  * @noimplement This interface is not intended to be implemented by clients.
  */
-public interface IMaven {
+public interface IMaven extends IComponentLookup {
+
+  static final String MVN_FOLDER = ".mvn";
```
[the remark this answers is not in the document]
laeubi: I don't see any issue here, at least maven don't have public ones for them, so we need to replicate them here, I don't want any other consumer forcing to replicate it as well :-)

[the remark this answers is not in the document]
mickaelistria: > I don't want any other consumer forcing to replicate it as well

Except that here, we should also avoid other consumers to use m2e value as a reference while it is actually not. It gives a false impression this is a reliable symbol.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
```diff
@@ -135,21 +139,23 @@ protected MavenExecutionRequest newExecutionRequest() throws CoreException {
...
+      request = createExecutionRequest(maven.getMavenConfiguration(), maven, maven);
+      request.setBaseDirectory(basedir);
     }
     request.setMultiModuleProjectDirectory(MavenImpl.computeMultiModuleProjectDirectory(basedir));
 
     return request;
   }
 
-  private MavenExecutionRequest createExecutionRequest() throws CoreException {
+  static MavenExecutionRequest createExecutionRequest(IMavenConfiguration mavenConfiguration,
```
mickaelistria, line 150: Can the same context be used wit different values for mavenConfiguraiton/lookup/maven? I was more expecting this context to be a stateful object more than a service.

mickaelistria (commented):
I really like the overall direction. Some questions/comments inline.
Can you please already submit a separate PR to rename `lookupExtensions` to `lookupCollections`; or extracting the PlexysContainerManager? This one could be merged promptly. As usual, all the internal stuff is pretty consensual; and more attention/discussion will be placed on APIs, so if some internal-only changes can be isolated, we will already have a good piece of the value and be more capable of reviewing APIs with more focus.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
laeubi: The code for creating a `MavenExecutionRequest` is also used inside the PlexusContainermanager that's the reason why I made it static here, actually we might later even move the code elsewhere as creating the request is not really bound to the execution at all.

mickaelistria: Is this the same code? ie should it really be factorized here? The point of IMavenExecutionContext is to be tied to the state more than being an utility class. Maybe it's one of the case where having 2 distinct methods, even with a bit a duplication, is actually better than trying to reach highest factorization possible.

laeubi: It requires the same setup, I don't think copy the code helps much here and we can do it later if we want as it is all package-private, so I think we should not complicate this too much.

mickaelistria: > it is all package-private

OK.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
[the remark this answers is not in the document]
laeubi: I have moved this to a "private" place now...

laeubi:
@mickaelistria I'm not sure if separating this more really helps much as the changes are closely related ...

laeubi:
@HannesWell I also applied your suggestions manually because I force-overwritten them :-\

If there are no concerns I'd like to merge this soon to be safe for the archetype refactorings....

org.eclipse.m2e.core/.classpath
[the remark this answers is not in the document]
HannesWell: > No it seems eclipse (or m2e?) has changed this automatically

Looks like it happened again.
Would be interesting to know, why this is happening.

mickaelistria:
> If there are no concerns I'd like to merge this soon to be safe for the archetype refactorings....

I have some concern about the IComponentLookup interface. I don't think it's something that the execution context should implement. Instead, having it on IMaven (as already suggested) seems to make a lot of sense. So maybe the IMavenExecutionContext should provide an accessor to get the underlying IMaven it's built from?

HannesWell (commented):
I not familiar enough with that part of the code, to give a qualified technical feedback, but in general it looks good to me.
So if Mickael is fine with it, I'm as well.

> @HannesWell I also applied your suggestions manually because I force-overwritten them :-\

No problem. for such small changes there is no need to track that in git from my perspective.

laeubi:
> > If there are no concerns I'd like to merge this soon to be safe for the archetype refactorings....
> 
> I have some concern about the IComponentLookup interface. I don't think it's something that the execution context should implement. 

Why not? It is just a generalization of the previous `lookupExtensions` (that was just an incomplete intermediate) and as described in the PR the execution is just one of three levels we need here to cover different cases.

> Instead, having it on IMaven (as already suggested) seems to make a lot of sense.

Please forget about that `IMaven` stuff that really only complicates things ;-)

>  So maybe the IMavenExecutionContext should provide an accessor to get the underlying IMaven it's built from?

`IMavenExecutionContext` is **not** build from an "underlying IMaven", this `IMaven` connection is really just for historic reasons and we are on a good way to break that wires.

laeubi:
> Looks like it happened again.

Fixed, this was simply not applied by me...

laeubi:
> I don't think it's something that the execution context should implement

I refactored this now to being a getter instead of extending the interface

mickaelistria:
> I refactored this now to being a getter instead of extending the interface

Thanks, that's much better.
And now there seems to be only 1 implementation for IComponentLookup (IMaven). Is there any obvious case that would highlight we'd need another implementation? If not, maybe we can drop the interface and get the IMavenExecutionContext return the IMaven directly?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IComponentLookup.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.Collection;
+
+import org.eclipse.core.runtime.CoreException;
+
+
+/**
+ * A {@link IComponentLookup} looks up a component in a given context
+ * 
+ * @noimplement This interface is not intended to be implemented by clients.
```
mickaelistria, line 21: needs `@since 2.0`

mickaelistria (changes requested):
One minor request and we're good to go.

laeubi:
> And now there seems to be only 1 implementation for IComponentLookup

Actually there are four of them:

1. `IMaven`
2. `IMavenProjectFacade`
3. (private) PlexusComponentLookup in `PlexusContainerManager`
4. (private) anonymous in `MavenExecutionContext`

> if the ultimate goal is to progressively kill IMaven.

We may not kill it completely, but greatly reduce the need for it.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IComponentLookup.java
laeubi: Done

mickaelistria:
I missed that IMavenProjectFacade implements the interface. Can it be replaced by a getter as well. As design patterns recommend, and we're fully into OO design here so this applies well, we should always prefer composition over inheritance.

laeubi:
> missed that IMavenProjectFacade implements the interface

I shouldn't have noted that :-P

mickaelistria:
> I shouldn't have mention that :-P

:) I would probably have noticed and changed it later though, but it's better if it can all be done now with a nice discussion and a seft-contained commit.

laeubi:
> Can it be replaced by a getter as well.

done

mickaelistria:
Thanks, that a very good step forward!

PR #758, opened 2022-06-07
Pull out the Archetype from the API interface by introduce a thin indirection layer

laeubi:


laeubi:
With this, there are only internal references to the Archetype...

laeubi:
This pulls away as much as possible of the archetype stuff with current state from the m2e-core, if tests are green this should be merged as an intermediate step for further refactoring.

HannesWell:
That's a reasonable change. I thought as well that we have to do something like this in order to have the flexibility to change archetype-story under the hood, without affecting the API. 👍🏽

PR #760, opened 2022-06-07
Convert plain data carriers classes to records

HannesWell:

Besides the actual conversion there is also potential to remove unused methods (at least in the m2e-code base) respectively to make some methods internal. Should that be done separately before/after this one?

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/configurator/MojoExecutionKey.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/configurator/MojoExecutionKey.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010 Sonatype, Inc.
+ * Copyright (c) 2010, 2022 Sonatype, Inc.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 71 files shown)


org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/configurator/MojoExecutionKey.java
```diff
@@ -14,108 +14,63 @@
...
-  }
-
-  public boolean match(MojoExecution mojoExecution) {
-    if(mojoExecution == null) {
-      return false;
-    }
-    return groupId.equals(mojoExecution.getGroupId()) && artifactId.equals(mojoExecution.getArtifactId())
-        && version.equals(mojoExecution.getVersion()) && goal.equals(mojoExecution.getGoal())
-        && executionId.equals(mojoExecution.getExecutionId());
+    return lifecyclePhase();
```
mickaelistria, line 74: what about just removing those?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/configurator/ProjectConfigurationRequest.java
```diff
@@ -39,16 +31,18 @@ public IProject getProject() {
   public ResolverConfiguration getResolverConfiguration() {
     return facade.getResolverConfiguration();
   }
+  //TODO: inline getters
 
   public MavenProject getMavenProject() {
-    return mavenProject;
+    return mavenProject();
   }
```
mickaelistria, line 38: remove

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/configurator/ProjectConfigurationRequest.java
```diff
@@ -22,15 +22,7 @@
...
-public class ProjectConfigurationRequest {
-  private final IMavenProjectFacade facade;
-
-  private final MavenProject mavenProject;
-
-  public ProjectConfigurationRequest(IMavenProjectFacade facade, MavenProject mavenProject) {
-    this.facade = facade;
-    this.mavenProject = mavenProject;
-  }
+public record ProjectConfigurationRequest(IMavenProjectFacade facade, MavenProject mavenProject) {
```
mickaelistria, line 25: facade -> mavenProjectFacade.

mickaelistria (commented):
I'm all for it; and as it's a breaking change, let's remove useless legacy methods by the way.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/configurator/ProjectConfigurationRequest.java
HannesWell: Done.

mickaelistria:
Is it rebased on top of master?

HannesWell:
> Is it rebased on top of master?

Yes.

HannesWell:
> It looks like PluginExecutionFilter is missing update to lately removed methods.

I updated this on my computer and was really confused that the error still showed up in the CI-build.
But now I noticed that those classes are generated. I updated the modello `lifecycle-mapping-metadata-model.xml` to use the new methods. Now it should work.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/configurator/MojoExecutionKey.java
```diff
@@ -9,113 +9,25 @@
...
     return groupId + ":" + artifactId + ":" + version + ":" + goal + ":" + executionId + ":" + lifecyclePhase;
   }
-
-  @Override
-  public String toString() {
-    return groupId + ":" + artifactId + ":" + version + ":" + goal + " (execution: " + executionId + ", phase: "
-        + lifecyclePhase + ")";
-  }
-
-  public boolean match(MojoExecution mojoExecution) {
```
laeubi, line 113: Please note that `match` has slightly different semantics than `equals` so unless the record is not `final` we should not just remove this method!

HannesWell: > Please note that `match` has slightly different semantics than `equals` so unless the record is not `final` we should not just remove this method! (or is this method not used anywhere in the code?)

Records are always final AFAIK. The only difference to equals is that `match()` ignores the `lifecyclePhase`.

It is only used in `MavenProjectFacade` and I therefore moved it there.

It's a bit hard to tell, if such helper methods should remain in the API or not. It depends on if there are real use-cases for downstream consumers, but I don't know that.

mickaelistria:
Looks like this does cause some regressions captured by tests.

HannesWell:
> Looks like this does cause some regressions captured by tests.

Yes, I will investigate it. My first guess is that this is because some of the records did not implement hashCode()/equals() before.

HannesWell:
Locally all tests were green. The reason for the failure was that some tests made assumptions about the specific value returned by `toString()` of some of the classes that are now records.
The reason for the build failure is that GH and Jenkins merge the branch with the current master and the adjustments for the Mavenarchiver are missing since I want to await the fixed merge.

Regarding a specific `toString()` implementation I'm not sure if we should have that. Often one is told to not rely on a specific toString(), but on the other hand consumer of M2E might do that nevertheless nobody forbids that and it is covered by tests. Additionally the specific implementation is IMHO more expressive and I don't want to find all places where it is used to replace it with a method that maybe makes a stronger promise about the returned value.

HannesWell:
The tests are green now (at least at the last pass) so we can talk about what we want to keep in the records and then submit this PR.

HannesWell:
@laeubi, @mickaelistria unless you have any more remarks/suggestions I think this is ready. I will interpret no reaction as consent. :)

.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "m2e-core-tests"]
 	path = m2e-core-tests
-	url = https://github.com/tesla/m2e-core-tests.git
-	branch = master
+	url = https://github.com/HannesWell/m2e-core-tests.git
```
laeubi, line 3: don't forget to revert this :-)

laeubi (approved):

.gitmodules
HannesWell: Thanks for the reminder. Not sure that I would not have missed it. :D
Done.

PR #765, opened 2022-06-09
Exceptions when failing to wrap Maven artifacts should be logged.

pzi-dsa:
Errors are only shown in the target editor. However, this is a volatile
medium and messages may disappear, when reloading the target platform or
after closing and re-opening the editor.

Writing the error also to the Error Log gives access to the stack trace,
making it easier to find the cause of the problem, in case something
goes wrong.

As suggested in #756 , any exceptions that are thrown when trying to
resolve Maven dependencies are also written to the Error Log.

### DIFF
--- a/org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/MavenTargetBundle.java
+++ b/org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/MavenTargetBundle.java
@@ -22,7 +22,9 @@ import java.util.jar.Manifest;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
 import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.core.runtime.ILog;
 import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.equinox.frameworkadmin.BundleInfo;
 import org.eclipse.pde.core.target.TargetBundle;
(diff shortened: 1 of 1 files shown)


org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/MavenTargetBundle.java
```diff
@@ -26,12 +26,15 @@
...
 import org.eclipse.pde.core.target.TargetBundle;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import aQute.bnd.osgi.Analyzer;
 import aQute.bnd.osgi.Jar;
 
 public class MavenTargetBundle extends TargetBundle {
 
+	private static final Logger LOGGER = LoggerFactory.getLogger(MavenTargetBundle.class);
```
laeubi, line 37: Would you mind using the Eclipselog (`ILog`) instead here, that would make it more accessible I think and users are more aware of that facility.

laeubi (approved):
This looks good and will be useful if we should see any issues.

laeubi:
@pzi-dsa if we merge this, do you htink you can capture a fresh stack-trace for the other issue or is it currently not reproducible?

pzi-dsa:
I just did some minor cleanup and removed unnecessary code modifications.

> do you htink you can capture a fresh stack-trace for the other issue or is it currently not reproducible?

That's the tricky part. I rarely run into this problem on my machine, but I hear about it quite often from my colleages.
Personally, it's not an urgent problem. So we can put the other pull request on hold and I can see whether I can reproduce it with the latest snapshot again.
Alternatively, I'll ask one of my colleagues to update their version and send me the error log for when it happens to them.

laeubi:
That would be great, now with the enhanced logging we might see even more errors logged so we can get a more complete view what is going on, so I'll merge this asap when the build passes (even though I don't expect any failures)

HannesWell:
In case you are interested, I proposed to add another method to Platform to obtain an `ILog` that does not require any argument:
https://github.com/eclipse-platform/eclipse.platform.runtime/pull/58

Feel free to add your remarks.

Issue #249, opened 2021-06-24
514993 - Maven archetype post generated script is not executed while creating from maven module wizard 


hd42:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=514993
Maven archetypes can include Groovy scripts which will be executed after the Maven project is created from that archetype. It works fine from the command prompt. But the script is not executed when the project is created using the same archetype via the Eclipse >> New >> Maven project wizard.
Apparently, license issues prevent(ed?) m2e's move from including the maven-artefact-plugin 2.* to 3.*.

This issue still persists in Eclipse 21.6.

mbooth101:
This is expected, m2e includes a very old version of maven archetype (2.4) and only maven-archetype 3.x supports the post-step groovy scripts.

I don't think the groovy license is a problem at all, there is already *lots* of ASL licensed software in m2e, Eclipse and Orbit.

More likely the blocker is the non-trivial porting effort to move to archetype 3.x.

As a starting point, you can look at the patch I carried in Fedora:

https://src.fedoraproject.org/rpms/eclipse-m2e-core/blob/f34/f/0005-Port-to-latest-version-of-maven-archetypes.patch

mickaelistria:
@mbooth101 you're right; going to archiver 2.4 to 3.x isn't trivial enough to be a priority at the moment. Is the Fedora patch functional? Would you be willing to submit it as a pull request?

mbooth101:
Err, sure. I'll rebase it and open a PR then we can take a closer look

HannesWell:
> > Err, sure. I'll rebase it and open a PR then we can take a closer look
> 
> Any updates here? I'd like to push this forward for the 2.x line as this is a good opportunity to change versions.

@laeubi if you plan to update the code I can offer to contribute the update of `org.eclipse.m2e.archetype.common/pom.xml`. It is a bit tedious work to find all the right exclusion but I have done locally already.

laeubi:
@HannesWell sure, every piece would help, I think the discussions about size, interactive and system exit are valid but don't have to be the primary driver here given that

1. It does not work at all (even not for 'gracefull' scripts
2. We are shipping ancient versions
3. It seems not profitable to first refactor m2e using ancient versions and then refactor it to use newer versions
4. 'size' is not a development goal of m2e and if we find opportunities to reduce the size can do so any time later

[...]

PR #767, opened 2022-06-10
Let maven do the hard work of archetype generation

laeubi:
Prerequisite for https://github.com/eclipse-m2e/m2e-core/issues/249

### DIFF
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeGenerator.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeGenerator.java
@@ -10,57 +10,46 @@
  * Contributors:
  *      Sonatype, Inc. - initial API and implementation
  *******************************************************************************/
+
 package org.eclipse.m2e.core.ui.internal.archetype;
 
-import java.util.ArrayList;
-import java.util.Collections;
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.util.Collection;
 import java.util.List;
 import java.util.Properties;
-import java.util.Set;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
-import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IWorkspaceRoot;
 import org.eclipse.core.resources.ResourcesPlugin;
-import org.eclipse.core.runtime.Adapters;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.osgi.util.NLS;
+import org.eclipse.swt.widgets.Display;
 
-import org.codehaus.plexus.util.StringUtils;
-
-import org.apache.maven.archetype.ArchetypeGenerationRequest;
-import org.apache.maven.archetype.ArchetypeGenerationResult;
-import org.apache.maven.artifact.Artifact;
-import org.apache.maven.artifact.repository.ArtifactRepository;
-
-import org.eclipse.m2e.core.embedder.IMaven;
-import org.eclipse.m2e.core.embedder.IMavenExecutionContext;
 import org.eclipse.m2e.core.embedder.MavenModelManager;
+import org.eclipse.m2e.core.internal.IMavenConstants;
 import org.eclipse.m2e.core.internal.Messages;
-import org.eclipse.m2e.core.internal.embedder.MavenImpl;
+import org.eclipse.m2e.core.internal.launch.IMavenLauncher;
 import org.eclipse.m2e.core.project.IArchetype;
-import org.eclipse.m2e.core.project.IMavenProjectImportResult;
 import org.eclipse.m2e.core.project.IProjectConfigurationManager;
-import org.eclipse.m2e.core.project.IProjectCreationListener;
 import org.eclipse.m2e.core.project.LocalProjectScanner;
 import org.eclipse.m2e.core.project.MavenProjectInfo;
-import org.eclipse.m2e.core.project.ProjectImportConfiguration;
 
 
 @Component(service = {ArchetypeGenerator.class})
+@SuppressWarnings("restriction")
 public class ArchetypeGenerator {
-
-  private static final Logger log = LoggerFactory.getLogger(ArchetypeGenerator.class);
-
-  @Reference
-  IMaven maven;
 
   @Reference
   IProjectConfigurationManager projectConfigurationManager;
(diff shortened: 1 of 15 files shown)


org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/IMavenLauncher.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.eclipse.core.runtime.CoreException;
+
+
+/**
+ * {@link IMavenLauncher} allows to trigger a maven run
+ */
+public interface IMavenLauncher {
+
+  CompletableFuture<?> runMaven(File basedir, String goals, Properties properties) throws CoreException;
```
mickaelistria, line 25: Couldn't this be added to `AbstractMavenRuntime` ?

laeubi: That's absolutely possible, sad enough I haven't noticed that class before :-\

HannesWell (commented):
This change looks very good.
Addressing the potential crashing of Eclipse by post scripts is IMHO important and this looks like a good step forward to maybe even get completely rid of the embedded archetype dependency?




I basically only have one remark: Could you please provide a more descriptive commit message, like: `Generate archtetype projects in separate Maven-build process`

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/IMavenLauncher.java
laeubi: @mickaelistria the runtime can be used to create launches but not to actually perform a launch or do I oversee something?

HannesWell: > @mickaelistria the runtime can be used to create launches but not to actually perform a launch or do I oversee something?

The actual launching, i.e. assembling all command line arguments and finally starting the Java-process, is done my JDT and I assume replicating that is in the end just more complicated than just creating a hidden launch as you did.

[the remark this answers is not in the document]
laeubi: By the way, that we get/join here is just bad but due to the current structure there is no better way I think atm and I don't want to change to much. Actually one would need to integrate the Job API with `CompletableFuture`... but that would likely produce more discussions and efforts than I currently can afford ;-)

laeubi: That's why I think I cant just add to AbstractMavenRuntime here... but it would make sense to move the interface there...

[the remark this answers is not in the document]
HannesWell: > `ILaunch` is a `jdt.debug` dependency
> 
Indeed. So just disregard that suggestion.



> By the way, that we get/join here is just bad but due to the current structure there is no better way I think atm and I don't want to change to much. Actually one would need to integrate the Job API with `CompletableFuture`... but that would likely produce more discussions and efforts than I currently can afford ;-)

I'm not sure what you mean exactly, but I share your assumption.^^

mickaelistria: > he runtime can be used to create launches but not to actually perform a launch or do I oversee something?

I'm not sure I get the difference. Can you please elaborate?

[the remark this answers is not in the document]
mickaelistria: ILaunch is org.eclipse.debug.core, totally fine to use it.
ILaunch can already be used to track completion, so it seems redundant with completablefuture.

laeubi:
I'm struggling a bit to get the test repo filled completely (I probably will enable online maven central as it otherwise becomes really cumbersome), but the code itself is now functional and thus is ready to be reviewed.

laeubi:
All tests are green now!

org.eclipse.m2e.launching/src/org/eclipse/m2e/internal/launch/EclipseMavenLauncher.java
```diff
@@ -0,0 +1,118 @@
...
+    ILaunchConfigurationWorkingCopy workingCopy = launchConfigurationType.newInstance(null,
+        launchManager.generateLaunchConfigurationName("mavenrun-" + UUID.randomUUID()));
+    workingCopy.setAttribute(MavenLaunchConstants.ATTR_POM_DIR, basedir.getAbsolutePath());
+    workingCopy.setAttribute(MavenLaunchConstants.ATTR_GOALS, goals);
+    workingCopy.setAttribute(MavenLaunchConstants.ATTR_SAVE_BEFORE_LAUNCH, false);
+    List<String> properties = new ArrayList<>();
+    for(String propertyKey : mavenProperties.stringPropertyNames()) {
+      String propertyValue = mavenProperties.getProperty(propertyKey);
+      if(propertyKey != null && propertyKey.trim().length() > 0) {
+        String prop = propertyKey.trim() + "=" + (propertyValue == null ? "" : propertyValue);
```
HannesWell, line 60:
```suggestion
        String prop = propertyKey.trim() + (propertyValue == null ? "" : "=" + propertyValue);
```
Besides that I wonder if there is a realistic chance that the `propertyKey` is null?

Furthermore what do you think about using a `Map<String, String>` instead of `Properties`?

org.eclipse.m2e.launching/src/org/eclipse/m2e/internal/launch/MavenLaunchDelegate.java
```diff
@@ -106,6 +106,14 @@ public void launch(ILaunchConfiguration configuration, String mode, ILaunch laun
     }
   }
 
+  protected boolean saveBeforeLaunch(ILaunchConfiguration configuration, String mode, IProgressMonitor monitor)
```
HannesWell, line 109: Missing `@Override`.

org.eclipse.m2e.tests.common/src/org/eclipse/m2e/tests/common/AbstractMavenProjectTestCase.java
```diff
@@ -90,24 +104,100 @@
...
   protected IWorkspace workspace;
 
   @Rule
   public TestName name = new TestName();
 
+  @Rule(order = Integer.MIN_VALUE)
+  public TestWatcher watchman = new TestWatcher() {
+    @Override
+    protected void failed(Throwable e, Description description) {
+      if(STOP_ON_FAILED_TEST) {
```
HannesWell, line 171: Is this opening a shell to abort the test?
If yes, should this really be submitted to the master? If the test case tends to dead-lock/run forever I suggest to assign a timeout to it.

HannesWell (commented):
In general this looks good to me.

I just have a few minor remarks.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeGenerator.java
```diff
@@ -10,178 +10,151 @@
...
-          .generateProjectFromArchetype(request);
-
-      Exception cause = result.getCause();
-      if(cause != null) {
-        String msg = NLS.bind(Messages.ProjectConfigurationManager_error_unable_archetype, archetype);
-        log.error(msg, cause);
-        throw new CoreException(Status.error(msg, cause));
+      String goals = "-U org.apache.maven.plugins:maven-archetype-plugin:2.4:generate";
+      if(emptyPom != null) {
+        goals += " -f " + emptyPom.getAbsolutePath();
```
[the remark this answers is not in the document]
laeubi: What do you mean by 'new'? This is to prevent m2e to first generate the code at a temporary place and then move it around, without the empty pom you will get higher build times and can't create a module of a parent that has pom-errors and at least there is a test for that (don't know if it uses intentionally an incomplete module pom).

org.eclipse.m2e.launching/src/org/eclipse/m2e/internal/launch/EclipseMavenLauncher.java
laeubi: `Properties` are actually `Map<String, String>` after generic erasure :-)
As maven itself uses `Properties` (and they are also named like that) it seems a good choice here even though its not that much relevant.

org.eclipse.m2e.tests.common/src/org/eclipse/m2e/tests/common/AbstractMavenProjectTestCase.java
laeubi: This open a small Shell with a button to wait and keep the test-eclipse open so you are able to inspect the state (e.g. project, console outputs etc.) as the tests run in the main thread, using a breakpoint don't work because the test-eclipse is frozen as long as the main thread is suspended.

This has become quite handy for me to debug the tests, as it is sometimes quite hard to guess just from the object state whats going on.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeGenerator.java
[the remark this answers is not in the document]
laeubi: By the way there are currently two restrictions one might report to maven and submit a patch, but this would then take some time until it is released:

1. If running `mvn` inside an empty folder / without a pom, it uses a "no-pom build", one can then force a different location of the pom by -f but if the folder contains no pom I can't force to ignore that file.
2. maven-archetype always generates the archetype at the folder `$outputDir/$artifactId` while here it would be good if we can just generate it in the current folder (what would be empty anyways).

org.eclipse.m2e.launching/src/org/eclipse/m2e/internal/launch/MavenLaunchDelegate.java
laeubi: Reminds me about inconsistent configurations across projects ...

laeubi:
@HannesWell I have applied your suggestion and fixed the compile issues, if you are fine I'll swash all content into one before merge...

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeGenerator.java
```diff
@@ -145,11 +143,9 @@ private File getEmptyPom(File basedir) {
...
         File tempFile = File.createTempFile("pom", ".xml", basedir);
         tempFile.deleteOnExit();
-        try (OutputStreamWriter writer = new OutputStreamWriter(new FileOutputStream(tempFile),
-            StandardCharsets.UTF_8)) {
-          writer.write(
-              "<project><modelVersion>4.0.0</modelVersion><groupId>empty</groupId><artifactId>empty</artifactId><version>1</version><name>Generating archetype</name></project>");
-        }
+        Files.writeString(tempFile.toPath(),
+            "<project><modelVersion>4.0.0</modelVersion><groupId>empty</groupId><artifactId>empty</artifactId><version>1</version><name>Generating archetype</name></project>",
+            StandardCharsets.UTF_8);
```
HannesWell, line 148: I just noticed that Files.writeString() uses UTF-8 by default so specifying the charset explicitly is actually not necessary.

org.eclipse.m2e.tests.common/src/org/eclipse/m2e/tests/common/AbstractMavenProjectTestCase.java
HannesWell: That's indeed a problem I just encountered yesterday.
What I have done in such case, is that I have restarted the test after I have changed the configuration to not run in the UI-thread. That may work for some tests but probably not for all.

So yes that can be handy. 👍🏽

HannesWell (approved):
Looks good to me. Thank you.

> @HannesWell I have applied your suggestion and fixed the compile issues, if you are fine I'll swash all content into one before merge...

Sure squash it, I see no need to keep the review history in git. :)

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/IMavenLauncher.java
HannesWell: > > I'm not sure I get the difference. Can you please elaborate?
> 
> I actually want to execute a maven run (with console) like when you do `Run As > Maven Build` and while the RunPomAction _uses_ the `AbstractMavenRuntime` to build all the JVM parameters, the actual execution happens as a jvm process and not through `AbstractMavenRuntime`

This also adds the 'feature' that the maven installation selected as default in the workspace preferences is used to create the archetype project. So I consider this a plus besides the fact that we want to run it in an extra process.

laeubi:
Okay, lets merge when build runs fine then (after reverting the modules link).

mickaelistria:
I still don't like much IMavenLauncher (but at least it's internal, so not a too big deal)

> I actually want to execute a maven run (with console) like when you do Run As > Maven Build and while the RunPomAction uses the AbstractMavenRuntime to build all the JVM parameters, the actual execution happens as a jvm process and not through AbstractMavenRuntime

Couldn't this be a new method to AbstractMavenRuntime then, something like `IProcess runExtenally(...)` ?

laeubi:
> Couldn't this be a new method to AbstractMavenRuntime then, something like `IProcess runExtenally(...)` ?

If you like, just try it out I won't mind adjusting that part of the code as it is covered by tests and not used elsewhere until now we can change relative easy here.

PR #771, opened 2022-06-13
[Maven-Target] Consider extra-repositories when fetching source-jars

HannesWell:
This PR fixes the fetch of source bundles when additional repositories are specified in a Maven-target.

In https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/282#issuecomment-1152122846 it was discovered that additional repositories are not considered when source-artifacts are downloaded.


laeubi (approved):
Looks good, thanks for taking a look into it!

HannesWell:
> Looks good, thanks for taking a look into it!

Great. Thanks for reviewing.

PR #772, opened 2022-06-14
[PDE] update biz.aQute.bndlib and -util to version 6.3.1

HannesWell:
Looks like I oversaw the recent service-release in #445.
But if it's available lets use it.

### DIFF
--- a/org.eclipse.m2e.pde.feature/forceQualifierUpdate.txt
+++ b/org.eclipse.m2e.pde.feature/forceQualifierUpdate.txt
@@ -1,0 +1,2 @@
+# To force a version qualifier update add the bug here
+[PDE] update biz.aQute.bndlib and -util to version 6.3.1 - https://github.com/eclipse-m2e/m2e-core/pull/772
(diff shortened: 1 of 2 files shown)


laeubi:
> Looks like I oversaw the recent service-release

Sadly 
- https://github.com/dependabot/dependabot-core/issues/4682

did not get much interest from the eclipse-community or otherwise dependabot might has mentioned it already :-)

HannesWell:
> Sadly
> 
>     * [Add support for maven-dependecies inside eclipse-pde target files dependabot/dependabot-core#4682](https://github.com/dependabot/dependabot-core/issues/4682)
> 
> 
> did not get much interest from the eclipse-community or otherwise dependabot might has mentioned it already :-)

Yes that would be really awesome. I saw your proposal already but since I'm neither familiar how dependabot works internally nor  with the languages it is written in I cannot contribute to that.

It would be even more awsome if dependabot could understand p2 InstallableUnit targets but this would likely require more work and has the problem that p2 repos don't have a unified schema to store different versions like Maven has.

HannesWell:
As suspected the test-failure is unrelated to this change because it occurs even if this PR effectively contains no change.
I wonder why the master-build succeeds.

PR #774, opened 2022-06-14
Mavenarchiver

mickaelistria:


laeubi:
@mickaelistria do you think we can filter out the "remoterepo" stuff and just fetch everything from central for now?

Beside the size, I'm not sure if we want to include that much stuff here where we maybe need IP-clearance?

mickaelistria:
The IP review is ongoing, so this doesn't matter much from IP perspective now.

laeubi:
Sure but if we prepare it as if we would submit I think its good to show the actual state, e.g. we can the update the CQ with a link to this PR.

mickaelistria:
> we can the update the CQ with a link to this PR.

I don't think that would change anything.

laeubi:
Still I thought it would be good to leave some comments just in case you are eager to merge this once CQ is approved :-)

laeubi:
Just another thing that came into my mind: If converting the tests is to cumbersome, we might want to add the test to the tesla/me2-core-tests repo instead (what already contain a test-repository). Beside that I'm a bit confused why it takes so long to actually approve this?

mickaelistria:
> If converting the tests is to cumbersome

I don't think it is.

> Beside that I'm a bit confused why it takes so long to actually approve this?

I don't know. Maybe @waynebeaton has some answer?

laeubi (changes requested):
I think we should remove the whole "remoterepo" stuff here, or move the test that strictly require such things to maven-core-tests repo.

laeubi (changes requested):
As far as I know we should squash all commits if importing code into a project

(this is also a test if one can request two changes at the same time in GH...)

mickaelistria:
@waynebeaton do you confirm that history needs to be dropped when merging external code?

waynebeaton:
> @waynebeaton do you confirm that history needs to be dropped when merging external code?

There is no such requirement. We removed this limitation in the 2019 update to the IP Policy (that is, it was a requirement at one point in the past).

laeubi:
Thanks for the update! Still we should then filter out the binary artifacts.

mickaelistria:
Keeping history is valuable.

mickaelistria:
It's possible to remove content from GitHub repo after the fact: https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/removing-sensitive-data-from-a-repository

laeubi:
This is for _accidental_ commit sensitive data, not for rushing content in and later try to fix that, destroying the history of commits just for some "urgent" demands and should have been done before the merge here...

HannesWell:
I share Christoph's opinion that we should not keep the binary data here.

Although it is actually bad practice it would be possible for a commiter to force push the master branch to the previous commit before this merge to make it undone.
Given the situation and that it not happened too long ago. I think that would be OK.

laeubi:
I have created a cleaned copy here: https://github.com/laeubi/tesla-m2eclipse-mavenarchiver/tree/master-cleanup

Please use that as a base for merging, it is 306,9 kB versus **17 mB (!)**:

grafik

Beside this, I really would appreciate to squash the commits, as the "valuable history" not really seems valuable here as the repo was split before (without history!) and then mainly some maintenance changes, I don't think we loose much here.

@fbricon as you are the project lead of m2e it would be good to give a comment here, currently I'm holding back changes to not make it more worse and we need a decision here.

For the record, here are the commands I used to clean the original repository:

1. create empty folder
2. `git clone --mirror git@github.com:laeubi/tesla-m2eclipse-mavenarchiver.git`
3. `git filter-branch --index-filter 'git rm -rf --cached org.eclipse.m2e.mavenarchiver.tests/remoterepo/*.*'`
4. push everything to new remote branch https://github.com/laeubi/tesla-m2eclipse-mavenarchiver/tree/maste-cleanup

HannesWell:
Thanks Christoph for your effort.

> @fbricon as you are the project lead of m2e it would be good to give a comment here, currently I'm holding back changes to not make it more worse and we need a decision here.

I support completely undoing this commit and force-pushing the master to the commit before this merge.

> Beside this, I really would appreciate to squash the commits, as the "valuable history" not really seems valuable here as the repo was split before (without history!) and then mainly some maintenance changes, I don't think we loose much here.

Instead of including the entire history a comment could be added to the commit's message containing the original repo URL and the commitId of the original commit in that repo from which the squashed content was created.

HannesWell:
And the first URL in https://github.com/eclipse-m2e/m2e-core/pull/774#issuecomment-1163909864 has one r too much. The branch is named [maste-cleanup](https://github.com/laeubi/tesla-m2eclipse-mavenarchiver/tree/maste-cleanup).

laeubi:
I renamed the branch but thing it dosen't matter much anyways ;-)

HannesWell:
> I renamed the branch but thing it dosen't matter much anyways ;-)

Definitely it doesn't matter, I just mentioned that because the first link pointed into nowhere. :)

fbricon:
The history we have for mavenarchiver is not that interesting, so I think we can live without.

However, the embedded remote repo actually has a value. Generally speaking (for m2e-*) it allows testing cases where you need to download dependencies, or need to check cases where you know the data is  present in your local repository. This ensure self-contained, idempotent repository state across builds. In that case, I really don't care that storing binaries in git is bad practice. The real issue with embedding jars is purely IP-related, you'd need to open CQs for pretty much each jar in the cached remote repo. That's why all the m2e, m2e-wtp tests are in separate repos, to not have to deal we Eclipse IP management.

Now, if you believe removing the jars helps with m2e maintenance, I'm not gonna oppose it, given my current level of involvement in the project.

laeubi:
> However, the embedded remote repo actually has a value. Generally speaking (for m2e-*) it allows testing cases where you need to download dependencies, or need to check cases where you know the data is present in your local repository. This ensure self-contained, idempotent repository state across builds.

I think we should better use other techniques than putting arbitrary content into the repository or do it in the m2e-tesla-repo then as already suggested.

> The real issue with embedding jars is purely IP-related, you'd need to open CQs for pretty much each jar in the cached remote repo.

I think this a much bigger issue, also I found *updating* such repository is really a pain... and given that if central is down we can't really build (as repos are cleared on each build) I also think we do not gain much here.

> Now, if you believe removing the jars helps with m2e maintenance

See above, there is literally no way to update/cleanup/enhance things without major efforts and it bloats the repository.

> given my current level of involvement in the project.

Still you are project lead ;-)

So how should we proceed here? Given that @HannesWell also seem to prefer not having the binary stuff here reset master to precious state and merge without them? Anyone @eclipse-m2e/technology-m2e-committers here want to join the discussion?

fbricon:
let's remove the jars then

HannesWell:
> So how should we proceed here? Given that @HannesWell also seem to prefer not having the binary stuff here reset master to precious state and merge without them? 

I suggest somebody, probably either you or Mickael, prepares an adjusted variant of this PR (with squashed or filtered history) to add Mavenarchiver based on the commit before this merge (i.e. 00d04b309c3cc600140c23e58372e197d7cd01a5) and then force pushes the master to that previous commit (i.e. again 00d04b309c3cc600140c23e58372e197d7cd01a5).
The PR should then be ready to merge or does your work require more adjustments?

> Anyone @eclipse-m2e/technology-m2e-committers here want to join the discussion?

Besides @mickaelistria everybody that I have seen here since I participate in m2e has expressed his opinions and there was only approval. I think it would be good if he shares his opinion too (I actually don't want to guess that).
Nevertheless, since I had the impression that Mickael's primary goal was to have the Mavenarchiver fast, I suggest to not wait to long for this because the earlier we start, the faster this is fixed and the faster Mavenarchiver can be used with joy.

laeubi:
I'll start this right now.

PR #776, opened 2022-06-15
Fix the GH action build and upgrade to Java 17

laeubi:
For a while we disabled the GH action build because of "strange errors", in the mean while we found out that it was because of the checkout depth and thus can enable this again now.

### DIFF
--- a/.github/workflows/junit.yml
+++ b/.github/workflows/junit.yml
@@ -1,0 +1,37 @@
+name: Publish Unit Test Results
+
+on:
+  workflow_run:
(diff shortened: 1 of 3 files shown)


HannesWell:
Not to forget that we also get a test-result overview at each PR, which is very nice in the Web-Interface.
@laeubi do you know if I can specifically turn of notifications via Email from a specific or all bots?

laeubi:
Not that I'm aware of. The best seems to install a mail-filter rule.

PR #779, opened 2022-06-15
Align version of m2e.maven.runtime with embedded maven version

HannesWell:
Probably we want a higher lower bound that Maven 3.0.0 in all plug-ins requiring m2e.maven.runtime respectively archetype.

Any suggestions? Should we just go with 3.8.0? or even 3.8.6?


HannesWell:
Given that no older 3.x or even 2.x version of m2e.maven.runtime exists a lower bound than 3.8.6 has no advantage so I think we should just go as high as possible here, meaning 3.8.6.
The same applies for the required m2e.archetype.common.

PR #783, opened 2022-06-15
Consider target-files in deps-caching and set build timeout to 45min

HannesWell:
The dependencies stored in the .m2 folder not only depend on the content of pom.xml files but also on the content of *.target-files. Therefore the latter should be considered when computing cache-hash.

Furthermore a regular build runs ~10-15min. So a time-out of 45min should be more than sufficient even on a slow builder.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -13,9 +13,10 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
+    timeout-minutes: 45
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
       with:
         fetch-depth: 0
     - name: Set up JDK 17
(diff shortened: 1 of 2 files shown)


.github/workflows/maven.yml
```diff
@@ -23,7 +24,14 @@ jobs:
       with:
         java-version: '17'
         distribution: 'temurin'
-        cache: 'maven'
+    - name: Cache local Maven repository
```
laeubi, line 27: I think it would be usefull to as the java-people to adjust their default to include target files as well.

HannesWell: That would be great, but I suspect they will not consider Eclipse target files important enough for the whole java-world.
But it's worth to ask. What would also help if there would be an option to adjust the pattern of included files.

laeubi: > What would also help if there would be an option to adjust the pattern of included files.

maybe this is a more acceptable change for generic java world then...

HannesWell: > > What would also help if there would be an option to adjust the pattern of included files.
> 
> maybe this is a more acceptable change for generic java world then...

I assume that this is at least more likely to happen.
Do you want to ask the setup-java action devs? Otherwise I can do that later.

HannesWell:
Also updated to actions-checkout v3

HannesWell:
The GH workflow succeeded. No need wait for Jenkins then.

.github/workflows/maven.yml
HannesWell: Created https://github.com/actions/setup-java/issues/345.

PR #784, opened 2022-06-15
Fix Pom-Editor's exclude dependency label

HannesWell:


HannesWell:
These labels are not checked in the verification build, so there is no need to wait for Jenkins too.

Issue #747, opened 2022-06-03
Use latest maven-4.0.0.xsd to validate pom.xml


brbog:
At the moment following snippets are considered faulty inside a pom.xml:
```
	<distributionManagement>
		<site child.site.url.inherit.append.path="false">
			<id>...</id>
			<name>...</name>
			<url>...</url>
		</site>
	</distributionManagement>
```
```
	<scm child.scm.connection.inherit.append.path="false" child.scm.developerConnection.inherit.append.path="false">
		<connection>...</connection>
		<developerConnection>...</developerConnection>
	</scm>
```
This is because the xsd used to validate the pom.xml is cached instead of fetched from its internet location and the cached version is outdated. These attributes were additions to later Maven versions.

You can get the latest xsd from Maven (http://maven.apache.org/xsd/maven-4.0.0.xsd). I think the fix is replacing content of ``m2e-core/org.eclipse.m2e.model.edit/src-xsd/org/eclipse/m2e/model/pom/pom.xsd`` with that of the latest ``maven-4.0.0.xsd``, but I can be wrong.

mickaelistria:
I couldn't reproduce it locally (that doesn't invalidate a caching issue). Can you provide steps to reproduce the issue with a fresh Eclipse IDE and installing m2e from https://download.eclipse.org/technology/m2e/snapshots/latest ?

HannesWell:
> You can get the latest xsd from Maven (http://maven.apache.org/xsd/maven-4.0.0.xsd). I think the fix is replacing content of `m2e-core/org.eclipse.m2e.model.edit/src-xsd/org/eclipse/m2e/model/pom/pom.xsd` with that of the latest `maven-4.0.0.xsd`, but I can be wrong.

That XSD indeed can be updated. But it is not only about updating the xsd, the code generated from it also has to be re-generated (and the project itself can be cleaned up and its documentation can be improved). I started some work on this a while ago but did not complete it yet (I'm not sure about the reasons), but I can have a look at that again soon.

But with Mickaels comment I'm not sure if this is the actual root of your problem.

brbog:
Reproduced using following steps:
[...]

PR #786, opened 2022-06-15
Update maven-v4_0_0.xsd with latest from https://maven.apache.org/xsd/maven-4.0.0.xsd

brbog:
Fix after discussing in issue https://github.com/eclipse-m2e/m2e-core/issues/747 .


brbog:
Creating a new commit so the eca check will succeed.

PR #787, opened 2022-06-15
Update maven-v4_0_0.xsd with latest from https://maven.apache.org/xsd/maven-4.0.0.xsd

brbog:
Fix after discussing in issue #747 .


mickaelistria:
Thank you. New snapshots of m2e including this change should be available in https://download.eclipse.org/technology/m2e/snapshots/latest/ in the next ~15 minutes. Please try upgrading to those and report on the issue ticket whether this is fixed or not.

Issue #682, opened 2022-05-09
[EPIC] allow m2e to process projects that don't have a nature


mickaelistria:
The most important parts of m2e are:

- Classpath/Dependency resolution injected into JDT
- Builder that allows to run some mojos as part of Eclipse build
- pom.xml edition

Ideally, those shouldn't require a nature. We should audit that those can work without a nature and try to fix them if necessary; and only rely on the classpath/builders/editor.

The only foreseen issue is for detection of m2e capabilities on some projects (eg building local project registry, enabling some actions...).  The nature is sometimes checked. More dynamic ways (eg trying to retried an IMavenProjectFacade via an adapter) could be used and Maven could be queried about its capacity to handle some projects or not. This may require progress with https://github.com/eclipse-m2e/m2e-core/issues/118 as pomless projects should be visible to m2e internals.

laeubi:
I think the problem is that a builder is needed to effectively participate in workspace builds.

mickaelistria:
I believe the builder doesn't mandate a particular nature.

laeubi:
I recently tried in another project and found no other reliable way, but yes maybe there are other natures one could attach to. The question remains, if really the pure presence of a pom.xml should trigger certain non trivial actions.

palminha:
> [@palminha](https://github.com/palminha) if this is crucial to your business and you likes to speed up the development of m2e in general a [sponsoring](https://github.com/sponsors/laeubi?frequency=one-time) would allow me to assign more time-slots to bug fixing or you can contact me for an individual contract for example fixing this specific issue.
> 
> Beside that contributions are of course always welcome.

@laeubi Sorry, but this is not a question of being critical of business... if that was the question, I would move to IntelliJ and stop worrying about the IDE community.

It's the first time I've received this reaction from a foundation-backed open-source project... ' Pay us more, and we'll fix it. '
This is typical in those hybrid projects when a company is behind. In this case, I find the "hussling" a bit too much for a project backed by an open-source foundation.

Your first goal should be to prioritize user needs, not those who pay more.

laeubi:
> It's the first time I've received this reaction from a foundation-backed open-source project... ' Pay us more, and we'll fix it. '
This is typical in those hybrid projects when a company is behind.

So if no company is behind you expect developers to work for free or what? 

As said, we always welcome any contribution in the sense of open-source that is YOU the user!

[...]

PR #791, opened 2022-06-18
Allow running maven from subfolder and non-maven-projects

laeubi:
Currently there are two ways to trigger a maven run:

1) right click on a pom.xml file
2) right click on a project with maven nature

in all other cases, an execution is not offered.

With this change, it is now possible to do this also for:

3) an arbitrary subfolder that contains a pom.xml
4) an arbitrary subfolder that is a polyglot (aka pomless)
5) any project with any nature that fulfills (3) or (4)

This is an enhancement towards 
- #682


laeubi:
As an example, one could import the [Equinox project](https://github.com/eclipse-equinox/), but only the root folder of the git:
grafik

As one can see, this projects has a plain java nature but a `pom.xml`

# Current behavior

## Nothing for maven is offered:
grafik

the same happens for every subfolder, regardless if it contains a `pom.xml `or not.

# Behavior after this change

## Right click on the projects offers the usual maven run options:
grafik

## Right click on a folder containing a `pom.xml` offers the usual maven run options
grafik

## Right click on a folder that contains a polyglot (aka pomless) project also offers the usual maven run options:
grafik

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IPomFacade.java
```diff
@@ -0,0 +1,25 @@
...
+ *      Christoph Läubrich - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.m2e.core.embedder;
+
+import java.io.File;
+
+import org.eclipse.core.resources.IFile;
+
+
+public interface IPomFacade {
```
mickaelistria, line 18: I dislike a new public undocumented API for this. Could it be made internal? (that's a rhetorical question)

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/MavenModelManager.java
```diff
@@ -146,6 +147,34 @@ public org.apache.maven.model.Model readMavenModel(File pomFile) throws CoreExce
     return null;
   }
 
+  public Optional<File> locatePom(File baseDir) {
```
mickaelistria, line 150: Please add @since and javadoc for this new API.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/PomFacadeAdapterFactory.java
```diff
@@ -0,0 +1,120 @@
...
+            .orElse(null);
+        return adapterType.cast(facade);
+      }
+      return null;
+    }
+    return null;
+  }
+
+  private IPomFacade getFacadeForPom(IFile file) {
+    System.out.println("file=" + file);
```
HannesWell, line 73: This debug print-out should be removed as well.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/MavenModelManager.java
```diff
@@ -146,6 +147,34 @@ public org.apache.maven.model.Model readMavenModel(File pomFile) throws CoreExce
...
+      //check the obvious case first...
+      return Optional.of(file);
+    }
+    ClassLoader ccl = Thread.currentThread().getContextClassLoader();
+    try {
+      PlexusContainer plexusContainer = containerManager.aquire(baseDir);
+      Thread.currentThread().setContextClassLoader(plexusContainer.getContainerRealm());
+      ModelProcessor modelProcessor = plexusContainer.lookup(ModelProcessor.class);
+      File pom = modelProcessor.locatePom(baseDir);
+      System.out.println(pom);
```
HannesWell, line 165: This should be removed.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/PomFacadeAdapterFactory.java
```diff
@@ -0,0 +1,120 @@
...
+    if(parent instanceof IProject) {
+      IProject project = (IProject) parent;
+      //a direct child of a project, lets check if the project itself could be used...
+      IMavenProjectFacade mavenProjectFacade = mavenProjectRegistry.getProject(project);
+      if(mavenProjectFacade != null) {
+        return mavenProjectFacade;
+      }
+    }
+    //a "detached" resource pom facade is needed
+    return new ResourcePomFacade(file);
```
HannesWell, line 97: What about just a lambda?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/PomFacadeAdapterFactory.java
```diff
@@ -0,0 +1,120 @@
...
+
+  @Override
+  public <T> T getAdapter(Object adaptableObject, Class<T> adapterType) {
+    if(adapterType == IPomFacade.class) {
+      if(adaptableObject instanceof IFile) {
+        IFile file = (IFile) adaptableObject;
+        if(file.getName().equalsIgnoreCase(IMavenConstants.POM_FILE_NAME)) {
+          return adapterType.cast(getFacadeForPom(file));
+        }
+        //TODO not a pom file, but actually pom files can have any name...
```
HannesWell, line 54: Maybe it should be looked into the file if it defines the corresponding XML header and starts with a suitable element to decide if it is good to go further. Or does the look-up fail gracefully if it is an unrelated file.

Alternatively the container could be used in this case. But then the 'Run as Maven build' would be available for all files in a (pomless) Module. This is probably not desired.

HannesWell (commented):
That's a handy improvement. Thanks for your work on that.
Besides Mickael's comments I only have a few minor remarks.

In general I wonder if we can also improve the default values for the build's basedir.
If a certain resource in the workspace is used as context I think the launch-config should be configured to build that resource. I.e. I'm think of if it is a folder that folder should be used as basedir or if it is a file the basedir could be the file's container and the -f option to the file should be specified.

But at the same time it would be great if we had a convenient way to build the all projects from the root that contains the resource (e.g. by providing a selection of good defaults). At the moment most of the time I specify for example `${git_work_tree:/m2e-core}`.
But the git-work-tree root is probably not always a reliable assumption for the root. Alternatively it could be checked for an `.mvn` folder but that is not always present.
Does Maven provides means to get the 'root' folder of a project already?

laeubi:
> Does Maven provides means to get the 'root' folder of a project already?

The is no root as maven projects are not required to be hierarchically organized (from file system POV). so actually the user has to know from where to run the build.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IPomFacade.java
laeubi: I'll add java doc for this, `IPomFacade` is needed for `IMavenProjectFacade` so we can differentiate between maven project related and "detached"  `IPomFacade` thus  `IPomFacade` needs to be public as well.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/MavenModelManager.java
laeubi: I might move this to another place instead here ...

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/PomFacadeAdapterFactory.java
laeubi: This will require more methods to support other use-cases as well so better not "optimize" here what must be reverted soon.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/PomFacadeAdapterFactory.java
laeubi: Using the container is bad as one can't decide then if this is really something that could be executed as maven, this should be avoided. Beside that m2e currently do not support other things than pom elsewhere so we should not add more magic here.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/PomFacadeAdapterFactory.java
HannesWell: Acknowledge.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/PomFacadeAdapterFactory.java
HannesWell: Sound's reasonable. These were just some thought I wanted to put on the table.

HannesWell:
> > Does Maven provides means to get the 'root' folder of a project already?
> 
> The is no root as maven projects are not required to be hierarchically organized (from file system POV). so actually the user has to know from where to run the build.

Yes, that makes it impossible to compute the 'right' root. But maybe should add some convenient support for the most frequent use-cases? E.g. a drop-down menu that provides entries like
- `${git_work_tree:/<current-context-project>}` 
- `${project_loc:/<current-context-project>}`
- more?

And if wanted a user can select them more easily than navigating through the variables? Of course it should be still possible to enter arbitrary paths.
But I think this is for another issue.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IPomFacade.java
laeubi: > Do clients of m2e (eg people developing extensions) have to know about IPomFacade?

If they want to support "traditional" pom.xml and polyglot projects, yes. Beside this, IPomFacade could be used as a more lightweight method parameter in cases where you just need "the pom" (as in loading the projects object model) in contrast to the full MavenProject, e.g. as shown here for the launching case.

> Do they need to implement it?

Probably yes, if a plugin has a mean of locating the pom beside the usual polyglot/maven defaults. but we can for sure add `@noimplement` for now and see if there are strong use cases for implementation.

> explains why API addition makes everything more complex and encourages to avoid creating public API is there is no obvious value offered by consumers.

While in general this might be true, I actually noticed that this often leads to very bad API design where everyone actually *is* using internal API (best example P2), and also m2e suffer from this. Actually we use all over the place internal APIs (even in so called plugins) and that's really bad as it hinders other to write "good" plugins. 
As often there it is told we should consume our own code in platform, I think we also should consume/respect our own API to prove its useful and complete. So actually no `x-friends` no `@SupressWarning(discouraged)` should ever be necessary beside some very rare circumstances, as each of those indicates that our API is suffering from actually being useful to (outside) extenders.

mickaelistria: "Probably yes" is not sufficient to add an API. We need real use-cases, requirement definitions and more than 1 example to achieve good API.

> but we can for sure add @noimplement for now and see if there are strong use cases for implementation.

Or just make it internal and everything would still work as expected.

> Everyone actually is using internal API (best example P2), and also m2e suffer from this

That's the problem of consumers if they use internal APIs without contributing user-story or workforce to the project to stabilize the API. If things get broken, it's their problem. So they are not preventing the project from innovating like API addition does.
An API is rarely on 1st shot. See this "internal" phase a a maturation phase. If later there are good use-cases reported for an API and feedback and associated resource, then it can be promoted from internal to official/stable API with all the drawbacks (as it became clear there is enough value and manpower to handle the related drawbacks).

> As often there it is told we should consume our own code in platform, I think we also should consume/respect our own API to prove its useful and complete. So actually no x-friends no @SupressWarning(discouraged) should ever be necessary beside some very rare circumstances, as each of those indicates that our API is suffering from actually being useful to (outside) extenders.

While I admire this, it reminds me my first years as an Eclipse contributor. But I learned this is not as easy as you just said. APIs are a brake to innovation, they're hard to design correctly and require some real requirements to achieve such correctness. Bundle internal vs Bundle external is something that is too simplistic in practice; so there is a world in-between (exported internal modules with x-friends) which has proven to be very helpful.

mickaelistria: What are the use-case you maintain for this API outside of m2e? Can you please showcase how consuming this API proposal will allow eg some connectors to work better?

PR #794, opened 2022-06-19
Fix local archetype-catalog id->URL conversion on Windows

HannesWell:
When Running the `ArchetypeManagerTest` locally on my Windows computer the `testBadLocalArchetypeCatalogFactory` case fails with the following error:
```
java.nio.file.InvalidPathException: Illegal char <:> at index 4: crap://nope
	at java.base/sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
	at java.base/sun.nio.fs.WindowsPath.parse(WindowsPath.java:92)
	at java.base/sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:232)
	at java.base/java.nio.file.Path.of(Path.java:147)
	at org.eclipse.m2e.core.ui.internal.archetype.ArchetypeCatalogFactory$LocalCatalogFactory.getArchetypeCatalog(ArchetypeCatalogFactory.java:177)
	at org.eclipse.m2e.tests.archetype.ArchetypeManagerTest.asserEmptyCatalog(ArchetypeManagerTest.java:135)
	at org.eclipse.m2e.tests.archetype.ArchetypeManagerTest.testBadLocalArchetypeCatalogFactory(ArchetypeManagerTest.java:128)
```

This PR fixes the error for me.
@laeubi is this in general a reasonable fix?

### DIFF
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeCatalogFactory.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeCatalogFactory.java
@@ -13,10 +13,10 @@
 
 package org.eclipse.m2e.core.ui.internal.archetype;
 
-import java.io.BufferedInputStream;
-import java.io.File;
 import java.io.InputStream;
+import java.net.URI;
 import java.net.URL;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
@@ -174,12 +174,7 @@ public abstract class ArchetypeCatalogFactory {
         //local but not embedded catalog
         IMavenExecutionContext context = maven.createExecutionContext();
         ArtifactRepository localRepository = new MavenArtifactRepository();
-        Path path = Path.of(getId());
-        File file = path.toFile().getAbsoluteFile();
-        if(file.isFile()) {
-          file = file.getParentFile();
-        }
-        localRepository.setUrl(file.toURI().toString());
+        localRepository.setUrl(getLocalRepositoryURL());
         context.getExecutionRequest().setLocalRepository(localRepository);
         return context.execute((ctx, m) -> {
           ProjectBuildingRequest buildingRequest = ctx.newProjectBuildingRequest();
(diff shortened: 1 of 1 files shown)


org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeCatalogFactory.java
```diff
@@ -194,13 +189,31 @@ public ArchetypeCatalog getArchetypeCatalog() throws CoreException {
...
+    }
+
     private ArchetypeCatalog getEmbeddedCatalog() throws CoreException {
       URL url = getEmbeddedUrl();
       if(url == null) {
         //Not an embedded catalog, nothing else to do
         return null;
       }
-      try (InputStream is = new BufferedInputStream(url.openStream())) {
+      try (InputStream is = url.openStream()) {
```
HannesWell, line 216: The ArchetypeCatalogXpp3Reader uses a buffered Stream/Reader already internally. Not need to buffer twice.

laeubi:
if it helps, why not :+1:

laeubi (approved):

HannesWell:
Good. Thanks for your review.

HannesWell:
> ## Unit Test Results
> 
> 50 tests   49 [✔️](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols)  2m 18s [⏱️](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols) 12 suites    1 [💤](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols) 12 files      0 [❌](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols)


Btw. I think the github build has to check-out submodules, which explains why it runs so much faster than the Jenkins build.
I'll prepare a PR.

laeubi:
> > ## Unit Test Results
> > 50 tests   49 [heavy_check_mark](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols)  2m 18s [stopwatch](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols) 12 suites    1 [zzz](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols) 12 files      0 [x](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols)
> 
> Btw. I think the github build has to check-out submodules, which explains why it runs so much faster than the Jenkins build. I'll prepare a PR.

we might use the GH action as a fast pre-check. Beside that I think whenever we write new code we should add tests to m2e directly and use me2-core-tests as legacy test suite.

HannesWell:
> we might use the GH action as a fast pre-check. Beside that I think whenever we write new code we should add tests to m2e directly and use me2-core-tests as legacy test suite.

We can use GH as fast pre-check (for compiling, version bumps etc.). But then I suggest to skip tests completely.
Otherwise the report gives a false impressions about the really executed tests and only provides half of the value.

PR #795, opened 2022-06-19
Checkout submodules in GH-workflow build

HannesWell:
By default submodules are not checked out, but the m2e-core repo has the 'm2e-core-tests' sub-module wihich contains many tests.

For a detailed explanation of the input parameters see: https://github.com/actions/checkout#usage


HannesWell:
The `its` profile was missing in the build command, which also explains why the build did not fail with absent submodules.

@laeubi suggested to use the GH-build as 'quick response' build. While this is not a bad idea, I think this is only reasonable if all tests are disabled and we use it only to check if the build compiles, versions are fine, etc. Only executing a part of all tests has only partial value. But we would then loose the test result comment, wouldn't we?

laeubi:
> I think this is only reasonable if all tests are disabled and we use it only to check if the build compiles, versions are fine, etc. Only executing a part of all tests has only partial value. 

At Tycho we for exampely onle excute the itests when the PR is not a draft, but still execute the unit test, so I don't think its bad per-se...

> But we would then loose the test result comment, wouldn't we?

This only works if tests are executed obviously :-)

HannesWell:
> > I think this is only reasonable if all tests are disabled and we use it only to check if the build compiles, versions are fine, etc. Only executing a part of all tests has only partial value.
> 
> At Tycho we for exampely onle excute the itests when the PR is not a draft, but still execute the unit test, so I don't think its bad per-se...

Hm. That's a point.

> 
> > But we would then loose the test result comment, wouldn't we?
> 
> This only works if tests are executed obviously :-)

That doesn't surprise me. Was more a rhetorical question. :D

However I think both approaches have their valid arguments, so I'm not sure how to proceed here.
If we want to keep it as it is I suggest to change the title to something like `Unit Test Results (excluding Integration-Tests)` to make it more clear (in my opinion):
https://github.com/EnricoMi/publish-unit-test-result-action#configuration


Besides that I think it would be better to add the test results even for failed runs. At least failed runs due to test failures are even more interesting.

HannesWell:
Looks like all tests are now executed and displayed by the JUnit test result publisher.

So how should we continue here?

HannesWell:
Actually I could think of building in two steps. First without the integration tests report that result and then resume the build and build and run the integration tests and update the results.
It could even be done in three steps:

1. Compile, version check etc., basically everything but -DskipTests
2. Run non-integration tests
3. Run integration tests.

If possible the test results could be updated after step 2 and 3.

This way we would get quick responses in case of failures but still get the full test coverage.

On the other hand. Since tests are usually sorted to the end (from my experience and IIRC this changed in past, so I think @laeubi or @mickaelistria you tweaked that are are aware of it) and can implicitly assume that if the build did not fail quickly that it at least passed the compilation and the longer it runs the more is passed.

HannesWell:
> Is there anything preventing this one from being merged? It would be nice to have integration tests executed by GH action and it looks to me that it's already working.

It is working. I just did not commit it because Christoph suggested to keep it as is to get faster results. But the discussion stalled since then.
But I would be favor to submit this, especially since if https://github.com/eclipse-m2e/m2e-core/pull/851 passes one day only the `org.eclipse.m2e.tests` remain in the sub-module and the difference should only be a few minutes.

laeubi:
Its fine from my side!

PR #796, opened 2022-06-19
Remove unused embedded archetype-dependencies

HannesWell:
In order to keep the size of the m2e.archetype.common bundle as small as possible, this removes the embedded dependencies that are never accessed at runtime.

I removed all dependencies that are now excluded one by and checked if the `ArchetypeManagerTest` still passes. If we are are really sure that this test covers all use-cases this should be fine. Nevertheless we should use the Archetype capabilities intensively in the near future so that we hopefully did not miss a case.

Size of `org.eclipse.m2e.archetype.common` bundle:
- current snapshot: 10.1MB
- latest release (1.20.1): 1.9MB
- after this change (sum of all embedded jars): 672 KB

### DIFF
--- a/m2e-maven-runtime/org.eclipse.m2e.archetype.common/pom.xml
+++ b/m2e-maven-runtime/org.eclipse.m2e.archetype.common/pom.xml
@@ -76,6 +76,55 @@
 					<groupId>org.apache.maven</groupId>
 					<artifactId>maven-aether-provider</artifactId>
 				</exclusion>
+				<exclusion>
+					<groupId>org.slf4j</groupId>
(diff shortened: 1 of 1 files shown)


laeubi (approved):
Looks good!

laeubi:
By the way if you like to do some manual test:

1. New > Maven Project
2. Type "faktorips"
3. create a project with that archetype

This includes:

- Interactive input
- Groovy script

HannesWell:
Just tried that and got the following error, but this looks just like https://github.com/eclipse-m2e/m2e-core/issues/799.
But I have to admit I did not check if this worked before.

grafik

PR #797, opened 2022-06-19
Use the maven standalone project instead of a custom string

laeubi:


org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeGenerator.java
```diff
@@ -146,9 +150,10 @@ private File getEmptyPom(File basedir) {
       try {
         File tempFile = File.createTempFile("pom", ".xml", basedir);
         tempFile.deleteOnExit();
-        Files.writeString(tempFile.toPath(),
-            "<project><modelVersion>4.0.0</modelVersion><groupId>empty</groupId><artifactId>empty</artifactId><version>1</version><name>Generating archetype</name></project>",
-            StandardCharsets.UTF_8);
+        URL standalone = DefaultMaven.class.getResource("project/standalone.xml");
```
HannesWell, line 153:
```suggestion
        URL standalone = DefaultMaven.class.getResource("/org/apache/maven/project/standalone.xml");
```

For me this would be a bit more clear where I have to look in the jar containing the class if I want to see what's in that file. But the current value also work, so maybe this is just my personal preference.

HannesWell (approved):
That's a good change. Using predefined resources instead of some 'magic' strings is definitely better.

I have just on suggestion to make the path IMO more clear. But since the current one also works this is not a blocker.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeGenerator.java
laeubi: I'm using this because its the way maven lookup this, just in case the `DefaultMaven` is moved:

https://github.com/apache/maven/blob/b3e7e57e4bc20c5c387fb1712be07847043bef9a/maven-core/src/main/java/org/apache/maven/graph/DefaultGraphBuilder.java#L395

PR #798, opened 2022-06-20
Add a MavenToolbox

laeubi:
Currently we have some scattered places where we put useful "shorthands"
that perform some maven specific tasks as it is cumbersome to write them
down each time. Often they are not really related to the object itself
but could be written as a function of an execution and/or session.

Because it is often arguable if we should add it to the given interface
to not bloat the API and on the other hand it might became useful on
other places as well, this change introduces a "MavenToolbox" that could
be a place for such methods, easily constructed by a static method but
not intended to be implemented by clients.

This will solve several issues:
- do not require to do the same thing on different places just to not
introduce API
- prevent code from referencing internal API
- can enhance this as we like
- can deprecate methods if we think they are no longer useful without
harming actual m2e implementations

I have only refactored a few places until now so we can first decide if this is a viable option.

### DIFF
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/wizards/MavenModuleWizardParentPage.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/wizards/MavenModuleWizardParentPage.java
@@ -46,6 +46,7 @@ import org.apache.maven.model.Model;
 
 import org.eclipse.m2e.core.MavenPlugin;
 import org.eclipse.m2e.core.internal.IMavenConstants;
+import org.eclipse.m2e.core.internal.IMavenToolbox;
 import org.eclipse.m2e.core.project.ProjectImportConfiguration;
 import org.eclipse.m2e.core.ui.internal.Messages;
 import org.eclipse.m2e.core.ui.internal.WorkingSets;
@@ -221,7 +222,7 @@ public class MavenModuleWizardParentPage extends AbstractMavenWizardPage {
       parentContainer = pom.getParent();
 
       try (InputStream pomStream = pom.getContents()) {
-        parentModel = MavenPlugin.getMavenModelManager().readMavenModel(pomStream);
+        parentModel = IMavenToolbox.of(MavenPlugin.getMaven()).readModel(pomStream);
         validateParent();
         parentProjectText.setText(parentModel.getArtifactId());
       } catch(CoreException | IOException e) {
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
@@ -42,6 +42,7 @@ import org.apache.maven.settings.Settings;
 import org.apache.maven.settings.building.SettingsProblem;
 import org.apache.maven.wagon.proxy.ProxyInfo;
 
+import org.eclipse.m2e.core.internal.IMavenToolbox;
 import org.eclipse.m2e.core.internal.embedder.MavenImpl;
 import org.eclipse.m2e.core.project.IMavenProjectFacade;
 
@@ -60,9 +61,17 @@ public interface IMaven extends IComponentLookup {
 
   // POM Model read/write operations
 
-  Model readModel(InputStream in) throws CoreException;
+  /**
+   * @deprecated use the {@link MavenModelManager} instead
+   */
+  @Deprecated(forRemoval = true)
+  default Model readModel(InputStream in) throws CoreException {
+    return IMavenToolbox.of(this).readModel(in);
+  }
 
-  void writeModel(Model model, OutputStream out) throws CoreException;
+  default void writeModel(Model model, OutputStream out) throws CoreException {
+    IMavenToolbox.of(this).writeModel(model, out);
+  }
 
   // artifact resolution
 
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/MavenModelManager.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/MavenModelManager.java
@@ -16,7 +16,6 @@ package org.eclipse.m2e.core.embedder;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Collection;
@@ -24,6 +23,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 
 import javax.xml.parsers.DocumentBuilder;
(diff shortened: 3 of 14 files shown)


mickaelistria:
> Currently we have some scattered places where we put useful "shorthands" that perform some maven specific tasks as it is cumbersome to write them down each time. 

A quick look at the proposal gives me there is now 1 more place that host "useful shorthands that perform some maven specific tasks" and adds indirections. So it actually makes things even more scattered.

> prevent code from referencing internal API

Internal APIs are not wrong.

> can enhance this as we like [...] can deprecate methods if we think they are no longer useful without harming actual m2e implementations

We can do it too if things remain internal.


Again, the reason why we struggle to change things here and there is because there are already (too?) many APIs, that were open to the wider world without a consolidated requirements to have them. So now we're stuck respecting this. Adding more APIs is definitely not helping the problem of too much APIs. IMO, this patch does the opposite of what it's aiming at.

laeubi:
> A quick look at the proposal gives me there is now 1 more place that host "useful shorthands that perform some maven specific tasks" and adds indirections. So it actually makes things even more scattered.

As explained, I have not yet started to refactor all places.

> Internal APIs are not wrong.

They are wrong when it leads to implementing the same thing multiple times as it is done with these methods, that are actually not a matter of where they are placed but its just feels random.

> We can do it too if things remain internal.

Honestly, we probably better should make the whole m2e 'internal' and don't allow any extensions then ...

mickaelistria:
> They are wrong when it leads to implementing the same thing multiple times as it is done with these methods,

This assumption seems weak to me. Can you please show me an example of why internal APIs (ie exported packages with x-friends) require to duplicate things?

mickaelistria:
> Honestly, we probably better should make the whole m2e 'internal' and don't allow any extensions then ...

That's how things should have started. But m2e decided to open many things as API, probably much more things than necessary, and now we're paying the price of it.

laeubi:
> > They are wrong when it leads to implementing the same thing multiple times as it is done with these methods,
> 
> This assumption seems weak to me. Can you please show me an example of why internal APIs (ie exported packages with x-friends) require to duplicate things?

Just search for `@SuppressWarnings("restriction")` in m2e source (around 200 hits) all these places uses internal API to not replicate some code from an "internal"  package.

> But m2e decided to open many things as API, probably much more things than necessary, and now we're paying the price of it.

That's why I currently try to consolidate things but if there is no replacement there is no way to migrate from all the different places. I think we already have made a big step forward and (from my opinion) the `IMavenToolbox` will even allow to reduce this more as most of the time it will be enough to have:

1. `IPomFacade` / `IMavenProjectFacade`
2. `IMavenExecutionContext`
3. `IComponentLokkup`
4. `IMavenToolbox`
5. `IMaven` in some rare cases to get 2 or 3 in a global context

that's the long term goal (for me) but surely will not happen over-night.

mickaelistria:
> Just search for @SuppressWarnings("restriction")

Those could be replaced by proper `x-friends` definition in the bundle defining the internal API.

> all these places uses internal API to not replicate some code from an "internal" package.

And what's the issue with that?

About your proposal, I think we can already reduce it to IMavenProjectFacade + IMavenExecutionContext + IComponentLokkup + MavenModelManager + IMaven without any functional loss for users and consumers.

laeubi:
> > Now there is unified access vie the MavenToolbox, using only the lookup, so any IComponetLookup can now be used to reliable read a maven model, there is only one implementation and we can actually throw away much code.
> 
> OK, but why not moving this to MavenModelManager instead of creating something new which doesn't have strong semantics ?

Because `MavenModelManager` do not really "manages" anything and Maven Model do not need any managing at all, actually we should be able to get rid of this completely and/or convert it to either Toolbox or static util methods, e.g `MavenModelManager.createMavenModel(IFile, Model)` could be static if we would passing them a MavenToolbox and thus it become independet of IMaven at all and could be used in cases where there are other Modelwriters...

mickaelistria:
>  these shows that none of these plugins are actually be able to be provided by someone else

The point of `x-friends/x-internal` is to tell consumers that this shouldn't be used at the moment as it can change without notice in future version, and that it's currently exported by the project because it's used by multiple bundles, but that have enough checks (main one being building together) to ensure consistency.

laeubi:
Alright I think I'd like to get some more feedback on this topic so please @eclipse-m2e/technology-m2e-committers participate here so we can process further, new features require new APIs, and marking everything as internal neither feels right nor makes it easier for consumer to migrate to new features. As this is a great opportunity to also remove the need for several competing API I think this is the way to go unless someone suggests a better alternative.

mickaelistria:
> new features require new APIs

Again a wrong assumption.

laeubi:
> > new features require new APIs
> 
> Again a wrong assumption.

Maybe your assumption, that my assumption is wrong, is wrong ;-)

laeubi:
I didn't state that *all* new features require new APIs so you base your argumentation on a false already so your whole claim is false from the beginning. Nerveless this all do not add any value to the discussion nor does it solves anything but gives a bad mood claiming that someone else is "again wrong" without giving any useful context nor alternative.

As said before, we are doing a major release so lets mark everything as x-internal and we have no "API" anymore so no problems... nerveless it seems we just disagree here and should then follow https://github.com/eclipse-platform/.github/wiki/PMC-project-guidelines#committer-disagreement-resolution if there is no chance to get consent here.

laeubi:
I have now moved the toolbox to an internal package and redirect at some places, so no new API added anymore...

laeubi:
@mickaelistria as there is no more new API I assume your major concerns are addressed and we can proceed with more enhancements in a separate PR if necessary?

mickaelistria (approved):
> as there is no more new API I assume your major concerns are addressed and we can proceed with more enhancements in a separate PR if necessary?

perfect, thanks.

mickaelistria:
Looks like the rebase made it cause test failures: https://ci.eclipse.org/m2e/job/m2e/job/master/335/ . @laeubi can you please have a look?

laeubi:
if the next build do not fix this I'll take a look at it.

mickaelistria:
Failure is persisting on master and also on PR builds.

mickaelistria:
Usually, those case show that some test would now require some extra dependency to work, which is not added by default in Tycho but that's usually part of the PDE Launch Configuration.

laeubi:
```
mvn clean verify -Dtest=org.eclipse.m2e.tests.internal.wizard.MavenImportWizardPageTest -Puts -Pits

....

Running org.eclipse.m2e.tests.internal.wizard.MavenImportWizardPageTest
Tests run: 6, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5.128 s - in org.eclipse.m2e.tests.internal.wizard.MavenImportWizardPageTest
test408042_multimoduleProject(org.eclipse.m2e.tests.internal.wizard.MavenImportWizardPageTest)  Time elapsed: 1.096 s
test408042_nestedproject(org.eclipse.m2e.tests.internal.wizard.MavenImportWizardPageTest)  Time elapsed: 1.747 s
test408042_noRootProject(org.eclipse.m2e.tests.internal.wizard.MavenImportWizardPageTest)  Time elapsed: 0.495 s
test533463_noArtifactId(org.eclipse.m2e.tests.internal.wizard.MavenImportWizardPageTest)  Time elapsed: 0.495 s
test408042_simpleProject(org.eclipse.m2e.tests.internal.wizard.MavenImportWizardPageTest)  Time elapsed: 0.464 s
test422008_preselectedWorkingSet(org.eclipse.m2e.tests.internal.wizard.MavenImportWizardPageTest)  Time elapsed: 0.789 s

Results:

Tests run: 6, Failures: 0, Errors: 0, Skipped: 0

```

PR #804, opened 2022-06-20
[Build] Remove duplicating 'uts' Maven profile

HannesWell:
It has the same content like the 'its' profile.


HannesWell:
I replayed the build with the new command. If that succeeds this change should be fine.

HannesWell:
This was too fast. I oversaw that the m2e-core-tests submodules adds each of its two test-projects with one of those two profiles.
So basically this disabled `org.eclipse.m2e.tests`.
-> Reverting.

PR #805, opened 2022-06-20
Revert "[Build] Remove duplicating 'uts' Maven profile"

HannesWell:
Reverts eclipse-m2e/m2e-core#804


HannesWell:
Just reverts the break of what has worked before. So no need to wait for the build.

PR #814, opened 2022-06-22
Support Pomless in ProjectRegistry/ConfigurationManager

laeubi:
Currently the use hard coded defaults, but should locate the real name for the pom instead.


mickaelistria:
This seems like a great change!

PR #818, opened 2022-06-24
Add a m2e-ide product

laeubi:
This serves different purpose:

- one could get a runnable ide from building m2e without the need for
using snapshot and ibuild sites or whatever
- we can use this to create tests using RCPTT
- we can make sure that it is possible to build own packages with m2e as
a basis

### DIFF
--- a/products/.project
+++ b/products/.project
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>m2e-products</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+	</buildSpec>
+	<natures>
+	</natures>
+</projectDescription>
--- a/products/m2e-ide/m2e-ide.product
+++ b/products/m2e-ide/m2e-ide.product
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?pde version="3.5"?>
+
+<product name="M2Eclipse IDE" uid="m2e-ide" id="org.eclipse.platform.ide" application="org.eclipse.ui.ide.workbench" version="2.0.0" useFeatures="true" includeLaunchers="true" autoIncludeRequirements="true">
+
+   <configIni use="default">
+   </configIni>
+
+   <launcherArgs>
+      <programArgs>-data @noDefault
+      </programArgs>
+      <vmArgs>--add-modules=ALL-SYSTEM
+--illegal-access=permit 
+--add-opens java.base/java.lang=ALL-UNNAMED
+-Djava.security.manager=allow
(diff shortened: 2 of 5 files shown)


products/m2e-ide/m2e-ide.product
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?pde version="3.5"?>
+
+<product name="M2Eclipse IDE for testing purpose" uid="m2e-ide" id="org.eclipse.platform.ide" application="org.eclipse.ui.ide.workbench" version="2.0.0" useFeatures="true" includeLaunchers="true" autoIncludeRequirements="true">
```
HannesWell, line 4: Maybe `M2Eclipse IDE Demo`?

HannesWell (commented):
In general this looks good and if it helps to get even better Testing or for users to start, why not. :)

If we add a product I suggest to associated the `Eclipse-M2E-IDE` launch-config currently located in `m2e-core/setup` with it and move that into the product project.

I think you should also add a .project file for a general purpose project to the m2e-ide to make it a Eclipse project visible in the workspace and not only nested within the m2e-core root project.

laeubi:
> If we add a product I suggest to associated the `Eclipse-M2E-IDE` launch-config currently located in `m2e-core/setup` with it and move that into the product project.

Currently the product could not be launched see: 
- https://github.com/eclipse-pde/eclipse.pde/issues/185

if everything works, we could simply use the product instead of a launch config I think that's more stable but for now I would just keep it as is.

products/.project
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
HannesWell, line 1: Do we will have more products?
If not I suggest to move it into the `releng` folder. The target and repository project can then reside there as well (when we move to a deeper project structure).

laeubi: > Do we will have more products?

don't know, but I'd like to follow structured layout here, a product has nothing to do with releng (=release engineering) stuff.

HannesWell: I wouldn't say nothing, but yes products if is of course the most suitable name.
However, would the .project file then better be in the folder containing the .product instead of the container?

laeubi: As a product file do not need a project at all I don't think we need (dedicated) projects per product here.

HannesWell: OK. I both is valid, but since this your change, you decide. :)

I just wonder in general if Tycho then builds multiple products in the same project as long as they are in separate directories? IIRC having multiple products in the same directory does not work (well).

laeubi: Tycho (pomless) builds each folder, but using multiple in one folder would also work...

PR #820, opened 2022-06-24
Clean up MavenArchiver sources and project configuration

HannesWell:
Lets wait until we have tests (https://github.com/tesla/m2e-core-tests/pull/166) included so this does not add bugs.


HannesWell:
Besides the clean-ups this fixes #827.

PR #821, opened 2022-06-24
Unify and improve Plug-in/Feature description

HannesWell:
Since every root Feature is optional I think there is no advantage in mentioning that in its name.
Furthermore I think it is not necessary to mention in the Feature name that m2e includes some incubating components since the majority of features is not incubation and some experimental preferences are labeled as such.

What do the others think?


HannesWell:
Since no objections were raised, I think this is good to go now.
The test-failure is unrelated since this only changes metadata that cannot influence the tests.

PR #824, opened 2022-06-28
No help available, add missing comma

kjsmita6:
The "Add Dependency" dialog of the pom editor has a help button that does nothing. This is to remove it. Also, add a missing comma.

Related: I notice that the indexer project was removed. I am investigating why we are unable to search for Maven dependencies to add, even including Maven projects in the workspace. Is this related? If so, are there plans to add this functionality back a different way?

`Eclipse.org - m2e	M2E Maven Integration for Eclipse Core	1.19.0.20211116-0804	org.eclipse.m2e.feature`


mickaelistria:
Thank you!

PR #828, opened 2022-07-02
Replace Arrays by suitable Collections in m2e.core APIs

HannesWell:
In today's Java programs it is more common to use Lists (or other suitable Collections) than Arrays. Changing the APIs to return/accept Lists therefore avoid conversion, thus making the code more readable and more performant (although the overall impact is likely low).


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "m2e-core-tests"]
 	path = m2e-core-tests
-	url = https://github.com/tesla/m2e-core-tests.git
-	branch = master
+	url = https://github.com/HannesWell/m2e-core-tests.git
```
HannesWell, line 3: This has to be reverted before submission.

HannesWell:
@mickaelistria, @laeubi since this is also a change in M2Es API, although not a big one, I would like to know that you consent to it?

mickaelistria:
m2e APIs have been severly broken compared to 1.x anyway and about all consumers are incompatible now. So one more little change that's easy to perform is acceptable.

laeubi:
No concerns from my side.

PR #829, opened 2022-07-02
Unify dev-data project names and make m2e-setup a project

HannesWell:
And fix warnings about missing project encoding and associate provided M2E launch-config with M2E product.

### DIFF
--- a/products/m2e-ide/m2e-ide.product
+++ b/products/m2e-ide/m2e-ide.product
@@ -7,8 +7,6 @@
    </configIni>
 
    <launcherArgs>
-      <programArgs>-data @noDefault
-      </programArgs>
       <vmArgs>--add-modules=ALL-SYSTEM
 --illegal-access=permit 
 --add-opens java.base/java.lang=ALL-UNNAMED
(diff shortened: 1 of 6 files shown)


products/m2e-ide/m2e-ide.product
```diff
@@ -7,8 +7,6 @@
    </configIni>
 
    <launcherArgs>
-      <programArgs>-data @noDefault
```
HannesWell, line 10: @laeubi when this property is present in the launch-config the eclipse launched from the IDE asks for a workspace.
Of course I can only remove it in the launch-config, but I would like to keep them consistent.
Do you have a non-trivial reason to have it in the product?

HannesWell: This program-argument has been replaced by the corresponding property that will be written to the config.ini:
`<property name="osgi.instance.area" value="@noDefault" />`
So this should behave the same for the build product but does not show-up/affect the launch-config.

HannesWell: Verified with the build product, that it works.

PR #832, opened 2022-07-03
Clean up and simplify m2e.lifecyclemapping and remove unused elements

HannesWell:


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "m2e-core-tests"]
 	path = m2e-core-tests
-	url = https://github.com/tesla/m2e-core-tests.git
-	branch = master
+	url = https://github.com/HannesWell/m2e-core-tests.git
```
HannesWell, line 3: This has to be reverted before submission!

PR #835, opened 2022-07-03
Remove unnecessary debugging .options files

HannesWell:
Does anybody know the purpose of these `.options` files and if we still need them?
IIRC (and from the content) they enable/disable certain debug output?


laeubi:
As far as I know they are used to tell eclipse what debugging options are available for the plugin if you go to `General > Tracing`

mickaelistria:
+1 with what @laeubi ; they're declaration files for tracing options.
See https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.user%2Freference%2Fref-tracing_preference_page.htm and https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.pde.doc.user%2Fguide%2Ftools%2Flaunchers%2Ftracing.htm . Those should not be removed without a clear audit that they are effectively useless, or without a better replacement.

laeubi:
> Given that the debug options for m2e.core only provide marginal information and are not working (likely for some time) I suggest to proceed with the removal as it is after the update.

In fact m2e behaves "bad" here and mostly uses a mix of different logging techniques, so actually it might be better to fix/use this more, e.g if we ask users to "provide more information" it might be better to tell them enabling different tracing options than looking at the maven console, error log and so on...

HannesWell:
> > Given that the debug options for m2e.core only provide marginal information and are not working (likely for some time) I suggest to proceed with the removal as it is after the update.
> 
> In fact m2e behaves "bad" here and mostly uses a mix of different logging techniques, so actually it might be better to fix/use this more, e.g if we ask users to "provide more information" it might be better to tell them enabling different tracing options than looking at the maven console, error log and so on...

Agree on that. But can a user of m2e that just starts its IDE with the launcher and not from a workspace even enable them? Maybe one can specify the -debug flag, but manually creating the options file seems cumbersome.
Therefore I think it is more convenient to use the Eclipse Error-Log or the Maven-Console because they are available in the IDE without additional setup. But if your entire Eclipse crashes that of course does not help.

laeubi:
> Agree on that. But can a user of m2e that just starts its IDE with the launcher and not from a workspace even enable them?

In your Eclipse IDE you go to `General > Tracing` and enable the options for your current IDE, no need to restart or pass any options on commandline.

HannesWell:
Thanks for the explanation.

I agree that we should have a more unified way of logging events, but I'm not fully convinced that this tracing is the best solution for M2E. I think we should keep the Maven-console to get a tracing of Maven events. It would be possible to use another slf4j binding an it could redirected wherever we want, but for Example the Eclipse log seems not to be a good chice for me because then the log view would be permanently spammed. Furthermore the Maven-Console is relatively close to what users are used to from the standard usage of Maven on the CLI.
In general my first impression of the Eclipse tracing is not so well, since it seems to be very verbose. And I wonder if it is really useful in practical use-cases to be able to specify the tracing in (possibly) great detail.

Anyway, since the tracing is broken at the moment (and the few uses are IMHO not very useful) are you OK with the removing the .options for now and submitting this PR as it is?
If we later decide to use the Tracing, we can still recreate a fixed version of them.

laeubi:
As you can see one can choose to redirect to a file or the "std" output, the maven log is really something "special" and serves different purpose.

> furthermore the Maven-Console is relatively close to what users are used to from the standard usage of Maven on the CLI.

Exactly, so m2e should actually not interfere with that, the tracing is a standardized way, we have a good Ui to enable/disable it and you can have fine grained control about what to show.

> In general my first impression of the Eclipse tracing is not so well, since it seems to be very verbose. And I wonder if it is really useful in practical use-cases to be able to specify the tracing in (possibly) great detail.

How verbose or how detailed it is depend on how we craft it, tracing is designed to be hierarchical, so you can e.g enable "m2e"  (->all tracing) or only "m2e/downloads" (-> show downloads) or "m2e/dowloads/sources" and "m2e/dowloads/javadoc" to see only tracing about sources+javadoc downloads.

> If we later decide to use the Tracing, we can still recreate a fixed version of them.

My personal opinion is that "later" is often another term for "never" :-)

Still I'm not an evangelist of tracing, but the current "write everything to the log" is also far from perfect and hard to use.

HannesWell:
> > furthermore the Maven-Console is relatively close to what users are used to from the standard usage of Maven on the CLI.
> 
> Exactly, so m2e should actually not interfere with that, 

That's right. On the other hand it makes it simpler to associate events (and their order) with each other if they are in the same View/File than in different ones.

> 
> > If we later decide to use the Tracing, we can still recreate a fixed version of them.
> 
> My personal opinion is that "later" is often another term for "never" :-)

You got me. :D
Let me say it this way, I don't plan to work on this for now and in the future (maybe this changes one day).

> 
> Still I'm not an evangelist of tracing, but the current "write everything to the log" is also far from perfect and hard to use.

Sure, but that's not the only thing. :)

But still, since the tracing is not operational at the moment and if would be fixed only with little value (IMHO), this is more a clean up. If anybody one day implements better tracing in M2E the `.options` files can be recreated and adjusted.
So I still think this is ok to merge. :)

HannesWell:
> But still, since the tracing is not operational at the moment and if it would be fixed only with (IMHO) little value, this is more a clean up. If anybody one day implements better tracing in M2E the `.options` files can be recreated and adjusted. So I still think this is ok to merge. :)

Since nobody objected I'm going to merge this removal of the not working tracing code and config.
If anybody want's implement a fixed and better crafted logging/tracing this is for sure welcome.

PR #836, opened 2022-07-03
Move M2E's default lifecycle-mapping-metadata into m2e.core plug-in

HannesWell:
The default lifecycle-mapping-metadata were located in `o.e.m2e.lifecyclemapping.defaults` and that plug-in served no other
purpose. Moving the default lifecycle-mapping-metadata.xml to m2e.core allows to remove that quite empty plug-in.

For reference the separate bundle was introduce in commit: c022f237d630ae17d8a0691dc21e007c0ff4c095

As a consequence the default `lifecycleMapping` for `packagingType` `war` is moved to m2e.jdt and since it uses the `org.eclipse.m2e.jdt.JarLifecycleMapping`. The mapping for `packagingType` `jar` resides there already so I think this is ok.


mickaelistria:
No objection from me. I also see it as a nice simplification.

PR #838, opened 2022-07-04
Bump remaining Plug-ins to version 2.0.0 and apply version ranges

HannesWell:
As discussed this bumps the versions of all Plug-ins in m2e.core to version 2.0.0 (except for the maven-runtime components, that use their own version).
Furthermore I applied the half open version range "[2.0.0,3.0.0) to all `Require-Bundle` dependencies in-between m2e plugins.

But for tests remove all dependency versions. For tests it is not crucial to have a version specified on a dependency since the available versions are usually controlled by the workspace TP. So specifying no version simplifies dependency updates.

@laeubi, @mickaelistria are you fine with that?


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "m2e-core-tests"]
 	path = m2e-core-tests
-	url = https://github.com/tesla/m2e-core-tests.git
-	branch = master
+	url = https://github.com/HannesWell/m2e-core-tests.git
```
HannesWell, line 3: This has to be reverted before submission!

mickaelistria:
I'm fine with that. However, one day we should consider using PDE API Tools to automate usage of the best range.

.gitmodules

HannesWell:
> I'm fine with that. However, one day we should consider using PDE API Tools to automate usage of the best range.

Yes absolutely. I think after the 2.0 release is a good point to start.

PR #843, opened 2022-07-06
Always configure JRE container to be the highest environment from

### LINKED ISSUE #842: Java version mismatch between JRE_CONTAINER and targetPlatform in M2E 2.0.0
steps to reproduce:

1. given sample project: https://github.com/mybatis/mybatis-3.git
2. open it in VS Code, with redhat.java extension 1.8.0 installed (which integrates m2e 2.0.0 snapshot).
3. after project imported, VS Code reports an error in `Item` record under package `org.apache.ibatis.submitted.record_type`.
> java.lang.Record cannot be resolved to a type.

as we know, record should be available in Java 16 but not in Java 8.

4. check the generated `.classpath` file, the JRE_CONTAINER is set to JavaSE-1.8 VM.
> classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"
5. check the generated `.settings/org.eclipse.jdt.prefs` file, the target platform is set to Java 16 (which would be the expect source level since it's defined in https://github.com/mybatis/mybatis-3/blob/9da3467e96a03e8bd6e040139581a22ce5297c29/pom.xml#L449-L459)
> org.eclipse.jdt.core.compiler.codegen.targetPlatform=16
org.eclipse.jdt.core.compiler.compliance=16
org.eclipse.jdt.core.compiler.source=16

if we rollback to redhat.java 1.6.0, which uses m2e 1.18.2, this issue disappears. The JRE_CONTAINER in .classpath file is JavaSE-16, and the value of pref `org.eclipse.jdt.core.compiler.codegen.targetPlatform` is also 16.

I think this issue might related to https://github.com/eclipse-m2e/m2e-core/pull/597 which introduces a new limitation from maven-enforcer-plugin. But I'm not sure it's caused by the plugin or m2e itself. When debugging, the two environments seem to mismatch:
image

cc: @fbricon @rgrunber

kwin:
m-compile-p and m-enforcer-p parametrisation

This closes #842

### DIFF
--- a/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
+++ b/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
@@ -22,6 +22,8 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -172,7 +174,8 @@ public abstract class AbstractJavaProjectConfigurator extends AbstractProjectCon
 
     String executionEnvironmentId = getExecutionEnvironmentId(options);
     String buildEnvironmentId = getMinimumJavaBuildEnvironmentId(request, monitor);
-    addJREClasspathContainer(classpath, buildEnvironmentId != null ? buildEnvironmentId : executionEnvironmentId);
+    // use the higher environment id as JREs are backwards compatible
+    addJREClasspathContainer(classpath, getHighestEnvironmentId(buildEnvironmentId, executionEnvironmentId));
 
     addMavenClasspathContainer(classpath);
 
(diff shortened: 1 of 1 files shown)


org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
```diff
@@ -198,6 +201,24 @@ public void configure(ProjectConfigurationRequest request, IProgressMonitor moni
     MavenJdtPlugin.getDefault().getBuildpathManager().updateClasspath(project, monitor);
   }
 
+  static String getHighestEnvironmentId(String... environmentIds) {
```
kwin, line 204: Where could I put a unit test for this (no IT necessary, no bundling into an Eclipse plugin either)? m2e.jdt.tests also only contains package `o.e.m.jdt.tests` and this test would need to be located in the same package.

kwin: The approach from https://stackoverflow.com/a/44926613 seems reasonable to me.

mickaelistria: o.e.m.jdt.tests would be the best place. As the class is internal you could make the method static.
Another approach, that m2e is not leveraging yet so it may require more work to get started, is https://github.com/eclipse/tycho/blob/master/RELEASE_NOTES.md#execute-unit-tests-with-eclipse-plugin-packaging

kwin:
As discussed in https://github.com/eclipse-m2e/m2e-core/issues/842#issuecomment-1177267445 different compiler settings for test and compile are not supported by Eclipse and having a lower enforced minimum Java version than compile target would be a bug in the `pom.xml` already there is no need for this approach.

PR #845, opened 2022-07-08
Try newer Wild Web Developer + LemMinX-Maven

mickaelistria:


target-platform/target-platform.target
```diff
@@ -34,7 +34,7 @@
 			<unit id="ch.qos.logback.slf4j" version="0.0.0"/>
 		</location>
 		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<repository location="https://download.eclipse.org/wildwebdeveloper/releases/0.13.5/"/><!--Keep in sync with repo-ref in org.eclipse.m2e.repository/category.xml-->
+			<repository location="https://download.eclipse.org/wildwebdeveloper/snapshots/"/><!--Keep in sync with repo-ref in org.eclipse.m2e.repository/category.xml-->
```
HannesWell, line 37: You have to update the referenced repo in the category.xml as well. Otherwise users have to install the snapshots separately.

mickaelistria: I'd rather avoid having snapshots leaking in categoy.xml because we're very likely to forget about them. Snapshots are for testers, and testers will probably have to add the Wild Web Developer snapshot p2 repo by themselves. I think it's safer than the risk of having Wild Web Developer snapshots installed on all m2e users' machine if we forget about it.

HannesWell: That's a point. 

I wonder if we could use variables for the specific URL that is defined at one Location (maybe the root pom.xml or the .mvn/maven.config?) and shared by the *.target and category.xml file and is understood in the build and the IDE.

laeubi: See https://github.com/eclipse-pde/eclipse.pde/issues/204

PR #847, opened 2022-07-09
Draft for review

karypid:


HannesWell:
What is the status here?
@laeubi do you want to review that as well?

laeubi:
Comment about `JdtPlugin` is still not solved...

mickaelistria:
Can you please retitle and add a PR message to tell what this change is intended to? Without such info, coming back to that ticket after a break is not so easy,

laeubi:
@karypid are you still interested in this or should we close it?

kwin:
This needs an update after https://github.com/eclipse-m2e/m2e-core/pull/1137.

laeubi:
CLsoing because of no response from original autor, @karypid let us know if you want to catch up on this any time later, it would probabbly easier to just start fresh from master.

Issue #274, opened 2021-07-13
Bug 534772 - m2e does not involve project's maven.config file when building the effective POM


BrannanH:
[See original bugzilla issue](https://bugs.eclipse.org/bugs/show_bug.cgi?id=534772) and [a related messaged](https://www.eclipse.org/lists/m2e-users/msg05711.html).

Maven allows arbitrary command line instructions to be specified in `.mvn/maven.config` [as detailed here.](https://maven.apache.org/configure.html#mvn-maven-config-file)

As in the above, this file should be able to be used to specify any arbitrary properties allowing various properties to be parameterised in POM files, and for instructions to be invoked by default when running maven commands.

One particular application of this is to externalise versions when using CI-friendly maven versions, and as such it can seen mentioned on [this page](https://maven.apache.org/maven-ci-friendly.html). The properties supported through `.mvn/maven.config` shouldn't be limited to `revision`, `sha1` and `changelist` however. Arbitrary properties and command line instructions should be supported as if manually specified in a POM, and/or as command line arguments.

mickaelistria:
Mostly, such extra properties should be added to the ProjectBuildingRequest used by `MavenImpl.readProject()` and similar.
It shouldn't be too hard to implement, at least partly. Feel free to provide a PR about that.

mickaelistria:
I don't think m2e should fix it. The proper project model is the pom.xml; default value for properties is expected to be set directly in pom.xml. The maven.config are not parameters for the project model per se, but more parameters for the CLI execution, which is a different topic.
To get what you want, just use what Maven model does provide as more standard about it: defining default values for the properties in pom.xml.

laeubi:
Requires
- https://github.com/eclipse-m2e/m2e-core/issues/546

I think we should consider this as the `maven.config` can contain other useful hints and will be the only option to use when configuring extension versions.

laeubi:
This should work now, please open specific bugs with a runnable example or better an junittest.

PR #848, opened 2022-07-09
Move the tycho-version into the maven.config and unify build arguments

HannesWell:
This avoids duplicated specification of build parameters in the different builds.

However I'm a bit concerned because now if one runs the build locally test failures and errors do not let the build fail and one has to look it up on the console.

Alternatively we could use --fail-at-end, but then only projects are build that don't depend on the failed once. On the other hand, a test project should not have dependents? So probably that would be the better option.

### DIFF
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -2,6 +2,7 @@
 --errors
 --update-snapshots
 --no-transfer-progress
--Dmaven.test.error.ignore=true
--Dmaven.test.failure.ignore=true
+--fail-at-end
+-Dtycho.localArtifacts=ignore
 -Dtycho.target.pomDependencies=consider
+-Dtycho-version=4.0.4
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -43,12 +43,12 @@ jobs:
     - name: Build m2e-core
       uses: coactions/setup-xvfb@b6b4fcfb9f5a895edadc3bc76318fae0ac17c8b3 # v1.0.1
       with:
-       run: mvn clean verify -Pits -Dtycho.p2.baselineMode=failCommon --batch-mode
+       run: mvn clean verify --batch-mode -Pits -Dtycho.p2.baselineMode=failCommon -Dmaven.test.error.ignore=true -Dmaven.test.failure.ignore=true
     - name: Upload Test Results
       uses: actions/upload-artifact@5d5d22a31266ced268874388b861e4b58bb5c2f3 # v4.3.1
       with:
         name: test-results-${{ matrix.os }}
-        if-no-files-found: error
+        if-no-files-found: warn
         path: |
           ${{ github.workspace }}/**/target/surefire-reports/*.xml
   event_file:
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -3,6 +3,6 @@
   <extension>
     <groupId>org.eclipse.tycho</groupId>
     <artifactId>tycho-build</artifactId>
-    <version>4.0.4</version>
+    <version>${tycho-version}</version>
   </extension>
 </extensions>
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -42,7 +42,7 @@ pipeline {
 				withCredentials([string(credentialsId: 'gpg-passphrase', variable: 'KEYRING_PASSPHRASE')]) {
 				xvnc(useXauthority: true) {
 					sh '''
-						mavenArgs="clean verify -Dtycho.p2.baselineMode=failCommon --batch-mode"
+						mavenArgs="clean verify --batch-mode -Dmaven.test.error.ignore=true -Dmaven.test.failure.ignore=true -Dtycho.p2.baselineMode=failCommon"
 						if [[ ${BRANCH_NAME} == master ]] || [[ ${BRANCH_NAME} =~ m2e-[0-9]+\\.[0-9]+\\.x ]]; then
 							mvn ${mavenArgs} -Peclipse-sign,its -Dgpg.passphrase="${KEYRING_PASSPHRASE}" -Dgpg.keyname="011C526F29B2CE79"
 						else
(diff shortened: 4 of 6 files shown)


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "m2e-core-tests"]
 	path = m2e-core-tests
-	url = https://github.com/tesla/m2e-core-tests.git
-	branch = master
+	url = https://github.com/HannesWell/m2e-core-tests.git
```
HannesWell, line 3: This has to be reverted before submission.

HannesWell:
Oh and it looks like M2E does not consider the `maven.config` and therefore shows many errors like this 
`org.eclipse.tycho:tycho-source-plugin:jar:${tycho-version} was not found ...`

I think this is https://github.com/eclipse-m2e/m2e-core/issues/274 and should not be too hard now?

.gitmodules

mickaelistria (dismissed):
I don't think it's actually an improvement. Many settings in the script are fine for CI but undesired for a local build (because the test failure/error needs to be visible on a local build).

HannesWell:
> I don't think it's actually an improvement. Many settings in the script are fine for CI but undesired for a local build (because the test failure/error needs to be visible on a local build).

That's right, but therefore I replaced the ignore.failure/error properties by '--fail-at-end' which seems suitable in both cases.

.mvn/maven.config
```diff
@@ -4,3 +4,6 @@
 --no-transfer-progress
 -Dmaven.test.error.ignore=true -Dmaven.test.failure.ignore=true
 -Dtycho.resolver.classic=false
+--fail-at-end
+-Dtycho-version=4.0.0-SNAPSHOT
+-s setup/tychoSnapshotSettings.xml -U
```
laeubi, line 9: please use one line for one command, Maven 3.9 will not support having everything in one line anymore.

laeubi:
FYI I have adjusted the build here:
- https://github.com/eclipse-m2e/m2e-core/pull/1285

so maybe the both approaches should be merged.

.mvn/maven.config
HannesWell: Done.

HannesWell:
> so maybe the both approaches should be merged.

Merged your changes to and move the specification of the maven.test.error/failure.ignore properties to the CI scripts so that local builds can fail.

HannesWell:
@laeubi I lost a bit track of your work in regards to support properties in the maven.config and are therefore not sure if it is supposed to work now? When I prepared the update to this PR m2e did not complain in my m2e-dev workspace so I assume this can be merged as soon as we agree upon it?

laeubi:
Status is it should work in all relevant cases but not sure about leminx and maybe there are still some areas uncovered :-D

HannesWell:
> Status is it should work in all relevant cases but not sure about leminx and maybe there are still some areas uncovered :-D

OK, I can get my workspace into a state where at least m2e shows no errors about a missing `tycho.version` property.
But as soon as I open a pom, the language server complains about a missing version.

@mickaelistria could you help add support for maven.configs in the language server?
Would it be sufficient to adjust the `InitializationOptionsProvider` respectively the `MavenRuntimeClasspathProvider`?
I belive some parts of the magic could happen in.
https://github.com/eclipse-m2e/m2e-core/blob/69a6425a9b638e1ced2b711cb902db89cb88dba1/org.eclipse.m2e.editor.lemminx/src/org/eclipse/m2e/editor/lemminx/MavenRuntimeClasspathProvider.java#L57-L59

HannesWell:
Now that we have https://github.com/eclipse-m2e/m2e-core/pull/1692, we can finally use this.
When trying this out locally I noticed that changes in the maven.config are not immediately picked up by the LS (at least when I hover over the tycho-version property the old value is present).

In m2e internally it seems to work, at least the effective pom shows the updated version.

laeubi:
You need to edit the file to get the changes reflected I think, or the LS must watch the file for changes but as one no really change this very often I think it is acceptable and the same will happen if you for example change a parent pom file that is currently not opened as a file in the editor.

@mickaelistria is there a way m2e can tell the language server to flush its cache?

mickaelistria:
>  is there a way m2e can tell the language server to flush its cache?

No, and as mentioned in previous discussion, the goal is that m2e decides smartly and automatically when to flush its cache by monitoring all the files that participate to resolution (including maven.config) and flush the cache on changes, like it already does for pom.xml files; without involving the client.

laeubi:
> No, and as mentioned in previous discussion, the goal is that m2e decides smartly and automatically when to flush its cache by monitoring all the files that participate to resolution (including maven.config) and flush the cache on changes

I assume you mean lemminx? Where is this "automatic watching" implemented in lemminx? The main point is if two components watching files this double work can maybe avoided.

mickaelistria:
Indeed, I mean lemminx*=maven*. There is no automatic watching implemeted in lemminx-maven at the moment, because LSP sends notifications for changed files that the clients configures as supported by this LS (currently only pom.xml files).

> The main point is if two components watching files this double work can maybe avoided.

lemminx-maven work at best on its own without m2e. If this means duplicated work from m2e to lemminx-maven, then it's worth it.

HannesWell:
> > The main point is if two components watching files this double work can maybe avoided.
> 
> lemminx-maven work at best on its own without m2e. If this means duplicated work from m2e to lemminx-maven, then it's worth it.

If there is a way to share things between m2e and the LSP I would consider it an advantage if this means less CPU-cycles are burned on a users computer and less memory is used. And having duplicated implementations is something I personally consider relatively bad and would only choose as last or at least later resort. I understand that the LemMinX-LSP want's to be standalone, but I still see an advantage in the things said before. Probably a good compromise has to be found.
I guess m2e can somehow poke the LSP using the approach asked in https://github.com/eclipse-m2e/m2e-core/pull/848#issuecomment-1657242453, can't it?

Anyways I think it would be best to discuss enhancements of the LSP integration in a dedicated ticket.

Now that https://github.com/eclipse-m2e/m2e-core/pull/1692 is integrated and tested, we can finally eat our own dog-food and use this in m2e.

PR #849, opened 2022-07-09
Remove legacyMavenProjects cache from ProjectRegistryManager

HannesWell:
That legacy cache was introduce with commit 561d6fbfd7c132dcdc29f4bd610aa2a590c498c7.
This was 9 years ago and I think with a major release like 2.0 it is a good opportunity to get rid of that heritage.

Does anybody know if there are more places where support for legacy usage outside the scope of a `MavenExecutionContext` exists?

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -17,6 +17,11 @@ The legacy textual editor is removed as it doesn't receive enough update/mainten
 
 m2e-apt plugins that were so far included in JBoss Tools were migrated into m2e and are shipped by default with m2e installations. So annotation processing should be better supported out of the box.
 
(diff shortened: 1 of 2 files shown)


mickaelistria (approved):

HannesWell:
@laeubi you suggested to remove this elsewhere too, so am I right assumimg you are finde with this?

laeubi:
Sure I have no clue for what this is/was useful, so if everything works it seems obsolete.

HannesWell:
> Sure I have no clue for what this is/was useful, so if everything works it seems obsolete.

I also only know what's in the removed doc.

Thank you to you two. I'll merge this if all builds are green.

PR #850, opened 2022-07-09
Make ProjectRegistryManager's project-cache size configurable

HannesWell:
... and increase the default size to 20.

As requested in https://github.com/eclipse-m2e/m2e-core/pull/507#issuecomment-1086624238 and https://github.com/eclipse-m2e/m2e-core/issues/157 this make the internal `ProjectRegistryManager` cache size configurable and increases the default size of the cache from 5 to 20.

With this one can adjust the cache size to the desired value by adding a VM argument `-Dm2e.project.cache.size=<desired-value>` for example the the `eclipse.ini`.

@mickaelistria, @laeubi any objections?

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
@@ -1128,7 +1128,7 @@ public class ProjectRegistryManager implements ISaveParticipant {
   }
 
   private Map<MavenProjectFacade, MavenProject> createProjectCache() {
-    int maxCacheSize = 5;
+    int maxCacheSize = Integer.getInteger("m2e.project.cache.size", 20);
     return Collections.synchronizedMap(new LinkedHashMap<>(maxCacheSize * 4 / 3 + 1, 0.75f, true) {
       private static final long serialVersionUID = 8022606648487974598L;
 


laeubi:
No objections, but as discussed before, a user configurable preference would be more useful

HannesWell:
> No objections, but as discussed before, a user configurable preference would be more useful

Indeed. I hoped I can be lazy here.
I can update this when I'm back from vacation after Friday.
We then also need a preference listener to react on changes in the cache size.

mickaelistria:
> a user configurable preference would be more useful

I'm not so certain about that. Best tools don't require users to deal with complex settings to work decently, keeping the setting hidden is actually more likely to make m2e behave better and better than letting users override it to some higher value and assume that this s not a bug.
Yet it is a performance bug and there would be way to get much better results with a relatively small cache.

So +1 to increase the value a bit, as it is nowadays very probably that a user has 20+ unrelated Maven projects in the workspace. But -1 for making a user facing preference a requirement here.

HannesWell:
> > a user configurable preference would be more useful
> 
> I'm not so certain about that. Best tools don't require users to deal with complex settings to work decently, keeping the setting hidden is actually more likely to make m2e behave better and better than letting users override it to some higher value and assume that this s not a bug. Yet it is a performance bug and there would be way to get much better results with a relatively small cache.

I fully agree that it would be better to avoid the need for this tuning Parameter.
However I just created this PR because I remembered the requests in the mentioned issued/PR and I thought I provide this simple fix, since the requstors obviously preferred to maintain their own patch than providing a PR.
I was not yet affected by the problem.

famod:
Nice! Any clue when this is gonna be released? Thanks!

PR #852, opened 2022-07-10
Consider maven.config in multimodule root

HannesWell:
Fixes https://github.com/eclipse-m2e/m2e-core/issues/274

With this PR properties and profiles specified in the `.mvn/maven.config` are considered within the M2E workspace build and one can use version properties in `.mvn/extension.xml`.

As far as I can tell this works, but at the moment I'm struggling because `MavenExecutionContext.newProjectBuildRequst()` is called for a context that does not have `basedir` and context project and thus does cannot copy the user-settings and profiles to the request, which has the consequence that the variables cannot be resolved.

Furthermore a test case has to be added.

I'm on vacation until Friday and then can continue to work on this one.
Maybe @laeubi or @mickaelistria you have a hint how to solve the mentioned problem? I suspect it is rooted in how the `ProjectRegistryManager` creates execution contexts, but that is just my first guess.

### DIFF
--- a/org.eclipse.m2e.core.tests/resources/projects/mavenConfig/.mvn/extensions.xml
+++ b/org.eclipse.m2e.core.tests/resources/projects/mavenConfig/.mvn/extensions.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<extensions>
+	<extension>
+		<groupId>org.eclipse.tycho</groupId>
+		<artifactId>tycho-build</artifactId>
+		<version>${tycho-version-for-extension-test}</version>
+	</extension>
+</extensions>
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2013, 2022 Igor Fedorenko
+ * Copyright (c) 2013, 2023 Igor Fedorenko and others
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
@@ -10,11 +10,10 @@
  * Contributors:
  *      Igor Fedorenko - initial API and implementation
  *      Christoph Läubrich - move creation MavenExecutionRequest from MavenImpl->MavenExecutionContext
+ *      Hannes Wellmann - Add support for properties and profile-activations from .mvn/maven.config
  *******************************************************************************/
 
 package org.eclipse.m2e.core.internal.embedder;
-
-import static org.eclipse.m2e.core.internal.M2EUtils.copyProperties;
 
 import java.io.File;
 import java.util.ArrayDeque;
@@ -44,6 +43,7 @@ import org.apache.maven.Maven;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.InvalidRepositoryException;
 import org.apache.maven.artifact.repository.ArtifactRepository;
+import org.apache.maven.cli.MavenCLIParser;
 import org.apache.maven.cli.configuration.SettingsXmlConfigurationProcessor;
 import org.apache.maven.eventspy.internal.EventSpyDispatcher;
 import org.apache.maven.execution.DefaultMavenExecutionRequest;
@@ -62,7 +62,6 @@ import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.project.DefaultProjectBuildingRequest;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.project.ProjectBuildingRequest;
-import org.apache.maven.properties.internal.EnvironmentUtils;
 import org.apache.maven.repository.RepositorySystem;
 import org.apache.maven.session.scope.internal.SessionScope;
 import org.apache.maven.settings.Settings;
@@ -147,19 +146,23 @@ public class MavenExecutionContext implements IMavenExecutionContext {
         throw new IllegalStateException(); //
       }
       request = DefaultMavenExecutionRequest.copy(parent);
+      request.setMultiModuleProjectDirectory(parent.getMultiModuleProjectDirectory());
     }
-    if(request == null) {
-      request = createExecutionRequest(IMavenConfiguration.getWorkspaceConfiguration(),
-          containerLookup, MavenPlugin.getMaven().getSettings());
+    File multiModuleRoot = MavenProperties.computeMultiModuleProjectDirectory(basedir);
+    if(multiModuleRoot == null) {
+      multiModuleRoot = multiModuleProjectDirectory;
+    }
+    if(request == null
+        || (multiModuleRoot != null && !multiModuleRoot.equals(request.getMultiModuleProjectDirectory()))) {
+      request = createExecutionRequest(IMavenConfiguration.getWorkspaceConfiguration(), containerLookup,
+          MavenPlugin.getMaven().getSettings(), multiModuleRoot);
       request.setBaseDirectory(basedir);
     }
-    request.setMultiModuleProjectDirectory(multiModuleProjectDirectory);
-
     return request;
   }
 
   static MavenExecutionRequest createExecutionRequest(IMavenConfiguration mavenConfiguration, IComponentLookup lookup,
-      Settings settings) throws CoreException {
+      Settings settings, File multiModuleProjectDirectory) throws CoreException {
     MavenExecutionRequest request = new DefaultMavenExecutionRequest();
 
     // this causes problems with unexpected "stale project configuration" error markers
@@ -197,10 +200,11 @@ public class MavenExecutionContext implements IMavenExecutionContext {
     request.setLocalRepositoryPath(localRepository.getBasedir());
     request.setOffline(mavenConfiguration.isOffline());
 
-    request.getUserProperties().put("m2e.version", MavenPluginActivator.getVersion()); //$NON-NLS-1$
-    request.getUserProperties().put(ConfigurationProperties.USER_AGENT, MavenPluginActivator.getUserAgent());
-
-    MavenExecutionContext.populateSystemProperties(request);
+    request.setMultiModuleProjectDirectory(multiModuleProjectDirectory);
+    Properties userProperties = request.getUserProperties();
+    userProperties.put("m2e.version", MavenPluginActivator.getVersion()); //$NON-NLS-1$
+    userProperties.put(ConfigurationProperties.USER_AGENT, MavenPluginActivator.getUserAgent());
+    MavenCLIParser.populateCLIOptions(request);
 
     request.setCacheNotFound(true);
     request.setCacheTransferError(true);
@@ -246,7 +250,6 @@ public class MavenExecutionContext implements IMavenExecutionContext {
       } catch(MavenExecutionRequestPopulationException ex) {
         throw new CoreException(Status.error(Messages.MavenImpl_error_read_config, ex));
       }
-      populateSystemProperties(request);
       setValue(CTX_LOCALREPOSITORY, request.getLocalRepository());
       final FilterRepositorySystemSession repositorySession = createRepositorySession(request,
           MavenPlugin.getMavenConfiguration(), lookup);
@@ -417,18 +420,6 @@ public class MavenExecutionContext implements IMavenExecutionContext {
   public static MavenExecutionContext getThreadContext(boolean innermost) {
     final Deque<MavenExecutionContext> stack = threadLocal.get();
     return stack != null ? (innermost ? stack.peekFirst() : stack.peekLast()) : null;
-  }
-
-  public static void populateSystemProperties(MavenExecutionRequest request) {
-    // temporary solution for https://issues.sonatype.org/browse/MNGECLIPSE-1607
-    // oddly, there are no unit tests that fail if this is commented out
-    if(request.getSystemProperties() == null || request.getSystemProperties().isEmpty()) {
-      Properties systemProperties = new Properties();
-      EnvironmentUtils.addEnvVars(systemProperties);
-      copyProperties(systemProperties, System.getProperties());
-      MavenProperties.setProperties(systemProperties);
-      request.setSystemProperties(systemProperties);
-    }
   }
 
   /*
--- a/m2e-maven-runtime/org.eclipse.m2e.maven.runtime/src/main/java/org/apache/maven/cli/MavenCLIParser.java
+++ b/m2e-maven-runtime/org.eclipse.m2e.maven.runtime/src/main/java/org/apache/maven/cli/MavenCLIParser.java
@@ -1,0 +1,72 @@
+/*******************************************************************************
+ * Copyright (c) 2022, 2023 Hannes Wellmann and others
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *      Hannes Wellmann - initial API and implementation
+ *******************************************************************************/
+
+package org.apache.maven.cli;
+
+import java.io.File;
+import java.util.regex.Pattern;
+
+import org.apache.commons.cli.CommandLine;
+import org.apache.maven.execution.MavenExecutionRequest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class MavenCLIParser {
+	private MavenCLIParser() {
+	}
+
+	private static final Logger LOGGER = LoggerFactory.getLogger(MavenCLIParser.class);
+
+	public static void populateCLIOptions(MavenExecutionRequest request) {
+		try {
+			File multiModuleProjectDirectory = request.getMultiModuleProjectDirectory();
+			CommandLine cl;
+			CliRequest cliRequest = new CliRequest(new String[0], null);
+			if (multiModuleProjectDirectory != null) {
+				cliRequest.multiModuleProjectDirectory = multiModuleProjectDirectory;
+				new MavenCli().cli(cliRequest);
+				cl = cliRequest.commandLine;
+			} else {
+				cl = new CLIManager().parse(cliRequest.getArgs());
+			}
+			MavenCli.populateProperties(cl, request.getSystemProperties(), request.getUserProperties());
+
+			populateProfilesProfiles(request, cl);
+		} catch (Exception e) {
+			LOGGER.error("Failed to populate request for", e);
+		}
+	}
+
+	private static final Pattern COMMA = Pattern.compile(",");
+
+	private static void populateProfilesProfiles(MavenExecutionRequest request, CommandLine mavenConfig) {
+		// Based on MavenCli.populateRequest() section 'Profile Activation'
+		String[] profileOptionValues = mavenConfig.getOptionValues(CLIManager.ACTIVATE_PROFILES);
+		if (profileOptionValues != null) {
+			for (String profileValue : profileOptionValues) {
+				COMMA.splitAsStream(profileValue).map(String::trim).forEach(profileAction -> {
+					if (profileAction.startsWith("-") || profileAction.startsWith("!")) {
+						request.addInactiveProfile(profileAction.substring(1));
+					} else if (profileAction.startsWith("+")) {
+						request.addActiveProfile(profileAction.substring(1));
+					} else {
+						request.addActiveProfile(profileAction);
+					}
+				});
+			}
+		}
+	}
+
+	// TODO: parse more?
+
+}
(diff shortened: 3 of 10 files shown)


laeubi:
This is not a real "issue" because anonymous executions are always possible, but if you have an extensions there should also be a basedir. So if no basedir you can not have any a `maven.config` so maybe I'm getting something wrong?

HannesWell:
> This is not a real "issue" because anonymous executions are always possible, but if you have an extensions there should also be a basedir. So if no basedir you can not have any a `maven.config` so maybe I'm getting something wrong?

That's right, but I had the impression that the RegistryManager uses anonymus 'too' often. But maybe I just have to parse the maven.config again.
But I'ill investigate that when I'm back.

laeubi:
I have not checked the code, but "extensions" are bound to a `IMavenFacade` so if you can start the execution from there you should get a `basedir`.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
```diff
@@ -114,7 +126,16 @@ public MavenExecutionContext(IComponentLookup lookup, IMavenProjectFacade projec
       }
       this.basedir = basedir == null ? null : (basedir.isDirectory() ? basedir : basedir.getParentFile());
     } else {
-      this.basedir = null;
+      ClassWorld world = null;
```
laeubi, line 129: This looks suspicious what is this good for?

laeubi:
This seems to add a lot of stuff to the execution context regarding configuration parsing, maybe we better extract this into some dedicated class?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
HannesWell: In same cases, for example `ProjectRegistryManager.readProjectsWithDependencies()` the context is created without a project facade but the lookUp has a basedir and is not the global `M2EClassWorld`.
Because the context's basedir is later used to search the `multiModuleProjectDirectory` it is important to get that basedir. Otherwise the content of the `.mvn` folder is not considered. Unfortunately the result of that call is used in the workspace build.

Of course that could be done when creating the execution request and the basedir of the context is null.

grafik

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
```diff
@@ -132,25 +153,30 @@ public MavenExecutionRequest getExecutionRequest() throws CoreException {
   protected MavenExecutionRequest newExecutionRequest() throws CoreException {
     MavenExecutionRequest request = null;
     Deque<MavenExecutionContext> stack = threadLocal.get();
+    File parentMultiModuleProjectDirectory = null;
```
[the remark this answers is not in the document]
HannesWell: I'm not sure anymore why I added that part. Maybe for the case of a nested `.mvn` folder?
I have to check it.

HannesWell:
> This seems to add a lot of stuff to the execution context regarding configuration parsing, maybe we better extract this into some dedicated class?

Yes sure, no problem.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
laeubi: I think we should pass the basedir as a constructor parameter as well instead of read it from the project facade indirectly.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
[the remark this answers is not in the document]
laeubi: It might be that there are still places that do use the "global" maven where the should have used a project scoped (or basedir scoped) execution instead. I think I already have catched some but not all, as I mostly focused o getting to a working state again. SO if you find out where this was required we can take a look if there are better ways to solve it.

In general one can say there are three cases:

1. whenever you can get an IMavenFacade --> Use it for execution as you most probably want project scope execution
2. whenever you do anything "external" (e.g. as we do in inital project setup or when reading external-to-the-workspace poms) create a basedir scoped execution (IMavenExecution.of(...))
3. when there is the rare case you just want do something "global" that is not bound to a project or directory (e.g. maven targets just want fetch artifacts using the global maven config) you can use IMaven as the global context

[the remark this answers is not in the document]
HannesWell: That makes sense.
This list should definitely be documented somewhere.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
HannesWell: > I think we should pass the basedir as a constructor parameter as well instead of read it from the project facade indirectly.

I have reworked the the constructors, but nevertheless I was not able to completely get rid of that indirect retrieval, although it is in more limited scope now.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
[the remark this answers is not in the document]
HannesWell: > I'm not sure anymore why I added that part. Maybe for the case of a nested `.mvn` folder? I have to check it.

I checked it and that is necessary when a execution request is created for a nested context, i.e. when a 'surrounding' context is already on the stack. Then the parent is just copied and often has a null `basedir/multiModuleProjectDir`.

This happens for example when importing projects. Then within the execution of a global Maven context a context with basedir is created. But since there is already a context on the stack, that parent context's request is copied, but that does not have basedir specific settings.

But maybe the check should be changed from `if(!Objects.equals(multiModuleProjectDirectory, parentMultiModuleProjectDirectory))` to `if(parentMultiModuleProjectDirectory==null)`.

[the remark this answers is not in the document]
laeubi: I'm not sure if nesting is the right approach then here. Nesting only is useful if you want to fully inherit the context of the caller (e.g. you perform a generic method) but I'm not sure if it is actually useful at all here, so the question is why there is a surrounding context at all and if it could not be replaced with one of the other alternatives (or most probably must!).

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
laeubi: Then we maybe need to refactor tings here, I thing the execution should always be well defined in regard to lookup + basedir. I'll take a look and maybe try to prepare an extra PR as a prerequisite to this.

HannesWell: Thanks for #1027 this part is now resolved.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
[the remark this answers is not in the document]
HannesWell: I reduced the number of cases where this happened by also copying the `multiModuleProjectDirectory` from the original to the copy when a parent is present. For whatever reason `DefaultMavenExecutionRequest.copy()` does not copy that.

However that case still happens when for example opening a closed project that has a `.mvn/maven.config`.
And as long as the thread-local context stack principle is used, I think it actually makes sense in such cases to copy the outer context and then loads again the properties, etc. from the specific `.mvn/maven.config` file for the inner context.
And such cases happen for example on project-refresh that is started with the global context.

Of course we could refactor the `ProjectRegistryRefreshJob` to only use basedir aware contexts, maybe even project specific contexts and should check for other locations where that could be a problem, that logic should remain.
And I don't know if it can happen, that within the execution of some basedir-aware context another context with a deeper nested basedir can be created in corresponding scenarios.
If that can happen we should keep that otherwise it could be hard for users to do it right.

HannesWell:
In general I noticed that with this change changes to the maven.config in the IDE should lead to the same model/project reload events like if all using projects have been modified.
Does anybody know where to start this?

@mickaelistria could you please give me some advice/help with the Language-Server problem before in https://github.com/eclipse-m2e/m2e-core/pull/852#issuecomment-1298613428?

laeubi:
It seems we are making some progress here :-)

You can use a ResourceChangelistener if it is okay to assume the file is in the IDE, or a filewatcher, but I think it is not crucial here as one usually don't change the file often and the this might be a case where a manual update is acceptable.

That said, m2e uses timestamps in `MavenProjectFacade` for the settings and the pom. I think one can enhance this to take Multimoduledirectory into account by using the latest modification stamp from any file in the .`mvn` folder.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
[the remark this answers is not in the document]
laeubi: > For whatever reason DefaultMavenExecutionRequest.copy() does not copy that.

Most probably it was forgotten and you probably want to propose a fix to maven.

> And as long as the thread-local context stack principle is used, I think it actually makes sense in such cases to copy the outer context and then loads again the properties, etc. from the specific .mvn/maven.config file for the inner context.

To be honest I'm not convinced. Mixing global with project context seems not right, its not only that directory, its also the container lookup and the class worlds are different.

> And such cases happen for example on project-refresh that is started with the global context.

Then this must be fixed (e.g. with a separate PR) similar to how the initial reading of projects is performed by grouping the projects and container appropriately.

> Of course we could refactor the ProjectRegistryRefreshJob

Actually this should be done. I think the intial idea here is/was that because refresh is maybe called multiple times, we can reuse the session and thus using a stacked execution. This won't work well with the new concept and one would either need to accepts that sessions are created multiple times (I think that's really fast) and let the manager do the grouping instead.

laeubi:
@HannesWell do you like to finish this soon or should I try to take over? I think everything is now prepared at m2e and this is ist the missing piece for further enhancements.

m2e-maven-runtime/org.eclipse.m2e.maven.runtime/pom.xml
```diff
@@ -163,6 +163,7 @@
 								META-INF.plexus;-noimport:=true;x-internal:=true,\
 								META-INF.sisu;-noimport:=true;x-internal:=true,\
 								org.apache.maven.*;provider=m2e;mandatory:=provider,\
+								org.apache.commons.cli;provider=m2e;mandatory:=provider,\
```
laeubi, line 166: does maven uses common.cli internally? I can't find a reference in your code about it.

HannesWell: It used to be used in `MavenConfigParser`. But I moved the 'Parser' to the m2e.maven.runtime. This way it is possible to access package private methods from MavenCLI and to reuse more of its code.
Then this package does not have to be exported as well.

org.eclipse.m2e.core.tests/resources/projects/mavenConfig/.mvn/extensions.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<extensions>
+	<extension>
+		<groupId>org.eclipse.tycho</groupId>
+		<artifactId>tycho-build</artifactId>
+		<version>${tycho-version}</version>
```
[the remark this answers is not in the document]
HannesWell: Sounds good. Will do that.

HannesWell:
> @HannesWell do you like to finish this soon or should I try to take over? I think everything is now prepared at m2e and this is ist the missing piece for further enhancements.

Yes I just updated it to the master branch and also improved the settings parser to reuse more from the MavenCLI. In this step I maybe also achieved to consider not only properties and profiles but also setting files specified in the maven.config. But I have to verify if that works.

I did not yet have the time to check if the properties adjustments for nested context in `MavenExecutionContext.newExecutionRequest()` is now obsolete in all cases. But I will continue that tomorrow.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
```diff
@@ -147,19 +146,24 @@ protected MavenExecutionRequest newExecutionRequest() throws CoreException {
...
     }
+    File multiModuleProjectDirectory = PlexusContainerManager.computeMultiModuleProjectDirectory(basedir);
     if(request == null) {
-      request = createExecutionRequest(IMavenConfiguration.getWorkspaceConfiguration(),
-          containerLookup, MavenPlugin.getMaven().getSettings());
+      request = createExecutionRequest(IMavenConfiguration.getWorkspaceConfiguration(), containerLookup,
+          MavenPlugin.getMaven().getSettings(), multiModuleProjectDirectory);
       request.setBaseDirectory(basedir);
+    } else if(!Objects.equals(multiModuleProjectDirectory, request.getMultiModuleProjectDirectory())) {
+      //TODO: this should have been removed, shouldn't it?
```
laeubi, line 157: I think this should be removed. If there are any obstacles, this is most likely because of a wrong call hierarchy and we should then fix this there.

laeubi (dismissed):
Beside the `TODO` this looks good and if the tests suceed I think it should be fine to start with this!

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
HannesWell: See this if still being hit, for example in:

```
Thread [Worker-8: Updating Maven Dependencies] (Suspended (breakpoint at line 159 in MavenExecutionContext))	
	MavenExecutionContext.newExecutionRequest() line: 159	
	MavenExecutionContext.getExecutionRequest() line: 135	
	ProjectRegistryManager.readMavenProjectFacades(Collection<IFile>, MutableProjectRegistry, IProgressMonitor) line: 742	
	ProjectRegistryManager.refresh(MutableProjectRegistry, DependencyResolutionContext, IProgressMonitor) line: 380	
	ProjectRegistryManager.refresh(MutableProjectRegistry, Collection<IFile>, IProgressMonitor) line: 360	
	ProjectRegistryRefreshJob.lambda$4(MutableProjectRegistry, Set, IMavenExecutionContext, IProgressMonitor) line: 154	
	0x0000000801941ef0.call(IMavenExecutionContext, IProgressMonitor) line: not available	
	MavenExecutionContext.executeBare(MavenProject, ICallable<V>, IProgressMonitor) line: 375	
	MavenExecutionContext.execute(MavenProject, ICallable<V>, IProgressMonitor) line: 279	
	MavenExecutionContext.execute(ICallable<V>, IProgressMonitor) line: 219	
	MavenImpl.execute(IMaven, boolean, boolean, ICallable<V>, IProgressMonitor) line: 1102	
	ProjectRegistryRefreshJob.lambda$0(MutableProjectRegistry, List, IProgressMonitor, IMavenExecutionContext, IProgressMonitor) line: 153	
	0x0000000801941cd0.call(IMavenExecutionContext, IProgressMonitor) line: not available	
	MavenExecutionContext.executeBare(MavenProject, ICallable<V>, IProgressMonitor) line: 375	
	MavenExecutionContext.execute(MavenProject, ICallable<V>, IProgressMonitor) line: 279	
	MavenExecutionContext.execute(ICallable<V>, IProgressMonitor) line: 219	
	ProjectRegistryRefreshJob.run(IProgressMonitor) line: 106	
	Worker.run() line: 63	
```

HannesWell: But I'm still not sure if this is really so bad. Narrowing from a 'global' to a 'mvn-folder'-scoped context seems not too bad in general. But I have not yet fully thought this through.

laeubi: Actually it *is* bad, because that means you are probabbly executing with the wrong settings and/or container, `ProjectRegistryManager` should actually already group projects by their config/basedir.

Also think about that currently the "subcontext" modify the session from the outer one, but never restores this! 
So probably if we think this shoudl be supported, then we probably need to do it a bit earlier, where it is checked if a parent context exist and first test there if the parent is compatible, if yes, we can join, if not just start a new execution.

HannesWell: > ProjectRegistryManager should actually already group projects by their config/basedir.

I'm not arguing about that, but if it does not work in this special case, isn't it then possible to not work in other cases as well? Should we all fix them? Therefore if possible I think it would be better to find a solution that hopefully works for all. 🤓 

> Also think about that currently the "subcontext" modify the session from the outer one, but never restores this!
So probably if we think this shoudl be supported, then we probably need to do it a bit earlier, where it is checked if a parent context exist and first test there if the parent is compatible, if yes, we can join, if not just start a new execution.

Indeed. Or could we create a new request, that can inherit values (it doesn't override) from the request of the parent?

laeubi: > Should we all fix them?

Absolutely! This is mandatory to fully support such advanced features, if you just use "anything" to execute your task your outcome will not be correct (as seen with the recent failure with the copy-resources) so even though it is some work I thin we cannot avoid it if we want things to work consistent.

laeubi:
I now made some experiments here:
- https://github.com/eclipse-m2e/m2e-core/pull/1182

but there are too much sideefects with that to easily resolve this. So I think we need to keep with that for now, and even though it might create inconsistent state when used wrong, a way would be to store the current settings/state and restore them when this run is leaved, so I think if thats implemented somehow and the test run we should go with this aproach for now as it is better than nothing at the moment and the impact won't be noticed by the user anyways.

HannesWell:
> I now made some experiments here:
> 
>     * [Directly pass offline/update status in ProjectRegistryRefreshJob #1182](https://github.com/eclipse-m2e/m2e-core/pull/1182)
> 
> 
> but there are too much sideefects with that to easily resolve this. So I think we need to keep with that for now, and even though it might create inconsistent state when used wrong, a way would be to store the current settings/state and restore them when this run is leaved, so I think if thats implemented somehow and the test run we should go with this aproach for now as it is better than nothing at the moment and the impact won't be noticed by the user anyways.

Understand. The current approach is to create a new request in case the parent request exists, but does not have a `MultiModuleProjectDirectory`. That would cover the narrowing case where one starts with a global context and then uses a scoped one.
In case a parent request exists but has a different `MultiModuleProjectDirectory` an exception is thrown, because I assume that is a case that should not occur or do you think that would be possible?

laeubi:
> In case a parent request exists but has a different MultiModuleProjectDirectory an exception is thrown, because I assume that is a case that should not occur or do you think that would be possible?

Can't it just be handled the same way? I don't think we must handle this in an asymetric way here unless it adds additional problems. because no `MultiModuleProjectDirectory` is equal to having one but with no `maven.config` file...

HannesWell:
> > In case a parent request exists but has a different MultiModuleProjectDirectory an exception is thrown, because I assume that is a case that should not occur or do you think that would be possible?
> 
> Can't it just be handled the same way? I don't think we must handle this in an asymetric way here unless it adds additional problems. because no `MultiModuleProjectDirectory` is equal to having one but with no `maven.config` file...

Actually yes, I change it that way.

HannesWell:
> > In case a parent request exists but has a different MultiModuleProjectDirectory an exception is thrown, because I assume that is a case that should not occur or do you think that would be possible?
> 
> Can't it just be handled the same way? I don't think we must handle this in an asymetric way here unless it adds additional problems. because no `MultiModuleProjectDirectory` is equal to having one but with no `maven.config` file...

Done.

Something seems to leak now. The `ExtensionsTest.testReloadExtensionAfterDeletion()` fails when executed with the other cases in the class but passes when executed alone... already at the first import...

laeubi:
I get the feeling that parsing this as part of the execution might not work that nice (even though first obvious choice here) but we better update the `IProjectConfiguration` (what already holds project specific profiles and properties) and then the update will already group projects appropriately.

The we have two cases:

1. It just works because m2e already act correctly
2. It shows some issues in this area, e.g. we have only one test that actually changes properties (and I can't find any UI code for this)

For the second case I think it would even be fine to merge this as it is already (partly) broken then and try to solve the issue with a testcase in a separate PR.

laeubi:
@HannesWell can you probably submit the **test-case** as a standalone PR? 
I'm currently work on this and like to include it in my PR with correct authorship.

laeubi:
I think we can close this for now as the test-case is on a seperate PR now!

Issue #854, opened 2022-07-12
Ignore lifecycle configuration dialog too small


kjsmita6:
When you select the quick-fix for "Plugin execution not covered...", and select "Mark goal as ignored in pom.xml, the popup looks like this:
![Screen Shot 2022-07-12 at 2 28 00 PM](https://user-images.githubusercontent.com/9405706/178567706-97c59b35-f34c-4cf0-838d-5b6537dfef58.png)

And the location to ignore isn't visible at all. It can still be selected but I could not determine any way to make this dialog bigger to make the text visible.

mickaelistria:
Would you be willing to submit a patch the fixes that?

kjsmita6:
![5ab7d599d9a0b2dc3880731d302ad319](https://user-images.githubusercontent.com/9405706/179382196-2ccfc18f-3c95-4add-a938-0778d7fcf9d4.png)
Was going to work on this issue this weekend on my Windows PC but funnily enough, the issue doesn't happen. So I guess I'll do it next week when I have my work Mac.

Or it's just something that was fixed already and not released yet.

kjsmita6:
Closed by #860.

PR #860, opened 2022-07-22
Pom editor dialog fixes

kjsmita6:
Fixed two things relating to some dialogs in the pom editor:

- **Changed the "Add..." button in the Dependency Management section of the pom editor to include "import" scope**
Screen Shot 2022-07-22 at 12 25 19 PM

- **Fixes PomHierarchyComposite drawing**
    - The tree control before was being drawn above its parent composite so it was getting cut off, see #854. This change replaced the control (which was a subclass of Composite) with a class that just wrapped a TreeViewer. It now draws properly:

**Before**:
Screen Shot 2022-07-22 at 12 27 05 PM
Screen Shot 2022-07-22 at 12 26 48 PM

**After**:
Screen Shot 2022-07-22 at 12 32 18 PM
Screen Shot 2022-07-22 at 12 26 22 PM


kjsmita6:
Let me know your thoughts on adding a packaging type combo to the "Select Dependency" dialog. Since sometimes you might want to add a dependency of another type, like bundle, pom, war, etc. The current way requires you to add the dependency and go in the pom and change the type manually.

Of course if the dialog finds the artifact (which rarely happens in my experience) it will add it correctly.

mickaelistria:
Rebased and merged as 487485116c5849ea3408e766429681b045d11fb3 . Thanks!

PR #865, opened 2022-07-29
Update TP to wildwepdevloper 0.15 and lsp4e 0.20.5

HannesWell:
Before the 2.0 release we should replace all snapshots from the TP with latest releases.
@mickaelistria is this fine?


mickaelistria:
More than fine! Thanks!

PR #868, opened 2022-08-09
Switch to tycho-build extension

akurtakov:
Instead of the deprecated tycho-pomless one.

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <extensions>
   <extension>
-    <groupId>org.eclipse.tycho.extras</groupId>
-    <artifactId>tycho-pomless</artifactId>
+    <groupId>org.eclipse.tycho</groupId>
+    <artifactId>tycho-build</artifactId>
     <version>2.7.4</version>
   </extension>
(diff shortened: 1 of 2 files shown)


akurtakov:
@HannesWell @laeubi Any idea why the build is broken ?

laeubi:
m2e currently uses the "maven mode" of tycho that is not supported by the tycho-build extension, I think we decided to rework that due to recent enhancements in m2e+tycho but it got a bit forgotten ...

akurtakov:
So the change should happen on m2e or on tycho side?

laeubi:
m2e needs to change here, but this might reveal some tasks for tycho as well (e.g. if there is something that can not be done).

HannesWell:
It is exactly as Christoph said. I started the work here https://github.com/eclipse-m2e/m2e-core/pull/466 but got distracted by other things in the meantime. I'll try to push that up in my list.
But at the moment I'm slowed down since I catched Covid as souvenir from my vacation and have to get well first.

akurtakov:
Fast recovery, Hannes!

HannesWell:
> Fast recovery, Hannes!

Thank you. :)

akurtakov:
Closing this one as there is further work to be done before that's feasible

PR #872, opened 2022-08-09
Adjust M2E_VERSION extraction in Jenkinsfile to m2e.sdk/feature.xml

HannesWell:
If one modifies the version of the m2e.sdk feature via the PDE-Feature editor the XML-element 'feature' gets its attributes always written in a fixed order. But the Snapshot-deployment in the Jenkinsfile relies on the attribute order to extract the current M2E-Version. 
In order to not require modifications in the source tab, this adjusts the version extraction logic in the Jenkinsfile to the default feature attribute order.

This fixes the currently broken master build.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -48,7 +48,7 @@ pipeline {
 							scp -r org.eclipse.m2e.repository/target/repository/* genie.m2e@projects-storage.eclipse.org:${1}
 						}
 						# Read M2E branding version
-						regex='<feature id="org\\.eclipse\\.m2e\\.sdk\\.feature" version="([0-9]\\.[0-9]\\.[0-9])\\.qualifier" '
+						regex='<feature id="org\\.eclipse\\.m2e\\.sdk\\.feature" label="%featureName" version="([0-9]\\.[0-9]\\.[0-9])\\.qualifier" '
 						content=$(echo $(<"org.eclipse.m2e.sdk.feature/feature.xml")) # replaces consecutive newline and tabs by single space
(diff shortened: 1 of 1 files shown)


HannesWell:
The PR verification does not execute the adjusted `Deploy-snapshot` Jenkins stage. But I verified with an adjusted replayed job execution that this fixes the version extraction:
https://ci.eclipse.org/m2e/job/m2e/job/master/375/

laeubi:
Tycho uses a version.properties file that is processed by maven and contains the current pom version. We could use similar here, that would be more reliable than greping xml files where

1. order of attributes is not guaranteed
2. number of whitespace is not guaranteed
3. line breaks can occur
4. ....

HannesWell:
> Tycho uses a version.properties file that is processed by maven and contains the current pom version. We could use similar here, that would be more reliable than greping xml files where
> 
Yes I could just add a profile activated for feature.xml files that writes a version file using the antrun plugin. We either had this already or it was at least a variant of a PR.
I have chosen the current approach because it did not require any assistance in any other file and until know it worked well and it should continue to work as long as only the version is changed (in the Editor-field or the source-tab). If more is changed then the regex indeed has to be adjusted or that other approach has to be chosen.

laeubi:
> Yes I could just add a profile activated for feature.xml files that writes a version file using the antrun plugin. We either had this already or it was at least a variant of a PR.

You don't need ant for this, this can be done all by maven itself, if you choose a default location for your resource you don't need to configure any special profile:

https://maven.apache.org/plugins/maven-resources-plugin/examples/filter.html

PR #874, opened 2022-08-17
[TP] Build against Eclipse 4.25-I-builds

HannesWell:
If we release the next version of M2E before the Eclipse 2022-09 release it will be definitely targeted to that version. Therefore I think it is OK to build against 4.25-I-builds already now.

@mickaelistria (if you are reading during your vacation), @laeubi what do you think?

@mihnita can the MavenLaunchDelegegate be simplified further? IIRC you said that there is already a check per console if coloring is supported/enabled? Or is that only planned?


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "m2e-core-tests"]
 	path = m2e-core-tests
-	url = https://github.com/tesla/m2e-core-tests.git
-	branch = master
+	url = https://github.com/HannesWell/m2e-core-tests.git
```
HannesWell, line 3: @mihnita this way we can test changes in the m2e-core-tests submodule before merging them.
It is a bit more complicated and requires at least two verification-builds (if one wants to always have green builds), but it works.
Of course this has to be revered before this PR is submitted.

target-platform/target-platform.target
```diff
@@ -4,13 +4,17 @@
...
+			<repository location="https://download.eclipse.org/eclipse/updates/4.25-I-builds/"/><!--Keep in sync with repo-ref in org.eclipse.m2e.repository/category.xml-->
 			<unit id="org.eclipse.sdk.feature.group" version="0.0.0"/>
 			<unit id="org.eclipse.equinox.p2.discovery.feature.feature.group" version="0.0.0"/>
 			<unit id="org.eclipse.ui.tests.harness" version="0.0.0"/>
 			<unit id="org.mockito.mockito-core" version="0.0.0"/>
 			<unit id="org.eclipse.equinox.executable.feature.group" version="0.0.0"/>
 		</location>
+		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
+			<repository location="https://download.eclipse.org/eclipse/updates/4.24/"/>
+			<unit id="org.eclipse.pde.ds.lib" version="0.0.0"/>
```
HannesWell, line 16: This is necessary since https://github.com/eclipse-pde/eclipse.pde/pull/269 is only available in the I-Builds and not even in a Milestone, but `org.eclipse.pde.ds.lib` already has been removed from the I-build repo.

mihnita:
> @mihnita can the MavenLaunchDelegegate be simplified further?
I can't imagine how. If we require a newer version of org.eclipse.ui.console, one that has ANSI, we can remove `isAnsiSupportInstalled()`, but nothing other than that.

> IIRC you said that there is already a check per console if coloring is supported/enabled? 

There is none.
Right now the enable / disable status is global (for same for all consoles).
And that is the one accessed with `preferencesService.getBoolean(...)` in `isAnsiProcessingEnabled()`

.gitmodules
mihnita: I know. That part I did, and worked to test / fix things.

But I was unable to make the changes in `m2e-core-tests` show up in the `m2e-core` PR.
When I changed stuff and pushed it went in mihnita/m2e-core-tests.git, and I was able to make a PR from there.
But the PR from `m2e-core` ignored the changes in  `m2e-core-tests`

mihnita (approved):

.gitmodules
HannesWell: > But the PR from `m2e-core` ignored the changes in `m2e-core-tests`

Changes of submodules are not shown in the container git, you only see that the ref-point to m2e-core-tests submodule has changed and you can only stage and commit that ref-pointer change. The staging view then looks like this:
grafik

But if you did that, maybe GH and Jenkins ignore changes in the .gitmodules of non-committers when checking out the submodules.

laeubi: For me it has worked to just edit the `.gitmodules` file and never changed the ref pointer....

HannesWell:
> > @mihnita can the MavenLaunchDelegegate be simplified further?
> 
> I can't imagine how. If we require a newer version of org.eclipse.ui.console, one that has ANSI, we can remove `isAnsiSupportInstalled()`, but nothing other than that.

OK, I just wanted to know if I missed something.

> 
> > IIRC you said that there is already a check per console if coloring is supported/enabled?
> 
> There is none. Right now the enable / disable status is global (for same for all consoles). And that is the one accessed with `preferencesService.getBoolean(...)` in `isAnsiProcessingEnabled()`

Then I understood that wrong and this is fine as it is.

PR #875, opened 2022-08-18
Merge SLF4J-binding slf4j-simple into o.e.m2e.maven.runtime and replace it with maven-slf4j-provider

HannesWell:
The purpose of having the slf4j-simple binding in the separated o.e.m2e.maven.runtime.slf4j.simple bundle is to make it inaccessible for the OSGi runtime of the Eclipse-IDE, to not use it when the embedded Maven-runtime is executed within the IDE. In that case the SLF4J-binding provided by Eclipse shall be used. Therefore o.e.m2e.maven.runtime.slf4j.simple simply did not export any packages making it completely inactive in OSGi.

However when a Maven build is launched from the IDE the embedded slf4j-simple binding is supposed to be used. This worked because the Maven-runtime is not a OSGi-runtime and does not care about Export-Package headers.

The new approach is to embed the jar of the SL4J-binding directly in the o.e.m2e.maven.runtime bundle and is made invisible for OSGi by simply not placing it on the Bundle-ClassPath. The MavenEmbeddedRuntime is adjusted to nevertheless add the SL4J-binding's jar on the classpath of a launched Maven-build runtime. To simplify this the 'M2E-SLF4JBinding' header is automatically generated, whose value is the path to the SLF4J-binding jar within the o.e.m2e.maven.runtime bundle.

@mickaelistria, @laeubi, @akurtakov what do you think?

@mihnita if all tests pass (but I'm confident) and this change makes in on the master, it should be simple to merge `sl4j-simple` into `o.e.m2e.maven.runtime`, on top of this it should be simple to exchange `slf4j-simple` by `maven-slf4j-provider`. Just replace both occurances of slf4j-simple accordingly and clean up the unused slf4j-version variable and licenses.


HannesWell:
> @mihnita if all tests pass (but I'm confident) and this change makes in on the master, it should be simple to merge `sl4j-simple` into `o.e.m2e.maven.runtime`, on top of this it should be simple to exchange `slf4j-simple` by `maven-slf4j-provider`. Just replace both occurances of slf4j-simple accordingly and clean up the unused slf4j-version variable and licenses.

@mihnita do you want/plan to provide a PR to move from `sfl4j-simple` to `maven-slf4j-provider` on top of this PR? If not I can do that, that wouldn't be a problem.

mihnita:
Ah, sorry, I thought you will take it "all the way" :-)
And was puzzled why `maven-slf4j-provider` is not mentioned.

To be fair, I am still a bit unclear about all the logging in Eclipse,
combined with "all kinds of maven" (external, embedded, workspace)
I've also seen some logback (plugins/ch.qos.logback.* & plugins/org.eclipse.m2e.logback.appender).
And plugins/org.apache.commons.logging :-)

So I would feel safer if you do it.
And I can look over your shoulder and learn :-)

But it is Friday night here (California).
Who knows, if you don't do it tonight (your Saturday) I might take a shot at it.
I get to learn something :-)

Thanks,
Mihai

mihnita (approved):
LGMT
(not that you need my approval, you understand much better what's going on here :-)

HannesWell:
> Ah, sorry, I thought you will take it "all the way" :-) And was puzzled why `maven-slf4j-provider` is not mentioned.
> 
> To be fair, I am still a bit unclear about all the logging in Eclipse, combined with "all kinds of maven" (external, embedded, workspace) I've also seen some logback (plugins/ch.qos.logback.* & plugins/org.eclipse.m2e.logback.appender). And plugins/org.apache.commons.logging :-)

No problem. I assumed that you already have a half ready commit and wanted to give you the opportunity to submit that. But since I already checked what needs to be done (and therefore had it almost ready) there is no need to make you extra work. :)
Nevertheless I mentioned you as Co-author. Thank you for this suggestion. This makes the colored print-out of Maven builds even nicer.

Yes indeed the logging topic with its different logging APIs bound to one logger-binding via bridges is itself not easy (in case it is valuable for you, https://www.slf4j.org/manual.html provides some documentation). The usage scenario in M2E with its two worlds Maven-runtime and Eclipse/OSGi-runtime that it sometimes connects and sometimes separates makes it even harder to understand.

HannesWell:
For me this change is ready.
The removal of the `o.e.m2e.maven.runtime.slf4j.simple` bundle is not a problem. It does not export any packages and actually nobody besides the o.e.m2e.maven.runtime bundle can really use it. And as long as one does not reference it directly in a product or launch config nobody should notice. And even in that cases the solution is to simply remove it.
So the merge (should) change nothing just that we have one bundle less and a therefore little bit simpler projects.

Nevertheless I would like to have some feedback of the other M2E committers, but I have the impression that everybody is on vacation. I'm also on vacation next week, but can press the submit button on my phone when the others have consented (or at least not objected).

laeubi (approved):
Seems good to me.

HannesWell:
> Seems good to me.

Thank you Christoph. I assume the others will have the same opinion, at least I don't see a reason that speaks against that.
So lets merge this now.

PR #878, opened 2022-08-22
Replace pde.ui dependency with pde.core in org.eclipse.m2e.pde.target

rgrunber:
- org.eclipse.pde.ui dependencies are not used in
  org.eclipse.m2e.pde.target
- org.eclipse.pde.ui 3.12.0 corresponds to org.eclipse.pde.core 3.14.0
- org.eclipse.m2e.pde.target should explicitly depend on pde.core rather
  than implicitly through pde.ui
- Bump minor version for plugin & feature

See https://download.eclipse.org/eclipse/updates/4.18/R-4.18-202012021800/plugins/ for reference of where org.eclipse.pde.ui 3.12.0 corresponds to org.eclipse.pde.core 3.14.0.

Signed-off-by: Roland Grunberg <rgrunber@redhat.com>

### DIFF
--- a/org.eclipse.m2e.pde.feature/feature.xml
+++ b/org.eclipse.m2e.pde.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.m2e.pde.feature"
       label="%featureName"
-      version="2.0.0.qualifier"
+      version="2.0.1.qualifier"
       provider-name="%providerName"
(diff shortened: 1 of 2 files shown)


HannesWell:
Thank for this PR. Look's like a reasonable change.
But a Micro-Version bump (2.0.1) seems sufficient in this case.
Can you please change that.

HannesWell:
Thank you.

PR #882, opened 2022-08-28
Update build plug-in versions

HannesWell:


mickaelistria:
Looks like something changed with the signer as it's now failing when there is apparently no error (just a note that there won't be a re-sign).

mickaelistria:
Is the signer error a regression in the jarsigner plugin 1.3.2 ? Can you please try sticking to older jarsigner so we can try bumping other Maven plugin versions?

HannesWell:
> Is the signer error a regression in the jarsigner plugin 1.3.2 ? Can you please try sticking to older jarsigner so we can try bumping other Maven plugin versions?

Yes I will try that.
But I would like to hold this back for a little bit, just in case some more hot-fixes for 2022-09 are needed. Then it is better to change as few as possible to prevent unintended changes. And from my side these updates are not urgent.

PR #883, opened 2022-08-28
Simplify extraction of M2E_VERSION and fix archiving of m2e-repo

HannesWell:
As discussed in https://github.com/eclipse-m2e/m2e-core/pull/872, this changes the build to read the value from the `pom.properties` created by the `maven-archiver-plugin` instead of the `m2e.sdk/feature.xm`l. The format of the former is much simpler.
Although this is not exactly what has been suggested in the linked PR this approach does not require any additional configuration in poms and does not depend on a specific XML attribute order.

For the Jenkins `archiveArtifacts` plug-in the artifacts property has been adjusted to the new name of the o.e.m2e.repository project.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,14 +19,17 @@ pipeline {
 		}
 		stage('Build') {
 			steps {
-				sh 'mvn clean generate-sources -f m2e-maven-runtime/pom.xml -B -Dtycho.mode=maven -Pgenerate-osgi-metadata '
+				sh 'mvn clean generate-sources -f m2e-maven-runtime/pom.xml -B -Dtycho.mode=maven -Pgenerate-osgi-metadata'
 				wrap([$class: 'Xvnc', useXauthority: true]) {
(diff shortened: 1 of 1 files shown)


mickaelistria:
Another approach would be to use `xmllint` to get the version from feature. Would it be better/worse?

HannesWell:
> Another approach would be to use `xmllint` to get the version from feature. Would it be better/worse?

You mean it should be something like the following?
```
version=$(xmllint --xpath /feature@version org.eclipse.m2e.sdk.feature/feature.xml)
if [[ $version =~ '([0-9]\\.[0-9]\\.[0-9])\\.qualifier' ]]
...
```
Keeping the regex to remove the `.qualifier` suffix also adds a consistency check.

Looks better to me. It is relatively simple, independent from a specific formatting and does not rely on the fact that the maven-archiver-plugin write a properties file (like the current approach).

mickaelistria:
Yes, that's what I have in mind. I let you go for what you think is better.

HannesWell:
Thank you Mickael for suggestion xmllint, using that is really a benefit.

Issue #896, opened 2022-09-08
apt preferences are broken on latest m2e release


martinlippert:
It looks like the latest m2e release contains a broken version of the apt ui bundle.

The problem is based on this commit:
https://github.com/eclipse-m2e/m2e-core/commit/8b876a6748d8858578ee5894c40a872affc0e052

It changed the return types of many methods. One of those methods is used here:
https://github.com/eclipse-m2e/m2e-core/blob/master/org.eclipse.m2e.apt.ui/src/org/eclipse/m2e/apt/ui/preferences/AnnotationProcessingSettingsPage.java#L290

While the source is still compatible with the change made to the return type (from array to collection), the compiled code is not compatible and breaks with:

```
java.lang.NoSuchMethodError: 'org.eclipse.m2e.core.project.IMavenProjectFacade[] org.eclipse.m2e.core.project.IMavenProjectRegistry.getProjects()'
	at org.eclipse.m2e.apt.ui.preferences.AnnotationProcessingSettingsPage.getImpactedProjects(AnnotationProcessingSettingsPage.java:290)
	at org.eclipse.m2e.apt.ui.preferences.AnnotationProcessingSettingsPage.updateImpactedProjects(AnnotationProcessingSettingsPage.java:245)
	at org.eclipse.m2e.apt.ui.preferences.AnnotationProcessingSettingsPage.performOk(AnnotationProcessingSettingsPage.java:205)
	at org.eclipse.jface.preference.PreferenceDialog$7.run(PreferenceDialog.java:905)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.jface.util.SafeRunnable.run(SafeRunnable.java:174)
	at org.eclipse.jface.preference.PreferenceDialog.okPressed(PreferenceDialog.java:889)
	at org.eclipse.ui.internal.dialogs.FilteredPreferenceDialog.okPressed(FilteredPreferenceDialog.java:461)
	at org.eclipse.jface.preference.PreferenceDialog.buttonPressed(PreferenceDialog.java:233)
	at org.eclipse.jface.dialogs.Dialog.lambda$0(Dialog.java:619)
	at org.eclipse.swt.events.SelectionListener$1.widgetSelected(SelectionListener.java:84)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:252)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
[...]

PR #906, opened 2022-09-12
Fail build on baseline differences and exclude expected differences

HannesWell:
This should prevent bugs like https://github.com/eclipse-m2e/m2e-core/issues/896.

As suggested in https://github.com/eclipse-equinox/p2/pull/113#issuecomment-1240392193.

One drawback of the initial suggestion is that when building on Windows all text-files are detected as different due to the default line-endings difference between Windows and Linux.
So eventueally the `baselineMode = failCommon` should either be wrapped in a profile (activated only on non-windows?) or could be set via the System-property `tycho.baseline`.


pom.xml
```diff
@@ -231,6 +232,19 @@
...
+						<repository>
+							<url>https://download.eclipse.org/technology/m2e/snapshots/latest/</url>
+						</repository>
+					</baselineRepositories>
+					<baselineMode>warn</baselineMode>
+					<ignoredPatterns>
+						<pattern>META-INF/ECLIPSE_.RSA</pattern>
+						<pattern>META-INF/ECLIPSE_.SF</pattern>
+						<pattern>**/org/fusesource/jansi/internal/native/Mac/*/libjansi.jnilib</pattern>
+						<pattern>org/eclipse/m2e/core/internal/lifecyclemapping/model/*.java</pattern>
```
HannesWell, line 245: I wonder, why these files are detected as different? They are generated at the build server, nevertheless they should be equal since the build environment did not change since this morning?

pom.xml
```diff
@@ -316,22 +330,6 @@
...
 					</configuration>
 				</plugin>
-				<plugin>
-					<groupId>org.eclipse.tycho</groupId>
-					<artifactId>tycho-p2-plugin</artifactId>
-					<version>${tycho-version}</version>
-					<configuration>
-						<baselineRepositories>
-							<repository>
-								<url>https://download.eclipse.org/technology/m2e/releases/latest/</url>
```
HannesWell, line 326: Is there a reason why we use releases/latest and snapshots/latest together? The latter should always be at least as recent as the former or more recent.

mickaelistria: You're right, you should be able to remove the release here.

pom.xml
```diff
@@ -391,6 +390,20 @@
...
 			</build>
+			<properties>
+				<p2.baselineMode>failCommon</p2.baselineMode>
+			</properties>
+		</profile>
+		<profile>
+			<id>baseline-comparison-strict</id>
+			<activation>
+				<property>
+					<name>CI</name>
```
mickaelistria, line 401: Why only on CI?

pom.xml
mickaelistria: Maybe there is some timestamp or other noise added?
Note that the build server is a kubernetes cluster, so every build is a different (virtual) host.

HannesWell: > Maybe there is some timestamp or other noise added? 

Indeed it is a timestamp generated. Thank you for that hint! I could have come up with that myself:

```
/*
 =================== DO NOT EDIT THIS FILE ====================
 Generated by Modello 1.5 on 2022-09-13 10:09:52,
 any modifications will be overwritten.
 ==============================================================
 */
```

> Note that the build server is a kubernetes cluster, so every build is a different (virtual) host.

I first assumed that this could be a reason, but for all other source files this wasn't an issue. And as long as all of them don't use a Windows OS line-endings should always be the same, just like encoding.

pom.xml
mickaelistria: > Because when one runs the build on Windows all text-based source files are different due to different line-endings.

OK. Can't we find some option to force the line-ending (and remove timestamps) in the code generator so this is portable?
If the issue is the OS only then instead of a property the activation should be based on the os.

HannesWell: > > Because when one runs the build on Windows all text-based source files are different due to different line-endings.
> 
> OK. Can't we find some option to force the line-ending (and remove timestamps) in the code generator so this is portable? 

I have not found an option to disable that in modello. But I since only the java source files are ignored and not the class files I think it is OK to ignore differences in the sources. If there are changes that lead to a binary difference they will be noticed in the class-files.
Regarding the text files, do you know a elegant way to achieve that? I only found some suggestions to use the resources plug-in and define different resource sets. But that seems a bit inconvenient. On the other hand just accepting that and only warn about differences like it is at the moment by default is also ok.


> If the issue is the OS only then instead of a property the activation should be based on the os.

I would prefer enable strict comparison explicitly, so that it does not suddenly disabled (without anybody noticing).
Enabling that with the `eclipse-sign` looks like a suitable place to me. This is only enabled on Jenkins which also creates the released bits. The chances of false positive differences are the lowest here.

mickaelistria (approved):
OK

HannesWell:
Thanks.

PR #922, opened 2022-09-21
Expose ArchetypeGenerator and ArchetypeManager

### LINKED ISSUE #921: API missing for creating archetype projects
Previously there was the public `IProjectConfigurationManager.createArchetypeProjects(...)` but now there is only the internal class `ArchetypeGenerator` (introduced with https://github.com/eclipse-m2e/m2e-core/commit/eb408339868ce4484e52b2bf9f5df59b5ea76229#diff-4742ec5a022dec0edc843e783c76a0a258224fcb273fb08ba3f107e0dc6887cb). How should consumers now trigger a project creation for archetypes?

Was the `ArchetypeGenerator` being put deliberately in package `org.eclipse.m2e.core.ui.internal.archetype`?

kwin:
This closes #921

### DIFF
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeGenerator.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeGenerator.java
@@ -1,0 +1,59 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Konrad Windszus
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *      Konrad Windszus
+ *******************************************************************************/
+package org.eclipse.m2e.core.ui.archetype;
+
+import java.util.Collection;
+import java.util.Map;
+
+import org.osgi.annotation.versioning.ProviderType;
+
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.IProgressMonitor;
+
+import org.eclipse.m2e.core.project.IArchetype;
+import org.eclipse.m2e.core.project.MavenProjectInfo;
+
+
+/**
+ * Implementations of this OSGi service interface allow to create new Maven projects based on Maven archetypes.
+ * 
+ * @since 2.1.0 (package version 1.0.0)
+ */
+@ProviderType
+public interface ArchetypeGenerator {
+
+  /**
+   * Creates a project structure using the given archetype in non-interactive mode.
+   *
+   * @return a list of created Maven projects.
+   */
+  default Collection<MavenProjectInfo> createArchetypeProjects(IPath location, IArchetype archetype, String groupId,
+      String artifactId, String version, String javaPackage, Map<String, String> properties, IProgressMonitor monitor)
+      throws CoreException
+  {
+    return createArchetypeProjects(location, archetype, groupId, artifactId, version, javaPackage, properties, false,
+        monitor);
+  }
+
+  /**
+   * Creates a project structure using the given archetype.
+   *
+   * @return a list of created Maven projects.
+   */
+  Collection<MavenProjectInfo> createArchetypeProjects(IPath location, IArchetype archetype, String groupId,
+      String artifactId, String version, String javaPackage, Map<String, String> properties, boolean interactive,
+      IProgressMonitor monitor) throws CoreException;
+
+}
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeManager.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeManager.java
@@ -22,7 +22,6 @@ import org.osgi.annotation.versioning.ProviderType;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 
-import org.apache.maven.archetype.catalog.Archetype;
 import org.apache.maven.archetype.catalog.ArchetypeCatalog;
 import org.apache.maven.archetype.metadata.RequiredProperty;
 
@@ -47,14 +46,6 @@ public interface ArchetypeManager {
    * @throws CoreException if no archetype can be resolved
    */
   List<RequiredProperty> getRequiredProperties(IArchetype archetype, IProgressMonitor monitor) throws CoreException;
-
-  /**
-   * Updates the local catalog to either add or refresh the given archetype.
-   * 
-   * @param archetype the archetype to add or refresh in the local catalog
-   * @throws CoreException
-   */
-  void updateLocalCatalog(Archetype archetype) throws CoreException;
 
   /**
    * Returns all archetypes catalogs from the active underlying factories configured in the workspace properties.
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeManager.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeManager.java
@@ -1,0 +1,67 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Konrad Windszus
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *      Konrad Windszus
+ *******************************************************************************/
+
+package org.eclipse.m2e.core.ui.archetype;
+
+import java.util.List;
+import java.util.Map;
+
+import org.osgi.annotation.versioning.ProviderType;
+
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+
+import org.apache.maven.archetype.catalog.Archetype;
+import org.apache.maven.archetype.catalog.ArchetypeCatalog;
+import org.apache.maven.archetype.metadata.RequiredProperty;
+
+import org.eclipse.m2e.core.project.IArchetype;
+
+
+/**
+ * Provides operations to manage the catalog.
+ *
+ * @since 2.1.0 (package version 1.0.0)
+ */
+@ProviderType
+public interface ArchetypeManager {
+
+  /**
+   * Gets the required properties of an {@link IArchetype}.
+   *
+   * @param archetype the archetype possibly declaring required properties
+   * @param remoteArchetypeRepository the remote archetype repository, can be null
+   * @param monitor the progress monitor, can be null
+   * @return the required properties of the archetypes, null if none is found
+   * @throws CoreException if no archetype can be resolved
+   */
+  List<RequiredProperty> getRequiredProperties(IArchetype archetype, IProgressMonitor monitor) throws CoreException;
+
+  /**
+   * Updates the local catalog to either add or refresh the given archetype.
+   * 
+   * @param archetype the archetype to add or refresh in the local catalog
+   * @throws CoreException
+   */
+  void updateLocalCatalog(Archetype archetype) throws CoreException;
+
+  /**
+   * Returns all archetypes catalogs from the active underlying factories configured in the workspace properties.
+   * 
+   * @return a map containing the catalog factory id as key and the actual archetype catalogs as value
+   * @throws CoreException
+   */
+  Map<String, ArchetypeCatalog> getActiveArchetypeCatalogs() throws CoreException;
+
+}
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/MavenArchetype.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/MavenArchetype.java
@@ -11,16 +11,21 @@
  * Contributors:
  *      Christoph Läubrich - initial API and implementation
  *******************************************************************************/
-package org.eclipse.m2e.core.ui.internal.archetype;
+package org.eclipse.m2e.core.ui.archetype;
 
 import org.apache.maven.archetype.catalog.Archetype;
 
 import org.eclipse.m2e.core.project.IArchetype;
 
 
-public class MavenArchetype implements IArchetype {
+/**
+ * Implementation of {@link IArchetype} based on {@link Archetype}
+ * 
+ * @since 2.1.0 (package version 1.0.0)
+ */
+public final class MavenArchetype implements IArchetype {
 
-  private Archetype archetype;
+  private final Archetype archetype;
 
   public MavenArchetype(Archetype archetype) {
     this.archetype = archetype;
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeGeneratorImpl.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeGeneratorImpl.java
@@ -51,7 +51,7 @@ import org.eclipse.m2e.core.project.MavenProjectInfo;
 import org.eclipse.m2e.core.ui.archetype.ArchetypeGenerator;
 
 
-@Component(service = {ArchetypeGeneratorImpl.class})
+@Component(service = {ArchetypeGenerator.class})
 @SuppressWarnings("restriction")
 public class ArchetypeGeneratorImpl implements ArchetypeGenerator {
 
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeGeneratorImpl.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeGeneratorImpl.java
@@ -48,11 +48,12 @@ import org.eclipse.m2e.core.project.IArchetype;
 import org.eclipse.m2e.core.project.IProjectConfigurationManager;
 import org.eclipse.m2e.core.project.LocalProjectScanner;
 import org.eclipse.m2e.core.project.MavenProjectInfo;
+import org.eclipse.m2e.core.ui.archetype.ArchetypeGenerator;
 
 
-@Component(service = {ArchetypeGenerator.class})
+@Component(service = {ArchetypeGeneratorImpl.class})
 @SuppressWarnings("restriction")
-public class ArchetypeGenerator {
+public class ArchetypeGeneratorImpl implements ArchetypeGenerator {
 
   @Reference
   IProjectConfigurationManager projectConfigurationManager;
@@ -63,19 +64,7 @@ public class ArchetypeGenerator {
   @Reference
   IMavenLauncher mavenLauncher;
 
-  /**
-   * Creates project structure using Archetype and then imports created project(s)
-   *
-   * @return a list of created projects.
-   * @since 1.8
-   */
-  public Collection<MavenProjectInfo> createArchetypeProjects(IPath location, IArchetype archetype, String groupId,
-      String artifactId, String version, String javaPackage, Map<String, String> properties, IProgressMonitor monitor)
-      throws CoreException {
-    return createArchetypeProjects(location, archetype, groupId, artifactId, version, javaPackage, properties, false,
-        monitor);
-  }
-
+  @Override
   public Collection<MavenProjectInfo> createArchetypeProjects(IPath location, IArchetype archetype, String groupId,
       String artifactId, String version, String javaPackage, Map<String, String> properties, boolean interactive,
       IProgressMonitor monitor) throws CoreException {
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypePlugin.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypePlugin.java
@@ -52,6 +52,7 @@ import org.codehaus.plexus.classworlds.ClassWorld;
 import org.codehaus.plexus.component.repository.exception.ComponentLookupException;
 
 import org.apache.maven.archetype.catalog.Archetype;
+import org.apache.maven.archetype.catalog.ArchetypeCatalog;
 import org.apache.maven.archetype.common.ArchetypeArtifactManager;
 import org.apache.maven.archetype.exception.UnknownArchetype;
 import org.apache.maven.archetype.metadata.ArchetypeDescriptor;
@@ -64,6 +65,8 @@ import org.apache.maven.project.ProjectBuildingRequest;
 import org.eclipse.m2e.core.embedder.IMaven;
 import org.eclipse.m2e.core.internal.MavenPluginActivator;
 import org.eclipse.m2e.core.project.IArchetype;
+import org.eclipse.m2e.core.ui.archetype.ArchetypeGenerator;
+import org.eclipse.m2e.core.ui.archetype.ArchetypeManager;
 import org.eclipse.m2e.core.ui.internal.M2EUIPluginActivator;
 import org.eclipse.m2e.core.ui.internal.archetype.ArchetypeCatalogFactory.LocalCatalogFactory;
 import org.eclipse.m2e.core.ui.internal.archetype.ArchetypeCatalogFactory.RemoteCatalogFactory;
@@ -74,8 +77,8 @@ import org.eclipse.m2e.core.ui.internal.archetype.ArchetypeCatalogFactory.Remote
  *
  * @author Eugene Kuleshov
  */
-@Component(service = ArchetypePlugin.class)
-public class ArchetypePlugin {
+@Component(service = {ArchetypeManager.class, ArchetypePlugin.class})
+public class ArchetypePlugin implements ArchetypeManager {
 
   public static final String ARCHETYPE_PREFIX = "archetype";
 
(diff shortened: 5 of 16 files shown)


kwin:
@laeubi The compilation inside Eclipse works fine and `org.eclipse.m2e.core.ui` list `org.osgi.annotation.versioning_1.1.2.202109301733.jar` as plug-dependency. Still Tycho cannot compile and fails with

```
Error:  Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:2.7.4:compile (default-compile) on project org.eclipse.m2e.core.ui: Compilation failure: Compilation failure: 
Error:  /home/runner/work/m2e-core/m2e-core/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/wizards/MavenDependenciesWizardPage.java: 
Error:  	/*******************************************************************************
Error:  	^
Error:  The type org.osgi.annotation.versioning.Version cannot be resolved. It is indirectly referenced from required .class files
Error: [ERROR] 1 problem (1 error)
Error:  -> [Help 1]
```

The target file though contains the bundle `org.osgi.annotation.versioning` so I am puzzled why Tycho does not pick up that dependency....

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeGenerator.java
```diff
@@ -0,0 +1,54 @@
...
+import org.eclipse.core.runtime.IProgressMonitor;
+
+import org.eclipse.m2e.core.project.IArchetype;
+import org.eclipse.m2e.core.project.MavenProjectInfo;
+
+
+/**
+ * Implementations of this OSGi service interface allow to create new projects based on Maven archetypes.
+ * 
+ * @since 2.0.1 (package version 1.0.0)
```
mickaelistria, line 32: Will be a 2.1 for new APIs, not 2.0.1

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/package-info.java
```diff
@@ -0,0 +1,17 @@
+/*******************************************************************************
```
mickaelistria, line 1: We usually don't use package-info.json. Is it something really needed and useful here? If not, I suggest we remove it and maybe add it later, when there is a compelling reason to add them everywhere in m2e.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeGeneratorImpl.java
```diff
@@ -64,19 +65,10 @@ public class ArchetypeGenerator {
...
   IMavenLauncher mavenLauncher;
 
-  /**
-   * Creates project structure using Archetype and then imports created project(s)
-   *
-   * @return a list of created projects.
-   * @since 1.8
+  /* (non-Javadoc)
+   * @see org.eclipse.m2e.core.ui.internal.archetype.ArchetypeGeneratorInterface#createArchetypeProjects(org.eclipse.core.runtime.IPath, org.eclipse.m2e.core.project.IArchetype, java.lang.String, java.lang.String, java.lang.String, java.lang.String, java.util.Properties, boolean, org.eclipse.core.runtime.IProgressMonitor)
    */
```
mickaelistria, line 70: This is not necessary and should be removed. @Override is enough here to reference documentation from parent type.

laeubi:
> The target file though contains the bundle `org.osgi.annotation.versioning` so I am puzzled why Tycho does not pick up that dependency....

I think the problem is that Tycho 2.4 is used here.

mickaelistria (changes requested):
Just moving this is IMO not enough. We need to audit what are the real operations necessary to be exposed for clients to efficiently use the Archetype API here. Many operations in the promoted APIs can probably remain implementation-internal and should then be so.

kwin:
> Just moving this is IMO not enough. We need to audit what are the real operations necessary to be exposed for clients to efficiently use the Archetype API here.

The reasoning is in https://github.com/eclipse-m2e/m2e-core/issues/921. This PR obviously just contains the code changes.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/package-info.java
kwin: This is `package-info.java` not `.json`. It is necessary to version on package level as recommended by OSGi. Otherwise you cannot do proper semantic versioning as outlined in https://docs.osgi.org/whitepaper/semantic-versioning/040-semantic-versions.html.

mickaelistria: > It is necessary to version on package level as recommended by OSGi. Otherwise you cannot do proper semantic versioning as outlined in https://docs.osgi.org/whitepaper/semantic-versioning/040-semantic-versions.html.

package version has to be declared in MANIFEST.MF, not with a package-info.java. There might be some tools to build a good MANIFEST with this file, but the MANIFEST-first approach used in m2e doesn't allow that and versions need to be set directly in the MANIFEST.MF Export-Package.

kwin: Is the ConsumerType/ProviderType annotations evaluated by PDE for automatically generating the correct import-version ranges? Are there any plans for Tycho/PDE to evaluate the official annotations?
I will update the PR accordingly to export this package in version 1.0.0 manually in MANIFEST.MF.

mickaelistria:
> The reasoning is in https://github.com/eclipse-m2e/m2e-core/issues/921.

The reasoning "it was this way before" is not viable. The API was removed in this major version because it wasn't appearing as useful to anyone involved regularly in the development of m2e. That is already a sign that the API state should be challenged and not cloned for the future of m2e. If we're to reopen an API, it can and should be a minimal one, just like if we were starting it from scratch.
What is the particular methods you're using in your code? We should just start by exposing as API and keep everything else internal until we have a good story to promote them.

kwin:
https://github.com/eclipse-m2e/m2e-core/issues/921#issuecomment-1253955753
> The use case is triggering new project creation based on a specific artifact coordinates (https://github.com/apache/sling-ide-tooling/blob/93bcb69fe7cfad6b68d5caf838a14d3bdf887b92/eclipse/eclipse-m2e-ui/src/org/apache/sling/ide/eclipse/ui/wizards/np/AbstractNewMavenBasedSlingApplicationWizard.java#L106).

mickaelistria:
So you would only need a way to run something like `archetypeGenerator.createProject(Archetype, ...)` ?

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/package-info.java
laeubi: > package version has to be declared in MANIFEST.MF, not with a package-info.java. There might be some tools to build a good MANIFEST with this file, but the MANIFEST-first approach used in m2e doesn't allow that and versions need to be set directly in the MANIFEST.MF Export-Package.

As noted above this is not true anymore PDE already handles these annotations and updates the manifest automatically.
Beside that, it allows other tools to find out what is the purpose of an interface if it is used as a dependency so using these annotation can (and should be) done to move forward towards best OSGi practice.

mickaelistria: Good to know that modification can be automated, but still the MANIFEST.MF does need to include the version as part of the commit as PDE doesn't run those "fixes" in regular Tycho build, does it?

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeGenerator.java
kwin: Fixed in https://github.com/eclipse-m2e/m2e-core/pull/922/commits/4685cbd46790097615112e789598bf0315e1d3a8.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/package-info.java
kwin: Thanks, I updated it in https://github.com/eclipse-m2e/m2e-core/pull/922/commits/4685cbd46790097615112e789598bf0315e1d3a8 to manually set the package export version in the manifest and removed the package-info.java for now.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeGeneratorImpl.java
kwin: Fixed in https://github.com/eclipse-m2e/m2e-core/pull/922/commits/4685cbd46790097615112e789598bf0315e1d3a8.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/package-info.java
laeubi: > Good to know that modification can be automated, but still the MANIFEST.MF does need to include the version as part of the commit as PDE doesn't run those "fixes" in regular Tycho build, does it?

Yes, that currently only a very rough first step to empower further enhancements.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeGenerator.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Implementations of this OSGi service interface allow to create new projects based on Maven archetypes.
+ * 
+ * @since 2.1.0 (package version 1.0.0)
+ */
+@ProviderType
+public interface ArchetypeGenerator {
+
+  /**
+   * Creates project structure using the given archetype and then imports created project(s)
```
kwin, line 38: This javadoc (although just copied from ArchetypeGeneratorImpl) seems to require an update, as projects are no longer created by this method.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeGenerator.java
```diff
@@ -0,0 +1,54 @@
...
+  default Collection<MavenProjectInfo> createArchetypeProjects(IPath location, IArchetype archetype, String groupId,
+      String artifactId, String version, String javaPackage, Properties properties, IProgressMonitor monitor)
+      throws CoreException
+  {
+    return createArchetypeProjects(location, archetype, groupId, artifactId, version, javaPackage, properties, false,
+        monitor);
+  }
+
+  Collection<MavenProjectInfo> createArchetypeProjects(IPath location, IArchetype archetype, String groupId,
+      String artifactId, String version, String javaPackage, Properties properties, boolean interactive,
```
HannesWell, line 51: As proposed in #949 for the existing implementation we should consider to use a Map<String, String> instead of Properties. With static Map factories those are much more convenient to handle than Properties if there is a fixed set of key-values.

HannesWell:
@kwin if you rebase on top of the current master, some (maybe all) of the CI failures should vanish.

kwin:
Before I rebase and adjust the tests I would like to know if you are fine with the proposed API exposure.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeGenerator.java
mickaelistria: +1, `Map<String, String>` is more modern.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/MavenArchetype.java
```diff
@@ -11,16 +11,21 @@
...
 import org.eclipse.m2e.core.project.IArchetype;
 
 
-public class MavenArchetype implements IArchetype {
+/**
+ * Implementation of {@link IArchetype} based on {@link Archetype}
+ * 
+ * @since 2.1.0 (package version 1.0.0)
+ */
+public final class MavenArchetype implements IArchetype {
```
mickaelistria, line 26: Why is this required as API? I don't see it used in the signatures of ArchetypeManager/ArchetypeGenerator

mickaelistria (commented):
This looks fresher and a well focused API! I just have a couple of questions/comments inline.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/MavenArchetype.java
kwin: IArchetype is used as method parameter, so someone would need to come up with an implementation.

laeubi: See my comment, beside that a custom implementation is really trivial...

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypePlugin.java
```diff
@@ -74,8 +77,8 @@
  *
  * @author Eugene Kuleshov
  */
-@Component(service = ArchetypePlugin.class)
-public class ArchetypePlugin {
+@Component(service = {ArchetypeManager.class, ArchetypePlugin.class})
```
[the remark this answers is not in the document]
kwin: Not possible, as the ArchetypePlugin is the internal API which is used. If you would only register as ArchetypePlugin you would need to cast in M2EUIPluginActivator. How do you want to get hold of this without casting?

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeManager.java
```diff
@@ -0,0 +1,67 @@
...
+   */
+  List<RequiredProperty> getRequiredProperties(IArchetype archetype, IProgressMonitor monitor) throws CoreException;
+
+  /**
+   * Updates the local catalog to either add or refresh the given archetype.
+   * 
+   * @param archetype the archetype to add or refresh in the local catalog
+   * @throws CoreException
+   */
+  void updateLocalCatalog(Archetype archetype) throws CoreException;
```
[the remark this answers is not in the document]
kwin: IArchetype does not give you the possibility to reference the artifact among other things. Why do you consider org.apache.maven.archetype.catalog.Archetype not suitable here?

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeGenerator.java
kwin: fixed by rebasing.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/MavenArchetype.java
kwin: Done in https://github.com/eclipse-m2e/m2e-core/pull/922/commits/c39ab6b3ee2ade42b2a7da710f82006791c1c38b

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeGeneratorImpl.java
```diff
@@ -48,11 +48,12 @@
 import org.eclipse.m2e.core.project.IProjectConfigurationManager;
 import org.eclipse.m2e.core.project.LocalProjectScanner;
 import org.eclipse.m2e.core.project.MavenProjectInfo;
+import org.eclipse.m2e.core.ui.archetype.ArchetypeGenerator;
 
 
-@Component(service = {ArchetypeGenerator.class})
+@Component(service = {ArchetypeGeneratorImpl.class})
```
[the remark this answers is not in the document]
kwin: Fixed in https://github.com/eclipse-m2e/m2e-core/pull/922/commits/c39ab6b3ee2ade42b2a7da710f82006791c1c38b.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeManager.java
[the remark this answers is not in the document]
laeubi: Then this is probably not suitable for an API method here, why does user code needs this? The inital proposal in the issue was that it should be possible to generate a project given an Archetype (that is GAV) updating catalogs seems not something user code work with.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypePlugin.java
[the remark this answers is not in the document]
laeubi: > DS allows to replace one service implementation with another one!
This has nothing to do with DS, anyways you would be most likely wrecked then anyways. If you want be extra you can assert the type with instanceof and throw an exception otherwise...

> Also remember that you previously registered only to the internal API ArchetypePlugin via DS, so that pattern is not something I introduced.

Many thing in m2e need improvements and some are just for migrating thing so we should not take this as good examples when refactoring/rework things.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/MavenArchetype.java
laeubi: > Done in [c39ab6b](https://github.com/eclipse-m2e/m2e-core/commit/c39ab6b3ee2ade42b2a7da710f82006791c1c38b)

This looks good to me!

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeManager.java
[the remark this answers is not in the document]
kwin: No longer necessary, removed from API in https://github.com/eclipse-m2e/m2e-core/pull/922/commits/50971b871cd650381d5744a9d114de1f3817aa87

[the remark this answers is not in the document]
laeubi: Great :+1:

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypePlugin.java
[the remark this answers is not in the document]
laeubi: In Softwaredevelopment "later" often means "never" :-)

I would prefer to have a cast on the cases where it is required (that is we use no API method) as this is already done elsewhere in m2e and shows where we have an API gap...

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeManager.java
```diff
@@ -0,0 +1,58 @@
...
+import org.eclipse.m2e.core.project.IArchetype;
+
+
+/**
+ * Provides operations to manage the catalog.
+ *
+ * @since 2.1.0 (package version 1.0.0)
+ */
+@ProviderType
+public interface ArchetypeManager {
```
laeubi, line 37: Why is ArchetypeManager public? Should ArchetypeGenerator not be enough?

laeubi:
Looks good so far, just wondering if we really need to make `ArchetypeManager` public?

laeubi:
> ArchetypeManager is used e.g. 

But all these should be m2e internal, so can we not make `ArchetypeManager` internal as well?

laeubi:
> I use it outside of m2e, so I prefer using it with an API. I don't want to depend on internal API for my use case which was the original driver for this PR.

Can you explain how you use this? Do you provide own UI wizard? The main problem with this (and why it was decided to make this "non API") is that the Archetype is a separate plugin, so exposing its classes directly has two problems:

1. We need to make people depend on a specific ArcheType version (that is the one embedded in m2e) and we only support it partially, e.g. some classes will throw CNF because we only embedd the bar minimum of stuff.
2. If we update the ArcheType version and that API changes, this is actually an API break for m2e even though nothing was changed here.

This has lead to the situation that previously m2e a very long time relied on outdated Archetype 2.x. and it was quite hard to upgrade to 3.x. The result is/was the `IArcheType` facade interface that exposed the bare minimum required.

Can you check if for your use-case something similar would be possible to have (read only) facade types with a reduced signature for `RequiredProperty` and `ArchetypeCatalog`?

laeubi:
About `ArchetypeParametersWizardPage` do you think you probably would reuse the m2e page? It also supports parameter validation through regexp now, otherwhise I think we can have an `IRequiredProperty` facade that wraps the maven `RequiredProperty`with just the getters required here.

TIBCOeddie:
Our product uses the ArchetypeGenerator and friends. Is there still time to look at something and provide comments?

laeubi:
I think best would be too finish this PR and then if there are still things missing start with a new PR.

HannesWell:
> I think best would be too finish this PR and then if there are still things missing start with a new PR.

If it directly about what is changed in this PR I think comments should be made now. Of course it can be started again afterwards, but if possible why not consider them now but later?
If it is more an extension I agree that doing it step by step is better. :)

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeGenerator.java
```diff
@@ -0,0 +1,59 @@
...
+ * Implementations of this OSGi service interface allow to create new Maven projects based on Maven archetypes.
+ * 
+ * @since 2.1.0 (package version 1.0.0)
+ */
+@ProviderType
+public interface ArchetypeGenerator {
+
+  /**
+   * Creates a project structure using the given archetype in non-interactive mode.
+   *
```
TIBCOeddie, line 39: please have javadoc for params, or you'll get all the javadoc warnings especially since j17+ have turned it up, and for sure it's always a pity to not have these! esp handy is what can be null, what is optional, etc.

TIBCOeddie: I remember our integration had to dive into source to understand what "interactive mode" was

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeManager.java
```diff
@@ -0,0 +1,67 @@
...
+
+  /**
+   * Gets the required properties of an {@link IArchetype}.
+   *
+   * @param archetype the archetype possibly declaring required properties
+   * @param remoteArchetypeRepository the remote archetype repository, can be null
```
TIBCOeddie, line 43: param is not in sync with actual

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/archetype/ArchetypeManager.java
```diff
@@ -0,0 +1,67 @@
...
+   * @param monitor the progress monitor, can be null
+   * @return the required properties of the archetypes, null if none is found
```
TIBCOeddie, line 45: when your signature has a list, isn't it eclipse code guidelines to not return null and return an empty list instead?

mickaelistria: +1, it's generally a good practice to not return null for Arrays or Collections but empty ones; not only for Eclipse but for Java in general.

laeubi: Better return an `Optional<...>` instead then. There is a difference between an Archetype not found and having no required properties... one might even throw a CoreException if not found here ...

mickaelistria: > Better return an Optional<...> instead then.

I don't think so.

>  There is a difference between an Archetype not found and having no required properties..

Maybe, but that's not the goal of this method to detect that an Archetype is not found.  But your comment here emphasize that documentation is ambiguous and that there are different ways to interpret `none`.

> one might even throw a CoreException if not found here ...

CoreException is just an abstract exception in Eclipse world; creating a custom exception type would be better for an API.

laeubi: > > Better return an Optional<...> instead then.
> 
> I don't think so.

Why? Either is is optional (because it is fine / could happen) but then it is not valid to just return an empty list if nothing is found as a user of that method then falsely assumes everything is fine and one could potentially call this archetype (what will just fail because it is not found)...

> Maybe, but that's not the goal of this method to detect that an Archetype is not found. But your comment here emphasize that documentation is ambiguous and that there are different ways to interpret none.

Obviously there is a way to detect this, so either it is an error (= exception must be thrown) or a user has to be able to distinguish the cases.

mickaelistria:
> Why? Either is is optional (because it is fine / could happen) but then it is not valid to just return an empty list if nothing is found as a user of that method then falsely assumes everything is fine and one could potentially call this archetype (what will just fail because it is not found)...

There is a lot of literature against null collections, the most "reference" one being the "Effective Java" book, Your comment is based IMO on a dogmatic vision of API development; but experience of the whole Java (and other languages as well) community has concluded that returning null collections is more an anti-pattern than a benefit.

> Obviously there is a way to detect this, so either it is an error (= exception must be thrown) or a user has to be able to distinguish the cases.

Or maybe we just don't care about distinguishing, do we? Or maybe since we have an `Archetype` object as parameter, then we can assume that this Archetype exists and is found already.

laeubi:
> There is a lot of literature against null collections,

And how does it relates to returning an empty Optional in case nothing could be computed here?

> Or maybe we just don't care about distinguishing, do we? Or maybe since we have an `Archetype` object as parameter, then we can assume that this Archetype exists and is found already.

The parameter is just a simple container object and obviously the creator of that method has taken into account that the archetype is not present.

mickaelistria:
Given the comments, it seems like the overall design of those APIs is then questionable as they come with very poor guarantee about validity and lifecycle of instances.

So the current IArchetype is not an IArchetype then, it's just a GAV which is more used as an ArchetypeRequest. If I have an IArchetype, in a good API, I expect the archetype exists. The issue is then not what to do with a non existing IArchetype, it's that the API should ensure an IArchetype maps to something existing.
In the current state, I'd suggest removing the IArchetype.of method which is just a GAV, and even deprecating the IArchetype interface; or to make the IArchetype interface host the method to query parameters and other stuff and let only the ArchetypeManager create the IArchetype instance if this one does exists.

HannesWell:
What's the status here?

laeubi:
@kwin please rebase and resolve conflicts.

laeubi:
Closing this for now as no progress and no response, but feel fre to just reopen or provide a new PR on the top of the master branch.

Issue #924, opened 2022-09-22
ConcurrentModificationException when downloading sources and/or javadoc


pcdavid:
Using Spring Tool Suite, recently updated to version 4.16 based on Eclipse 2022-09, and:

`org.eclipse.m2e.feature.feature.group version 2.0.5.20220912-1211`

I switched to a different git branch (from the cli), and while the workspace was rebuilding I got a series of error dialogs. The error log afterwards shows several instances of this:

```
eclipse.buildId=4.16.0.202209151144
java.version=18.0.2
java.vendor=Eclipse Adoptium
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=fr_FR
Framework arguments:  -product org.springframework.boot.ide.branding.sts4 -cssTheme none
Command-line arguments:  -os linux -ws gtk -arch x86_64 -product org.springframework.boot.ide.branding.sts4 -cssTheme none

This is a continuation of log file XXX/wks/.metadata/.bak_0.log
Created Time: 2022-09-22 09:01:41.642

org.eclipse.m2e.core
Error
Thu Sep 22 15:55:38 CEST 2022
Unexpected internal Error occured

java.util.ConcurrentModificationException
	at java.base/java.util.HashMap.computeIfAbsent(HashMap.java:1221)
	at java.base/java.util.Collections$SynchronizedMap.computeIfAbsent(Collections.java:2760)
	at org.eclipse.m2e.core.internal.project.registry.ProjectRegistryManager.lambda$25(ProjectRegistryManager.java:1036)
	at java.base/java.util.Map.computeIfAbsent(Map.java:1054)
	at org.eclipse.m2e.core.internal.project.registry.ProjectRegistryManager.getMavenProject(ProjectRegistryManager.java:1036)
	at org.eclipse.m2e.core.internal.project.registry.MavenProjectFacade.getMavenProject(MavenProjectFacade.java:269)
	at org.eclipse.m2e.jdt.internal.DownloadSourcesJob.downloadMaven(DownloadSourcesJob.java:248)
	at org.eclipse.m2e.jdt.internal.DownloadSourcesJob.downloadFilesAndPopulateToUpdate(DownloadSourcesJob.java:220)
	at org.eclipse.m2e.jdt.internal.DownloadSourcesJob.lambda$0(DownloadSourcesJob.java:149)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.executeBare(MavenExecutionContext.java:350)
[...]

PR #928, opened 2022-09-23
Use Guava cache for MavenProjectCache

mickaelistria:
Easier to maintain, and thread safe.


laeubi:
Just wanted to note that @HannesWell previously has replaced Guava : https://github.com/eclipse-m2e/m2e-core/commit/55b47da6847baa69fd8410dcfcbdbf92cf0a5a51 but no idea if there was any issue with that.

laeubi:
> [Use Maven 3.0.0-SNAPSHOT](https://github.com/eclipse-m2e/m2e-core/pull/928/commits/5fd9c80c73c2e81dca752ce3653b3afea1a78d3e)

Isn't maven 3.0 a bit outdated ;-)

mickaelistria:
> Isn't maven 3.0 a bit outdated ;-)

I've used it to more easily switch to 3.0 when it's released.

laeubi:
> > Isn't maven 3.0 a bit outdated ;-)
> 
> I've used it to more easily switch to 3.0 when it's released.

**Maven** 3.0 was already released 2010 :-)
 https://maven.apache.org/docs/history.html#maven-3-0-x

mickaelistria:
Ah got it!

HannesWell:
> Just wanted to note that @HannesWell previously has replaced Guava : [55b47da](https://github.com/eclipse-m2e/m2e-core/commit/55b47da6847baa69fd8410dcfcbdbf92cf0a5a51) but no idea if there was any issue with that.

No particular problem IIRC. Mainly just to get one step closer to have less Guava.

Nevertheless, as noted in https://github.com/eclipse-m2e/m2e-core/issues/924#issuecomment-1256032691 I don't think this solves the problem.

mickaelistria:
I'm merging this one as I have high enough hope it improves things; and the effect on code maintainability seems overall positive. If this doesn't fix the initial issue and someone has concerns with this approach, we can revert later.

Issue #738, opened 2022-05-24
logback RCE CVE-2021-42550


freedom1b2830:
https://github.com/advisories/GHSA-668q-qrv7-99fm

[scanner](https://github.com/logpresso/CVE-2021-44228-Scanner)
args:[--scan-log4j1 --scan-logback --scan-zip]

[?] Found CVE-2021-42550 (logback 1.2.7) vulnerability in $ECLIPSE_DIR/plugins/ch.qos.logback.classic_1.2.3.v20200428-2012.jar, logback N/A
[?] Found CVE-2021-42550 (logback 1.2.7) vulnerability in $HOME/.p2/pool/plugins/ch.qos.logback.classic_1.1.2.v20171220-1825.jar, logback N/A

https://github.com/eclipse-equinox/equinox.bundles/issues/46

HannesWell:
Thanks for your report.
At the moment the logback dependencies are provided by Eclipse Orbit.
The version either has to be updated there or we could obtain logback-classic/core directly from Maven-Central.
But since orbit also provides a `ch.qos.logback.slf4j` fragment to wire `ch.qos.logback.classic/core` and some dependencies to the slf4j bundle I first have to check how that can be achieved with the bundles from Maven-Central. They already provide an OSGi compliant Manifest.MF but I have to check the details since this is a tricky area.

freedom1b2830:
can you provide a link to issues Eclipse Orbit ?

akurtakov:
https://bugs.eclipse.org/bugs/enter_bug.cgi?product=Orbit

freedom1b2830:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=580000

HannesWell:
I have checked how the slf4j-api bundle from Maven-Central is wired to its bindings:
For version<2 it imports a package `org.sfl4j.impl` that the binding has to export.
For version 2 it looks like the OSGi Service Loader Mediator is used.

Any way the original logback from Maven-Central can only be used if also slf4j from Maven-Central is used. And although Eclipse Platform does not use slf4j I suspect that many other Eclipse projects do, so such change should be coordinated across SimRel.

akurtakov:
As it's CVE just update and it should contributed to M1 for September simrel, that should give enough time for others to adjust accordingly.

HannesWell:
[...]

PR #931, opened 2022-09-24
Migrate to SLF4J + Logback and Gson from Maven-Central instead of Orbit

HannesWell:
Build against slf4j and logback from Maven-Central and include them into m2e's p2 repo to contribute it to SimRel.
The version at Maven-Central is more recent and not jar-signed by the Eclipse-Foundation which helps to address the difficulties encountered in https://github.com/eclipse-m2e/m2e-core/pull/926.

Logback from Orbit uses a different approach to wire `org.slf4j.api` and `ch.qos.logback.classic` (both from Orbit) . Besides logback-core and classic it is split additionally into a third fragment `ch.qos.logback.slf4j` whose host is `org.slf4j.api` (from Orbit) and that contains and exports the `org.slf4j.impl` package of logback-classic and imports packages from the latter. This way effectively `org.slf4j.api` requires `ch.qos.logback.classic`.

In order to now have two slf4j bindings on the classpath of Maven-build launched from within the IDE and thus to avoid corresponding warnings, I have adjusted the code in `MavenEmbeddedRuntime` to not pull in dependencies of slf4j (with old and new bundle name).

Fixes https://github.com/eclipse-m2e/m2e-core/issues/738

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/launch/MavenEmbeddedRuntime.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/launch/MavenEmbeddedRuntime.java
@@ -120,7 +120,9 @@ public class MavenEmbeddedRuntime extends AbstractMavenRuntime {
       addBundleClasspathEntries(allEntries, mavenRuntimeBundle, true);
       allEntries.add(getEmbeddedSLF4JBinding(mavenRuntimeBundle));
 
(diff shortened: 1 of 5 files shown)


HannesWell:
I also created https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/588 to coordinate the logback migration through the whole Eclipse Platform/IDE.

HannesWell:
I wonder why those Eclipse artifacts are not inherently vetted? is there an issue with the lookup logic?

PR #932, opened 2022-09-24
Build with Tycho 3.0

HannesWell:
Build with Tycho 3.0.

This PR also is a draft to test Tycho's build extension with Tycho 3.0.
It is intended to be submitted after the Tycho 3.0.0 GA release withouth SNAPSHOT versions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ src-gen/
 .DS_Store
 .polyglot*
 .tycho-consumer-pom.xml
+pom.tycho
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -3,6 +3,6 @@
   <extension>
     <groupId>org.eclipse.tycho.extras</groupId>
     <artifactId>tycho-pomless</artifactId>
-    <version>2.7.5</version>
+    <version>3.0.0</version>
   </extension>
 </extensions>
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,7 +22,7 @@ pipeline {
 			steps {
 				sh 'mvn clean generate-sources -f m2e-maven-runtime/pom.xml -B -Dtycho.mode=maven -Pgenerate-osgi-metadata'
 				wrap([$class: 'Xvnc', useXauthority: true]) {
-					sh 'mvn clean verify -f pom.xml -B -Dmaven.test.error.ignore=true -Dmaven.test.failure.ignore=true -Peclipse-sign,its -Dtycho.surefire.timeout=7200 -Dtycho.debug.artifactcomparator=true'
+					sh 'mvn clean verify -f pom.xml -B -Dmaven.test.error.ignore=true -Dmaven.test.failure.ignore=true -Peclipse-sign,its -Dtycho.surefire.timeout=7200'
 				}
 			}
 			post {
--- a/org.eclipse.m2e.logback/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.logback/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.m2e.logback;singleton:=true
 Bundle-Version: 2.1.0.qualifier
-Bundle-Name: M2E Logback Appender
+Bundle-Name: M2E Logback Appender and Configuration
 Bundle-Vendor: Eclipse.org - m2e
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Fragment-Host: ch.qos.logback.classic;bundle-version="1.2.0"
--- a/pom.xml
+++ b/pom.xml
@@ -26,7 +26,7 @@
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-		<tycho-version>2.7.5</tycho-version>
+		<tycho-version>3.0.0</tycho-version>
 
 		<tycho.testArgLine>-Xmx800m</tycho.testArgLine>
 		<tycho.surefire.timeout>7200</tycho.surefire.timeout>
@@ -239,6 +239,7 @@
 					</baselineRepositories>
 					<baselineReplace>common</baselineReplace>
 					<baselineMode>${tycho.p2.baselineMode}</baselineMode>
+					<writeComparatorDelta>true</writeComparatorDelta>
 					<ignoredPatterns>
 						<pattern>META-INF/ECLIPSE_.RSA</pattern>
 						<pattern>META-INF/ECLIPSE_.SF</pattern>
@@ -311,7 +312,7 @@
 							<version>${tycho-version}</version>
 						</dependency>
 						<dependency>
-							<groupId>org.eclipse.tycho.extras</groupId>
+							<groupId>org.eclipse.tycho</groupId>
 							<artifactId>tycho-buildtimestamp-jgit</artifactId>
 							<version>${tycho-version}</version>
 						</dependency>
@@ -367,11 +368,6 @@
 		<pluginRepository>
 			<id>cbi-releases</id>
 			<url>https://repo.eclipse.org/content/repositories/cbi-releases/</url>
-		</pluginRepository>
-		<pluginRepository>
-			<id>tycho-snapshots</id>
-			<url>https://repo.eclipse.org/content/repositories/tycho-snapshots/</url>
-      <snapshots><enabled>true</enabled></snapshots>
 		</pluginRepository>
 	</pluginRepositories>
 


HannesWell:
I assumed due to https://github.com/eclipse-tycho/tycho/issues/676 that the `tycho.mode=maven` should still work?

HannesWell:
With Tycho 3.0.0-SNAPSHOT the license vetting fails (not only in the CI where it fails because the snapshot repo is not available to obtain the extension).
One can reproduce this by checking out this PR and adding the following `pluginRepository` in the `pluginRepositories` section of M2E's root `pom.xml`:
```
		<pluginRepository>
			<id>dash-licenses-snapshots</id>
			<url>https://repo.eclipse.org/content/repositories/dash-licenses-snapshots/</url>
		</pluginRepository>
```
Then run:
`mvn -U -ntp org.eclipse.dash:license-tool-plugin:license-check -Ddash.fail=true`

And the build fails because of unvetted dependencies:
```
[INFO] --- license-tool-plugin:0.0.1-SNAPSHOT:license-check (default-cli) @ m2e-core ---
[INFO] Querying Eclipse Foundation for license data for 500 items.
[INFO] Found 349 items.
[INFO] Querying Eclipse Foundation for license data for 373 items.
[INFO] Found 365 items.
[INFO] Querying ClearlyDefined for license data for 42 items.
[INFO] Found 42 items.
[INFO] License information could not be automatically verified for the following content:
[INFO] 
[INFO] p2/orbit/p2.eclipse.plugin/org.osgi.service.repository/1.1.0.201505202024
[INFO] p2/orbit/p2.p2.installable.unit/org.eclipse.equinox.executable_root.cocoa.macosx.x86_64/3.8.1800.v20220720-1949
[INFO] p2/orbit/p2.p2.installable.unit/org.eclipse.equinox.executable_root.gtk.linux.x86_64/3.8.1800.v20220720-1949
[INFO] p2/orbit/p2.p2.installable.unit/org.eclipse.equinox.executable_root.win32.win32.x86_64/3.8.1800.v20220720-1949
[INFO] p2/orbit/p2.p2.installable.unit/org.eclipse.platform_root/4.25.0.v20220831-1800
[INFO] p2/orbit/p2.p2.installable.unit/org.eclipse.rcp_root/4.25.0.v20220831-1800
[INFO] 
[INFO] This content is either not correctly mapped by the system, or requires review.
[INFO] Summary file was written to: C:\dev\git\eclipse.m2e-core\target\dash\summary
[ERROR] Dependency license check failed. Some dependencies need to be vetted.
[INFO] ------------------------------------------------------------------------
```

I already requested a review of those (although they should be vetted inherently), but the vetting system does not understand them:
https://github.com/eclipse-m2e/m2e-core/pull/931#issuecomment-1256949793

When comparing the vetting-summary under target/dash/summary one can see that much more dependencies are checked with Tycho 3.0. Therefore I wonder if Tycho 3.0 is providing insufficient information or if simply more information are provided than before and the dash-license  tools simply can't handle them?

HannesWell:
And the test failures on Jenkins and GH also happen with Tycho 3.0.0-SNAPSHOT.
With Tycho 2.7.5 every build is green:
https://github.com/eclipse-m2e/m2e-core/pull/942

laeubi:
> When comparing the vetting-summary under target/dash/summary one can see that much more dependencies are checked with Tycho 3.0. Therefore I wonder if Tycho 3.0 is providing insufficient information or if simply more information are provided than before and the dash-license tools simply can't handle them?

The license vetting tool is depending on Tycho internals and *guess* things, so we can't really fix such "abuse" and the license tool might need adjustments here. Especial without not a single [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) covering the dash-license tooling ...

mickaelistria:
The `p2.p2.installable.unit` seems strange, doesn't it?

laeubi:
> The `p2.p2.installable.unit` seems strange, doesn't it?

See 
- https://github.com/eclipse-tycho/tycho/issues/784

at least not strange enough for anyone bothering to fix this (yet) :-)

HannesWell:
> > When comparing the vetting-summary under target/dash/summary one can see that much more dependencies are checked with Tycho 3.0. Therefore I wonder if Tycho 3.0 is providing insufficient information or if simply more information are provided than before and the dash-license tools simply can't handle them?
> 
> The license vetting tool is depending on Tycho internals and _guess_ things, so we can't really fix such "abuse" and the license tool might need adjustments here. Especial without not a single [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) covering the dash-license tooling ...

Makes sense. I created https://github.com/eclipse/dash-licenses/issues/183 so that the dash license tool hopefully addresses this problems.

> > The `p2.p2.installable.unit` seems strange, doesn't it?
> 
> See
> 
>     * [Duplicate p2 prefix for p2-installable-unit eclipse-tycho/tycho#784](https://github.com/eclipse-tycho/tycho/issues/784)
> 
> 
> at least not strange enough for anyone bothering to fix this (yet) :-)

I was wondering as well, but about the entire prefix. For comparison other deps start with `maven/mavencentral/...`:
https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/2692
Or is this simply because those are from a p2-repo and not from Maven-Central?

laeubi:
> Or is this simply because those are from a p2-repo and not from Maven-Central?

Yes, P2 items do not have a groupId and thus getting one **internally generated** by Tycho (that is `p2.<packaging type replaced hyphens by dots>`)

HannesWell:
> > Or is this simply because those are from a p2-repo and not from Maven-Central?
> 
> Yes, P2 items do not have a groupId and thus getting one **internally generated** by Tycho (that is `p2.<packaging type replaced hyphens by dots>`)

Makes sense. 

But lets put the license-vetting aside, do you know why the tests fail? Have there been some changes in tycho-surefire?

laeubi:
Tycho includes various upgrades of dependencies and changes/improvement in general but no problems are reported regarding surefire until now so I have no real guess here.

mickaelistria:
> https://download.eclipse.org/releases/2022-09/202209141001/plugins/

I suspect the new `-s` in Jenkinsfile cause that the default settings provided by the CI are overridden and that can cause some failures later. Can you please try without changing those settings?

HannesWell:
> > https://download.eclipse.org/releases/2022-09/202209141001/plugins/
> 
> I suspect the new `-s` in Jenkinsfile cause that the default settings provided by the CI are overridden and that can cause some failures later. Can you please try without changing those settings?

At least the Archetype related tests are failing because they try to use the Tycho 3.0.0-SNAPSHOT extension.
IIRC we already had that issue during previous Tycho-SNAPSHOT tests, but I don't remember why the archetype build launched to create a archetype project uses the Tycho-extension from the M2E build if executed in a surefire test? If I run the test in the IDE, Tycho is not used.
@laeubi can you explain that (again)? For now I added the builds console print-out in case the archetype process exited with non-zero. Should we keep that? At least this makes it simpler to understand why the build is failing when trying out Tycho snapshots.
Ideally the tests would not be affected by that at all, but I assume that if it where simple that would already be the case.

waynebeaton:
I've tweaked the EF IP Service to be a bit more forgiving. AFAICT, it is correctly identifying all of the identified components.

laeubi:
> It's the double "p2" that's messing up the license tool.
> 
> Specifically, when the tool asks the EF IP service about the entry, the service doesn't know what to do about the "p2.p2". It's relatively easy to have it handle the case. But... I'll admit that when I saw it, I assumed that it was bogus.
> 
> Should I extend it to support an arbitrary number of "p2"s? (sorry... couldn't resist)

Well actually nothing has changed since last time we had issues with the license tool and the group-id:

1. This is Tycho internals so one should not make too much assumptions about the format
2. The only thing one _might_ relay on is that a P2 maven dependency start with `p2.` following any arbitrary string, even "bogus" ones, because that the only thing that Tycho uses internally as well to find out if it is a "true" maven dependency or not.

So I don't know what is tries with whatever comes after the `p2.` or why this makes it fails, one better should not assume anything about that or should at least extra-defensive-programming style there.

@HannesWell  The best solution would be to ask EF to add the snapshot repro to the global config of the Jenkins

HannesWell:
> @HannesWell The best solution would be to ask EF to add the snapshot repro to the global config of the Jenkins

That would fix the symptom, but why does Tycho even participate in the archetype build? If I understand that correctly it actually has nothing to do with it? And then it even uses the version of Tycho that is running in the build of M2E, but that should be irrelevant for a maven build launched in a surefire-test execution that only runs the `archetype:general` goal.
Or do I misunderstand something?

laeubi:
I can only assume Tycho is somewhere referenced using a version property and this is slipping into the test run from the commandline...

waynebeaton:
> one better should not assume anything about that or should at least extra-defensive-programming style there.

One has to assume something and it has nothing to do with defensive programming styles.

Getting matches correct is a bit of an art form. Too narrow and you end up dealing with a lot of false negatives. Too wide and you end up missing stuff. Tycho is not the only project using this service.

FWIW, there is a way to tell the Eclipse Dash License Tool to to ignore certain group ids... `-DexcludeGroupIds=...`

HannesWell:
> I've tweaked the EF IP Service to be a bit more forgiving. AFAICT, it is correctly identifying all of the identified components.

Thank you Wayne for fixing the license check.

> I can only assume Tycho is somewhere referenced using a version property and this is slipping into the test run from the commandline...

I suspected the same but hoped for a hot hint. :)
Nevertheless after some debugging I found out that the `maven.multiModuleProjectDirectory` of the maven-build launched to generate the archetype-project is set to the root of my m2e git-repo (at it is likely the same for the CI). The reason is that starting at the `MavenLaunchConstants.ATTR_POM_DIR` the `MavenLaunchDelegate` searches for the `.mvn` folder in the parent. And since in the Maven build of M2E the archetype project is created within the targets-folder of the test project it finds the `.mvn` folder for m2e itself.
I added a little hack to stop the search immediately, but that should be improved before it can be considered for submission.

HannesWell:
Looks like all build failures vanished with the release. Great, so this can be submitted.
I will create another PR to adjust the archetype-project creation to make the build more robust when it comes to use snapshot versions.

Issue #927, opened 2022-09-23
Lots of fakerepo folders because of AbstractMavenArchiverConfigurator


laeubi:
Since upgrade to the new m2e release my temporary folder is getting spammed with hundreds of "fakerepoNNNNN" folders, becaus `AbstractMavenArchiverConfigurator` tries to do some "workaround" i don't fully understand.

Nerveless this creates exhaustive garbage in the temp directory as it seems each call generates even a new "fake" and it is never cached and they are even never cleared!

As the files inside the fakerepo are even never seem to contain any content, I like to propose the following fix:

1. Always use only one "fake" directory for the whole life-time of m2e
2. Delete this "fake" after shutdown of eclipse
3. Use the Bundle private area `org.osgi.framework.Bundle.getDataFile(String)` instead of the temp directory...

laeubi:
after a few hours working I got > 2000 "fakerepo" folders now!

PR #934, opened 2022-09-24
Create a single fake repo in the bundles private area

laeubi:
Fix https://github.com/eclipse-m2e/m2e-core/issues/927

### DIFF
--- a/org.eclipse.m2e.feature/feature.xml
+++ b/org.eclipse.m2e.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.m2e.feature"
       label="%featureName"
-      version="2.0.5.qualifier"
+      version="2.0.6.qualifier"
       provider-name="%providerName"
       plugin="org.eclipse.m2e.core"
       license-feature="org.eclipse.license"
--- a/org.eclipse.m2e.mavenarchiver/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.mavenarchiver/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: m2e connector for the mavenarchiver and pom properties
 Bundle-SymbolicName: org.eclipse.m2e.mavenarchiver;singleton:=true
-Bundle-Version: 2.0.1.qualifier
+Bundle-Version: 2.0.2.qualifier
 Bundle-Vendor: Eclipse.org - m2e
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Require-Bundle: org.eclipse.core.resources;bundle-version="[3.4.0,4.0.0)",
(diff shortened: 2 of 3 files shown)


laeubi:
@HannesWell @mickaelistria any ideas? GH+Jenkins show different failures and all these don't seem directly related here (or I don't understand the connection ... )

HannesWell:
> @HannesWell @mickaelistria any ideas? GH+Jenkins show different failures and all these don't seem directly related here (or I don't understand the connection ... )

Looks similar to the failures in https://github.com/eclipse-m2e/m2e-core/pull/931 and the master build is also failing with the same issue:
https://ci.eclipse.org/m2e/job/m2e/job/master/397/

Maybe it is related to that we use openjdk-latest (which is Java-19). I'm about to create a PR to check that.
Or it is the Tycho 3.0-SNAPSHOT. Besides that nothing changed since the last successful builds as far as I now.

HannesWell:
> Maybe it is related to that we use openjdk-latest (which is Java-19). I'm about to create a PR to check that. Or it is the Tycho 3.0-SNAPSHOT. Besides that nothing changed since the last successful builds as far as I now.

Lets see what https://github.com/eclipse-m2e/m2e-core/pull/936 does.

HannesWell (commented):
Regarding this change, it looks good to me.

I just wonder if it could be problematic if the file is used concurrently?

laeubi:
@HannesWell thanks for the build fix. The file is only empty and I actually think we should be able to get rid of this completely but for now I'm fine with not getting spammed all the day ;-)

HannesWell:
> @HannesWell thanks for the build fix. The file is only empty and I actually think we should be able to get rid of this completely but for now I'm fine with not getting spammed all the day ;-)

You're welcome. Sound good 👍🏽 I have to admit I usually don't monitor my temp-dir. But I agree it is annoying if the file-system is spammed.

laeubi:
Well I don't really "monitor" the temp dir but as Linux has a single temp dir `/tmp` that is easily accessible (and that is cleared on restart) I often use this to download... well temporary stuff ... like test-cases, build-logs and so on, so having thousands(!) of folders there makes it really hard to find anything :-)

HannesWell:
> Well I don't really "monitor" the temp dir but as Linux has a single temp dir `/tmp` that is easily accessible (and that is cleared on restart) I often use this to download... well temporary stuff ... like test-cases, build-logs and so on, so having thousands(!) of folders there makes it really hard to find anything :-)

Understand. 👍🏽 I'm often on Windows and therefore not so much used to activly use the temp-dir. But I would be annoyed as well if e.g. my Downloads folder would be spammed. :)

laeubi:
> But I would be annoyed as well if e.g. my Downloads folder would be spammed. :)

I have recently delete around 3.000 fakerepo dirs after half of a developers day :-\

Even if users don't notice this, this also take up space and creates disk I/O ...

sman-81:
> I have recently delete around 3.000 fakerepo dirs after half of a developers day :-\

same here

`rm -rf /tmp/fakerepo*` amongst my most frequently used commands these days ;)

PR #943, opened 2022-09-26
Merge m2e.logback.configuration into .appender

HannesWell:
and name the resulting fragment 'o.e.m2e.logback'.

Make the `LogPlugin` from m2e.logback.configuration a logback Configurator that is called by the Logback configuration during initialization.


HannesWell:
Both Plugins involved don't export any API and are usually installed via features, so it isn't an issue to delete/rename them since updating the feature will adjust existing installations accordingly.

I tested this change from the IDE and it worked well and will test this after submission in my dev IDE.
In the past there was not much interest in the logback area so I don't export more comments here and think this is good to go.

PR #949, opened 2022-09-28
Replace Properties by Map in IMavenLauncher and ArchetypeGenerator

HannesWell:
And simplify ArchetypeGenerator a little bit.

@laeubi you created that class. What do you think?

### DIFF
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeGenerator.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeGenerator.java
@@ -20,8 +20,9 @@ import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
 import java.util.Collection;
+import java.util.LinkedHashMap;
 import java.util.List;
-import java.util.Properties;
+import java.util.Map;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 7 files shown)


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "m2e-core-tests"]
 	path = m2e-core-tests
-	url = https://github.com/tesla/m2e-core-tests.git
-	branch = master
+	url = https://github.com/HannesWell/m2e-core-tests.git
```
HannesWell, line 3: Has to be reverted before this can be merged.
Requires https://github.com/tesla/m2e-core-tests/pull/185

laeubi:
> And simplify ArchetypeGenerator a little bit.
> 
> @laeubi you created that class. What do you think?

I'm not sure if I'm the *original author* of that class or just *moved and refactored* it, but Map seems fine and I don't see any drawback with that.

The only improvement I could think of would be using a `LinkedHashMap` instead of `HashMap` so one has some deterministic order later on (just in case we would print that out somewhere).

.gitmodules

HannesWell:
> > And simplify ArchetypeGenerator a little bit.
> > @laeubi you created that class. What do you think?
> 
> I'm not sure if I'm the _original author_ of that class or just _moved and refactored_ it, but Map seems fine and I don't see any drawback with that.

You were at least the last one that worked intensively in that area and your name is in the copyright header at least of some of the affected interfaces. Therefore I assumed you are maybe interested in this. :) But I'm glad you like it.

> 
> The only improvement I could think of would be using a `LinkedHashMap` instead of `HashMap` so one has some deterministic order later on (just in case we would print that out somewhere).

Good suggestion. I changed it at the two places where a Map is created. The static Map factories intentionally have a changing iteration order, but it is then up to the caller to use a implementation that preserves insertion order if it is of interest.

PR #851, opened 2022-07-09
Migrate o.e.m2e.editor.tests from m2e-core-tests submodule to this repo

HannesWell:
As discussed in https://github.com/tesla/m2e-core-tests/pull/175#issuecomment-1174655604, this PR moves the `o.e.m2e.editor.tests` and their resources from the `tesla/m2e-core-tests` sub-module to the project with the same name in this repo and adapt to removal of project 'o.e.m2e.editor.tests' in the 'm2e-core-tests' submodule.

I'm going to create a CQ to get a clearance from the IP-team for this move.

PR #950, opened 2022-09-28
[Build] Remove now obsolete 'uts' Maven profile

HannesWell:
Since o.e.m2e.editor.tests is part of this repo (https://github.com/eclipse-m2e/m2e-core/pull/851) the 'uts' profile has no effect anymore. Therefore it can be removed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -435,12 +435,6 @@
 			</build>
 		</profile>
 		<profile>
-			<id>uts</id>
-			<modules>
-				<module>m2e-core-tests</module>
-			</modules>
-		</profile>
-		<profile>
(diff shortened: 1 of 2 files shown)


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "m2e-core-tests"]
 	path = m2e-core-tests
-	url = https://github.com/tesla/m2e-core-tests.git
-	branch = master
+	url = https://github.com/HannesWell/m2e-core-tests.git
```
HannesWell, line 3: This has to be reverted before merg, but requires
https://github.com/tesla/m2e-core-tests/pull/186

PR #956, opened 2022-09-30
fix: workspace resolution broken

### LINKED ISSUE #946: Workspace resolution don't work anymore
Seems this is base on the workspacestate.properties file which is not generated anymore.

jeffmaury:
Fixes the proper bindings between components to get WorkspaceStateWriter registered as a IMavenProjectChangedListener

Fixes #946

Signed-off-by: Jeff MAURY <jmaury@redhat.com>

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/WorkspaceStateWriter.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/WorkspaceStateWriter.java
@@ -46,7 +46,7 @@ import org.eclipse.m2e.workspace.MutableWorkspaceState;
  * Maintains map file of maven artifacts present in workspace.
  */
 
-@Component(service = {IMavenProjectChangedListener.class})
+@Component(service = {IMavenProjectChangedListener.class}, immediate = true)
 public class WorkspaceStateWriter implements IMavenProjectChangedListener {
   private static QualifiedName PPROP_EXTENSION = new QualifiedName(WorkspaceStateWriter.class.getName(), "extension"); //$NON-NLS-1$
 
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/WorkspaceStateWriter.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/WorkspaceStateWriter.java
@@ -38,6 +38,7 @@ import org.eclipse.m2e.core.embedder.ArtifactKey;
 import org.eclipse.m2e.core.internal.project.registry.MavenProjectManager;
 import org.eclipse.m2e.core.project.IMavenProjectChangedListener;
 import org.eclipse.m2e.core.project.IMavenProjectFacade;
+import org.eclipse.m2e.core.project.IMavenProjectRegistry;
 import org.eclipse.m2e.core.project.MavenProjectChangedEvent;
 import org.eclipse.m2e.workspace.MutableWorkspaceState;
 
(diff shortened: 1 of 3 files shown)


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectManager.java
```diff
@@ -47,7 +47,7 @@
  * dependency between ProjectRegistryRefreshJob and ProjectRegistryManager. Otherwise, all requests are forwarded to
  * ProjectRegistryManager as is.
  */
-@Component(service = {IMavenProjectRegistry.class})
+@Component(service = {MavenProjectManager.class, IMavenProjectRegistry.class})
```
laeubi, line 50: Is MavenProjectManager used anywhere as a service? Actually we should avoid registering implementations as services and instead use e.g. Whiteboardpattern instead.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/WorkspaceStateWriter.java
```diff
@@ -46,7 +46,7 @@
  * Maintains map file of maven artifacts present in workspace.
  */
 
-@Component(service = {IMavenProjectChangedListener.class})
+@Component(service = {IMavenProjectChangedListener.class}, immediate = true)
```
laeubi, line 49: I think we better should have a component consuming hte service than making it immedate..

laeubi (changes requested):

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectManager.java
jeffmaury: Is is used by WorkspaceStateWriter so that's why I have to add it there

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/WorkspaceStateWriter.java
jeffmaury: Then which one ? The bundle ? Better defining is as an extension that through DS then

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectManager.java
laeubi: Then probably this has to be reversed, `WorkspaceStateWriter` should not use the `MavenProjectManager` but `MavenProjectManager` should collect all `IMavenProjectChangedListener`

laeubi: Then I don't see how `MavenProjectManager` is ever needed and we probobly better find an alternative way to get workspace state file location (even duplicating some code or using a static method to build its name).

laeubi (approved):
Looks good, but what happens to the testcase from previous commit?

jeffmaury:
> Looks good, but what happens to the testcase from previous commit?

Not sure to get it ? It should be part of the PR as well and should be ok (waiting for the Jenkins job)

laeubi:
> > Looks good, but what happens to the testcase from previous commit?
> 
> Not sure to get it ? It should be part of the PR as well and should be ok (waiting for the Jenkins job)

Never mind I only looked at the last commit... :-\

mickaelistria:
Thanks guys!

PR #958, opened 2022-09-30
Make archetype-project creation more robust

HannesWell:
Use a separate working-directory for archetype-project creation and ensure that it is used as the maven.multiModuleProjectDirectory. The parents of the working-directory are searched for a .mvn-directory and if one is found, it is used as maven.multiModuleProjectDirectory.

Consequently if one creates an archetype-project nested inside an existing multimodule-project the parent that contains the .mvn-folder is used as multiModuleProjectDirectory of the archetype-project creation build. This has the consequences that the maven-extensions defined in that mmPDir participate in the archetype build.
This is unnecessary and can cause problems if those extensions are loaded from a repository that is specified in a specified maven-settings file. This happens for example (and causes build failures) when we test Tycho-snapshots from the tycho-snapshot repo.

And add printout of archetype-project creation build in case of failure.

@laeubi any objections?
I can demonstrate/test that this works by temporarily using the Tycho 3.1 snapshot.

### DIFF
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeGenerator.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypeGenerator.java
@@ -13,12 +13,11 @@
 
 package org.eclipse.m2e.core.ui.internal.archetype;
 
+import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
-import java.io.InputStream;
-import java.net.URL;
 import java.nio.file.Files;
-import java.nio.file.StandardCopyOption;
+import java.nio.file.Path;
 import java.util.Collection;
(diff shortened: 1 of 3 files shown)


laeubi:
Looks like a good idea, even better would if we can make the archetype-project have an option to specify an output folder, but in the meanwhile I think it is a good idea.

HannesWell:
> Looks like a good idea, even better would if we can make the archetype-project have an option to specify an output folder, but in the meanwhile I think it is a good idea.

Good, then lets see if that really works.
Can't one already specify the `outputDirectory`? I have to admit I never used it from the UI but there is the location parameter that controls where the project is created. Or do you mean let the user specify the workingDir? Then I wonder what benefits this would have?

HannesWell:
Looks good. All builds (except for the License-Vetting) succeeded.
So in the future at least the archetype-generator should not make problems anymore when testing Tycho snapshots.

PR #961, opened 2022-10-01
Extract referenced slf4j bundle for Maven runtime classpath again

HannesWell:
This is necessary because even the slf4j-bundle from Maven-Central is jar-singed as part of the m2e build when the m2e-repo is assembled.

If the slf4j-jar is not extraced launching the embedded maven-runtime will fail with errors like:
```
class "org.slf4j.MavenSlf4jFriend"'s signer information does not match
signer information of other classes in the same package
```
The reason for such an error is that the 'slf4j-api' jar and the 'maven-slf4j-provider' jar both provide classes for the package 'org.slf4j'. But all jars that provide classes for the same package must have the same jar-signer. And while the 'slf4j-api' jar is referenced as bundle and therefore signed in the m2e build (although it is originated from Maven-Central), the 'maven-slf4j-provider' jar embedded into the m2e.maven.runtime is not jar signed (because nested jars are not signed).
Extracting the jar and providing them from a directory This check is not performed if a jar is extracted and added to the classpath as directory. Therefore only the slf4j jar is extracted again. Effectively this reverts #926 for slf4j.

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/Bundles.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/Bundles.java
@@ -44,7 +44,7 @@ public class Bundles {
 
   private static final Logger log = LoggerFactory.getLogger(Bundles.class);
 
-  public static List<String> getClasspathEntries(Bundle bundle) {
+  public static List<String> getClasspathEntries(Bundle bundle, boolean extractJar) {
     log.debug("getClasspathEntries(Bundle={})", bundle);
     Set<String> cp = new LinkedHashSet<>();
     if(inDevelopmentMode()) {
@@ -53,13 +53,7 @@ public class Bundles {
     cp.addAll(parseBundleClasspath(bundle));
     List<String> entries = new ArrayList<>();
     for(String cpe : cp) {
-      String entry;
-      if(".".equals(cpe)) {
-        entry = FileLocator.getBundleFileLocation(bundle)
-            .orElseThrow(() -> new NoSuchElementException("Unable to locate bundle:" + bundle)).toString();
-      } else {
-        entry = getClasspathEntryPath(bundle, cpe);
-      }
+      String entry = getClasspathEntryPath(bundle, cpe, extractJar);
       if(entry != null) {
         log.debug("\tEntry:{}", entry);
         entries.add(entry);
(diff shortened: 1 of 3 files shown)


HannesWell:
I also considered to disable the signing of the third-party dependencies embedded into the m2e-repo (with are all from Maven-Central), but we have discussed this at the beginning of this year in #520 at not signing them was not an option back then. 
However I noticed that for example [com.google.guava_31.1.0.jre](https://download.eclipse.org/releases/2022-09/202209141001/plugins/com.google.guava_31.1.0.jre.jar) is not jar signed in the current SimRel release repo. Therefore I wonder if that rule has changed in the meantime. I know a lot is different now, but I'm not sure abou that particular rule.
@akurtakov can you say more?

mickaelistria:
Embedded dependencies shouldn't be signed but should instead remain equals to upstream ones. jarsigner is required for bundles that are installable units in p2 and are built at Eclipse.org. This doesn't apply to slf4j or other embedded jars.
Do you have any idea why embedded jars got signed in 1st place?

HannesWell:
> Embedded dependencies shouldn't be signed but should instead remain equals to upstream ones. jarsigner is required for bundles that are installable units in p2 and are built at Eclipse.org. This doesn't apply to slf4j or other embedded jars. Do you have any idea why embedded jars got signed in 1st place?

The embedded jars are not signed.
But the slf4j jar (and the other deps) which we get from Maven-Central and that is/are part of m2e's p2-repo as 'standalone' bundles are jarsigned because as part of the m2e build all bundles in the m2e-repo are signed.
That repo contains the slf4j and logback bundles since PR #931 in order to contribute them to the SimRel.

HannesWell:
I'm going to submit this now in order to fix launching Maven builds using M2Es embedded Maven.
If we can lift the requirement to sign third-party dependencies in the m2e-repo and if we are sure they are not singed as part of the SimRel repo build, then I can again remove the work around.

mickaelistria:
> But the slf4j jar (and the other deps) which we get from Maven-Central and that is/are part of m2e's p2-repo as 'standalone' bundles are jarsigned because as part of the m2e build all bundles in the m2e-repo are signed.

It sounds like the dual symptom of what I expressed in https://github.com/eclipse-m2e/m2e-core/pull/520#issuecomment-1019462837 . If we sign the artifacts we produce, and not the sign content, then we wouldn't erroneously sign 3rd party content.

HannesWell:
> > But the slf4j jar (and the other deps) which we get from Maven-Central and that is/are part of m2e's p2-repo as 'standalone' bundles are jarsigned because as part of the m2e build all bundles in the m2e-repo are signed.
> 
> It sounds like the dual symptom of what I expressed in [#520 (comment)](https://github.com/eclipse-m2e/m2e-core/pull/520#issuecomment-1019462837) . If we sign the artifacts we produce, and not the sign content, then we wouldn't erroneously sign 3rd party content.

Yes, if we would only sign each Maven-module/project we build and not the entire repo in the end, then this work-around would not be necessary. But at least in https://github.com/eclipse-m2e/m2e-core/pull/520#issuecomment-1020872217 it was stated that the latter is necessary. But if that is what I'm wondering about.

laeubi:
I think it this no longer required to sign third party jars (and maybe not third party ones?) so we can remove this from the m2e build and instead "sign" them using PGP...
Beside that, one could even think to drop this "embedding of jars" and instead wrap them with manifest infos using the target support and make them a fragment of the m2e.maven.runtime, with extensible API header for PDE, and a p2.inf for always install them together with that fragment.

PR #962, opened 2022-10-01
Remove pom for repository

mickaelistria:
More importantly, remove the extra signing step that was erroneously modifying external bundles; and uselessly signing bundles that are built in the same reactor, but which are already jarsigned as part of their own build.

PR #963, opened 2022-10-02
Dont extract slf4j (again)

HannesWell:
Now that dependencies are not signed anymore (thanks to #962), the workaround to prevent `signer information does not match` exceptions from #961 can be removed again.

But keep the part of #961 about canonicalizing the paths of referenced bundle-jars.
Furthermore improve the Maven-Console logs about the commands of launched builds and advance the debug-output about the assembled classpath of a maven-runtime from debug to info, so that it shows up once in the Maven-Console. Since the story about non-extracted dependencies in conjunction with deps from Maven or Orbit seems to be a bit sensitive, I think it is good to have another source of information relatively easy accessible.

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/Bundles.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/Bundles.java
@@ -44,7 +44,7 @@ public class Bundles {
 
   private static final Logger log = LoggerFactory.getLogger(Bundles.class);
 
-  public static List<String> getClasspathEntries(Bundle bundle, boolean extractJar) {
+  public static List<String> getClasspathEntries(Bundle bundle) {
     log.debug("getClasspathEntries(Bundle={})", bundle);
     Set<String> cp = new LinkedHashSet<>();
     if(inDevelopmentMode()) {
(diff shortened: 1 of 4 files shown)


HannesWell:
For those that have downloaded the latest snapshot since https://github.com/eclipse-m2e/m2e-core/pull/931, you have to make sure that the slf4j 1.7.36 jar used by your Eclipse-installation is replaced by the original one from Maven-Central not signed by Eclipse.

So if you get a `signer information does not match` exceptions when launching a maven build and the signer information in the Installation details looks like this:
grafik


... you have to replace it by the original.
If you are using Oomph you can try to use Oomph's Bundle Pool analyzer and repair the slf4j.api bundle from https://download.eclipse.org/technology/m2e/snapshots/latest/

I first tried to delete it using the Bundle Pool Analyzer and did not manage to make my m2e working again. Even a few uninstalls of m2e features did not help. In the end I simply deleted and re-installed my Eclipse.
Therefore it is probably saver to download the [slf4j-api-1.7.36.jar](https://repo1.maven.org/maven2/org/slf4j/slf4j-api/1.7.36/slf4j-api-1.7.36.jar) from M-C and replace the slf4j jar from your installation, wherever it is.

PR #606, opened 2022-03-11
[Core] delete now obsolete nexus-cache directory

HannesWell:
As suggested in https://github.com/eclipse-m2e/m2e-core/issues/509#issuecomment-1018922265 this PR enhances the Activator of `o.e.m2e.core` to delete the now obsolete nexus-cache directory on start-up (if it exists).
This can free a lot of disk-space for users that had downloaded the index (like me, where the nexus folder occupied almost 5GB per workspace!).

Besides that this PR fixes a little error in the m2e.core Manifest and bumps the `m2e-tests` sub-module reference.

PR #964, opened 2022-10-02
[Core] delete obsolete ".m2/.cache/m2e/" directory and old logback logs

HannesWell:
Similar to https://github.com/eclipse-m2e/m2e-core/pull/606 the M2E-cache at `<user-home>/.m2/.cache/m2e/` is not used anymore and can be deleted, since we don't sue the indexer anymore. 
Anyone can check the size of that folder. This usually removes a few GB of unused files that where created by m2e in the past.
Furthermore the old logs in the state location of the now renamed o.e.m2e.lockback.configuration can be deleted, which removes up to another GB of useless data.

### DIFF
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/M2EUIPluginActivator.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/M2EUIPluginActivator.java
@@ -14,6 +14,11 @@
 
 package org.eclipse.m2e.core.ui.internal;
 
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.ServiceReference;
 import org.osgi.util.tracker.ServiceTracker;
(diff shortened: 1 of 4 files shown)


fbricon:
-1 you might have users still using older versions of eclipse/m2e, that are simply evaluating a newer version of m2e. That'd be really bad.

You should show a popup asking users whether they want to delete something on their machine, that's goes beyond the scope of a given eclipse instance.

HannesWell:
> -1 you might have users still using older versions of eclipse/m2e, that are simply evaluating a newer version of m2e. That'd be really bad.
> 
> You should show a popup asking users whether they want to delete something on their machine, that goes beyond the scope of a given eclipse instance.

Good point.
I added a pop-up as you suggested. It is not very nice, but does the job and should not be visible very often.

grafik

HannesWell:
@fbricon, are you fine with this now?
Or does anybody else has more remarks?

fbricon:
How about :

```
A cache directory used by previous M2E versions was detected: <user-home>/.m2/.cache/m2e/ 
It's no longer used newer M2E versions and, unless older Eclipse installations need it, can be safely deleted.
 
                                        Delete Cache / Keep Cache
```

HannesWell:
> How about :
> 
Sounds good. Thank you.
I just swapped the `Delete` and `Keep` Buttons so that `Keep` is the default:
grafik

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/M2EUIPluginActivator.java
```diff
@@ -180,4 +199,25 @@ public ArchetypePlugin getArchetypePlugin() {
...
 
+  private static void deleteLegacyCacheDirectory(File m2eCache) {
+    new UIJob("Delete legacy M2E cache") {
+      @Override
+      public IStatus runInUIThread(IProgressMonitor monitor) {
+        int selection = MessageDialog.open(MessageDialog.QUESTION, getDisplay().getActiveShell(),
+            "Delete obsolete M2E cache?",
+            "A cache directory used by previous M2E versions was detected:\n" + m2eCache + "\n\n"
+                + "It's no longer used by newer M2E versions and, unless older Eclipse installations need it, can be safely deleted.",
+            SWT.NONE, "Keep Cache", "Delete Cache");
```
fbricon, line 210: you probably need a don't ask again choice as well, else it'll nag users who decide to keep the cache on every start

laeubi: That's really a huge effort just for users switching back-and.forth between old and new m2e ... So probably simply let them waste theri disk-space than wasting development and maintenance effort here... @mickaelistria

HannesWell: > That's really a huge effort just for users switching back-and.forth between old and new m2e ... So probably simply let them waste theri disk-space than wasting development and maintenance effort here... @mickaelistria

It was indeed not the intended to end up with that much code. :/


> I would also suggest storing the result globally for the user, so that they won't be asked for every new eclipse installation again.

They should at least be ask for each Eclipse installation, because otherwise it will probably forgotten if the last installation started to use M2E 2.

cpfeiffer: Well, IMHO that's still too much in certain environments, where people deploy new eclipse installations weekly rather than yearly. With oomph, things are set up automatically, and that popup would interrupt the automatic installation process (git clone, project import, etc.). These modal dialogs on startup (takari stats, jboss usage and now this) are not helpful in that regard.

There could be a system property to disable the popup. Or make it a passive notification instead of a modal dialog.

HannesWell: OK. I actually never wanted to make it that complicated, but before we get many complaints and since I had started it ...
You can now set the Java-system property `m2e.keep.legacy.cache` to `true` to just keep the cache.
Furthermore you can set a check-box to be not asked again.
In that case a marker file `DELETE_ME.txt` is created in the directory that explains why/when this can be deleted and that prevents the dialog from popping up subsequently.

If that is not sufficient I will just add the file to the directory so one can delete it manually if stumbling upon it.
This way maybe gigabytes of storage are kept unnecessarily occupied but at least no pop-up shows up.

HannesWell:
I consider your absent of objections as silent consent.

If you think more has to be done, it can be done in a follow up PR.
But as it was said it was never the intend to make it that complex, so if this is not fine as it is, the clean up of `~/.m2/cache/m2e` probably should be down-graded to just add a information file to the repo, which does not bother any workflow.

PR #967, opened 2022-10-05
Add project configurator for tycho packagings

### LINKED ISSUE #605: Migrate m2e-tycho to m2e
Now that we have m2e containing some PDE integration, and that a CQ was approved for m2e-tycho ( https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23882 ), it would probably make sense to migrate m2e-tycho directly into m2e.

kwin:
This closes #605


org.eclipse.m2e.pde.connector/META-INF/MANIFEST.MF
```diff
@@ -13,5 +13,7 @@ Require-Bundle: org.eclipse.m2e.core;bundle-version="[2.0.0,3.0.0)",
  org.eclipse.core.resources;bundle-version="3.16.0",
  org.eclipse.m2e.maven.runtime;bundle-version="[3.8.6,4.0.0)",
  org.eclipse.pde.core,
- org.eclipse.jdt.core
+ org.eclipse.jdt.core,
+ org.eclipse.pde.ds.annotations;bundle-version="1.2.500",
```
kwin, line 17: don't know which version range to take here

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
```diff
@@ -0,0 +1,84 @@
...
+            // see org.eclipse.pde.internal.ui.wizards.site.NewSiteProjectCreationOperation
+            if (!project.hasNature(PDE.SITE_NATURE)) {
+                CoreUtility.addNatureToProject(project, PDE.SITE_NATURE, monitor);
+            }
+        }
+    }
+
+    void applyDsConfiguration(ProjectConfigurationRequest request, IProgressMonitor monitor) throws CoreException {
+        List<MojoExecution> mojoExecutions = getTychoDsPluginMojoExecutions(request, monitor);
+        for (MojoExecution mojoExecution : mojoExecutions) {
```
kwin, line 61: if multiple executions exist we should probably just take the first one.

org.eclipse.m2e.pde.connector/META-INF/MANIFEST.MF
HannesWell: Since you are using mainly internal classes any version range that does allow a future version has the potential to break.
IIRC in that case it is recommended to just specify a lower bound as you already did.

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
HannesWell: Ideally the one 'closest' to the current project's pom.
But I don't know if there a reasonable way to know that.
So just the first one sounds reasonable.

org.eclipse.m2e.pde.connector/lifecycle-mapping-metadata.xml
```diff
@@ -51,6 +51,53 @@
...
+                <artifactId>tycho-packaging-plugin</artifactId>
+                <versionRange>[1.0.0,)</versionRange>
+                <goals>
+                    <goal>package-plugin</goal>
+                    <goal>package-feature</goal>
+                </goals>
+            </pluginExecutionFilter>
+            <action>
+                <configurator>
+                    <id>org.apache.m2e.pde.connector.configurator.tycho-java</id>
```
HannesWell, line 66:
```suggestion
                    <id>org.apache.m2e.pde.connector.configurator.tycho</id>
```
The plugin.xml does not declare a configurator with that id. Either this Id is wrong or the `plugin.xml`.

org.eclipse.m2e.pde.connector/plugin.xml
```diff
@@ -6,9 +6,21 @@
       <configurator
             class="org.eclipse.m2e.pde.connector.PDEMavenBundlePluginConfigurator"
             id="org.eclipse.m2e.pde.connector.configurator.bundle"
-            name="PDE integration for maven plugin projects">
+            name="PDE integration for Maven bundle plugin projects">
       </configurator>
+      <configurator
+            class="org.eclipse.m2e.pde.connector.TychoPackagingsConfigurator"
+            id="org.apache.m2e.pde.connector.configurator.tycho"
```
HannesWell, line 13:
```suggestion
            id="org.apache.m2e.pde.connector.configurator.tycho-java"
```
I suspect you indented to add a `-java` suffix here?

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
```diff
@@ -0,0 +1,84 @@
...
+            PDEProjectHelper.configurePDEBundleProject(project, mavenProject, monitor);
+            applyDsConfiguration(request, monitor);
+        } else if ("eclipse-feature".equals(packaging)) {
+            // see org.eclipse.pde.internal.ui.wizards.feature.AbstractCreateFeatureOperation
+            if (!project.hasNature(PDE.FEATURE_NATURE)) {
+                CoreUtility.addNatureToProject(project, PDE.FEATURE_NATURE, monitor);
+            }
+        } else if ("eclipse-repository".equals(packaging)) {
+            // see org.eclipse.pde.internal.ui.wizards.site.NewSiteProjectCreationOperation
+            if (!project.hasNature(PDE.SITE_NATURE)) {
```
HannesWell, line 53: PDE `Update-Site` projects are 'deprecated' and Tycho does not supported them anymore (IIRC, at least since Tycho 3, maybe even 2).
The `eclipse-repository` packaging is applied for projects that contain PDE products and `category.xml` files, which define a p2-repo (the successor of the PDE Update-Site). 
The site-nature is not useful here. In general in both cases a dedicated nature is not necessary because the PDE-builder doesn't process product- and category-files.

We should not add support for a legacy format here.

HannesWell (changes requested):
Thank you for this PR.
Additionally this should also help for https://github.com/eclipse-m2e/m2e-core/discussions/929.

I applied a few comments below, but in general I think this is on a good way.

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
laeubi: Yes site nature is not required here.

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
laeubi: There is always one as default and having more than one does not really make sense (you don't get better results generating the data twice).

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    void applyDsConfiguration(ProjectConfigurationRequest request, IProgressMonitor monitor) throws CoreException {
+        List<MojoExecution> mojoExecutions = getTychoDsPluginMojoExecutions(request, monitor);
+        for (MojoExecution mojoExecution : mojoExecutions) {
+            // apply PDE configuration for DS
+            IEclipsePreferences prefs = new ProjectScope(request.mavenProjectFacade().getProject())
+                    .getNode(org.eclipse.pde.ds.internal.annotations.Activator.PLUGIN_ID);
+            Xpp3Dom dom = mojoExecution.getConfiguration();
+            Xpp3Dom dsEnabled = dom.getChild("tycho.ds.enabled");
```
laeubi, line 66: This seems to be the user property but you want the xml element name (`enabled`), same below with tycho.ds.version (`version`)...

Also I think we must check for the path and should set (if not already configured)

```
generateBundleActivationPolicyLazy=true
validationErrorLevel=error
validationErrorLevel.missingImplicitUnbindMethod=error
```

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
```diff
@@ -0,0 +1,84 @@
...
+            if (dsEnabled != null && Boolean.valueOf(dsEnabled.getValue())) {
+                prefs.putBoolean(org.eclipse.pde.ds.internal.annotations.Activator.PREF_ENABLED, true);
+            } else {
+                prefs.putBoolean(org.eclipse.pde.ds.internal.annotations.Activator.PREF_ENABLED, false);
+            }
+            Xpp3Dom dsVersion = dom.getChild("tycho.ds.version");
+            if (dsVersion == null && !dsVersion.getValue().isEmpty()) {
```
laeubi, line 73: Looks wrong. maybe `||` instead?

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
```diff
@@ -0,0 +1,84 @@
...
+        for (MojoExecution mojoExecution : mojoExecutions) {
+            // apply PDE configuration for DS
+            IEclipsePreferences prefs = new ProjectScope(request.mavenProjectFacade().getProject())
+                    .getNode(org.eclipse.pde.ds.internal.annotations.Activator.PLUGIN_ID);
+            Xpp3Dom dom = mojoExecution.getConfiguration();
+            Xpp3Dom dsEnabled = dom.getChild("tycho.ds.enabled");
+            if (dsEnabled != null && Boolean.valueOf(dsEnabled.getValue())) {
+                prefs.putBoolean(org.eclipse.pde.ds.internal.annotations.Activator.PREF_ENABLED, true);
+            } else {
+                prefs.putBoolean(org.eclipse.pde.ds.internal.annotations.Activator.PREF_ENABLED, false);
```
laeubi, line 70: this should only be set if dsEnabled is explicitly given, currently one would disable any user config values

```
if (dsEnabled == null && !dsEnabled.getValue().isEmpty()) {
   ...your code goes here ...
}
```

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
```diff
@@ -0,0 +1,84 @@
...
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.m2e.core.project.configurator.AbstractProjectConfigurator;
+import org.eclipse.m2e.core.project.configurator.ProjectConfigurationRequest;
+import org.eclipse.pde.internal.core.natures.PDE;
+import org.eclipse.pde.internal.core.util.CoreUtility;
+
+public class TychoPackagingsConfigurator extends AbstractProjectConfigurator {
+
+    private static final String TYCHO_DS_PLUGIN_GROUP_ID = "org.eclipse.tycho";
```
laeubi, line 33:
```suggestion
    private static final String TYCHO_GROUP_ID = "org.eclipse.tycho";
```

org.eclipse.m2e.pde.connector/META-INF/MANIFEST.MF
kwin: In fact I only use constant which are interned at build time, so this is not a real run time requirement, but I haven't found a way to tell PDE to only load a package in the build classpath....

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
kwin: good catch regarding XML element names.
For the default values, I leave that up to https://github.com/eclipse-pde/eclipse.pde/blob/8312046eb6c566df0846f4125c2a5d41dcf0d0a3/ds/org.eclipse.pde.ds.annotations/src/org/eclipse/pde/ds/internal/annotations/PreferenceInitializer.java#L23. Path should in no case be set to the value from this plugin as this plugins writes to /target/OSGI-INF and PDE writes to /OSGI-INF.

laeubi: > writes to /target/OSGI-INF and PDE writes to /OSGI-INF.

Still the config option is always "OSGI-INF" ...

kwin: Not necessarily if someone gives an absolute path. This is not a read-only param and may be changed.

laeubi: > Not necessarily if someone gives an absolute path. This is not a read-only param and may be changed.

Then this is a configuration error. An absolute path is not useful here.

kwin: Which path do you consider useful apart from the default "OSGI-INF".

laeubi: It doesn't matter what I consider, per the specification you can place it in any folder, (e.g. one named 'components'), and you can configure this in PDE, there is an option in Tycho, so why not simply apply it as the other options as well? :-)

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
```diff
@@ -0,0 +1,131 @@
...
+        IEclipsePreferences prefs = new ProjectScope(request.mavenProjectFacade().getProject())
+                .getNode(org.eclipse.pde.ds.internal.annotations.Activator.PLUGIN_ID);
+        Xpp3Dom dsEnabled = dom.getChild("enabled");
+        final boolean isDsEnabled;
+        if (dsEnabled != null && !dsEnabled.getValue().isEmpty()) {
+            isDsEnabled = Boolean.valueOf(dsEnabled.getValue());
+            prefs.putBoolean(org.eclipse.pde.ds.internal.annotations.Activator.PREF_ENABLED, isDsEnabled);
+        } else {
+            isDsEnabled = false;
+        }
```
HannesWell, line 81:
```suggestion
		boolean isDsEnabled = false;
		if (dsEnabled != null && !dsEnabled.getValue().isEmpty()) {
			isDsEnabled = Boolean.valueOf(dsEnabled.getValue());
			prefs.putBoolean(org.eclipse.pde.ds.internal.annotations.Activator.PREF_ENABLED, isDsEnabled);
		}
```
No need for an extra else-block.

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
```diff
@@ -0,0 +1,131 @@
...
+    private boolean containsExplicitConfigurationValue(Xpp3Dom config) {
+        if (config == null) {
+            return false;
+        }
+        String value = config.getValue();
+        // check if value is a property name
+        if (value.startsWith("${")) {
+            return false;
+        }
+        return true;
```
HannesWell, line 106:
```suggestion
		// check if value is a property name
		return config != null && !config.getValue().startsWith("${");
```
Although the goal isn't to make everything as compact as possible, I think in this case a one-liner is more readable here.

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    void applyDsConfiguration(ProjectConfigurationRequest request, IProgressMonitor monitor) throws CoreException {
+        List<MojoExecution> mojoExecutions = getTychoDsPluginMojoExecutions(request, monitor);
+        if (mojoExecutions.isEmpty()) {
+            return;
+        }
+        if (mojoExecutions.size() > 1) {
+            log.error("Found more than one execution for plugin {}:{} and goal {}, only consider configuration of first one", TYCHO_GROUP_ID, TYCHO_DS_PLUGIN_ARTIFACT_ID, GOAL_DECLARATIVE_SERVICES);
```
HannesWell, line 67: This error message should not be logged to a slf4j-logger (that only logs to the Maven-Console and log-files).
Instead a warning-problem-marker should be created in the pom. Maybe at the location of the selected execution or at all ignored ones (I'm not certain what is better).

You can have a look at PDEMavenBundlePluginConfigurator.createWarningMarker(), how such markers are created. Probably that method has to be generalized.

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+        Version osgiCompliantVersion = Version.parseVersion(version);
+        if (osgiCompliantVersion.getMajor() == 1 && osgiCompliantVersion.getMinor() == 1) {
+            return DSAnnotationVersion.V1_1;
+        } else if (osgiCompliantVersion.getMajor() == 1 && osgiCompliantVersion.getMinor() == 2) {
+            return DSAnnotationVersion.V1_2;
+        } else if (osgiCompliantVersion.getMajor() == 1 && osgiCompliantVersion.getMinor() == 3) {
+            return DSAnnotationVersion.V1_3;
+        } else {
+            log.error("Unsupported DS spec version {} found, using default instead", version );
```
HannesWell, line 121: The same as above. For this problem a error/warning-marker should be created instead of just logging it.

HannesWell (commented):
I have a few more remarks below.

Furthermore please format the class with the standard Eclipse built-in formatter to align it with the other classes in this plugin. Mainly to use tabs instead of spaces for indentation. If you use the M2E-setup in a clean workspace, that should be set up by default.

In general it would be good to have tests (for all configurations in this Plug-in).
If you want to create some that would be great, otherwise I can do that in the next days.

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
```diff
@@ -0,0 +1,131 @@
...
+        IProject project = request.mavenProjectFacade().getProject();
+
+        String packaging = mavenProject.getPackaging();
+        if ("eclipse-plugin".equals(packaging) || "eclipse-test-plugin".equals(packaging)) {
+            PDEProjectHelper.configurePDEBundleProject(project, mavenProject, monitor);
+            applyDsConfiguration(request, monitor);
+        } else if ("eclipse-feature".equals(packaging)) {
+            // see org.eclipse.pde.internal.ui.wizards.feature.AbstractCreateFeatureOperation
+            if (!project.hasNature(PDE.FEATURE_NATURE)) {
+                CoreUtility.addNatureToProject(project, PDE.FEATURE_NATURE, monitor);
```
HannesWell, line 56:
```suggestion
                addNature(project, PDE.FEATURE_NATURE, monitor);
```
You could use the `addNature()` method of the super-class here. Looks like this does the same and is an API method in contrast to CoreUtility.

kwin: Done in https://github.com/eclipse-m2e/m2e-core/pull/967/commits/b30f5013acbf8e7e38c8e6166de8e9ac876f278a.

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
kwin: Done in https://github.com/eclipse-m2e/m2e-core/pull/967/commits/b30f5013acbf8e7e38c8e6166de8e9ac876f278a

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
kwin: Done in https://github.com/eclipse-m2e/m2e-core/pull/967/commits/b30f5013acbf8e7e38c8e6166de8e9ac876f278a.

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
kwin: Done in https://github.com/eclipse-m2e/m2e-core/pull/967/commits/b30f5013acbf8e7e38c8e6166de8e9ac876f278a.

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
```diff
@@ -0,0 +1,129 @@
...
+	private static final String TYCHO_GROUP_ID = "org.eclipse.tycho";
+	private static final String TYCHO_DS_PLUGIN_ARTIFACT_ID = "tycho-ds-plugin";
+	private static final String GOAL_DECLARATIVE_SERVICES = "declarative-services";
+
+	@Override
+	public void configure(ProjectConfigurationRequest request, IProgressMonitor monitor) throws CoreException {
+		MavenProject mavenProject = request.mavenProject();
+		IProject project = request.mavenProjectFacade().getProject();
+		String packaging = mavenProject.getPackaging();
+		if ("eclipse-plugin".equals(packaging) || "eclipse-test-plugin".equals(packaging)) {
```
kwin, line 45: probably there should also test class folders be added to the Java Build Path (compare with https://github.com/eclipse-tycho/tycho/issues/1491). Is there any recommended code layout for unit tests inside `eclipse-plugin`?

HannesWell (approved):
Looks good now, thank you.

I just applied a few minor adjustments and clean-ups myself.
The most relevant are probably that the warning about multiple executions is added to the first and chosen one (which is then more clear for the user) and wrapping the Version creation into a try-catch block to handle invalid version strings.

org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java
HannesWell: > Is there any recommended code layout for unit tests inside `eclipse-plugin`?

You mean in general or for testing this connector?
At the moment I don't think there is a general recommendation since Plug-in projects cannot (yet) specify test-dependencies and the usual pattern is to separate tests into fragments.
However I would make it similar to the maven structure and have a separate `src-test` or `test-src` folder where the sources are placed.

> probably there should also test class folders be added to the Java Build Path (compare with [eclipse-tycho/tycho#1491](https://github.com/eclipse-tycho/tycho/issues/1491)). 

For a `eclipse-test-plugin` that would be reasonable, yes.
Do you want to include that into this PR or create a follow-up?

laeubi: `eclipse-plugin-test` is nothing we should specifically support as it is actually nothing special and has only a meaning when executing Tycho. Even there I always would recommend to use `eclipse-test` instead.

Whether or not a folder is meant as a test-folder can not be derived from the Tycho configuration, you can *guess* it from the standard maven layout, but m2e already does that.

HannesWell: > `eclipse-plugin-test` is nothing we should specifically support as it is actually nothing special and has only a meaning when executing Tycho. Even there I always would recommend to use `eclipse-test` instead.
> 

Is  `eclipse-test` a new packaging type? I don't know about that.

However since that topic seems not yet settled, lets merge this PR now and improve/enhance in with follow up PRs.

laeubi: > Is eclipse-test a new packaging type? I don't know about that.

Should be `eclipse-plugin` with `test` folder ;-)

PR #973, opened 2022-10-07
Improved service tracker which throws meaningful exceptions in case

### LINKED ISSUE #972: Use enhanced service tracker to allow to debug issues while getting services provided by DS components
The regular ServiceTracker as e.g. used in https://github.com/eclipse-m2e/m2e-core/blob/0aa663f75337f846f35a610bfb893f9c7c07605f/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/M2EUIPluginActivator.java#L176 does not throw a meaningful exception which would allow further debugging.

There should be an enhanced service tracker being used which uses `org.osgi.service.component.runtime.ServiceComponentRuntime` to throw more meaningful messages.

kwin:
mandatory services are missing

This closes #972


laeubi (changes requested):
As mentioned already in the Ticket m2e is not a place for reinventing the wheel, Apache Felix already contains commands to inspect the service component runtime.

kwin:
> Apache Felix already contains commands to inspect the service component runtime.

AFAIK Apache Felix tools are not contained in Eclipse, so this won't help with reports of consumers.

laeubi:
grafik

```
WARNING: This console is connected to the current running instance of Eclipse!
____________________________
Welcome to Apache Felix Gogo

g! list
...
org.eclipse.m2e.core.internal.project.WorkspaceStateWriter in bundle 287 (org.eclipse.m2e.core:2.0.3.20220904-1703) enabled, 1 instance.
    Id: 101, State:UNSATISFIED REFERENCE
...
```

HannesWell:
> grafik
> 

I have to admit, that I didn't know that. But that looks very handy to analyze the running OSGi instance.
But I agree with Christoph that, although better error-messages are good, M2E is the wrong place to address this problem.
However I already thought in the past that the PDE Bundle Spy could be enhanced to also offer information about (declarative) Services. AFAIK they only offer BSN, Version and state at the moment.

laeubi:
> although better error-messages are good

I must disagree here that this makes better error messages here. We can't expect from the user that SCR, Service, Reference or even an interface name has **any** meaning to the user! Even though its rather old you can for example take a look at:

https://learn.microsoft.com/en-us/archive/msdn-magazine/2009/january/usability-in-practice-when-things-go-wrong

A better error message would therefore be something similar to:

laeubi:
@kwin @HannesWell I think we can close this for now? I think m2e is not the right place and if this is considered useful equinox might be much better so it would be useful for many projects.

PR #975, opened 2022-10-08
Move createProblemMarker() to AbstractProjectConfigurator and clean up

HannesWell:
This makes createProblemMarker() part of the `AbstractProjectConfigurator` API and therefore available to downstream implementors and makes it simpler for them to create problem markers.


laeubi:
I think its a good idea.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/configurator/AbstractProjectConfigurator.java
```diff
@@ -154,28 +162,56 @@ public void setInitializationData(IConfigurationElement config, String propertyN
...
+  /**
+   * Creates a problem marker in the pom.xml file at the specified attribute of the given execution with the passed
+   * message and severity.
+   * <p>
+   * If the attribute of the execution is specified in a parent pom.xml outside of the workspace the marker is created
+   * at the parent-element of the request's project.
+   * </p>
+   * 
+   * @param execution the execution
+   * @param attribute the attribute to mark
```
kwin, line 193: It would be good to elaborate which values are valid here

HannesWell: I assume that you can use any attribute of the plug-ins XML configuration, but I can check that.

HannesWell:
@mickaelistria what do you think?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/configurator/AbstractProjectConfigurator.java
kwin: Maven pom.xml doesn’t use XML attributes. I have seen that element names have been given here.

mickaelistria:
Are there some connectors requesting this change? I don't mind growing the API if it does solve real problems; but growing the API without clear value should be avoided as it only prevents from further innovation.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/configurator/AbstractProjectConfigurator.java
mickaelistria: There are at least `combine.*` attribute that Maven uses: https://maven.apache.org/pom.html#advanced_configuration_inheritance

kwin: Those are not meant here IIUC.

HannesWell: Actually (XML) `elements` are meant her, not `attributes`. Sorry for the confusion.
I just blindly copied that from the internal method.

HannesWell:
> Are there some connectors requesting this change? I don't mind growing the API if it does solve real problems; but growing the API without clear value should be avoided as it only prevents from further innovation.

Yes of course API additions should be well considered.
At the moment only the Tycho-/Bundle-Connector use them, which are now part of m2e-core. But since they used to be external think there could be more connectors that could benefit from this method.
In general I think the connectors that are now part of m2e-core should use as much as possible of the official m2e API and should only rely on internals if there is a good reason that nobody else could use it. IMHO this helps us to provide a good API for down-stream/external connectors and can function as show-case how to use it.

But personally I don't have a need for that API at the moment.

kwin:
I have a need and would love to use the new methods in the configurators of https://github.com/apache/sling-ide-tooling/tree/master/eclipse/eclipse-m2e-ui/src/org/apache/sling/ide/eclipse/m2e/internal.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/configurator/AbstractProjectConfigurator.java
```diff
@@ -154,28 +162,56 @@ public void setInitializationData(IConfigurationElement config, String propertyN
...
+    String[] gav = location.getResourceId().split(":");
+    IMavenProjectFacade facade = projectManager.getMavenProject(gav[0], gav[1], gav[2]);
+    if(facade == null) {
+      // attribute specifying project (probably parent) is not in the workspace.
+      // The following code returns the location of the project's parent-element.
+      location = SourceLocationHelper.findLocation(request.mavenProject(), new MojoExecutionKey(execution));
+      facade = request.mavenProjectFacade();
+    }
+    MavenProblemInfo problem = new MavenProblemInfo(message, severity, location);
+    markerManager.addErrorMarker(facade.getPom(), IMavenConstants.MARKER_LIFECYCLEMAPPING_ID, problem);
```
kwin, line 212: the marker should be made configurable (given as additional argument) as not every consumer uses that ID

kwin: Also `IMavenConstants.MARKER_CONFIGURATION_ID` seems to be the more reasonable default for error with the configurator due to some mojo execution configuration.

HannesWell: > Also `IMavenConstants.MARKER_CONFIGURATION_ID` seems to be the more reasonable default for error with the configurator due to some mojo execution configuration.

Good point.

> the marker id should be made configurable (given as additional argument) as not every consumer uses that ID

Do you want to use custom problem markers? 

One problem I see here, is that in most cases consumers are probably fine with the default problem marker. But `IMavenConstants.MARKER_CONFIGURATION_ID` is not part of the M2E's API so we should not encourage users to use it. Instead a overload of the method without marker-type that passes that type/id is probably more suitable. It should then make no claims about the type. If necessary we can just change the default in the future.

laeubi: If users want custom marker why can't they create them on their own?

HannesWell: > If users want custom marker why can't they create them on their own?

Most of the methods/classes used in that method are internal. So to create a custom marker one a consumer would have to rely on M2E internals or would have to copy a lot of other code to rely only on Eclipse/Maven API.

laeubi: Well custom markers are... custom ... so what from m2e would be required? The only thing I can think of is determine the line/column in the XML and that could be public.

HannesWell: > Well custom markers are... custom ... so what from m2e would be required? The only thing I can think of is determine the line/column in the XML and that could be public.

I think it is about the type/id String being custom. And from M2E the convenient way to add markers to a pom is wanted.
We could just make a method public that computes the location. But since the classes that carry that information are internal as well, we end up with a much larger API than just adding a overload with one additional parameter.
But of course the location computation maybe is interesting for other task as well?

laeubi: > I think it is about the type/id String being custom. And from M2E the convenient way to add markers to a pom is wanted.

But why? Custom marker only make sense if you add custom attributes and handle those custom, otherwise one can simply reuse the m2e one.

> we end up with a much larger API than just adding a overload with one additional parameter.

There are already many parameters and adding even more for unclear use-case do not make it better I think. All one needs would probably be in `SourceLocationHelper` already...

HannesWell: OK, that's a point.
Maybe @kwin can elaborate a bit more about his use case and what would fit?

Nevertheless a `createProblemMarker()` method that creates `IMavenConstants.MARKER_CONFIGURATION_ID` problems still seems reasonable for me.

So if suitable we can still create an API around `SourceLocationHelper` and `SourceLocation` later.

HannesWell:
@kwin do you have any more comments on this?
As said, above I think this is a reasonable start and is ready to go.
If more fine tuned marker creation is needed we can add more in subsequent PRs.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/configurator/AbstractProjectConfigurator.java
kwin: > But why? Custom marker only make sense if you add custom attributes and handle those custom, otherwise one can simply reuse the m2e one.

Custom markers appear in a dedicated section. There is not necessarily the need to set attributes on those. Also custom markers can be easily queried or cleared. So IMHO just exposing the marker ID as additional argument doesn't do any harm and allows this API to cover more use cases.

HannesWell: OK sounds like there is a use-case for that.
Instead of having more and more arguments to that method, what do you think about encapsulating the problem data into a record with static factories?

The components would be `int problemSeverity, String problemMessage, String markerId` and two static factory methods
- `create(int problemSeverity, String problemMessage)`
- `create(int problemSeverity, String problemMessage, String markerId)`

The first one uses `IMavenConstants.MARKER_CONFIGURATION_ID` as `markerId`.
The java-doc of the record should then state that only canonical constructor is considered internal and should not be called by clients (maybe with `@noinstanciate`) and only the static factories should be used.

kwin: Sounds good to me.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/configurator/AbstractProjectConfigurator.java
```diff
@@ -167,15 +175,43 @@ public static void addNature(IProject project, String natureId, int updateFlags,
...
+   * </p>
+   * 
+   * @param execution the execution
+   * @param element the XML-element to mark
+   * @param request the request of the project being build
+   * @param problemSeverity the problems severity, one of {@link IMarker#SEVERITY_INFO SEVERITY_INFO},
+   *          {@link IMarker#SEVERITY_WARNING SEVERITY_WARNING} or {@link IMarker#SEVERITY_ERROR SEVERITY_ERROR}
+   * @param problemMessage the message of the created problem marker
+   */
+  protected void createProblemMarker(MojoExecution execution, String element, ProjectConfigurationRequest request,
```
mickaelistria, line 199: Could the IMarker be returned here?
What if one wants to create the marker on the value of the element? Can't we give it a text position instead, and an helper to locate an element in a mojoExecution?

HannesWell: > Could the IMarker be returned here? What if one wants to create the marker on the value of the element? 

In general `IMavenMarkerManager` could be changed to return the marker too.
Creating the marker at the value is not yet supported, but in general could be considered.

> Can't we give it a text position instead, and an helper to locate an element in a mojoExecution?

Yes Christoph suggested that as well before.
And after some rethinking about it, it is probably the better way for a generalization because IIRC for markers of custom Ids m2e just creates them using the Eclipse Resources API. So there is no benefit in providing an extra API for that. But I have to check that again.
However I think for a simple case a convenient API method like this would be useful nevertheless.

mickaelistria: > However I think for a simple case a convenient API method like this would be useful nevertheless.

There is uncertainity with this API that needs to be covered before it can be claimed "API-quality" (ie we won't want to change it for a long time), and this comes by nature of using the "element" name:
* How to place a marker on the value or a subset of the value? Or on some attribute (some mojo configuration may like it?
* What happens is the element is repeated in the configuration?
* What happens if there is no such element in the configuration?
* ...

All those questions wouldn't be raised if we clearly separate locating the marker vs creating it. If it is clear that locating the marker is the important part, and not creating it (because Platform API is enough), then instead of API to create a marker, we should have an API to reoslve locations and use already existing satisfying APIs from Platform to create it.

HannesWell:
Yes I do, altough it isn't my no1 priority at the moment.

G-Ork:
Looks like i could use this to improve the Maven Settings dialog in eclipse preferences to list all validation problems with the settings or toolchain files. Actually only the first error is squeezed into the upper text output in the dialog. All info is lost after closing the dialog. But in this case i need more a factory like `IMavenMarkerManager`. But as the context an config is loaded for execution if needed it may possible to hook in there.

PR #977, opened 2022-10-08
PGP-sign third-party dependencies in M2E repo

HannesWell:
Based on the desciussion in https://github.com/eclipse-m2e/m2e-core/pull/962#issuecomment-1264577708 and described in
https://docs.google.com/document/d/1MnDBvOUwKvKacB-QKnH_PzK88dUlHkjs-D-DWEKmvkY/edit
this PR enhances the M2E build to pgp-sign third-party dependencies that are not jar-signed.

The gpg-keyring and passphrase still need to be set up by the EFN IT-Infrastructure Team but, I have requested that via:
https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2056

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,9 +20,11 @@ pipeline {
 		}
 		stage('Build') {
 			steps {
-				sh 'mvn clean generate-sources -f m2e-maven-runtime/pom.xml -B -Dtycho.mode=maven -Pgenerate-osgi-metadata'
+				sh 'mvn clean generate-sources -f m2e-maven-runtime/pom.xml -B -V -Dtycho.mode=maven -Pgenerate-osgi-metadata'
 				wrap([$class: 'Xvnc', useXauthority: true]) {
-					sh 'mvn clean verify -f pom.xml -B -Dmaven.test.error.ignore=true -Dmaven.test.failure.ignore=true -Peclipse-sign,its -Dtycho.surefire.timeout=7200'
+					sh 'mvn clean verify -B -V \
+						-Dmaven.test.error.ignore=true -Dmaven.test.failure.ignore=true \
+						-Peclipse-sign,its'
 				}
 			}
 			post {
@@ -48,7 +50,7 @@ pipeline {
 						{
 							echo Deploy m2e repo to ${1}
 							ssh genie.m2e@projects-storage.eclipse.org "\
-								rm -rf  ${1}/* && \
+								rm -rf ${1}/* && \
 								mkdir -p ${1}"
 							scp -r org.eclipse.m2e.repository/target/repository/* genie.m2e@projects-storage.eclipse.org:${1}
 						}
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,14 +18,28 @@ pipeline {
 				sh 'git submodule update --init --recursive --remote'
 			}
 		}
+		stage('initialize PGP') {
+			steps {
+				withCredentials([file(credentialsId: 'secret-subkeys.asc', variable: 'KEYRING')]) {
+					sh 'gpg --batch --import "${KEYRING}"'
+					sh '''
+						for fpr in $(gpg --list-keys --with-colons | awk -F: \'/fpr:/ {print $10}\' | sort -u)
+						do
+							echo -e "5\ny\n" | gpg --batch --command-fd 0 --expert --edit-key ${fpr} trust 
+						done
+					'''
+				}
+			}
+		}
 		stage('Build') {
 			steps {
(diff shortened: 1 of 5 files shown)


Jenkinsfile
```diff
@@ -18,12 +18,28 @@ pipeline {
 				sh 'git submodule update --init --recursive --remote'
 			}
 		}
+		stage('initialize PGP') {
+			steps {
+				withCredentials([file(credentialsId: 'secret-subkeys.asc', variable: 'KEYRING')]) {
+					sh 'gpg --batch --import "${KEYRING}"'
+					sh '''
+						for fpr in $(gpg --list-keys --with-colons | awk -F: \'/fpr:/ {print $10}\' | sort -u)
```
HannesWell, line 26: @mickaelistria I tried to unroll the one-liner form TM4E https://github.com/eclipse/tm4e/blob/4e5ffed415a0b1a5504f42765beb803b4280a7c1/Jenkinsfile#L18
Is that correct?

Is it correct that this is to trust all keys in the ring?

org.eclipse.m2e.repository/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+						<artifactId>tycho-gpg-plugin</artifactId>
+						<version>${tycho-version}</version>
+						<executions>
+							<execution>
+								<id>pgpsigner</id>
+								<goals>
+									<goal>sign-p2-artifacts</goal>
+								</goals>
+								<configuration>
+									<keyname>${gpg.keyname}</keyname>
```
HannesWell, line 44: If the `tycho-gpg-plugin` would have a skip-property this pom could be replaced by a profile in the root pom that is activated for a category.xml file and skips this execution if the eclipse-sign profile is not activated (by setting a property like `pgp.signing.skip` that is true by default and false in that profile.

org.eclipse.m2e.repository/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+									<skipIfJarsigned>true</skipIfJarsigned>
+								</configuration>
+							</execution>
+						</executions>
+					</plugin>
+					<plugin>
+						<groupId>org.eclipse.tycho</groupId>
+						<artifactId>tycho-p2-repository-plugin</artifactId>
```
HannesWell, line 52: Does the property [tycho-gpg-plugin:addPublicKeysToArtifacts](https://tycho.eclipseprojects.io/doc/3.0.0/tycho-gpg-plugin/sign-p2-artifacts-mojo.html#addPublicKeysToArtifacts) change the included third-party deps?
If yes we have to keep this.

Jenkinsfile
mickaelistria: Yes, it's correct. The keyring is provided by the Eclipse Infra team, and what they provide can be trusted.
See https://wiki.eclipse.org/Jenkins#How_can_artifacts_be_deployed_to_OSSRH_.2F_Maven_Central.3F as reference documentation about PGP signatures at Eclipse.org.

org.eclipse.m2e.repository/pom.xml
mickaelistria: To be honest, I don't think it would be really better...

org.eclipse.m2e.repository/pom.xml
mickaelistria: `true` is the default value anyway.
This property just adds an extra property to each artifact, it doesn't change *what* is included.

org.eclipse.m2e.repository/pom.xml
HannesWell: > To be honest, I don't think it would be really better...

Yes, after thinking about it again, I agree. Since it is no problem to for example adjust the parent's version in this pom, it is not worth to try hard to get rid of it.

org.eclipse.m2e.repository/pom.xml
HannesWell: So it does only change the metadata in the repo's `artifacts.xml` and not the jar itself, doesn't it?

At least when I just checked the jar's in the repo build on Jenkins the jars of the pgp-signed maven deps where identical to the jars from Maven-Central, so the answer is probably yes.

When looking at the build artifacts.xml I assume that `addPublicKeysToArtifacts` controls if the `pgp.publicKeys` is added to each pgp signed artifact and `addPublicKeyToRepo` controls if `pgp.publicKeys` is added to the 'root' properties.
Can you tell me what's the advantage of adding the pgp.publicKeys to each artifact instead of just adding it to the repo root? To me it just looks like a lot of redundant data.

HannesWell:
I checked the m2e repo build by Jenkins and now the pgp entries are present for third party dependencies as the are supposed to be:
grafik

Furthermore a *.asc signature file is added for each pgp-signed artifact.

Looks like everything works as intended and this is ready to go.

@merks FYI, is about to deliver properly PGP-signed artifacts.

PR #978, opened 2022-10-08
Configure test folders in "eclipse-plugin" projects

### LINKED ISSUE #976: PDE Tycho connector should support identifying and adding source test folders
As outlined in https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#execute-unit-tests-with-eclipse-plugin-packaging it is supported to have unit tests in packaging `eclipse-plugin`. Those folders should be automatically detected and added by https://github.com/eclipse-m2e/m2e-core/blob/master/org.eclipse.m2e.pde.connector/src/org/eclipse/m2e/pde/connector/TychoPackagingsConfigurator.java based on some heuristics.

I would suggest to add any top-level folders whose name start with "test" or end with "test" as test source folder. The logic must be similar to https://github.com/eclipse-m2e/m2e-core/blob/0223a44258a6c92ac4a1943721a983743809ac37/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java#L294 but not using MavenProject.getTestCompileSourceRoots() (https://github.com/eclipse-m2e/m2e-core/blob/0223a44258a6c92ac4a1943721a983743809ac37/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java#L373, https://maven.apache.org/ref/3-LATEST/maven-core/apidocs/org/apache/maven/project/MavenProject.html#getTestCompileSourceRoots--).

kwin:
This closes #976

### DIFF
--- a/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
+++ b/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
@@ -409,7 +409,7 @@ public abstract class AbstractJavaProjectConfigurator extends AbstractProjectCon
     return true;
   }
 
-  private IPath[] toPaths(String[] values) {
+  public static IPath[] toPaths(String[] values) {
     if(values == null) {
       return new IPath[0];
     }
@@ -422,7 +422,7 @@ public abstract class AbstractJavaProjectConfigurator extends AbstractProjectCon
     return paths;
   }
 
-  private void addSourceDirs(IClasspathDescriptor classpath, IProject project, List<String> sourceRoots,
+  public static void addSourceDirs(IClasspathDescriptor classpath, IProject project, List<String> sourceRoots,
       IPath outputPath, IPath[] inclusion, IPath[] exclusion, String sourceEncoding, IProgressMonitor monitor,
       boolean addTestFlag) throws CoreException {
 
@@ -466,7 +466,7 @@ public abstract class AbstractJavaProjectConfigurator extends AbstractProjectCon
 
   }
 
-  private IClasspathEntryDescriptor getEnclosingEntryDescriptor(IClasspathDescriptor classpath, IPath fullPath) {
+  private static IClasspathEntryDescriptor getEnclosingEntryDescriptor(IClasspathDescriptor classpath, IPath fullPath) {
     for(IClasspathEntryDescriptor cped : classpath.getEntryDescriptors()) {
       if(cped.getPath().isPrefixOf(fullPath)) {
         return cped;
@@ -475,7 +475,7 @@ public abstract class AbstractJavaProjectConfigurator extends AbstractProjectCon
     return null;
   }
 
-  private IClasspathEntryDescriptor getEntryDescriptor(IClasspathDescriptor classpath, IPath fullPath) {
+  private static IClasspathEntryDescriptor getEntryDescriptor(IClasspathDescriptor classpath, IPath fullPath) {
     for(IClasspathEntryDescriptor cped : classpath.getEntryDescriptors()) {
       if(cped.getPath().equals(fullPath)) {
         return cped;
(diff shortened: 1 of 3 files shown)


kwin:
@laeubi Can you comment on the heuristics used here? Also not sure if dedicated test resource folders should be supported.

laeubi:
Tycho contains some code that tries to detect the used imports/bundles to decide what JUnit version is used, even though thats not perfect its often a strong indication, so probably one can assume that if a source folder (how do we know it is one? Maybe because it contains a .java file?) and it has "test" in its name and it contains at least one JavaFile that import e.g. `org.junit.*` this is really a strong indication for a test-folder, but that's maybe to complicated here?

I'm more wondering if we really say there is no `.project`, no `.classpath` (what will contain the test attribute already) and just a `pom.xml` should we not require (or recommend?) it using a standard maven layout as well?

laeubi:
I think for `eclipse-test-plugin` one must assume it is always test folder even in "src" folder (what is very common in platform code). For `eclipse-plugin` there are actually only two cases:

1. there is a `.classpath` file already where it is configured what is test/no test
2. you use maven standard layout (even though I think this needs to be supported by Tycho but seems reasonable)

kwin:
I think at least some projects use pom.xml only approaches without `.project` or `.classpath` in source code management. Using Maven Standard Layout currently conflicts with PDE project setup as everything inside `src` ends up in the bundle by default. Is PDE is so flexible that it can deal with Maven Standard layout?

laeubi:
> I think at least some projects use pom.xml only approaches without `.project` or `.classpath` in source code management

Yes this is valid setup, but technically, this would imply standard maven layout (even though we sometimes try to 'guess' that something else is/was desired. But in the end its really "bad" because only Tycho/PDE+2e can understand such setup.

> Using Maven Standard Layout currently conflicts with PDE project setup as everything inside `src` ends up in the bundle by default.

Yes this is really bad it would be good if JDT/PDE would have an extension point to supply alternative layouts beyond the usual ones but currently this is hardcoded, so if you like to improve this  `org.eclipse.jdt.ui.wizards.NewJavaProjectWizardPageOne.LayoutGroup` seems the place to start. Once there is an extension point, we can implements this in m2e

> Is PDE is so flexible that it can deal with Maven Standard layout?

PDE does not care about the name of the source-folder you can use any layout you want.

kwin:
> Once there is an extension point, we can implements this in m2e

I agree but this is currently beyond my capability. Would you be fine with first supporting this PDE specific layout and later on add support for standard Maven layout? Can you comment directly on the currently used heuristics in the code?

laeubi:
> Would you be fine with first supporting this PDE specific layout

The problem is there is no PDE layout for test-code (actually PDE does not suppor this and JDT offers that), so assume I have a folder in my project named "testdata" then your current approach would assume it is a source folder that contains test.

So I think at least one needs to take `build.properties` into account if the folder is mentioned there, but if I remember correctly for test-source only folder one can't add them to `build.properties` but to `.classpath` only ... so probably it would be good to first add a test project (so one see what layout and files are there to be expected), similar to the ones in https://github.com/eclipse-m2e/m2e-core/tree/master/org.eclipse.m2e.core.ui.tests/resources/projects

HannesWell:
Btw. I started to create tests for the M2E-PDE connectors here https://github.com/eclipse-m2e/m2e-core/pull/979.
I think it would be good to complete that first, so that you can add new cases for this scenario.

laeubi:
@kwin can you rebase and resolve conflicts?

laeubi:
Closing this as there is no progress or feedback yet, @kwin if you want to revivew that, I think the best is to include a sample project in the core.tests as this would make it simpler to decide what exact layout should be supported.

In general I think there are two cases:

1. configuration with a `.classpath` in the source control -> everything already works
2. standard maven layout -> m2e should detect this automatically but seems it needs some help here for tycho projects at the moment.

PR #979, opened 2022-10-08
Add tests for M2E PDE connectors

HannesWell:
Adds tests for M2E's PDE Connectors.

Additionally this PR also fixes 
- Computation of a 'PDE-project' output folder, which is set to the .classpath 
- Computation of the Bundle-Root for Maven-projects that use the `bnd-maven-plugin`
- Resolves values of configuration parameters in `TychoPackagingsConfigurator` so that even variables can be used as well.


org.eclipse.m2e.pde.connector.tests/projects/tycho/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+		<module>pde.tycho.pomless.feature</module>
+	</modules>
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.eclipse.tycho</groupId>
+				<artifactId>tycho-maven-plugin</artifactId>
+				<version>${tycho-version}</version>
+				<extensions>true</extensions>
```
kwin, line 30: Either you use the extensions mechanism with `.mvn` folder (https://github.com/eclipse-m2e/m2e-core/pull/979/files#diff-5a68c98f891e1dee709d63ab749b455b6ac70f77f4fb51ab74214fd57d2388ff) or use `<extensions>true</extensions>` (not both)

laeubi: Nope, that are completely different things, `.mvn` folder contains maven core extensions while the later is a project scoped extension. Both can work together or independent of each other.

kwin: Any idea then where the error `Caused by: org.codehaus.plexus.classworlds.realm.DuplicateRealmException: coreExtension>org.eclipse.tycho:tycho-build:3.0.0` is coming from in org.eclipse.m2e.pde.connector.tests.TychoConnectorTest

kwin: Would be good to document somewhere which extensions are actually provided by `tycho-build` if it is not lifecycle mappings and artifact handlers.

laeubi: Extension handling in m2e is quite new and untested so there seem to be a bug in that handling, the message is from maven-core complaining that we create more than one real for the same name.

I plan to improve the extension support in this release, it would be probably the best to test the connector using traditional pom.xml than tycho-pomless unless the feature is made more robust.

laeubi: > Would be good to document somewhere which extensions are actually provided by `tycho-build` if it is not lifecycle mappings and artifact handlers.

These are tycho-internals and might change anytime. `tycho-build` extension is currently required for:

1. [Enhanced CI Friendly Versions Support](https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#enhanced-support-for-maven-ci-friendly-versions)
2. [Pomless Builds](https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#tycho-pomless-will-become-a-tycho-core-extension)
3. [Enhanced Reactor Make behavior](https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#tycho-specific-maven-graphbuilder-to-support---also-make--am-and---also-make-dependents--amd)

HannesWell: > Any idea then where the error `Caused by: org.codehaus.plexus.classworlds.realm.DuplicateRealmException: coreExtension>org.eclipse.tycho:tycho-build:3.0.0` is coming from in org.eclipse.m2e.pde.connector.tests.TychoConnectorTest

The main problem was that I didn't add the `pom.xml` to the path when importing and that the parent pom-file was missing because the `importProject()` method first copies the projects into the workspace.

> I plan to improve the extension support in this release, it would be probably the best to test the connector using traditional pom.xml than tycho-pomless unless the feature is made more robust.

That's good to hear, however when importing the pomless project into the IDE it already works well, so that is probably also a issue of the test-setup. Probably the .polyglot-poms have to be generated before the import is started. But I will further investigate that tomorrow.

HannesWell:
The tests (should) now succeed. Besides adding tests for the connectors, this PR also fixes 
- Computation of a 'PDE-project' output folder, which is set to the .classpath 
- Computation of the Bundle-Root for Maven-projects that use the `bnd-maven-plugin`
- Resolves values of configuration parameters in `TychoPackagingsConfigurator` so that even variables can be used as well.

@laeubi, @kwin do you have any more remarks for this? If not, I'm going to submit this.

kwin (approved):

PR #981, opened 2022-10-11
Only evaluate major and minor version of enforcer java rule

### LINKED ISSUE #842: Java version mismatch between JRE_CONTAINER and targetPlatform in M2E 2.0.0

kwin:
Fix evaluation of expressions in m-enforcer-p parameter Only select ExecutionEnvironments with at least one compatible VM installed.
This closes #842

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
@@ -158,6 +158,17 @@ public interface IMaven extends IComponentLookup {
       throws CoreException;
 
   /**
+   * Resolves a configuration parameter from the given {@code mojoExecution}. It coerces from String to the given type
+   * and considers expressions and default values.
+   * 
+   * @param <T>
+   * @param project the Maven project
+   * @param mojoExecution the mojo execution from which to retrieve the configuration value
+   * @param parameter the name of the parameter (may be nested with separating {@code .})
+   * @param asType the type to coerce to
+   * @param monitor the progress monitor
+   * @return the parameter value or {@code null} if the parameter with the given name was not found
+   * @throws CoreException
    * @since 1.4
    */
   <T> T getMojoParameterValue(MavenProject project, MojoExecution mojoExecution, String parameter,
--- a/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
+++ b/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
@@ -44,11 +44,10 @@ import org.eclipse.jdt.launching.JavaRuntime;
 import org.eclipse.jdt.launching.environments.IExecutionEnvironment;
 import org.eclipse.jdt.launching.environments.IExecutionEnvironmentsManager;
 
-import org.codehaus.plexus.util.xml.Xpp3Dom;
-
 import org.apache.maven.artifact.versioning.ArtifactVersion;
 import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
 import org.apache.maven.artifact.versioning.InvalidVersionSpecificationException;
+import org.apache.maven.artifact.versioning.Restriction;
 import org.apache.maven.artifact.versioning.VersionRange;
 import org.apache.maven.model.Resource;
 import org.apache.maven.plugin.MojoExecution;
@@ -56,6 +55,7 @@ import org.apache.maven.project.MavenProject;
 
 import org.eclipse.m2e.core.MavenPlugin;
 import org.eclipse.m2e.core.internal.M2EUtils;
+import org.eclipse.m2e.core.internal.embedder.MavenImpl;
 import org.eclipse.m2e.core.project.IMavenProjectFacade;
 import org.eclipse.m2e.core.project.IProjectConfigurationManager;
 import org.eclipse.m2e.core.project.configurator.AbstractProjectConfigurator;
@@ -254,11 +254,12 @@ public abstract class AbstractJavaProjectConfigurator extends AbstractProjectCon
       return null;
     }
     IExecutionEnvironmentsManager manager = JavaRuntime.getExecutionEnvironmentsManager();
-    for(IExecutionEnvironment environment : manager.getExecutionEnvironments()) {
-      if(environment.getId().equals(environmentId)) {
-        return environment;
-      }
+    IExecutionEnvironment environment = manager.getEnvironment(environmentId);
+    if(environment != null && environment.getCompatibleVMs().length > 0) {
+      return environment;
     }
+    log.error("Failed to find a compatible VM for environment id '{}', falling back to workspace default",
+        environmentId);
     return null;
   }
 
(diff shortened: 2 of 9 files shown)


org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
```diff
@@ -149,11 +152,42 @@ MojoExecution setupMojoExecution(MavenProject project, MojoExecution execution,
...
+   * @param parameterPath the path of the parameter to look up, the first item is the name of the element directly below
+   *          {@code <configuration>} and the last one is the element containing the actual value
+   * @param asType the type to coerce to
+   * @param monitor the progress monitor
+   * @return the parameter value or {@code null} if the parameter with the given name was not found
+   * @throws CoreException
+   * @since 2.1.0
+   * @see IMaven#getMojoParameterValue(MavenProject, MojoExecution, String, Class, IProgressMonitor)
+   */
+  <T> T getMojoParameterValue(MavenProject project, MojoExecution mojoExecution, List<String> parameterPath,
```
HannesWell, line 188: I assume that actually one would like to get an object of the 'real' configuration field type, but that is often not reasonably doable for complex types because they are in a Maven plugin associated library that one actually does not want to import.
In this case it would be the `EnforcerRule` class.

Therefore I wonder if instead of specifying a parameter path, it would be reasonable to define some kind of `ConfigurationProxy` that that is then returned and a caller can query for (child) elements:
```
public interface ConfigurationElementProxy {
  ConfigurationElementProxy get(String childName);
  <T> T as(Class<T>type);
}
```
With that the returned value can be treated more similar like an instance of the 'real' configuration field type.
In the case of this PR one could then do something like:
```
var rules = maven.getMojoParameterValue(mavenProject, mojoExecution, "rules", ConfigurationProxy.class, monitor)
String version = rules.get("requireJavaVersion").get("version").as(String.class)
```

The returned proxy would then just have to keep a reference to the instances needed to perform the final conversion in `as()`.

What do you and the others think?

HannesWell (changes requested):
In general that looks reasonable, but I have a general remark about the API enhancement.

Furthermore could you please add a test-case to ensure the associated issue is fixed.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
kwin: This method only allows to return value objects (https://maven.apache.org/guides/mini/guide-configuring-plugins.html#mapping-value-objects). For retrieving beans it is not suited! That requires much more refactoring and I currently don't see a use case for that. Your proposal would require to load the intermediate beans which is not necessary in this proposal here.

HannesWell: I'm not sure if we have the same idea. Created https://github.com/eclipse-m2e/m2e-core/pull/996 as draft to demonstrate it.
Comments are welcome. :)

kwin: @HannesWell Any idea how to move forward here given that there was no consensus for #996 yet?

HannesWell: > I am also fine with removing the API exposure of the new method in IMaven and just use the implementation directly. Then there is time afterwards to decide on the proper API and migrate this usage to it.

Yes please continue that way. Changing internales later is no problem.

Can you please also add a Test case for the issue?

laeubi: Just one note, please don't add "convenience" methods to `IMaven` this methods do not add any value in terms of object instance and cann be static methods on a ConfigurationHelper or alike or you could use the MavenToolBox instead that already has things like session so one do not needs to pass it as a parameter directly.

So please what ever is choose here take the opportunity to move the code out of the `MavenImpl` to a more suitable place.

kwin: Fixed in https://github.com/eclipse-m2e/m2e-core/pull/981/commits/36e1c4a4c3759ad35394fa097fa1afd60ba2a6c5

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
```diff
@@ -845,34 +847,48 @@ private String getMinimumJavaBuildEnvironmentId(ProjectConfigurationRequest requ
...
-    Xpp3Dom requireJavaVersion = rules.getChild("requireJavaVersion");
-    if(requireJavaVersion == null) {
-      return null;
-    }
-    Xpp3Dom version = requireJavaVersion.getChild("version");
-    if(version == null) {
-      return null;
+  private String getMinimumJavaBuildEnvironmentId(MavenProject mavenProject, MojoExecution mojoExecution,
+      IProgressMonitor monitor) throws InvalidVersionSpecificationException, CoreException {
+    MojoUtils mojoUtils = new MojoUtils(maven, ((MavenImpl) maven).getPlexusContainer().getContainerRealm(),
```
kwin, line 863: @laeubi Any idea how to access the classLoader to use from outside without doing potentially invalid casting??

laeubi: It depends on what class you want to access, do you have a specific error/problem here?

kwin: Everything works, but the pattern `((MavenImpl) maven).getPlexusContainer().getContainerRealm()` assumes that the service `IMaven` is always implemented by `MavenImpl`. Is it ok to to push `getPlexusContainer()` to IMaven?

laeubi: Actually using the plexus container should not be required at all, this is a bit an anti-pattern. Next, using the `IMaven` should be considered a potential code-smell, you are configuring a project here, so most likely you don't want the global context but a project context.

kwin: Please have a closer look at the code. I need a ClassLoader for conversion/expression interpolation. That was the `getPlexusContainer().getContainerRealm()` before. I don't want to change everything in this PR (you can do that subsequently) but to extract the Mojo value calculation to a dedicated class this is necessary.

laeubi: > Please have a closer look at the code. I need a ClassLoader for conversion/expression interpolation.

There is a difference between "a classloader" and "the correct classloader", it seems there are two "getParameterValue" methods one uses the plugin realm and the other is using (currently) the container realm, why not use the plugin-realm in both cases? And if there is a compelling reason why use the container realm on not the maven project realm as you have a MavenProject anyways?

> but to extract the Mojo value calculation to a dedicated class this is necessary.

Just extracting code from one place to another often does not gives better results, so we always should check if the code does the right things "later" often never happens. Looking at the code I'm not sure if it is even a good diea to have this separated, a Mojo and its executions are closely coupled to a project, so maybe it would better be placed the the IProjectFacade eg like this one:
https://github.com/eclipse-m2e/m2e-core/blob/9416e758afe1a405fe6616ce110e7d974e73858e/org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/IMavenProjectFacade.java#L149-L153

kwin: Sorry but this is another issue and I don’t want to solve it in this PR. All I need is a method to retrieve nested mojo parameter values.

laeubi: Then it might be better to implement it in your plugin directly, m2e already contains some ugly hacks and we should not add another one. Using the global context and the container realm most likely giving wrong results in some setups that are hard to track.

kwin: > in your plugin directly…

The issue and also this PR fixes an issue in m2e. But this kind of discussion doesn’t really foster contributions. Please don’t require bugfixes to improve everything. It is easy to follow up with a subsequent PR to make it a proper API. Since you committers couldn’t agree on an API please don’t let external contributors suffer from it.

laeubi: Well you asked what the proper way would be to access the classloader and I told you that the current approach is probably broken. So it does not help to "fix" broken code with some other broken code. If you think it is better fixed with separate PRs that fine. Its often the case that considerable amount of time and effort has to be done before the actual issue can be fixed, and the main goal is that it works under all circumstances we currently have.

This is not "improve everything", but these changes are already beyond trivial single one-liners and thus likely deserve closer review what might reveal related issues to be fixed for a solid soloution.

kwin: Ok, done now.

kwin:
> Furthermore could you please add a test-case to ensure the associated issue is fixed.

Unfortunately not easily due to big amount of test failures with Mac OS.

HannesWell:
> > Furthermore could you please add a test-case to ensure the associated issue is fixed.
> 
> Unfortunately not easily due to big amount of test failures with Mac OS.

Have you tried to create a new test case for this? Does that fail? You can simply ignore the failing tests on your computer for now. As long as the new test passes in the CI that's fine.

And could you please just squash all your changes to one commit with suitable message. This hopefully also makes the eclipsefdn-eca check pass.

If we manage to complete this change over the weekend I'm willing to wait for this so that it will be included in the next bug-fix release and eventually Eclipse 2022-12.

org.eclipse.m2e.feature/feature.xml
```diff
@@ -2,7 +2,11 @@
 <feature
       id="org.eclipse.m2e.feature"
       label="%featureName"
+<<<<<<< Upstream, based on master
```
HannesWell, line 5: You have to remove the conflict markers.

HannesWell (approved):
Given that I want to perform another bug-fix release tomorrow and that the discussion slowed this down I have added the tests by myself for this one time.
For the next time please provide a test case that covers the change. Fixes without tests are only half the battle, because without them you can't ensure that the issue does not happen again in the future.

Furthermore I reworked the fix itself a bit.
@kwin check if this is still correct.

@laeubi, @mickaelistria do you want to review this as well?

laeubi:
@HannesWell As you have taken a deeper look on this and my previous concerns are addressed I would leave it up to you how to proceed here.

HannesWell:
> @HannesWell As you have taken a deeper look on this and my previous concerns are addressed I would leave it up to you how to proceed here.

Alright, then I think this is good to go.
Thanks.

PR #983, opened 2022-10-11
MvnProtocolHandlerService: Implement lastModfified()

### LINKED ISSUE #982: Many 'Server returned lastModified <= 0' warnings when using the 'mvn:' syntax for update site URLs
When using the `mvn:` syntax for update site URLs, many `Server returned lastModified <= 0 for ...` are produced.

E.g.:
image

The reason is that `MvnProtocolHandlerService` does not override `java.net.URLConnection.getLastModified()` and always returns `0`.

I suppose we could simply return `lastModified()` of the resolved `File` referenced within `artifactResult`?

sratz:
Determine last-modified timestamp based on the referenced file.

This silences warnings such as
  ...
  [INFO] Adding repository mvn:org.eclipse.jetty:jetty-p2:11.0.12:zip:p2site
  [WARNING] [a7f3f9cb-793c-4f52-8e8e-b74719152388][extension>org.eclipse.tycho:tycho-maven-plugin:3.0.0] Server returned lastModified <= 0 for mvn:org.eclipse.jetty:jetty-p2:11.0.12:zip:p2site/content.xml
  ...

Fixes #982.

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MvnProtocolHandlerService.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MvnProtocolHandlerService.java
@@ -129,6 +129,19 @@ public class MvnProtocolHandlerService extends AbstractURLStreamHandlerService {
       String urlSpec = "jar:" + location.toURI() + "!" + subPath;
       return new URL(urlSpec).openStream();
     }
+
+    @Override
+    public long getLastModified() {
+      try {
+        connect();
+      } catch (IOException e) {
+        return 0;
+      }
+      if(artifactResult == null || artifactResult.isMissing()) {
+        return 0;
+      }
+      return artifactResult.getArtifact().getFile().lastModified();
+    }
   }
 
 }


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MvnProtocolHandlerService.java
```diff
@@ -129,6 +129,19 @@ public InputStream getInputStream() throws IOException {
       String urlSpec = "jar:" + location.toURI() + "!" + subPath;
       return new URL(urlSpec).openStream();
     }
+
+    @Override
+    public long getLastModified() {
+      try {
+        connect();
+      } catch (IOException e) {
+        return 0;
```
laeubi, line 138: I think -1 has to be returned for "cant determine last modified"

laeubi:
@sratz The log output don't seem to be related to m2e even though an implementation of lastModified is fine it might require an equal one for Tycho as well.

sratz:
The warning is written by p2:
https://git.eclipse.org/r/plugins/gitiles/equinox/rt.equinox.p2/+/master/bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/CacheManager.java#112

laeubi:
But m2e do not use P2 ;-)

sratz:
> But m2e do not use P2 ;-)

No, but P2 uses m2e-core. Indirectly, that is, if you use the `mvn:` URI scheme in a `.target` definition

laeubi:
> > But m2e do not use P2 ;-)
> 
> No, but P2 uses m2e-core. Indirectly, that is, if you use the `mvn:` URI scheme in a `.target` definition

[WARNING] [a7f3f9cb-793c-4f52-8e8e-b74719152388][extension>**org.eclipse.tycho:tycho-maven-plugin**:3.0.0] 

Still the message is from **tycho-maven-plugin** ;-)

Tycho+m2 both implement the mvn protocol...

sratz:
But you're right, I mixed up some things.

This issue is about the error being visible in Eclipse:

image


The Warning in Tycho is a different issue and needs to be solved over there.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MvnProtocolHandlerService.java
sratz: JavaDoc says:
`@return  the date the resource referenced by this {@code URLConnection} was last modified, or 0 if not known.`
Also `0` is what was returned if not overriding the method at all.

laeubi: Alright, then forget about that :-)

laeubi (approved):
A corresponding change for Tycho would be useful.

sratz:
> A corresponding change for Tycho would be useful.

Done: https://github.com/eclipse-tycho/tycho/pull/1514

PR #986, opened 2022-10-12
Fix multiple loads of Maven Core Extension

HannesWell:
Currently each Maven Core extension (with identical GAV) can only be loaded one time from one location.

Therefore loading Maven extensions fails in the following scenarios:
- A pomless project (and its extensions) is loaded, then delete from the Workspace and file-system and then re-created.
Then the project+extension is loaded again and loading the extension fails.
- A pomless project (and its extensions) is loaded from a first location.
Then at a disjunct pomless project, whose `.mvn`-folder resides at a different location, but references an extension with the same GAV is loaded and loading the extension fails.

In both cases the reason is that the `CLASS_WORLD` in the `PlexusContainerManager` contains the extension ClassRealm from the first load. In the first case it is not cleaned up correctly and in the second case it simply is still in use, when in `PlexusContainerManager.loadCoreExtensions()` the method `BootstrapCoreExtensionManager.loadCoreExtensions()` is called:

https://github.com/eclipse-m2e/m2e-core/blob/03b015240f32cd4c6feece22c1b8a2830b43da1c/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/PlexusContainerManager.java#L307-L310

A first approach to solve this would be to first ensure that also the extension realms are disposed and second to make the extension realm-ids unique by appending a running number, similar like it is done in `setupContainerRealm()`.

@laeubi, @mickaelistria what do you think?

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/PlexusContainerManager.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/PlexusContainerManager.java
@@ -17,6 +17,7 @@ package org.eclipse.m2e.core.internal.embedder;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.InputStream;
+import java.net.MalformedURLException;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -24,7 +25,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.atomic.AtomicLong;
 
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Deactivate;
@@ -37,6 +37,8 @@ import com.google.inject.AbstractModule;
 import org.eclipse.core.resources.IWorkspace;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.ILog;
+import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.Status;
 
 import org.codehaus.plexus.ContainerConfiguration;
@@ -47,6 +49,7 @@ import org.codehaus.plexus.PlexusContainer;
 import org.codehaus.plexus.PlexusContainerException;
 import org.codehaus.plexus.classworlds.ClassWorld;
 import org.codehaus.plexus.classworlds.realm.ClassRealm;
+import org.codehaus.plexus.classworlds.realm.DuplicateRealmException;
 import org.codehaus.plexus.classworlds.realm.NoSuchRealmException;
 import org.codehaus.plexus.component.repository.exception.ComponentLookupException;
 import org.codehaus.plexus.logging.LoggerManager;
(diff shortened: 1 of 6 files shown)


laeubi:
Lets try that out...

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/PlexusContainerManager.java
```diff
@@ -119,23 +150,35 @@ public class PlexusContainerManager {
...
       if(nonRootedContainer != null) {
-        nonRootedContainer.dispose();
+        disposeContainer(nonRootedContainer);
         nonRootedContainer = null;
       }
     }
   }
 
+  private static void disposeContainer(PlexusContainerWithExtensions container) {
+    //TODO: dispose  container.container().getContainerRealm() as well?
```
HannesWell, line 163: @laeubi do you think that would be useful reasonable? Or can this cause problems?
I'm not very familiar with the ClassRealm lifecycle, but since it is created by M2E below, I wonder if it should be cleaned up by M2E as well, when the container is disposed.

HannesWell (commented):
> Lets try that out...

Alright. I'm positively surprised that everything is working, at least all tests pass. My past experience when it comes to ClassRealms, were just problems.

I fine tuned it a bit, but I think then this is actually ready.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/PlexusContainerManager.java
```diff
@@ -86,7 +92,27 @@ public class PlexusContainerManager {
 
   private static final String PLEXUS_CORE_REALM = "plexus.core";
 
-  private static final ClassWorld CLASS_WORLD = new ClassWorld(PLEXUS_CORE_REALM, ClassWorld.class.getClassLoader());
+  private static final StackWalker STACK_WALKER = StackWalker.getInstance(Option.RETAIN_CLASS_REFERENCE);
+
+  private static final ClassWorld CLASS_WORLD = new ClassWorld(PLEXUS_CORE_REALM, ClassWorld.class.getClassLoader()) {
```
laeubi, line 97: Thinking a bit more about it, the soloution might be not to use one classworld for all containers, but an own classworld for each  container....

HannesWell: Yes, I thought about that as well but assumed there is a reason to have only one class-world, probably less `ClassCastException`s.
But as long as objects created by Plug-in executions are not saved that should not be an issue.

laeubi: The reason was that I though it might not be necessary too have multiple "Worlds" but before we need to special hack the implementation it more seems there should be one world for each "maven container with extension" thing ...

HannesWell: With one world per container there are definitely less changes necessary.
So lets see if that works. The extension tests passed. Lets see if all other tests pass as well.

HannesWell: Looks good. All tests have passed.
So should we try it?

laeubi (approved):

PR #996, opened 2022-10-18
Improve accessibility of Mojo configuration parameters and (partially) use it in M2E

HannesWell:
This PR aims to improve the accessibility of Mojo configuration parameters by introducing the method `IMaven.getMojoConfiguration(MavenProject , MojoExecution , IProgressMonitor )`, which returns a `ConfigurationElement` that allows to traverse the tree that can be formed by a configuration in general.

The advantage are:
- Enables access to nested elements, not only top-level elements
- Enables accessing multiple elements of same name
- In general enables to selectively walk the configuration element tree

The new method is intended as a replacement for the method `IMaven.getMojoParameterValue(MavenProject, MojoExecution, String, Class, IProgressMonitor)`, which returns only the value of a top-level configuration element.
The method `IMaven.getMojoParameterValue(MavenProject, String, Class, Plugin, ConfigurationContainer, String, IProgressMonitor)` can probably replaced in the same manner.

The new method can for example be used in https://github.com/eclipse-m2e/m2e-core/pull/981 or if we introduce for example a connector for the [Maven clean-plugin to handle additional files to delete](https://maven.apache.org/plugins/maven-clean-plugin/examples/delete_additional_files.html).

This PR is a draft that is not yet completed, method names have to/can be adjusted, java-doc has to be added as well as tests.


org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
```diff
@@ -148,11 +149,42 @@ MavenExecutionPlan calculateExecutionPlan(MavenProject project, List<String> goa
   MojoExecution setupMojoExecution(MavenProject project, MojoExecution execution, IProgressMonitor monitor)
       throws CoreException;
 
+  /**
+   * Instances should not be cached
+   *
+   * @since 2.1
+   */
+  public interface ConfigurationElement {
+    boolean exists();
```
kwin, line 158: This seems weird, I would rather throw some exception when invalid names are passed. The top level element should always be valid (even if no explicit configuration is given in the underlying effective pom).

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
```diff
@@ -148,11 +149,42 @@ MavenExecutionPlan calculateExecutionPlan(MavenProject project, List<String> goa
...
+
+    // or firstChild() ? That's what it is actually but it does not look so nice when chained
+    ConfigurationParameter get(String name) throws NoSuchElementException;
+
+    // TODO: or getAll() or similar.
+    Stream<ConfigurationParameter> children(String name) throws NoSuchElementException;
```
kwin, line 164: name is always unique except for collections (https://maven.apache.org/guides/mini/guide-configuring-plugins.html#mapping-collections-and-arrays). I don't think that we need to support that here, I one is interested in getting the collection she/he can call get("rootElementForCollection).as(Collection.class).

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
```diff
@@ -148,11 +149,42 @@ MavenExecutionPlan calculateExecutionPlan(MavenProject project, List<String> goa
...
+
+  /**
+   * @since 2.1
+   */
+  ConfigurationElement getMojoConfiguration(MavenProject project, MojoExecution mojoExecution, IProgressMonitor monitor)
+      throws CoreException;
+
   /**
    * @since 1.4
    */
```
kwin, line 181: deprecated tag also in javadoc with a reference to the method to be used instead.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
```diff
@@ -714,56 +713,104 @@ private Properties loadLastUpdated(ArtifactRepository localRepository, Artifact
...
+      ConfigurationConverter typeConverter = lookup.lookupConverterForType(clazz);
+      Object value = typeConverter.fromConfiguration(lookup, configuration, clazz, mojo.getImplementationClass(),
+          pluginRealm, evaluator, null);
+      return clazz.cast(value);
+    }
   }
 
-  private String formatAsDirectory(String directory) {
-    return directory.replace(GROUP_SEPARATOR, PATH_SEPARATOR);
+  private class ConfigurationElementImpl implements ConfigurationElement {
```
kwin, line 755: I would put this and related classes to a dedicated file.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
```diff
@@ -714,56 +713,104 @@ private Properties loadLastUpdated(ArtifactRepository localRepository, Artifact
...
+    final PlexusConfiguration configuration;
+
+    final String name;
```
kwin, line 758: name -> path

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
```diff
@@ -714,56 +713,104 @@ private Properties loadLastUpdated(ArtifactRepository localRepository, Artifact
...
-      ExpressionEvaluator expressionEvaluator = new PluginParameterExpressionEvaluator(session, mojoExecution);
-      ConfigurationConverter typeConverter = converterLookup.lookupConverterForType(asType);
-      Xpp3Dom dom = mojoExecution.getConfiguration();
-      if(dom == null) {
-        return null;
+    @Override
+    public <T> T as(Class<T> clazz) throws NoSuchElementException {
+      if(!exists()) {
+        throw new NoSuchElementException(
+            "Plugin execution " + valueComputer.mojo.getId() + "does not have a configuration parameter " + name);
```
kwin, line 803: this should emit the full path to the parameter

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
```diff
@@ -714,56 +713,104 @@ private Properties loadLastUpdated(ArtifactRepository localRepository, Artifact
...
+      }
+    }, monitor);
   }
 
-  private static final char PATH_SEPARATOR = '/';
+  private record ValueComputer(ConverterLookup lookup, MojoDescriptor mojo, ClassLoader pluginRealm,
+      ExpressionEvaluator evaluator) {
 
-  private static final char GROUP_SEPARATOR = '.';
+    private <T> T convertConfigurationValueToType(PlexusConfiguration configuration, Class<T> clazz)
```
kwin, line 745: this method name is a bit misleading as it also does resolve expression, maybe just name `compute`, or rather rename this record to `ValueFactory` with name `T create(configuration, clazz)`

kwin (commented):
Looks good overall, added some minor implementation remarks.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
HannesWell: I agree that the top-level element should always be valid and if actually absent should be treated as empty.
But since (some) elements can be present or absent, it is also a valid workflow to check if a certain element exists and only handle it if it exists. Therefore I think throwing an exception in that case is not so convenient and wanted ConfigurationElement to be similar to Optional.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
HannesWell: That is a good point. But if one calls `get("rootElementForCollection).as(Collection.class)` the elements of the returned collection are (probably) of the type used in the Mojo, which are usually not available in a connector. I think it is better to have a collection/stream of ConfiguraitonElements here so that a caller can further query on this abstract/proxy level.

However, I have to admit I never actually had such case yet and therefore should create a test-case for that. Nevertheless I think it is relevant for the example given initially.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
HannesWell: Good suggestion.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
HannesWell: Done.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
HannesWell: The name/path is already the full path. When a new `ConfigurationParameterImpl` is created the parents path concatenated with the elements name is passed as path.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
HannesWell: Indeed. Thanks for the suggestions.
Done.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
HannesWell: I have no strong opinion about that.
But the `MavenImpl` class is already quite large and moving the configuration related classes into a dedicated class makes it simpler to overview that story.

mickaelistria:
I'm skeptical in general about the idea of m2e building a different model and new classes for concepts that are mostly already expressed by Maven API. This makes m2e much harder to understand IMO, while sticking to Maven API is actually keeping things simple. For this case, I believe that instead of a IConfigurationElement model, some "MojoConfigurationHelper" that has static method to manipulate Maven's MojoExecution and other plain Maven types would in the end be simpler.

kwin:
To give more context: In #981 there was the need to evaluate nested Mojo parameters and currently there is no support for that. I proposed the simpler approach https://github.com/eclipse-m2e/m2e-core/pull/981/files#diff-367681037e5adcf9592ed08e4e4ae7e8e7372a00b38d77e4d66bcb7f8e60b59dR733. Maybe there is a third alternative you have in mind @mickaelistria...

laeubi:
As mentioned elsewhere, I don't think Mojo Parameters are a concern of `IMaven` and we should get rid of it here, move it to the `MavenToolbox` or a dedicated place.

kwin:
Mojo parameters can be evaluated with m2e 2.0 via `IMaven` already. I am fine with deprecating that method but please stay backwards compatible this time.

laeubi:
> Mojo parameters can be evaluated with m2e 2.0 via `IMaven` already.

But this does not mean we should add new ones...

> I am fine with deprecating that method but please stay backwards compatible this time.

There is no backward-compatible way for deprecation.

kwin:
@laeubi It seems you don’t get the point about deprecation. It gives downstream consumers time to migrate and an indication where to migrate to. Final removal of deprecated items should only be done in m2e 3.0 earliest

laeubi:
> Final removal of deprecated items should only be done in m2e 3.0 earliest

It _should not_, it _will happen_ at 3.0 ...

HannesWell:
> > Final removal of deprecated items should only be done in m2e 3.0 earliest
> 
> It _should_ not, it _will_ happen at 3.0 ...

Yes, at with 3.0 (whenever this will be) not earlier and not later. And with #996 there will be a proper deprecation notice including the path for migration. So this time the migration should be clear. :)

HannesWell:
> As mentioned elsewhere, I don't think Mojo Parameters are a concern of `IMaven` and we should get rid of it here, move it to the `MavenToolbox` or a dedicated place.

A note about the remarks in https://github.com/eclipse-m2e/m2e-core/pull/981:
- Plug-in executions are usually bound to a project, so the new methods could be part of the MavenProjectFacadade or in any other project bound api.

PR #1000, opened 2022-10-21
Use Wild Web Developer 1.0 stream

mickaelistria:


laeubi:
@mickaelistria is this ready to be merged?

mickaelistria:
I think it's better to wait a couple of more days, Wild Web Developer should be released tomorrow and we could then use the release URL direclty.

mickaelistria:
Actually, it's probably better to merge it right now to leave more time for testing.

PR #1019, opened 2022-10-30
Enhance the testcase to make sure extensions are not looked up from projects context

laeubi:
It seems the testcase is loading from the core extension instead of from the project context as desired :-\

### DIFF
--- a/org.eclipse.m2e.core.tests/src/org/eclipse/m2e/core/ExtensionsTest.java
+++ b/org.eclipse.m2e.core.tests/src/org/eclipse/m2e/core/ExtensionsTest.java
@@ -20,6 +20,7 @@ import org.apache.maven.AbstractMavenLifecycleParticipant;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.Adapters;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.m2e.core.embedder.IComponentLookup;
 import org.eclipse.m2e.core.project.IMavenProjectFacade;
 import org.eclipse.m2e.core.project.ResolverConfiguration;
 import org.eclipse.m2e.tests.common.AbstractMavenProjectTestCase;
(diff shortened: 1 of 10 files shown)


laeubi:
This locally shows a very strange behavior for me:

- if no `.mvn` folder is there then project extensions are looked up, but also for non-project context
- If `.mvn` folder is there this do not work anymore, regardless of context where it is called if the folder is just empty
- - If `.mvn` folder is there and it contains an extension, than only these are looked up

Something is wrecked in the container setup... but this plexus stuff is just horrible to debug ... so lets see if one can see the problem also in the CI .

laeubi:
FYI @HannesWell my current assumption is that somewhere the project is loaded by the default container, while then the facade uses the "mvn folder scoped" one, then it could happen that the class realms are not compatible because now we create different class worlds! I'll instigate this tomorrow.

laeubi:
I think one of the problems is `MavenImpl.readMavenProjects(Collection<File>, ProjectBuildingRequest)` that reads the models with the default container.

HannesWell:
> FYI @HannesWell my current assumption is that somewhere the project is loaded by the default container, while then the facade uses the "mvn folder scoped" one, then it could happen that the class realms are not compatible because now we create different class worlds! I'll instigate this tomorrow.

With other test-cases I noticed that if a project does not have a 'own' `.mvn` folder latest the .mvn folder in the m2e git-repo root is used because the search for it starts at the project folder and goes up until it finds one in any parent or if the file-system root is reached. Therefore all test cases that are sensitive for extensions need to have a `.mvn` folder. Otherwise the one of m2e-core itself is used.
Maybe that explains your strange observations.

HannesWell:
Can you point me to the point where extension plug-ins are loaded? Or is it the point to find that out and have a project scoped context/component lookup for extension projects.

laeubi:
Actually it **should** work like this:

1. You use `IMaven` => Global Context without any project scoped or core extension, just "bare maven"
2. You use `IMavenProjectFacade` **without** an execution => depending on the basedir you get the global context or one that contains the defined core-extensions but **without** project extensions
3. You use IMavenProjectFacade with an execution => you get 1 or 2 but **with project scoped extension**

Problem currently is, that m2e calls many things in IMaven e.g. to read proejcts, then those projects use the global realm even though they are supposed to use the project realm.

So unless 
- https://github.com/eclipse-m2e/m2e-core/pull/1021
- https://github.com/eclipse-m2e/m2e-core/pull/1020

is fixed we better leave this alone and see if it "just works" after getting the lookup-chaos solved :-)

laeubi:
Lets see if the recent refactoring make anything better here...

laeubi:
The changes didn't fix the problem, but I was now able to better understand *why* it happens, this is because of a very strange behavior of plexus bean lookup (and what has drive me insane a bit):

- If you do a single lookup everything "works" and you might get no item
- If I debug inside the lookup code an empty list is returned
- If one iterates afterwards over the list there are suddenly other items!

The reason is, that plexus do not return a copy of the lookup, but returns a dynamic collection! So it depends on where you call it, and strange enough if calling outside any context, simply return all beans from all context! So the solution here would be that we always set a context and always return a copy of the returned list!

laeubi:
This finally fixes also the problem that the pomeditor sometimes show too many packaging types.

PR #1020, opened 2022-10-31
Deprecate (and remove usages) of IMaven.readMavenProject

laeubi:
IMaven.readMavenProject(File, ProjectBuildingRequest) is dangerous to use in global context (aka IMaven) as it do not take basedir / extensions into account. Actually reading the raw maven model should not be exposed to consumers at all as it is very easy to get wrong and do not support caching.

This deprecates the method and converts usages to internal toolbox instead that are used in the correct context.

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
@@ -106,7 +106,10 @@ public interface IMaven extends IComponentLookup {
    * @since 1.4
    * @see {@link #readMavenProjects(File, ProjectBuildingRequest)} to group requests and improve performance (RAM and
    *      CPU)
+   * @deprecated this method should never have been API and is prone to errors, if you still need this method please
+   *             contact the m2e team to provide better alternatives for your use-case
    */
+  @Deprecated(forRemoval = true)
   MavenExecutionResult readMavenProject(File pomFile, ProjectBuildingRequest configuration) throws CoreException;
 
   /**
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/actions/SelectionUtil.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/actions/SelectionUtil.java
@@ -58,6 +58,7 @@ import org.eclipse.m2e.core.embedder.ArtifactKey;
 import org.eclipse.m2e.core.embedder.IMaven;
 import org.eclipse.m2e.core.embedder.IMavenExecutionContext;
 import org.eclipse.m2e.core.internal.IMavenConstants;
+import org.eclipse.m2e.core.internal.IMavenToolbox;
 import org.eclipse.m2e.core.project.IMavenProjectFacade;
 import org.eclipse.m2e.core.project.IMavenProjectRegistry;
 import org.eclipse.m2e.core.ui.internal.Messages;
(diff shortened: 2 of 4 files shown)


laeubi:
@HannesWell this shows that there is something broken with current plexus container handling ... need to investigate what could be the cause...

HannesWell:
> @HannesWell this shows that there is something broken with current plexus container handling ... need to investigate what could be the cause...

Thanks. I can/will have a look at that as well.
I'm not sure, but can this also be the reason why #852 does not always work in a workspace build?


Btw. that method was introduced with 293a1c09ed3ec5401093b7b3b783ea161838b6f0.

laeubi:
Probably, I'm currently working on getting rid of some of the `IMaven` references that might improve the situation already, also most probably one needs to change how readprojects works, currently it groups by the mapping config, but I think we need to group by PlexusContainer as well to get it right.

laeubi:
There is even `readProject` method that returns a project as well ... :-\

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
```diff
@@ -106,7 +106,10 @@ boolean isUnavailable(String groupId, String artifactId, String version, String
    * @since 1.4
    * @see {@link #readMavenProjects(File, ProjectBuildingRequest)} to group requests and improve performance (RAM and
    *      CPU)
+   * @deprecated this method should never have been API and is prone to errors, if you still need this method please
+   *             contact the m2e team to provide better alternatives for your use-case
    */
+  @Deprecated(forRemoval = true)
   MavenExecutionResult readMavenProject(File pomFile, ProjectBuildingRequest configuration) throws CoreException;
```
HannesWell, line 113: Shouldn't this changed to be a default implementation that calls the implementation from the IMavenToolBox with a custom/standalone (just how it is done a.t.m) component-lookup?

laeubi: Actually I don't want to change the deprecated stuff as in the past ti was much more work to delegate "right" to the new impl than just keep the old impl alone.

HannesWell:
> There is even `readProject` method that returns a project as well ... :-\

And `readMavenProjects()` that returns multiple projects...

Since those methods accept the pomFile, can't his be enhanced to aquire a `PlexusContainerManager` from that file and use that?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
HannesWell: Understand. Keeping the old one as it is also prevents from accidental behavior changes due to differences in the new method.

laeubi:
Yes thats essentially what I try to archive with
- https://github.com/eclipse-m2e/m2e-core/pull/1021

because currently that is bound to the `MavenImpl` ... its just a mess :-D

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
laeubi: Yep, especially if we want to change the new method in an incompatible way this is just too much hassle. This way we simply can delete the method at m2e 4.0 and until then no much need to touch it anyways.

HannesWell:
> 
> because currently that is bound to the `MavenImpl` ... its just a mess :-D

A good time to clean this up 👍🏽 :)

laeubi:
This is now part of 
- https://github.com/eclipse-m2e/m2e-core/pull/1022

HannesWell:
> This is now part of
> 
>     * [Move MojoExecutions to the ProjectFacade #1022](https://github.com/eclipse-m2e/m2e-core/pull/1022)

Can this then be closed?

PR #1021, opened 2022-10-31
Get rid of MavenImpl in MavenExecutionContext

laeubi:
FYI @HannesWell lets see if we can get rid of this here to containers are used more consistently when executing requests.

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
@@ -29,6 +29,7 @@ import java.util.Properties;
 import java.util.Set;
 
 import org.eclipse.aether.ConfigurationProperties;
+import org.eclipse.aether.DefaultRepositorySystemSession;
 import org.eclipse.aether.transfer.TransferListener;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.runtime.CoreException;
@@ -36,8 +37,10 @@ import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Status;
 
+import org.codehaus.plexus.PlexusContainer;
 import org.codehaus.plexus.classworlds.realm.ClassRealm;
 
+import org.apache.maven.DefaultMaven;
 import org.apache.maven.Maven;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.InvalidRepositoryException;
@@ -64,9 +67,9 @@ import org.apache.maven.repository.RepositorySystem;
 import org.apache.maven.session.scope.internal.SessionScope;
 import org.apache.maven.settings.Settings;
 
+import org.eclipse.m2e.core.MavenPlugin;
 import org.eclipse.m2e.core.embedder.ICallable;
 import org.eclipse.m2e.core.embedder.IComponentLookup;
-import org.eclipse.m2e.core.embedder.IMaven;
 import org.eclipse.m2e.core.embedder.IMavenConfiguration;
 import org.eclipse.m2e.core.embedder.IMavenExecutionContext;
 import org.eclipse.m2e.core.internal.MavenPluginActivator;
@@ -89,7 +92,7 @@ public class MavenExecutionContext implements IMavenExecutionContext {
 
   private static final ThreadLocal<Deque<MavenExecutionContext>> threadLocal = new ThreadLocal<>();
 
-  private final MavenImpl maven;
+  private final PlexusContainer container;
 
   private MavenExecutionRequest request;
 
@@ -100,8 +103,11 @@ public class MavenExecutionContext implements IMavenExecutionContext {
 
   private IMavenProjectFacade projectFacade;
 
-  public MavenExecutionContext(MavenImpl maven, IMavenProjectFacade projectFacade) {
-    this.maven = maven;
+  private IComponentLookup containerLookup;
+
+  public MavenExecutionContext(PlexusContainer container, IMavenProjectFacade projectFacade) {
+    this.container = container;
+    this.containerLookup = PlexusContainerManager.wrap(container);
     this.projectFacade = projectFacade;
     if(projectFacade != null) {
       File basedir;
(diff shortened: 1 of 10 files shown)


laeubi:
@HannesWell only the two felix tests fail, is there anything special with those?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
```diff
@@ -476,7 +485,7 @@ public IComponentLookup getComponentLookup() {
       throw new IllegalStateException();
     }
     if(projectFacade == null) {
-      return maven;
+      return containerLookup;
```
HannesWell, line 488: Since we have a containerLookup at hand here, why isn't that always returned?
Is it because in case a projectFacade is present the project's classrealm is used as classloader?

HannesWell (commented):
The intention of this change is to bind a MavenExecutionContext to a PlexusContainer and therefore one '.mvn' folder and consequently the associated extensions, isn't it?

I think that is a good improvement and I assumed that this were the case already.

I just have a few suggestions for minor refactoring that I'm about to push as separate commit to this branch. Feel free t squash them into you commit or to discard them as you think it is appropriated.

HannesWell:
> @HannesWell only the two felix tests fail, is there anything special with those?

Actually not really. They don't even have extensions... I have a look.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
laeubi: The passed lookup is only used for the case where we have not a project and otherwise depend on the project state, but probably can later be simplified... for now I just want to do it a bit step-by-step than one big bang :-)

laeubi:
> > @HannesWell only the two felix tests fail, is there anything special with those?
> 
> Actually not really. They don't even have extensions... I have a look.

If I start m2e from my eclipse I can import the project and it has the pde nature so everything looks fine....

HannesWell:
> > > @HannesWell only the two felix tests fail, is there anything special with those?
> > 
> > 
> > Actually not really. They don't even have extensions... I have a look.
> 
> If I start m2e from my eclipse I can import the project and it has the pde nature so everything looks fine....

Do the projects also show up in the `Target Platform State` view?
Because the failure actually says that the projects are not found in PDE's PluginRegistry. The import itself succeeds.
I can reliably reproduce the failure locally when executing them as Plugin-Tests. The previous commit works fine.
Although I don't know why this change affects PDE's plugin registry it looks like it does so.

laeubi:
Yes I can reproduce it as well, but why only felix has a problem but not bnd? It seems for some reason the mojo is not executed here and thus no manifest is written.

HannesWell:
> Yes I can reproduce it as well, but why only felix has a problem but not bnd? It seems for some reason the mojo is not executed here and thus no manifest is written.

I can confirm that the manifest is not written with this change.
Can it be that it is because the maven-bundle-plugin is configured as extension? That's the only difference I see a.t.m.
```
<plugin>
	<groupId>org.apache.felix</groupId>
	<artifactId>maven-bundle-plugin</artifactId>
	<extensions>true</extensions>
</plugin>
```

laeubi:
Good catch, is extension needed here? I'm not aware of it being required?

HannesWell:
> Good catch, is extension needed here? I'm not aware of it being required?

😎 It is required for the project with `<packaging>bundle</packaging>`. That's why there are two cases, one jar-project and one bundle-project.

laeubi:
I checked this and it works for the jar packaging then, but complains about unknown `bundle` packaging, would you be fine if we for now ignore that test until project scoped extensions are fixed? I think this the same issue I have described here:
- https://github.com/eclipse-m2e/m2e-core/pull/1019

that currently project extensions leak into the default context and project extensions actually don't work because we load them always with the default context. I just don't know how fast we can fix this ...

laeubi:
BTW  BND also has a mode as an extension:

https://github.com/bndtools/bnd/blob/master/maven/bnd-maven-plugin/README.md#jar-goal

I think the best would be two have two configurations then one with extension=true and one without...

HannesWell:
> would you be fine if we for now ignore that test until project scoped extensions are fixed? I think this the same issue I have described here:
> 
>     * [Enhance the testcase to make sure extensions are not looked up from projects context #1019](https://github.com/eclipse-m2e/m2e-core/pull/1019)

I would be ok, if we temporarily disable that test, but we should definitely fix that soon before the next release then. Otherwise projects that use the same setup like in the tests are effectively broken. 


> I think the best would be two have two configurations then one with extension=true and one without...

That's a good suggestion. Similar like for the felix-plugin. Do you plan to contribute such test case?

laeubi:
> That's a good suggestion. Similar like for the felix-plugin. Do you plan to contribute such test case?

Not yet, but I noticed that we only test the "good" case, so actually one would need one test case:

1. felix with extension = true and check that jar works but bundle gives an error marker
2. felix with extension = true and check that both works
3. same for BND jar / bnd-process...

laeubi:
> would be ok, if we temporarily disable that test, but we should definitely fix that soon before the next release then.

Sure I'll just disable they now and see if any of the integration test fail, if yes I think I need to broaden the scope of this change :-\

laeubi:
Great build successful except those felix stuff, lets merge this and build on top of it ...

HannesWell:
> > That's a good suggestion. Similar like for the felix-plugin. Do you plan to contribute such test case?
> 
> Not yet, but I noticed that we only test the "good" case, so actually one would need one test case:
> 
>     1. felix with extension = true and check that jar works but bundle gives an error marker
> 
>     2. felix with extension = true and check that both works
> 
>     3. same for BND jar / bnd-process...

Do you plan to add such test case? If not I can do that because I think we should test as much as possible in that area.

laeubi:
> > > That's a good suggestion. Similar like for the felix-plugin. Do you plan to contribute such test case?
> > 
> > 
> > Not yet, but I noticed that we only test the "good" case, so actually one would need one test case:
> > ```
> > 1. felix with extension = true and check that jar works but bundle gives an error marker
> > 
> > 2. felix with extension = true and check that both works
> > 
> > 3. same for BND jar / bnd-process...
> > ```
> 
> Do you plan to add such test case? If not I can do that because I think we should test as much as possible in that area.

That would be very helpful I'm currently debugging the Plexus stuff and more tests are always better.

PR #1022, opened 2022-10-31
Move MojoExecutions to the ProjectFacade

laeubi:
MojoExecutions are actually a concern of the project facade and require the execution in the project context than in the global context.

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
@@ -70,6 +70,8 @@ import org.eclipse.core.runtime.jobs.ISchedulingRule;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.osgi.util.NLS;
 
+import org.codehaus.plexus.PlexusContainer;
+
 import org.apache.maven.artifact.repository.MavenArtifactRepository;
 import org.apache.maven.execution.DefaultMavenExecutionResult;
 import org.apache.maven.execution.MavenExecutionRequest;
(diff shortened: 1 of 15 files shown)


HannesWell:
With all the changes I wonder what will be eventually left for the `IMaven` interface? Only context free tasks?
But if one is pedantic, actually nothing much is left. 
For example the results of the `resolve()`, `isUnavailable()` and even `getSettings()` or `getArtifactRepositories()` or alike can vary with if we finally support `maven.config` (PR #852) files and one has the `--settings` argument or `-Dmaven.repo.local=<path to alternate local repo>` specified in the `maven.config`.

And we should improve the documentation of the new methods.

Furthermore it looks like the new PDE-connector tests are very valuable. 😅

laeubi:
> With all the changes I wonder what will be eventually left for the `IMaven` interface?

At best we can delete it completely :-)

> Furthermore it looks like the new PDE-connector tests are very valuable. sweat_smile

At least I was able to break them all :-D

HannesWell:
> > With all the changes I wonder what will be eventually left for the `IMaven` interface?
> 
> At best we can delete it completely :-)
> 
I have more and more the impression that eventually this will be the result, if we really want to be fully .mvn-folder aware. 


> > Furthermore it looks like the new PDE-connector tests are very valuable. sweat_smile
> 
> At least I was able to break them all :-D

And it is good to have some thing that notices your breaking changes :P

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
```diff
@@ -740,56 +739,66 @@ private Map<IFile, MavenProjectFacade> readMavenProjectFacades(Collection<IFile>
...
+      Map<PlexusContainer, List<IFile>> pomFiles = mapToContainer(entry.getValue());
+      SubMonitor containerMonitor = subMonitor.split(pomFiles.size());
+      containerMonitor.setWorkRemaining(pomFiles.size());
+      for(var containerEntry : pomFiles.entrySet()) {
+        List<IFile> fileList = containerEntry.getValue();
+        MavenExecutionContext context = new MavenExecutionContext(containerEntry.getKey(), null);
+        configureExecutionRequest(context.getExecutionRequest(), state, fileList.size() == 1 ? fileList.get(0) : null,
+            resolverConfiguration);
+
+        result.putAll(context.execute((ctx, mon) -> {
```
HannesWell, line 751: Looks like you forgot to set the execution result into the 'result' map.
Lets see if that fixes the tests.

laeubi:
@HannesWell thanks, seems it was too late for coding yesterday... but it seems GH is fine now and Jenkins has a testfailure that seems a bit strange, maybe a race condition?

laeubi:
I now cleaned up the last reference and then we should hopefully be back to a usable state!

HannesWell:
> I now fixed the NPEs but still test fails on Jenkins.... I'll disable the test for now.

I wonder if something is still not done right if the NPE only shows up with this change and never happened before?


To recap the API changes done with this PR:
- Added:
  - `IMavenExecutionContext.of(File)`
  - `IMavenProjectFacade.calculateExecutionPlan()`
  - `IMavenProjectFacade.setupExecutionPlan()`
- Deprecated:
  - IMaven.calculateExecutionPlan() (replaced by `IMavenProjectFacade.calculateExecutionPlan()` and `IMavenProjectFacade.setupExecutionPlan()`)
  - `IMaven.setupMojoExecution()` - no replacement (internal use only)
  - `IMaven.readMavenProject()` - no replacement yet (if there is a need m2e should be contacted)
  - `IMaven.readMavenProjects()` - no replacement yet (if there is a need m2e should be contacted)

Given that we have to make M2E more context/ .mvn folder aware all those changes are reasonable for me and they even make the API better because the new methods have less arguments and especially no boolean flags.
Nevertheless maybe @mickaelistria want's to assess them as well?

laeubi:
> I wonder if something is still not done right if the NPE only shows up with this change and never happened before?

I have no clue I only saw this in the Jenkins CI log... local build is fine, and GH actions as well so I'm kind of lost. Also the called API states it will never return null... but there are some places where it is checked for null ...

PR #1024, opened 2022-11-01
Clean up and improve o.e.m2e.core internals

HannesWell:
Various clean ups, simplifications and some (expected) performance improvements in `MutableProjectRegistry`.
@laeubi or @mickaelistria do you want to review this?

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/BasicProjectRegistry.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/BasicProjectRegistry.java
@@ -94,18 +94,17 @@ abstract class BasicProjectRegistry implements Serializable {
    */
   @SuppressWarnings({"unchecked", "rawtypes"})
   private static void copy(Map from, Map to) {
-    for(Map.Entry entry : (Set<Map.Entry>) from.entrySet()) {
-      Object value = entry.getValue();
+    from.forEach((key, value) -> {
       if(value instanceof Map mapValue) {
         Map map = new LinkedHashMap();
         copy(mapValue, map);
+        to.put(key, map);
         value = map;
       } else if(value instanceof Set setValue) {
-        Set set = new LinkedHashSet(setValue);
-        value = set;
+        value = new LinkedHashSet(setValue);
       }
-      to.put(entry.getKey(), value);
-    }
+      to.put(key, value);
+    });
   }
 
   public MavenProjectFacade getProjectFacade(IFile pom) {
@@ -148,13 +147,7 @@ abstract class BasicProjectRegistry implements Serializable {
   }
 
   public boolean isValid() {
-    return workspaceArtifacts != null //
-        && workspacePoms != null //
-        && workspacePomFiles != null //
-        && requiredCapabilities != null //
-        && projectCapabilities != null //
-        && projectRequirements != null //
-        && areFacadesValid();
+    return areFacadesValid();
   }
 
   private boolean areFacadesValid() {
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenCapability.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenCapability.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008-2010 Sonatype, Inc.
+ * Copyright (c) 2008-2022 Sonatype, Inc.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
@@ -53,7 +53,7 @@ public class MavenCapability extends Capability {
 
   @Override
   public String toString() {
-    return getVersionlessKey().toString() + "/" + version; //$NON-NLS-1$
+    return getVersionlessKey() + "/" + version; //$NON-NLS-1$
   }
 
   @Override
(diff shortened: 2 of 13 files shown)


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/builder/MavenBuilderImpl.java
```diff
@@ -113,9 +113,8 @@ public Set<IProject> build(MavenSession session, IMavenProjectFacade projectFaca
     Map<Throwable, MojoExecutionKey> buildErrors = new LinkedHashMap<>();
     MavenProjectMutableState snapshot = MavenProjectMutableState.takeSnapshot(mavenProject);
     try {
-      for(Entry<MojoExecutionKey, List<AbstractBuildParticipant>> entry : participants.entrySet()) {
-        MojoExecutionKey mojoExecutionKey = entry.getKey();
-        for(InternalBuildParticipant participant : entry.getValue()) {
+      participants.forEach((mojoExecutionKey, executionParticipants) -> {
```
laeubi, line 116: I don't seeing any improvement using foreach with lamda over explicit loop beside its harder to read, harder to debug and produces additional runtime cost.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/BasicProjectRegistry.java
```diff
@@ -94,18 +94,16 @@ protected final void replaceWith(BasicProjectRegistry other) {
    */
   @SuppressWarnings({"unchecked", "rawtypes"})
   private static void copy(Map from, Map to) {
-    for(Map.Entry entry : (Set<Map.Entry>) from.entrySet()) {
-      Object value = entry.getValue();
+    from.forEach((key, value) -> {
```
laeubi, line 97: I don't no what this method does but the lamda do not make it more clear I think... should it be some kind of deepcopy of a map?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenCapability.java
```diff
@@ -85,13 +85,7 @@ public static MavenCapability createMavenParent(ArtifactKey key) {
...
 
   static String getId(ArtifactKey key) {
-    StringBuilder sb = new StringBuilder();
-    sb.append(key.groupId());
-    sb.append(':').append(key.artifactId());
-    if(key.classifier() != null) {
-      sb.append(':').append(key.classifier());
-    }
-    return sb.toString();
+    return key.groupId() + ":" + key.artifactId() + (key.classifier() != null ? ":" + key.classifier() : "");
```
laeubi, line 88: I find this hard to read to have concatenation and trigraph mixed in one expression. Objects.RequireNonNullOrElse might be more expressive.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
```diff
@@ -634,7 +597,7 @@ public IMavenExecutionContext createExecutionContext() {
       }
       return new MavenExecutionContext(PlexusContainerManager.wrap(container, mavenProject.getClassRealm()), this);
     } catch(Exception ex) {
-      throw new RuntimeException("Aquire container failed!", ex);
+      throw new IllegalStateException("Aquire container failed!", ex);
```
laeubi, line 600: Why is this an illegal state?

laeubi (commented):
Just some quick remarks.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
HannesWell: Sonar-lint complained that one should use a more specific exception than `RuntimeException`.
Since if the container creation failed, I would say that the facade is in an illegal state. Therefore IMHO `IllegalStateException` is most suitable of the existing RuntimeExceptions.

But I can revert that if that is only my opinion.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/BasicProjectRegistry.java
HannesWell: > I don't no what this method does but the lamda do not make it more clear I think... should it be some kind of deepcopy of a map?

I have no details about that either, but yes I think that is a deep copy of Maps/Sets that does not copy the elements itself.
But personally I think the forEach makes it more clear, because no casting and key/value getting is necessary, if one knows/is used to that method.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
laeubi: Don't trust sonar lint... that's just useless to invent new exceptions (or use unrelated ones) that are not used just to make tools happy. Actually we once had an (checked) exception but @mickaelistria has explicitly changed that because we mostly never would be able to react than throwing an error.

 Whatever makes the container failing its not because of an illegal state (what would be for example accessing a method after an object was disposed) but either an bug in eclipse or a wrong pom configuration or whatever.

HannesWell: > Don't trust sonar lint... 

In general I wouldn't say that because often the complaints/suggestions are valid and I even learned some things from that. Nevertheless they are not always right, so one should probably see that more as a suggestion than as mandatory and of course nobody is forced to follow that.

I'll revert that.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/builder/MavenBuilderImpl.java
HannesWell: Reverted it here.

PR #1027, opened 2022-11-02
Adjust MavenExecutionContext to be more well-defined

laeubi:
Currently we do some computations regarding the basedir in MavenExecutionContext that later influences the session created, but actually this should be the concern of the creator to decide if and what basedir has to be used, e.g. even if there is currently no project a basedir might still be known.

This changes the MavenExecutionContext to accept a basedir on construction time and changes the ImavenFacade to a more generic supplier.

FYI @HannesWell that should fix the obstacles you see regarding basedir/multiProjectBasedir

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
@@ -24,13 +24,14 @@ import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
+import java.util.function.Function;
 
 import org.eclipse.aether.ConfigurationProperties;
 import org.eclipse.aether.DefaultRepositorySystemSession;
 import org.eclipse.aether.transfer.TransferListener;
-import org.eclipse.core.resources.IFile;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.NullProgressMonitor;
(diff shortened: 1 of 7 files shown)


mickaelistria (approved):
Allowing MavenExecutionContext to work with plain directory instead of IMavenProjectFacade is going to allow m2e features to work on pom files and folders that do not have m2e nature; this is very welcome.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
```diff
@@ -92,29 +92,30 @@ public class MavenExecutionContext implements IMavenExecutionContext {
...
+   * @param lookup the lookup that is used to query for components, this lookup must support to lookup at least a
+   *          {@link PlexusContainer} for session setup
+   * @param baseDir the basedir of the execution, this is effectively the folder where one normally would execute the
+   *          maven command, also often refereed to as the working directory of the process. If it is <code>null</code>
+   *          we assume an anonymous execution, some feature might not work in such a context.
+   * @param projectSupplier if an execution is started without an explicit project and the projectSupplier is not
+   *          <code>null</code> it is queried to supply the MavenProject used for session setup.
+   */
+  public MavenExecutionContext(IComponentLookup lookup, File baseDir,
+      Function<? super MavenExecutionContext, MavenProject> projectSupplier) {
```
HannesWell, line 115: Is there a reason that `projectSupplier` is a Function and not a real supplier or simply passing the `MavenProject` directly?
Because the only caller that does not pass null, just passes a function that always returns the same project.
So I wonder if the last argument could just be a MavenProject?
Or is this required for some later work?

HannesWell (commented):
Thank you Christoph.
The change looks good. I tought about something similar in this direction as well, but didn't want to change that much in #852.

HannesWell:
And since there are many callers that pass null as last argument, a second two-argument constructor for `MavenExecutionContext` that delegates to the three-arg one could also be considered.

PR #1033, opened 2022-11-02
Clear reference to the ProjectBuildingRequest after loading the project

laeubi:
Currently a MavenProject holds a reference to the ProjectBuildingRequest (which is big) and thus retains the object in memory as long as the MavenProject is in memory.

The method itself is deprecated in maven and explained as

"This is an utility method that is meant to assist integrators of Maven, it must not be used by Maven plugins."

As m2e do not used this method it seems valid to clear the reference always to conserve space.

FYI @mickaelistria lets see if this breaks anything ...


mickaelistria:
It doesn't seem bad to remove it; but I'm not sure we'll save that much.
Also, you may want to remove it from the whole parency chain.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
```diff
@@ -974,13 +974,6 @@ MutableProjectRegistry newMutableProjectRegistry() {
    *           been created
    */
   void applyMutableProjectRegistry(MutableProjectRegistry newState, IProgressMonitor monitor) throws CoreException {
-    // don't cache maven sessions
-    for(MavenProjectFacade facade : newState.getProjects()) {
```
HannesWell, line 978: I wonder if it is a problem if this is not done anymore?
Or is it irrelevant because clearProjectBuildingRequest() is called in the IMavenToolbox?

laeubi:
I used the Apache camel as a test project and raised the cache size to 1000 this results in 2481 items (so it seems the cache is not really cleared) and a retained size of 4.5 MB...

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/ProjectRegistryManager.java
laeubi: Yes IMavenToolbox clears the reference now always.

laeubi:
Alright, clearing out the parent reference as well made things much better. I see MavenProject Size of max 4MB per instance but still more instance than cache size are kept around.

laeubi:
I poked around a bit and I think the problem is that items are not deduplicated....For example there 2 **Million (!)** `Dependency` instance (**using 720 MB** alone!) and even if maven tend to pull in much, this seems a bit too much even for a large project like camel.

The other what I currently can not proof is, that I have the feeling that parent projects are not deduplicated, so lets say we have 100 projects and each has 1 parent that reference another parent we might end up with 300 items in the cache! e.g I see 1018 project facades (what also seem to much!) but 3182 MavenProject instance!

laeubi:
There are 798.535 String instance of `org.apache.camel` :-O

laeubi:
If there are no objections I think we can merge this already, I'll then propose more improvement in separate PRs.

laeubi:
Just a summary of my duplicate analysis with the came project:

- There are 1147 items in the cache.. searching for duplicates...
- There are 574 unique facades in the cache (according to GAV)
- There are 575 unique projects in the cache (according to GAV) and a total of 3606

so it seems there are actually **each facade two times** in the cache, and about **six times duplicated Maven models** (according to GAV) ...

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
```diff
@@ -484,15 +484,10 @@ public Map<File, MavenExecutionResult> readMavenProjects(Collection<File> pomFil
     return result;
   }
 
-  /**
-   * Makes MavenProject instances returned by #readProject methods suitable for caching and reuse with other
-   * MavenSession instances.<br/>
-   * Do note that MavenProject.getParentProject() cannot be used for detached MavenProject instances. Use
-   * #resolveParentProject to resolve parent project instance.
-   */
+  @Deprecated
```
HannesWell, line 487: The deprecation annotation should have the for removal flag set and java-doc should be added that this method is not necessary anymore and one can safely cache all projects (but should somebody do that?).

HannesWell:
> 
> The other what I currently can not proof is, that I have the feeling that parent projects are not deduplicated, so lets say we have 100 projects and each has 1 parent that reference another parent we might end up with 300 items in the cache! e.g I see 1018 project facades (what also seem to much!) but 3182 MavenProject instance!

Wouldn't a per parent+project level caching as suggested by Mickael in https://github.com/eclipse-m2e/m2e-core/discussions/1032#discussioncomment-4039939 help in that regard?



> There are 798.535 String instance of `org.apache.camel` :-O

That's a lot. `String.intern()` could help here. The question is, why so many different instances are created? I would have expected that they are created on model read and then just passed around? Or are parts of Maven happily constructing and deconstructing keys like `groupId:artifactId`?

Would could help without any code change would be passing `-XX:+UseStringDeduplication` to the JVM. Can you try if this helps? Altough it does not reduce the String instances but the underlying byte-arrays.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
laeubi: It is already in the interface, Eclipse just complained that if its deprecated in the interface it should be deprecated in the impl...

laeubi:
> Would could help without any code change would be passing `-XX:+UseStringDeduplication` to the JVM.

The problem is that we then need to add this to the Eclipse EPPs (don't know if this option has any drawbacks) but I'll try that to gather some statistics.

mickaelistria:
> The other what I currently can not proof is, that I have the feeling that parent projects are not deduplicated, so lets say we have 100 projects and each has 1 parent that reference another parent we might end up with 300 items in the cache! e.g I see 1018 project facades (what also seem to much!) but 3182 MavenProject instance!

That is exactly one of the thing I had explained in https://github.com/eclipse-m2e/m2e-core/discussions/1032#discussioncomment-4039939 .

HannesWell:
> > Would could help without any code change would be passing `-XX:+UseStringDeduplication` to the JVM.
> 
> The problem is that we then need to add this to the Eclipse EPPs (don't know if this option has any drawbacks) but I'll try that to gather some statistics.

That's right. But I also considered this as an option for https://github.com/eclipse-equinox/equinox/pull/33, but haven't tried it yet. Nevertheless it would be interesting to know if there is a significant improvement with that.

PR #216, opened 2021-06-05
One way sync for jpms args from maven compiler args to .classpath file

treilhes:
This commit :

- get the maven-compiler-plugin compiler arguments if any
- use regex to parse them and extract jpms arguments (--add-exports,--add-opens,--add-reads,--patch-module)
- dispatch the arguments in the right container (if the target module is part of JRE then in JreContainer else in M2eContainer)
- transform them into classpath attribute (add-exports, add-opens, add-reads, patch-module)

+ 1 junit test

Issue #129, opened 2021-03-30
Bug 543631 - Eclipse - Maven - JPMS (m2e ignores compilerArgs in pom.xml)


LAlves91:
Originally reported by Imanuel Rohlfing at [bug.eclipse.org](https://bugs.eclipse.org/bugs/show_bug.cgi?id=543631). There's also a [Stack Overflow](https://stackoverflow.com/questions/55917386/eclipse-maven-compiler-not-supporting-compiler-plugin-compilerargs) discussion on the subject.

> Currently im migrating legacy-projects from java 8 to java 11. There are some issues using vm args like add-modules, add-> > > exports and add-opens:
> 
> Some base informations:
> - eclipse 2018-12 for Java developers 
> - buildin maven (3.5.3)
> - java 11 (open-jdk-hotspot_11.0.1_13-windows-x64)
> 
> Within the pom im adding modules and exports:
> ```
> <plugin>
> 	<groupId>org.apache.maven.plugins</groupId>
> 	<artifactId>maven-compiler-plugin</artifactId>
> 	<version>${compiler.plugin.version}</version>
> 	<configuration>
> 		<source>11</source>
> 		<target>11</target>
> 		<compilerArgs>
> 			<arg>--add-modules=java.desktop</arg>
> 			<arg>--add-exports=java.desktop/com.sun.java.swing.plaf.motif=ALL-UNNAMED</arg>
> 		</compilerArgs>
> 	</configuration>
> </plugin>
> ```
> Thats working fine. Using maven to compile the project ends with success. 
> But eclipse shows problems (the type xxx is not accessable). 
> 
> To avoid these problems i can modify the project configurtaion: 
> Java Build Path -> Libraries -> JRE System Library [JavaSE-11] -> Is modular -> Edit
> 
> BUT: m2e is controlling the project settings. maven update command removes the changes in the project settings every time.

In the project I work on, we've managed to find a temporary fix by adding the modules we need at the project's configuration (or by adding them in our .classpath file). It stays functional **as long as we don't forget to turn off the option 'Update project configuration from pom.xml' when running 'Maven > Update project...'**.

leerho:
This bug has been open for 3 years now.  I thought this was going to be fixed in 4.22, (2021-12) but the problem is still there. In fact, it not only impacts compiler JPMS args specified with the compiler plugin, it also impacts JPMS args specified in the surefire plugin.

This bug severely impacts smooth migration from Java 8 to Java 9+.

[...]

PR #1034, opened 2022-11-03
Support sync compilerArgs with delimiter

CsCherrYY:
Signed-off-by: Shi Chen <chenshi@microsoft.com>

related to #129 

sync for jpms args has been supported in https://github.com/eclipse-m2e/m2e-core/pull/216, but there is still one way to pass arguments to compilerArgs with `=` as the delimiter between the argument type and argument value, e.g., `<arg>--add-exports=javafx.graphics/com.sun.javafx.geometry=somemodule</arg>`. If you directly run `java` in the command line, you can find the following description at the end of the help description:
```
To specify an argument for a long option, you can use --<name>=<value> or
--<name> <value>.
```

which indicates both of the two styles are supported. but currently argument within `=` can be correctly recognized by maven compiler but m2e can't pass them to the classpath entry attribute. This PR will fix it.

### DIFF
--- a/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/ModuleSupport.java
+++ b/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/ModuleSupport.java
@@ -522,7 +522,17 @@ public class ModuleSupport {
             continue;
           }
 
-          String value = it.next();
+          String value = null;
+          if(argType.getArgumentName().equalsIgnoreCase(argumentName)) {
+            // full argument match
+            value = it.next();
+          } else {
+            // argument with delimiter match
+            String argumentWithDelimiter = argType.getArgumentName().toLowerCase() + JpmsArgType.argumentDelimiter;
+            if(argumentName.toLowerCase().startsWith(argumentWithDelimiter)) {
+              value = argumentName.substring(argumentWithDelimiter.length());
+            }
+          }
           JpmsArgValue argValue = argType.parse(value);
 
           if(argValue == null) {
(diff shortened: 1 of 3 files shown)


laeubi (approved):
Looks good, thanks for improving JPMS support!

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/ModuleSupport.java
```diff
@@ -682,14 +698,27 @@ public List<JpmsArgValue> getFromArgs(JpmsArgs args) {
...
 
-      return result.get();
+      // try to match argument with delimiter
+      result = Arrays.stream(JpmsArgType.values())
+          .filter(at -> argumentName.toLowerCase().startsWith(at.getArgumentName().toLowerCase() + argumentDelimiter))
+          .findFirst();
+
+      if(!result.isEmpty()) {
+        return result.get();
+      }
```
mickaelistria, line 719:
```suggestion
      return result.or(() -> Arrays.stream(JpmsArgType.values())
          .filter(at -> argumentName.toLowerCase().startsWith(at.getArgumentName().toLowerCase() + argumentDelimiter))
          .findFirst()).orElse(null);
```

CsCherrYY: let me create a new PR to refactor it.

mickaelistria: No need to if you're busy with other things; it's more food for thoughts than an actual requirement or request.

PR #1038, opened 2022-11-03
More fine grained info about times used to update Project Configuration

laeubi:
Currently only the total time is displayed, but intermediate steps are usefull as well.


mickaelistria:
That's IMO a good thing to add. Just a not that you may like to try using the Instant and Duration APIs for that; they are quite comfortable to use and provide good default printing/toString.

laeubi:
I once tried the new API but was not that convinced it really helps much in simple cases. But if you like you can rewrite it to show its usage,I might mostly lack experience with that to see its beauty :-)

PR #1039, opened 2022-11-03
Perform a regular cache cleanup and print some statistics

laeubi:


laeubi:
This do not perform any reload, but gives guava cache some time to cleanup (and prints some statistics on debug), my idea is to later perform some more tasks that save some memory.

mickaelistria:
Isn't the cache supposed to self clean-up in a smart enough way? What is the benefit of calling this explicitly?
As I read this patch, I see that it's more or less a workaround for the issue that MavenProjectFacades are not unique (what you mention about hashCode/equals being missing; or maybe the MavenProjectFacade not being the best key for the cache after all when a simple GAV could maybe be a better fit). I don't think this patch is "production-ready".

laeubi:
> Isn't the cache supposed to self clean-up in a smart enough way?

It depends a bit guava says that the cache performs *some* (or maybe none) cleanup actions on access/modification:

https://github.com/google/guava/wiki/cachesexplained#when-does-cleanup-happen

PR #1048, opened 2022-11-08
Add a Quick-Fix to add a dependency on missing types

laeubi:
Currently if a type is missing a user needs to search the workspace, then probably find the project and then go to the pom and add the desired coordinates.

This now adds a Quick-Fix to automate this and offers the user to add a dependency to a workspace project that contains the missing type.

## Missing type
grafik
## show quickfix
grafik
## compare pom with previous state from local history
grafik


mickaelistria:
Excellent!

mickaelistria (approved):

PR #1053, opened 2022-11-10
Remove references to deprecated method of project facade

laeubi:


laeubi:
We have one "evil" test here (`MavenPropertiesTest`) that modifies the project state directly ... this should be fixed.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/IProjectConfigurationManager.java
```diff
@@ -62,6 +63,7 @@ void enableMavenNature(IProject project, IProjectConfiguration configuration, IP
 
   void updateProjectConfiguration(IProject project, IProgressMonitor monitor) throws CoreException;
 
+  @Deprecated(forRemoval = true)
```
HannesWell, line 66: Is there an alternative users should use? I think it would be good to add one or if not make clear why it should not be used and what should be done instead.

PR #1058, opened 2022-11-13
Incooperate the multi-module-base dir into the project configuration

laeubi:
Currently we do not consider the mmbd in the project configuration that is used to group projects "with the same config" what requires some special considerations.

This changes the behavior in a way that the mmpb is part of the persistent configuration


laeubi:
@HannesWell this should help with your maven.config story as well....

HannesWell:
> @HannesWell this should help with your maven.config story as well....

Great. Thanks! I will try that as soon as I have completed my current other work.

PR #1059, opened 2022-11-13
Refresh projects in build order

laeubi:
Currently refreshed projects are returned in random (map) order, this might create noise if we later try to find other projects to refresh and also when debugging as order is not deterministic.

This changes the returned map of a refresh operation to use the project build order if possible.

@cpfeiffer I think this is something you like to archive with what you have mentioned here:
- https://github.com/eclipse-m2e/m2e-core/discussions/1032#discussioncomment-4050203

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/ProjectProcessingTracker.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/ProjectProcessingTracker.java
@@ -29,6 +29,8 @@ import org.eclipse.core.resources.IFile;
  */
 public class ProjectProcessingTracker {
 
+  static final int MAX_INTERATIONS = Integer.getInteger("m2e.project.refresh.maxiterations", 5);
+
   static final Logger log = LoggerFactory.getLogger(ProjectProcessingTracker.class);
 
   private Set<IFile> processed = new LinkedHashSet<>();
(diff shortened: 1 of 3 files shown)


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/ProjectProcessingTracker.java
```diff
@@ -29,6 +29,8 @@
  */
 public class ProjectProcessingTracker {
 
+  static final int MAX_INTERATIONS = Integer.getInteger("m2e.project.refresh.maxiterations", 5);
```
szarnekow, line 32: Nit: Typo in field name

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/ProjectProcessingTracker.java
```diff
@@ -60,6 +64,11 @@ public boolean needsImprovement() {
       log.debug("Seed is empty.");
       return false;
     }
+    iterations++ ;
+    if(iterations > MAX_INTERATIONS) {
+      log.debug("Max iterations reached!");
```
szarnekow, line 69: Please add the current value of max iterations to the log message

laeubi:
@szarnekow you can just hit the edit button for the file on Github to propose a change :-)

szarnekow:
> you can just hit the edit button for the file on Github to propose a change :-)

I would love to.

image

laeubi:
You need to start at the master version of the file:
https://github.com/eclipse-m2e/m2e-core/blob/master/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/ProjectProcessingTracker.java

PR #1061, opened 2022-11-13
Remove unused dependencies from M2E Plug-ins

HannesWell:


org.eclipse.m2e.pde.connector.tests/META-INF/MANIFEST.MF
```diff
@@ -7,7 +7,7 @@ Bundle-Vendor: Eclipse.org - m2e
 Fragment-Host: org.eclipse.m2e.pde.connector
 Require-Bundle: org.eclipse.m2e.tests.common,
  org.junit,
- org.eclipse.pde.ds.core,
- org.eclipse.equinox.preferences
+ org.eclipse.pde.ds.core;bundle-version="1.2.500",
+ org.osgi.service.component.annotations
```
laeubi, line 11: `org.osgi.service.component.annotations` should really not be required anymore...

HannesWell: It turned that the service component annotations are actually required. My explanation is that otherwise they are not pulled into the Eclipse runtime that is launched by Tycho-surefire and therefore the PDE DS processing does not work there which lets the test fail.
I now moved that 'runtime' dependency to the `dependency-resolution`, `extraRequirements` element of the `target-platform-configuration` to indicate that it is not a compile dependency of this tests.

laeubi: Why doe m2e need to test PDE annotation processing???

HannesWell:
Finally the build succeeded and all tests passed, so this can be submitted.
The tests can still be improved later if possible/necessary.

PR #1068, opened 2022-11-17
[PDE-Connector] Derive .classpath "lib" entries from Bundle-Classpath

HannesWell:
This requires
- https://github.com/eclipse-m2e/m2e-core/pull/1067 (currently also included).

and allows to remove the profile from the `o.e.m2e.maven.runtime/pom.xml` that is responsible to generate the .classpath file for the sub-project within the Eclipse+M2E-IDE, in:
- https://github.com/eclipse-m2e/m2e-core/pull/466

On the long run this could probably be implemented faster by performing the BND-processing directly as part of the PDE-Connector's BuildParticipant execution instead of just executing the mojo. This would allow to skip the copying to to the `project.build.outputDirectory` and instead reference the original artifacts and to only write the Manifest.MF. This would also allow to support the `bnd-maven-plugin`'s jar goal nicely, because that only generates the jar, that we would have to extract in order to have a Manifest (and resources available).

However this is for the moment sufficient for m2e's demands to complete https://github.com/eclipse-m2e/m2e-core/pull/466 withouth too many hacks. Once we can use the bnd:jar goal for the maven-runtime (which is as soon as bnd 6.4) is out, the mentioned enhancements can be done.

### DIFF
--- a/org.eclipse.m2e.pde.connector.tests/projects/bnd/pom.xml
+++ b/org.eclipse.m2e.pde.connector.tests/projects/bnd/pom.xml
@@ -4,7 +4,6 @@
 	<artifactId>bnd</artifactId>
 	<version>0.0.1-SNAPSHOT</version>
 	<packaging>jar</packaging>
-
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 	</properties>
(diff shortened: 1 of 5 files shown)


HannesWell:
@laeubi do you want to have a look at this?
But please consider this is likely not yet the ideal solution, but it should work for #466 for now. Once we can use the BND:jar goal this has to be reworked anyway.

PR #1072, opened 2022-11-20
Execute Maven plug-ins to generate sources required for M2E compilation

HannesWell:
The connectors for the 'maven-dependency-plugin' and the 'modello-maven-plugin' are currently broken because they are not yet adapted to M2E 2.0. Threrefore command m2e explicitly to execute those goals.

In order to make this work in the IDE the o.e.m2e.core/pom.xml had to be re-created.

@merks this should fix your problem from https://github.com/eclipse-m2e/m2e-core/discussions/1071.
You can try it by deleting the content of the src-gen folder and performing a workspace clean+full build afterwards.

### DIFF
--- a/org.eclipse.m2e.core/.project
+++ b/org.eclipse.m2e.core/.project
@@ -21,12 +21,18 @@
 			</arguments>
 		</buildCommand>
 		<buildCommand>
+			<name>org.eclipse.m2e.core.maven2Builder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
 			<name>org.eclipse.pde.ds.core.builder</name>
 			<arguments>
 			</arguments>
 		</buildCommand>
 	</buildSpec>
 	<natures>
+		<nature>org.eclipse.m2e.core.maven2Nature</nature>
 		<nature>org.eclipse.pde.PluginNature</nature>
 		<nature>org.eclipse.jdt.core.javanature</nature>
 	</natures>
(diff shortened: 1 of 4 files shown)


fbricon:
FYI m2eclipse-modello compatible with m2e 2.x is available from https://github.com/tesla/m2eclipse-modello/releases/download/latest/
The m2e catalog has been updated

HannesWell:
> FYI m2eclipse-modello compatible with m2e 2.x is available from https://github.com/tesla/m2eclipse-modello/releases/download/latest/ The m2e catalog has been updated

Thanks for that hint. This makes most of of this change obsolete. I will revert tomorrow and update the M2Es Oomph setup accordingly.

laeubi:
For me signing do not mean anything ... everyone can buy a signing certificate for a few dollars and the eclipse one is accessible to every eclipse-committer...

HannesWell:
That's all right. But not everybody knows that and has the same attitude in that regard. and therefore I'm a bit concerned that this could scare/annoy new contributors and maybe even existing ones.

fbricon:
eclipse signing is only available to projects built on the EF infra, which doesn't apply to m2eclipse-modello.
Now I can't setup jar signing for Takari projects, nor do I care enough. Maybe @jvanzyl would be interested to do it.

HannesWell:
> eclipse signing is only available to projects built on the EF infra, which doesn't apply to m2eclipse-modello. Now I can't setup jar signing for Takari projects, nor do I care enough. Maybe @jvanzyl would be interested to do it.

That's right. An alternative would be to use PGP signing.
Maybe that's simpler to set up.

PR #1078, opened 2022-11-22
Add m2e lifecycle metadata for maven-dependency-plugin

### LINKED ISSUE #1077: Add m2e lifecycle metadata for maven-dependency-plugin 3.4.0+
Following the support of the plexus-build-api from https://issues.apache.org/jira/browse/MDEP-674 the m2e lifecycle metadata needs to be added for `maven-dependency-plugin` to https://github.com/eclipse-m2e/m2e-core/blob/master/org.eclipse.m2e.core/lifecycle-mapping-metadata.xml.

Compare with the discussion at https://github.com/apache/maven-dependency-plugin/pull/257#issuecomment-1300330087

kwin:
This closes #1077

### DIFF
--- a/org.eclipse.m2e.core/lifecycle-mapping-metadata.xml
+++ b/org.eclipse.m2e.core/lifecycle-mapping-metadata.xml
@@ -17,6 +17,75 @@
     <pluginExecution>
       <pluginExecutionFilter>
         <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <goals>
+          <goal>copy</goal>
+          <goal>copy-dependencies</goal>
+          <goal>properties</goal>
+          <goal>unpack</goal>
+          <goal>unpack-dependencies</goal>
+        </goals>
+        <versionRange>[3.4,)</versionRange>
+      </pluginExecutionFilter>
+      <action>
+        <execute>
+          <runOnIncremental>true</runOnIncremental>
+        </execute>
+      </action>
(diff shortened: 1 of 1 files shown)


mickaelistria (approved):

HannesWell:
Thanks @kwin for this update. Your dedication to spread the BuildContextAPI is really helpful!

Do you plan to have a release of the maven-dependency-plugin soon?
Because otherwise users get warnings telling them they should update the dependency-plugin but that update is not available.
So if the 3.4 release will take some more time I suggest to at least remove the warning until then.

kwin:
The release should happen soon: https://lists.apache.org/thread/xo0m4dztj0v67rdv4n843ls3dxny66xg

HannesWell:
Great. 👍🏽

laeubi:
I'm not sure if an error is appropriate here, why not a warning instead? I assume even with a release there will be some projects that do not update immediately and some older projects that probably never update so all will get an error marker now the user will not really be able to act upon.

kwin:
The behaviour should be the same for all Maven plugins, and the existing ones emit an error already (https://github.com/kwin/m2e-core/blob/dbfe0c1fc2538694edc66a61fa404073b5cffd3b/org.eclipse.m2e.core/lifecycle-mapping-metadata.xml#L116). But I agree that most probably all should be changed to a warning.

laeubi:
I can't think of a use-case where error is anything useful, why should I error the user about my mojo in an IDE? So probably the "error" can just be treated as a warning marker...

PR #1086, opened 2022-11-25
Restore Maven Runtime compatibility with Java 8

kwin:
Add animal-sniffer to check for wrong API usage during build

### DIFF
--- a/m2e-maven-runtime/org.eclipse.m2e.maven.runtime/src/main/java/org/eclipse/m2e/internal/maven/listener/M2EMavenBuildDataBridge.java
+++ b/m2e-maven-runtime/org.eclipse.m2e.maven.runtime/src/main/java/org/eclipse/m2e/internal/maven/listener/M2EMavenBuildDataBridge.java
@@ -17,10 +17,12 @@ import java.io.IOException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
+import java.nio.Buffer;
 import java.nio.ByteBuffer;
 import java.nio.channels.ServerSocketChannel;
 import java.nio.channels.SocketChannel;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
@@ -144,9 +146,9 @@ public class M2EMavenBuildDataBridge implements EventSpy {
 			this.groupId = Objects.requireNonNull(data.get("groupId"));
 			this.artifactId = Objects.requireNonNull(data.get("artifactId"));
 			this.version = Objects.requireNonNull(data.get("version"));
-			this.projectBasedir = Path.of(data.get("basedir"));
-			this.projectFile = Path.of(data.get("file"));
-			this.projectBuildDirectory = Path.of(data.get("build.directory"));
+			this.projectBasedir = Paths.get(data.get("basedir"));
+			this.projectFile = Paths.get(data.get("file"));
+			this.projectBuildDirectory = Paths.get(data.get("build.directory"));
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


mickaelistria:
What about targeting compatibility with Java 17 as minimal requirement like it's done in most other places of m2e?

laeubi:
> What about targeting compatibility with Java 17 as minimal requirement like it's done in most other places of m2e?

I think this is uesd as "embedded maven" also for executing external processes where we want allow the user to user other jvms

m2e-maven-runtime/org.eclipse.m2e.maven.runtime/src/main/java/org/eclipse/m2e/internal/maven/listener/M2EMavenBuildDataBridge.java
```diff
@@ -144,9 +146,9 @@ public static final class MavenProjectBuildData {
 			this.groupId = Objects.requireNonNull(data.get("groupId"));
 			this.artifactId = Objects.requireNonNull(data.get("artifactId"));
 			this.version = Objects.requireNonNull(data.get("version"));
-			this.projectBasedir = Path.of(data.get("basedir"));
-			this.projectFile = Path.of(data.get("file"));
-			this.projectBuildDirectory = Path.of(data.get("build.directory"));
+			this.projectBasedir = Paths.get(data.get("basedir"));
```
HannesWell, line 149: Don't know why the CI build did not catch this, but the service version of this Plugin should be incremented by +1.

HannesWell (commented):
Thanks a lot for this fix @kwin!

> > What about targeting compatibility with Java 17 as minimal requirement like it's done in most other places of m2e?
> 
> I think this is uesd as "embedded maven" also for executing external processes where we want allow the user to user other jvms

Exactly.

I'll try to investigate the test failure this evening (or tomorrow).
For now I have only one remark.

m2e-maven-runtime/org.eclipse.m2e.maven.runtime/src/main/java/org/eclipse/m2e/internal/maven/listener/M2EMavenBuildDataBridge.java
kwin: Done in https://github.com/eclipse-m2e/m2e-core/pull/1086/commits/9d7df73a0f124a25669c9e13ef35ca4c07132dee.

m2e-maven-runtime/org.eclipse.m2e.maven.runtime/src/main/java/org/eclipse/m2e/internal/maven/listener/M2EMavenBuildDataBridge.java
```diff
@@ -180,7 +182,7 @@ public static MavenBuildConnection prepareConnection(String label, Consumer<Mave
 			try (ServerSocketChannel s = server; SocketChannel readChannel = server.accept()) {
 				ByteBuffer buffer = ByteBuffer.allocate(512);
 				StringBuilder message = new StringBuilder();
-				while (readChannel.read(buffer.clear()) >= 0) {
+				while (readChannel.read(buffer) >= 0) {
```
HannesWell, line 185: The buffer had to be cleared at the end of the loop, not only ones before (the initial clear is actually not needed, but was there because it fit in the control flow).

org.eclipse.m2e.feature/feature.xml
```diff
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.m2e.feature"
       label="%featureName"
-      version="2.0.6.qualifier"
+      version="2.1.1.qualifier"
```
HannesWell, line 5: Aligned the version with the m2e.sdk. Actually that could have be done earlier.

HannesWell (approved):
Found the reason for the failing test.
To speed things up a bit and to no bother you with more version bumbs, I pushed it directly onto your branch.
In my first attempt to push I forgot the prefix and therefore accidentally created another branch (I'm surprised GH allows this).

Thanks again for this fix!

PR #1100, opened 2022-11-29
ArchetypeGenerator javadoc improvements

kjsmita6:
Discussion in #1096 

Clarify the javadoc that the created projects are not imported automatically as that functionality was removed. This needs to be moved to the correct file once #922 is merged.


HannesWell:
@laeubi do you want to have a look at this?

@kjsmita6 you have to increment the version of the `o.e.m2e.core.ui` plugin by one. At the moment also the o.e.m2e.feature and o.e.m2e.sdk would have to be bumped as well, but there are a few other PRs pending that included that change too.

laeubi (approved):

Issue #1097, opened 2022-11-29
The 2.1.1 release repository contains unsigned source features


merks:
The 2.1.1 release repository:

https://download.eclipse.org/technology/m2e/releases/2.1.1

contains unsigned source features:

https://download.eclipse.org/oomph/archive/reports-extra/m2e-releases-2.1.1/download.eclipse.org/technology/m2e/releases/2.1.1/index.html

HannesWell:
Sigh. :/ Looks like I have to run the third release within five days.
Nevertheless thanks for making us aware of this problem.
Just created #1102 to address the issue.

Is it sufficient to have a fix today (Wednesday) evening (UTC+1)?

merks:
@HannesWell Yes, that will be just fine.

HannesWell:
The latest report does not list m2e under unsigned artifacts, so I think this is resolved successfully:

https://download.eclipse.org/staging/2022-12/buildInfo/archive/download.eclipse.org/staging/2022-12/

merks:
Thanks!

PR #1102, opened 2022-11-30
Sign IUs in a later build phase

HannesWell:
Fixes https://github.com/eclipse-m2e/m2e-core/issues/1097

The (prepare-) package phases are unfortunately really full already and IIRC the `feature-source` goal is sensitive in regard to its execution in relation to the `p2-metadata` goal.
An alternative could be to bind the `feature-source` goal to the `prepare-package` phase.
However the important part is that the eclipse-jar-signer runs after the sources jar has been created.

### DIFF
--- a/m2e-parent/pom.xml
+++ b/m2e-parent/pom.xml
@@ -117,6 +117,7 @@
 								<goals>
 									<goal>sign</goal>
 								</goals>
+								<phase>pre-integration-test</phase>
 							</execution>
 						</executions>
 					</plugin>
--- a/org.eclipse.m2e.lemminx.feature/feature.xml
+++ b/org.eclipse.m2e.lemminx.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.m2e.lemminx.feature"
       label="%featureName"
-      version="2.0.1.qualifier"
+      version="2.0.2.qualifier"
       provider-name="%providerName"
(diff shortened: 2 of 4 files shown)


HannesWell:
I also noticed that the commitId and build-qualifier computation of the maven-runtime components was flawed and always the last commit was used instead of the last commit touching the corresponding project.
I wonder why the `tycho-p2-extras-plugin:compare-version-with-baselines` mojo did not catch that the qualifier was increased withouth a service version bump?

laeubi:
> An alternative could be to bind the `feature-source` goal to the `prepare-package` phase.

The `feature-source` is a bit special as it depends on data from the feature-packaging phase and you have to regenerate the P2 metadata

HannesWell:
> I also noticed that the commitId and build-qualifier computation of the maven-runtime components was flawed and always the last commit was used instead of the last commit touching the corresponding project.

Extracted that part into
- https://github.com/eclipse-m2e/m2e-core/pull/1104

> > An alternative could be to bind the `feature-source` goal to the `prepare-package` phase.
> 
> The `feature-source` is a bit special as it depends on data from the feature-packaging phase and you have to regenerate the P2 metadata

Yes I noticed that as well. Therefore instead I went with the first approach to move the `eclipse-jarsigner-plugin:sign` goal to the `pre-integration-test`. It's not nice, but it seems to work, but I will check that.

PR #1106, opened 2022-11-30
Enforce m2e.feature qualifier update to align to maven-runtime component

HannesWell:
Because the maven-runtime components currently use the committers time-zone for their qualifier and don't align with the other projects that use UTC.


HannesWell:
This is not ideal, but for the moment I don't know a better solution.
In general we should use Tycho's [build-qualifier](https://tycho.eclipseprojects.io/doc/3.0.0/tycho-packaging-plugin/build-qualifier-mojo.html) mojo, but that currently does not work with jar projects.

PR #1107, opened 2022-12-01
Fix build path

Bananeweizen:
On first checkout the project has build path errors because the generated source path folder doesn't exist. Therefore please always check in all generated source folders with a gitignore, and then exclude the gitignore in the classpath settings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 target/
 bin/
-src-gen/
 .DS_Store
 .polyglot*
 .tycho-consumer-pom.xml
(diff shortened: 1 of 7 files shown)


HannesWell:
In general the change is reasonable.
You only have to bump the version of the `org.eclipse.m2e.core` plugin as well as of the `org.eclipse.m2e.feature` and `org.eclipse.m2e.sdk` feature.

Furthermore you have to exclude the `.gitignore` in the `build.properties` otherwise it is probably included into the build jars. Or is the exclusion in the .classpath considered by Tycho in this regard?

Bananeweizen:
> Furthermore you have to exclude the `.gitignore` in the `build.properties` otherwise it is probably included into the build jars. Or is the exclusion in the .classpath considered by Tycho in this regard?

Neither. Tycho has default excludes for gitignore files.

HannesWell:
> > Furthermore you have to exclude the `.gitignore` in the `build.properties` otherwise it is probably included into the build jars. Or is the exclusion in the .classpath considered by Tycho in this regard?
> 
> Neither. Tycho has default excludes for gitignore files.

Didn't know that. Thanks for that info.

HannesWell:
Thanks!

PR #1116, opened 2022-12-03
Enable the Tycho Baselineplugin for m2e

laeubi:
This enables the new [tycho-baseline-plugin](https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#new-tycho-baseline-plugin) for m2e to see how/if this works out and where Tycho must improve and test it on a real world example.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -36,7 +36,7 @@ jobs:
     - name: Build m2e-core
       uses: GabrielBB/xvfb-action@v1
       with:
-       run: mvn clean verify -B -V -Pits -Dmaven.test.error.ignore=true -Dmaven.test.failure.ignore=true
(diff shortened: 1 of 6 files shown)


laeubi:
Revealed a bug in Tycho 4.0
- https://github.com/eclipse-tycho/tycho/pull/1792

HannesWell:
Looks good 😃

But I think it would be good to split this into two commits, one to update to Tycho 4.0.0-SNAPSHOT+build adjustments and the second one to introduce the baseline repo. Just to have concerns a bit separated.

> Revealed a bug in Tycho 4.0
> 
>     * [InstallableUnitGenerator must check for null manifest eclipse-tycho/tycho#1792](https://github.com/eclipse-tycho/tycho/pull/1792)

I wonder which jar in the M2E build does not have a Manifest?

laeubi:
@HannesWell any idea why the build fails with version complaint?

laeubi:
I assume the test-failure is not really related to the maven build so we can merge this, all Tycho changes should be available now!

PR #1119, opened 2022-12-04
Prevent infinite recursion when cleaning projects

laeubi:


HannesWell (commented):
Looks reasonable.
But but would be good to have a test case for this. Also to verify that this fixes the issue.

laeubi:
> Looks reasonable. But but would be good to have a test case for this. Also to verify that this fixes the issue.

I would leave it up to you, at least that change should not harm anything... so feel free to merge this whenever you think is suitable.

laeubi:
as there is no further feedback I think this can be merged and we will see if it improves anything...

PR #1124, opened 2022-12-05
Use tycho-packaging-plugin:build-qualifier to compute build-qualifiers

HannesWell:
and build with Tycho 4.0.0-SNAPSHOT.

### DIFF
--- a/m2e-maven-runtime/pom.xml
+++ b/m2e-maven-runtime/pom.xml
@@ -152,6 +152,22 @@
 				</executions>
 			</plugin>
 			<plugin>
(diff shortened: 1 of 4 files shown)


HannesWell:
Since with this change the computed qualifier of the maven-runtime bundles is now again unified and UTC-based it is decreased by one hour from `20221130-2149` to `20221130-2049`.
The compare-version-with-baselines somehow does not complain about this, but as far as I can tell, this is not ideal, but also should not harm. I don't want to have another service-version bump which makes users download the jars again that actually did not change. And since there is Maven-3.9 probably coming soon, a new version will be necessary anyway.

PR #1125, opened 2022-12-05
Build with Tycho 4.0.0-SNAPSHOT

HannesWell:
The build-extension remains at Tycho 3.0.0

### DIFF
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,0 +1,1 @@
+-Dtycho.resolver.classic=false
--- a/m2e-parent/pom.xml
+++ b/m2e-parent/pom.xml
@@ -24,7 +24,7 @@
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<tycho-version>3.0.0</tycho-version>
+		<tycho-version>4.0.0-SNAPSHOT</tycho-version>
 		<tycho.p2.baselineMode>warn</tycho.p2.baselineMode>
 		<scmConnectionRaw>scm:git:https://github.com/eclipse-m2e/m2e-core.git</scmConnectionRaw>
 	</properties>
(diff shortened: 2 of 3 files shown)


HannesWell:
@laeubi any idea, why Tycho 4.0.0-SNAPSHOT fails while Tycho 3.0.0 succeeds for this mixed reactor?

laeubi:
See
- https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#mixed-reactor-setups-require-the-new-resolver-now
- https://github.com/eclipse-m2e/m2e-core/pull/1116/files#diff-9b8baedf521e6c7f80ad647b88afe601b947eca1f7a331b3e03e9b25184029ff

HannesWell:
> See
> 
>     * https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#mixed-reactor-setups-require-the-new-resolver-now
> 
>     * https://github.com/eclipse-m2e/m2e-core/pull/1116/files#diff-9b8baedf521e6c7f80ad647b88afe601b947eca1f7a331b3e03e9b25184029ff

My mistake. Thanks for the hint!

Do you mind if we merge this before https://github.com/eclipse-m2e/m2e-core/pull/1116 to have the Tycho update in a distinct commit?

This also removes the `<pomDependencies>consider</pomDependencies>` element from the tp-configuration plugin.
IIRC you also intended to make the specification of non `eclipse-plugin` typed projects as dependencies in the pom obsolete?
When I removed the dependencies to `m2e.maven.runtime` respectively `m2e.archetype.common` from the root `pom.xml`, the TP resolution failed in my local build, so I assume that is not yet working?

laeubi:
If you think Tycho 4.0 is okay for the build why not,

- https://github.com/eclipse-m2e/m2e-core/pull/1116

is likely to take a bit longer and is currently a bit experimental, so I won't wait for this if you want Tycho 4 :-)

HannesWell:
> If you think Tycho 4.0 is okay for the build why not,
> 
I would like to have https://github.com/eclipse-m2e/m2e-core/pull/1124 soon and also the result of https://github.com/eclipse-tycho/tycho/issues/1794 (in which ever form this is solved). The main goal is to get rid of the `antrun-plugin` (I have the impression that this makes m2e 'nervous' since it is generating the ant-xml and causing repetitive automatic builds).
And that requires Tycho-4.0. So if you think with the 4.0-SNAPSHOT the M2E build is not broken too often I'm fine with it. 👍🏽

PR #1130, opened 2022-12-08
[Build] Temporarily ignore baseline discrepancy for maven-runtime and bump qualifiers due to changed class-files

HannesWell:
Due to changes in Tycho the metadata of the maven-runtime components are now different when compared on textual level, but for consumers that change is not relevant and therefore the difference can be ignored once.


HannesWell:
@laeubi the build should now succeed again.

PR #1131, opened 2022-12-09
Accept version constraints without major version passed to m-enforcer-p

### LINKED ISSUE #1120: Proper Selection of Execution Environment for Java
This ticket is an effort to bring together many other tickets that have existed since m2e 2.0 release which has negatively impacted java determination within Eclipse.  Example simple projects used here to test out the process.

Related Tickets
- #1099
- #1050
- #992
- #842
- and probably others

Versions used
- Eclipse 2022-09 (4.25.0)
- m2e 2.1.3.20221203-1403
- No execution environments directly set before JavaSE-1.8
- No execution environment directly set for no longer supported versions: 9, 10, 12, 13, 14, 15, 16, 18
- Not using the beta jdk 19 support on this copy
- Execution environment JavaSE-1.8 set to jdk 1.8.0.351
- Execution environment JavaSE-11 set to jdk 11.0.17
- Execution environment JavaSE-17 set to jdk 17.0.5
- Installed jdks: 1.8.0.351, 11.0.17, 17.0.5, 19.0.1, 20ea (from end of october)
- All plugins are up to date in samples used with some minor variation on plugin updates in last 3 months or so.

Test project to be used (small and simple)
- https://github.com/mybatis/caffeine-cache with pom set to java 11 throughout when using its parent pom
-  or https://github.com/mybatis/ehcache-cache with pom left to defaults of the parent pom (compiler at 8, but 11 required due to plugins, tests)
- Project uses super pom project https://github.com/mybatis/parent
- Use 'master' branch for each as up to date within days.

```
Settings involved with relative usage

    <!-- Maven compiler options -->
    <java.version>8</java.version>
    <java.release.version>8</java.release.version>
    <java.test.version>11</java.test.version>
    <java.test.release.version>11</java.test.release.version>

    <!-- source/target are deprecated, remove in future release -->
    <maven.compiler.source>${java.version}</maven.compiler.source>
    <maven.compiler.target>${java.version}</maven.compiler.target>
    <maven.compiler.testSource>${java.test.version}</maven.compiler.testSource>
    <maven.compiler.testTarget>${java.test.version}</maven.compiler.testTarget>

    <maven.compiler.release>${java.release.version}</maven.compiler.release>
    <maven.compiler.testRelease>${java.test.release.version}</maven.compiler.testRelease>

Enforcer plugin is used along with extra enforcer rules to do the following.
- Force byte code compliance based on the 'java.version' argument - That is to say 8 for ehcache-cache sample and 11 for caffeine-cache sample.
- Require java version is set to a range of 11, 17, 18, 19, and 20.  Note the parent pom here was released on Sept 5 and thus 18 was viable then.

A profile is auto activated for m2e when m2e.version is present thus in Eclipse workspace.  This is done to add lifecycle mappings plugins lack or adjustments from the default meta data as well as telling eclipse to use the split 'test' variation of compiler legacy args (so older eclipses and m2e are supported and not sure m2e moved to use release yet).  This logic using the test side split was to overcome fact that older m2e (and possibly latest) did not support it and would not choose the right one in 1.x.  In both examples, this forces to 11.

    <profile>
      <id>eclipse</id>
      <activation>
        <property>
          <name>m2e.version</name>
        </property>
      </activation>

....all the lifecycle stuff followed by compiler work around.

            <!-- Force Eclipse to use test side in order to ensure no compilation errors (generally these are same and make no difference) -->
            <plugin>
              <configuration>
                <source>${maven.compiler.testSource}</source>
                <target>${maven.compiler.testTarget}</target>
              </configuration>
            </plugin>
```

The super pom is used across 20 or so underlying projects in separate repos.  Mybatis current policy is that we support java 8 runtimes unless override higher on downstream projects.  The default is telling the compiler to be java 8.  Its using both the legacy source/target and newer release args split for source vs test.  Legacy ones used as some plugins still require it.  M2e must be concerned of highest java version to use that will correctly work to build.  This means across all projects in mybatis it would be 11 because both tests set to that level and many plugins require that level which enforcer further states.  The sample caffine-cache project aligns that directly to 11 only overall.

Test import requirements
- No prior .settings folder
- No .project file
- No .classpath file
- Import as a maven project using the pom

Current m2e import results
- Setting JRE System Library to JRE-1.1

m2e should not consider execution environments that have not be setup which would fix the JRE-1.1 issue.  If intent is to always use a execution environment, then alert a warning or error on import that no match is set and direct user to set them up first.

m2e should not longer consider these.  In fact, Eclipse really should remove them.  While I could potentially still see java 1.5 being used in extreme cases, its implausible that others are and really if those old ones are used, I don't think user should be using any modern IDE.

- CDC-1.0/Foundation-1.0
- OSGi/Minimum-1.0
- CDC-1.1/Foundation-1.1
- JRE-1.1
- OSGi/Minimum-1.1
- J2SE-1.2
- OSGi/Minimum-1.2
- J2SE-1.3
- J2SE-1.4
- J2SE-1.5

In the samples provided, expectation is that the following is final resolution
- caffeine-cache -> JavaSE-11
- ehcache-cache -> JavaSE-11

Rules
- Enforcer plugin is just a hint on what java versions are acceptable to use from a potential list.  This addressed the dreaded java class error that is not as nice as telling user to use correct java version.  That could be just 1 or more listed versions that are acceptable.  The lowest version listed is the version that 'must' be used if the project fails to reach that level with compiler setup.  
- Note that source and test can be split.  The higher required java version must be used to set execution environment after reviewing the rules of enforcer if present.
- Thus, if enforcer says its ok for 11, 17, 18, 19, or 20 in samples provided that is the hint list on what is acceptable.  The actual determination then comes after looking at the arguments set on final resolved compiler setup.  Those can be split although seen here it was mitigated by accepting long term that m2e didn't handle such a split.  But lets say it does now, it needs to determine what is the higher requirement for the build to function between source and test needs.  In the examples given, that version is 11 in both cases.  With that version, it should then look in that enforcer list if provided and check to see that 11 is there.  Lets say 11 is not there, then it should fail with error that enforcer requires version of list X,X,X and build expects X.  That way user can fix the issue.  Further, once it knows it needs to use 11 for example, that better be set as an execution environment.  If it is not, it should error telling user to setup execution environment for 11 so its not guessing as I suspect that is what its doing now since I got JRE-1.1.

kwin:
This closes #1120

### DIFF
--- a/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
+++ b/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
@@ -907,6 +907,10 @@ public abstract class AbstractJavaProjectConfigurator extends AbstractProjectCon
   }
 
   private static ArtifactVersion getMajorMinorOnlyVersion(ArtifactVersion lower) {
+    if(lower.getMajorVersion() == 0) {
+      // not every version constraint has a major version, in case it doesn't just rely on compareTo semantics as is
+      return lower;
+    }
     return new DefaultArtifactVersion(lower.getMajorVersion() + "." + lower.getMinorVersion());
   }
 


kwin:
Superseded by #1137

PR #1133, opened 2022-12-10
[Build] Use Eclipse ECJ Java Compiler to build M2E

HannesWell:
Lets see if this changes the results.

@laeubi do you know if more has to be done to tell the tycho-compiler-plugin to use ecj? [compilerId](https://tycho.eclipseprojects.io/doc/master/tycho-compiler-plugin/compile-mojo.html#compilerId)?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -201,6 +201,13 @@
 					<groupId>org.eclipse.tycho</groupId>
 					<artifactId>tycho-compiler-plugin</artifactId>
 					<version>${tycho-version}</version>
+					<dependencies>
+						<dependency>
+							<groupId>org.eclipse.jdt</groupId>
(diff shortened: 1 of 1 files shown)


HannesWell:
Or does this even make a difference?

laeubi:
`tycho-compiler-plugin` uses ecj by default. but you can configure `maven-compiler-plugin` to use ecj instead of javac... as far as I know.

HannesWell:
> `tycho-compiler-plugin` uses ecj by default. but you can configure `maven-compiler-plugin` to use ecj instead of javac... as far as I know.

Thanks for that info. So this change is actually useless.
Yes I just noticed that for my SWT changes 😄

Issue #497, opened 2022-01-16
Incorrect JDK version activation when running project within Eclipse/m2e


harbulot:
It seems that the JDK version number used to activate a profile doesn't match the version of the project execution environment / JDK version.

* Eclipse Version: 2021-12 (4.22.0) (running on JDK 17).
* M2E Maven Integration for Eclipse Core 1.19.0.20211116-0804

Sample `pom.xml`:

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>
    <groupId>test003</groupId>
    <artifactId>test003</artifactId>
    <version>0.0.1-SNAPSHOT</version>
    <build>
        <plugins>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-compiler-plugin</artifactId>
                <version>3.9.0</version>
                <configuration>
                    <source>1.8</source>
                    <target>1.8</target>
                    <testSource>1.8</testSource>
                    <testTarget>1.8</testTarget>
                </configuration>
            </plugin>
        </plugins>
    </build>

    <profiles>
        <profile>
            <id>jdk-15-and-above</id>
            <activation>
                <jdk>[15,)</jdk>
            </activation>
            <dependencies>
                <!-- JavaScript -->
                <dependency>
                    <groupId>org.openjdk.nashorn</groupId>
                    <artifactId>nashorn-core</artifactId>
                    <version>15.3</version>
                </dependency>
            </dependencies>
        </profile>
    </profiles>
</project>
```

Sample class:

```java
package test003;

import javax.script.ScriptEngine;
import javax.script.ScriptEngineManager;
import javax.script.ScriptException;

public class MyClass {
    public MyClass() throws ScriptException {
        ScriptEngine scriptEngine = new ScriptEngineManager().getEngineByName("nashorn");
        Object v = scriptEngine.eval("1;");
        System.out.println("Result: " + v);
    }

    public static void main(String[] args) throws Exception {
        new MyClass();
    }
}
```

Output:

[...]

PR #1137, opened 2022-12-12
Configure Maven JRE separately

### LINKED ISSUE #1120: Proper Selection of Execution Environment for Java

### LINKED ISSUE #1134: No "project JRE" option in JRE tab for Maven Launches
For regular Java Launches (e.g. for Java Application) the JRE tab exposes the project JRE as first option (whenever the launch is associated with a Java Project)
Screenshot 2022-12-10 at 21 22 02

The Maven Launches only expose the Workspace Default as the default option (and don't expose the project JRE as option at all). This is surprising as the project JRE is used for all default configs (compare with #1099).

Instead for all Maven Launches the project JRE should be exposed as first option as well (instead of the Workspace default).

kwin:
The value is automatically derived from maven-enforcer rule requireJava

This closes #1134
This closes #1120


kwin:
@mickaelistria @laeubi @HannesWell Can you quickly comment, whether that extension makes sense from your PoV. Then I would adjust the test cases accordingly, although this probably requires some more refactoring as the tested Eclipse instance only has one installed JRE....

laeubi:
@kwin can you enhance the PR description a bit more with:

1. How does it solve the referenced issues
2. maybe a screenshot that shows previous / new state
3. add an entry to the [release notes](https://github.com/eclipse-m2e/m2e-core/blob/master/RELEASE_NOTES.md)

kwin:
Yes, but only after we have consensus that this is the way forward. Here are some screenshots of the PR:
Screenshot 2022-12-14 at 13 53 32
Screenshot 2022-12-14 at 13 54 51

kwin:
> Can you explain what a "Maven JRE" is? The one to run maven commandline build in contrast to the "JRE System Library"?

Yes, exactly. Maybe we should relabel as "Maven Execution JRE"

> If yes, maybe this should have two levels:

There is already two levels:
a) on project level (by default equal to the Java Build JRE if not stated otherwise by the maven-enforcer-plugin)
b) on launch config level (allows to override a))

I don't think we need a third level here TBH. Also not sure about the minimum enforced by Maven itself, this is currently not enforced at all, but we could improve later on to also consider this.

laeubi:
> Yes, exactly. Maybe we should relabel as "Maven Execution JRE"

Do you think it would be possible to show the one currently active on the left like it is done for Java Projects?

> I don't think we need a third level here

Alright then start with this one.

kwin:
> Do you think it would be possible to show the one currently active on the left like it is done for Java Projects?

Where? Please share some screenshot because I don't know what you are referring to.

laeubi:
grafik

kwin:
I don't think the Maven Execution JRE should be listed in the project explorer as it has zero impact on the (compile or test) classpath!

laeubi:
If it has no impact why do we configure it fro the project? I don't think this is reserved to "classpath" items and as it is a persistent property of the project it seems important enough.

kwin:
> If it has no impact why do we configure it fro the project?

I didn't say that! 

> I don't think this is reserved to "classpath" items

For me it is, it makes no sense to list anything else being configured on the project properties there.

laeubi:
Yes please go on...

HannesWell (commented):
In general this change looks good and I agree that it is the right direction to let the enforcer plugin only influence the JRE to run a Maven-build process and the adjust the Maven launch config to use that EE by default, but not to consider the enforcer-p when setting the JRE_CONTAINER to build.

But I'm not so happy to handle all of that in m2e.core and therefore introducing a dependency to jdt for m2e.core.
Wouldn't it be possible to handle the selection of the JRE to launch completely in the m2e.launching plugin? Letting m2e.core.ui depend on m2e.launching is probably less problematic.

To test it, wouldn't it be sufficient to check the selected EE and to test the VMInstall algorithm in a unit test.

org.eclipse.m2e.core/META-INF/MANIFEST.MF
```diff
@@ -53,12 +53,15 @@ MavenArtifact-ArtifactId: org.eclipse.m2e.core
...
  javax.inject;version="1.0.0",
+ org.eclipse.jdt.core,
+ org.eclipse.jdt.launching,
  org.slf4j;version="[1.7.0,3.0.0)",
  org.slf4j.spi;version="[1.7.31,3.0.0)",
  com.google.gson;version="[2.9.1,3.0.0)",
  org.apache.commons.codec.digest;version="[1.14.0,2.0.0)"
 Automatic-Module-Name: org.eclipse.m2e.core
 Service-Component: OSGI-INF/org.eclipse.m2e.core.embedder.MavenModelManager.xml,
+ OSGI-INF/org.eclipse.m2e.core.internal.*.xml,
```
HannesWell, line 64: Please don't use wild-cards here and just let PDE generate the explicit value.

kwin: Not possible due to https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#d0e30931. The service from the test fragment needs to be loaded.

kwin:
> But I'm not so happy to handle all of that in m2e.core and therefore introducing a dependency to jdt for m2e.core. Wouldn't it be possible to handle the selection of the JRE to launch completely in the m2e.launching plugin? Letting m2e.core.ui depend on m2e.launching is probably less problematic.

JDT Launching is necessary when updating projects to calculate the default Maven Execution JRE (and that happens in core)

> To test it, wouldn't it be sufficient to check the selected EE and to test the VMInstall algorithm in a unit test.

The Maven Execution JRE is not based on EEs but on concrete JREs.

RELEASE_NOTES.md
```diff
@@ -1,5 +1,19 @@
...
 
+## Next version
+
+### Configuration of Maven Execution JRE
+
+In the past the project's build JRE was also used by default to execute Maven itself. This is now separately configurable in `Project -> Properties -> Maven`:
+
+![Maven Project Properties](https://user-images.githubusercontent.com/185025/208966384-816570f8-8c02-4578-bd67-588f7d8e2c27.png)
+
+The default value is automatically set according to the configuration of the `maven-enforcer-plugin` rule [`requireJavaVersion`](https://maven.apache.org/enforcer/enforcer-rules/requireJavaVersion.html) when creating or updating the Maven configuration. This value is no longer considered for configuring the project's build JRE.
```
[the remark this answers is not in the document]
kwin: Technically the Maven Execution JRE falls back to project build JRE or workspace JRE when no enforcer rule is found!

laeubi:
I think we already have had the demand to inspect JRE here:
- https://github.com/eclipse-m2e/m2e-core/pull/847

I therefore thing it would be good to have some kind of `JavaToolchainManager` interface that abstracts this part and the JDT bundle implements the service.

kwin:
Then we can not use Interfaces like IVMInstall in that service interface but must define our own one…

laeubi:
> Then we can not use Interfaces like IVMInstall in that service interface but must define our own one…

because this is "owned" by JDT and has much more requirements than we need here. So we should decouple m2e.core from jdt.

laeubi:
Beside that I think we can implement most of required stuff in the interface itself, e.g. we can have a method:

`setBestMatchingJre(IProject, String);`

HannesWell:
> > But I'm not so happy to handle all of that in m2e.core and therefore introducing a dependency to jdt for m2e.core. Wouldn't it be possible to handle the selection of the JRE to launch completely in the m2e.launching plugin? Letting m2e.core.ui depend on m2e.launching is probably less problematic.
> 
> JDT Launching is necessary when updating projects to calculate the default Maven Execution JRE (and that happens in core)
> 
But why does this have to happen in core? The Maven Execution JRE is eventually just used to select a JRE to run a Maven build?
It is even just the default, based on the compiler- and if present enforcer-plugin configuration, if no JRE/EE was chosen explicitly.
To me that does not sound like that information is required in m2e.core. And therefore I think that is sufficient to have the selection logic in m2e.launching?

> > To test it, wouldn't it be sufficient to check the selected EE and to test the VMInstall algorithm in a unit test.
> 
> The Maven Execution JRE is not based on EEs but on concrete JREs.

Why is it based on JREs/VMInstalls and not EEs anymore? Previously in `AbstractJavaProjectConfigurator` it was based on EEs. Wouldn't it be better to keep that and use the JRE that is mapped to that version in the workspace?
Or the other ways round, what is the advantage of operating on JRE's directly?

Furthermore I want to challenge the need for a separate preference. All the preference is doing is saving the default Maven Execution JRE of a project, isn't it?
But it is already possible to specify another JRE or EE in a Maven launch config, if the value derived from the compiler- and enforcer-plugin config is not suitable. For me the launch config seems to be the right place to save such choice. And since you can also save a launch-config in a project and put it under version control I see only a little benefit in introducing the preference.
It would only be beneficial in cases where I want to use another JRE than derived from compiler+enforcer (but why is the enforcer-rule not sufficient?) and want to create new launch-configs that I don't want to save in the project?
In general if a project is so special that you need to use a specific JRE then the chances are high that you want to configure other aspects of the build as well and it is therefore reasonable to save the launch config into git.

Therefore my suggestion is to drop the extra preference and just use compiler+enforcer as default. If it later turns out that there is a great demand for such preference we can still add it. :)

kwin:
> But why does this have to happen in core?

Somewhere you need to persist the information which JRE to use by default for executing Maven. Currently other bundles don't configure anything about the project. Calculating the value on demand whenever you launch Maven is IMHO too much overhead.

> Why is it based on JREs/VMInstalls and not EEs anymore?

Because very often the requireJavaVersion rule requires specific minor versions which cannot be expressed with EE.

> But it is already possible to specify another JRE or EE in a Maven launch config, if the value derived from the compiler- and enforcer-plugin config is not suitable. 

I often use predefined Maven Launches (e.g. for install) which always take the default JRE with no possibility to overwrite.

> Therefore my suggestion is to drop the extra preference and just use compiler+enforcer as default. If it later turns out that there is a great demand for such preference we can still add it. :)

That is all already implemented in the PR, the maintenance effort for that code is IMHO pretty low and there are some use cases for it, so why do you want to throw away that code?

HannesWell:
> > But why does this have to happen in core?
> 
> Somewhere you need to persist the information which JRE to use by default for executing Maven. Currently other bundles don't configure anything about the project. Calculating the value on demand whenever you launch Maven is IMHO too much overhead.

Do you have an estimate how long the calulation runs? My guess would be maybe a few milliseconds, if not less. That would not be a noticeable delay in a launch, that takes a short moment any way. But to be sure you could at least make some basic measurements using `System.currentMillis()` to get an idea about the magnitude.

On the other hand if it is a so long running operation that it causes a noticeable delay in a Maven build launch, say half a second, then it should definitely not be done for each project in the workspace whenever it is configured. Because then you multiply that runtime by the number of projects in the workspace (which can be a range from 1 to multiple hundreds , maybe even a thousand? I'm not sure about the upper limit). And most of the projects probably don't serve as base for a launch, so that time is wasted.

So one way or the other I think that is better done in m2e.launching.

> 
> > Why is it based on JREs/VMInstalls and not EEs anymore?
> 
> Because very often the requireJavaVersion rule requires specific minor versions which cannot be expressed with EE.

Acknowledged. Makes sense.

> 
> > But it is already possible to specify another JRE or EE in a Maven launch config, if the value derived from the compiler- and enforcer-plugin config is not suitable.
> 
> I often use predefined Maven Launches (e.g. for install) which always take the default JRE with no possibility to overwrite.

Understand. Yes that's right. But is the default that now derived from the compiler- and enforcer-plugin not sufficient? I can only think of if you not only need a JRE of a certain version but a very specific one, e.g. the GraalVM or some preview builds of the OpenJDK or alike. But do you have that case or know that this is a real use case for others?
Because that would mean that you are using a maven-plugin in your build that requires that. If it is just about compiling projects, then Maven-toolchains are a more reliable way.

> 
> > Therefore my suggestion is to drop the extra preference and just use compiler+enforcer as default. If it later turns out that there is a great demand for such preference we can still add it. :)
> 
> That is all already implemented in the PR, the maintenance effort for that code is IMHO pretty low and there are some use cases for it, so why do you want to throw away that code?

I don't want to throw away that code and I understand that it can be frustrating if others are slowing you down. But in general just because some code does not need to be touched doesn't mean it does not cause any effort. More code means that somebody that tries to read and understand that class later has to read/understand more code. Furthermore that adds an extra option in the preferences that a User probably tries to understand if seeing it. So even code that does not have to be touched can cause workload (of course less than code that as to be actively maintained) and should not be added just because it already exists and it should still be weighed if it is necessary and worth it or not.

kwin:
In the future resolving `java.home` in the pom.xml also requires to evaluate the Maven Execution JRE (compare with https://github.com/eclipse-m2e/m2e-core/issues?q=is%3Aissue+is%3Aopen+java.home) and that happens definitely outside of `m2e.launching`.

HannesWell:
> Resolving `java.home` in the pom.xml also requires to evaluate the Maven Execution JRE (compare with https://github.com/eclipse-m2e/m2e-core/issues?q=is%3Aissue+is%3Aopen+java.home) and that happens definitely outside of `m2e.launching`.

OK, that makes a difference. The analysis in https://github.com/eclipse-m2e/m2e-core/issues/497#issuecomment-1015468495 looks quite correct. And you are right, that this can't be done outside of m2e.core.

laeubi:
As mentioned above I think this should be decoupled by a service, also the linked PR has some comments how such things can be archived, also I think we should not mix things just to judge a certain direction.

The launching JRE is clearly a concern of m2e.launching and should be handled there. m2e-core might set a project preference (e.g. `m2e.runtime.jre = 11.2.4``) derived from the enforcer plugin but m2e-launching should interpret this value and probably match this to an available JRE befor performing the actual launch.

Injecting system.properties (like `java.home`) is a different concern and is irrelevant here because when you launch the maven build it is already set correctly, it only becomes a concern when executing the mojo inside m2e, where one might argue that setting java home to something project specific is wrong because it obviously is **not** the home of the jvm that is currently running, I would even say any project that relies one a `java.home` is probably broken and would better use something like toolchains instead.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/MavenProjectPreferencePage.java
```diff
@@ -98,21 +134,151 @@ protected Control createContents(Composite parent) {
...
+        IVMInstall left = (IVMInstall) o1;
+        IVMInstall right = (IVMInstall) o2;
+        return left.getName().compareToIgnoreCase(right.getName());
+      }
+
+      @Override
+      public boolean equals(Object obj) {
+        return obj == this;
+      }
+    });
```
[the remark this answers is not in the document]
laeubi: `Comparator.comparing(IVMInstall::getName, String.CASE_INSENSITIVE_ORDER)`

kwin:
I am wondering how to proceed here and come to some consensus. I followed the advice from https://github.com/eclipse-m2e/m2e-core/pull/1137#issuecomment-1362460896 and removed all JDT dependencies from m2e.core. As outlined before I refrain from removing the persistence of the Maven Execution JRE (both for performance reasons and to also allow to leverage that setting for other purposes (`java.home`)).

HannesWell:
> I am wondering how to proceed here and come to some consensus. I followed the advice from [#1137 (comment)](https://github.com/eclipse-m2e/m2e-core/pull/1137#issuecomment-1362460896) and removed all JDT dependencies from m2e.core. As outlined before I refrain from removing the persistence of the Maven Execution JRE (both for performance reasons and to also allow to leverage that setting for other purposes (`java.home`)). I would prefer some direct code related comments now to bring this into the final shape over just general remarks.

I'm torn in this case, but since #497 is probably another difficult topic that takes more time, I think we should not consider it in this PR even if that means that we might move the code again. But the other way round but be bad either, if we now add it in m2e.core assuming we will need it there one day, but eventually never use it. Both is not ideal, but the latter case is IMHO worse.

So my suggestion is to handle the selection of the default-JVM only in m2e.launching on every launch. I don't expect any noticeable performance regression from that. If you do, please make at least some very basic runtime measurements of the method that selects the default JVM from maven-compiler+enforcer-plugin.

laeubi:
> If you do, please make at least some very basic runtime measurements of the method that selects the default JVM from maven-compiler+enforcer-plugin.

What I don't get here is why not the maven-compiler+enforcer-plugin can extract the versions from the pom and store that wherever and **at the time of the launch** the mapping to an installed JRE is performed? That should be a simple loop/match over all installed JREs what can't be that slow.

kwin:
> We already have had the demand to inspect JRE here where it also was quite complex to not do it in core directly:
> 
> * [Draft for review #847](https://github.com/eclipse-m2e/m2e-core/pull/847)
> 
> I therefore thing it would be good to have some kind of `JavaToolchainManager` interface that abstracts this part and the JDT bundle implements the service.

So this is no longer relevant? Should the service be removed then again? There is a lot back and forth here which makes it difficult for me…

laeubi:
> So this is no longer relevant? 

@HannesWell  convinced me that for this purpose it is not required and could be simplified to these steps:

1. `maven-compiler-plugin`+`maven.enforcer-plugin` settings are evaluated an stored as raw values
2. on launching (if not a custom JRE is selected by the user) these raw values are mapped to the ones configured in JDT

this should reduce the scope of this PR to the minimum and allows it to be merged soon.

> Should the service be removed then again?

I think such a service can be useful in the future, but for the moment you might want to save this effort in  a dedicated branch so we can decide over this independently (or probably delay until it is strictly required).

> There is a lot back and forth here which makes it difficult for me…

Sorry for the confusion here, but it often shows that what initially was meant as a good idea later turned out to be not optimal due to others raising concerns, so this has evolving a bit here, but in the end I think that's the purpose of reviews, to explore different ways and find the one that fits best as of today knowledge. An alternative would be that before introducing a new feature we plan/discuss it, drawing diagrams and so on... I just not sure if this would really be beneficial and find "discussion on code" a lot better suited.

kwin:
Ok, let me try to summarise the necessary changes then again from my PoV:
1. remove UI for maintaining a Maven Execution JRE from project properties
1. only store m-execution-p configuration for java version in project preferences (evaluating m-compiler-p configuration is not necessary, as this is already done elsewhere in the code and can still act as fallback. The m-compiler-c is persisted in the project already via Build JRE)
1. I will keep the service for dealing with installed JREs, but keep it private inside the `m2e.launching` module, as that eases testing

laeubi:
> remove UI for maintaining a Maven Execution JRE from project properties

I think it would be fine to have the UI as part of the m2e.jdt.ui for example that could then just be a separate entry (like we have for life-cycle-mapping there could be "launching")

> I will keep the service for dealing with installed JREs, but keep it private inside the m2e.launching module, as that eases testing

If we can improve testing that's good for sure!

kwin:
>  think it would be fine to have the UI as part of the m2e.jdt.ui for example that could then just be a separate entry

Then the selected JRE and the enforced java version (via m-enforcer-p configuration) can easily get out of sync (i.e. the former does no longer satisfy the latter). If anyone has a good idea how to solve that one can enhance this later on.

laeubi:
What do you mean by "getting out of sync"?

HannesWell:
>     1. remove UI for maintaining a Maven Execution JRE from project properties
> 

I would leave that decision to you if you think that is necessary. I never even had that use-case to need to adjust the execution JRE at all, so it is hard for me to judge if the cases where the combination of compiler+enforcer-p can happen so often that a preference is worth it. So I'll leave that decision up to you.

>     2. only store m-execution-p configuration for java version in project preferences (evaluating m-compiler-p configuration is not necessary, as this is already done elsewhere in the code and can still act as fallback. The m-compiler-c is persisted in the project already via Build JRE)

I still think that it isn't a noticable performance issue to read the enforcer-p configuration when launching a build-process. That would also ensure that this value is always up to date.

laeubi:
> I still believe that it probably isn't a noticeable performance issue to read the enforcer-p configuration each time when launching a build-process. That would also ensure that this value is always up to date.

I just want to note that at m2e we now always update the project if the pom changes automatically, so I really don't understand why things should get out of sync at all?

kwin:
> So I really don't understand why things should get out of sync at all?

With persisting a Maven Execution JRE in the UI we would have 3 configurations:
a) m-enforcer-p, requireJava configuration value (in the pom.xml)
b) same value as a) persisted in the Eclipse project context
c) the selected JRE from the UI which must match a) and b)

a) and b) cannot get out of sync, but c) will never be updated automatically when another Java version is now required. That is what I meant (and though I even wrote) above.

laeubi:
> a) and b) cannot get out of sync, but c) will never be updated automatically when another Java version is now required.

If the user forces a given java version I think it is his own responsibility to keep this updated regardless of what m2e can do automatically.

kwin:
For now I removed the additional UI for project level Maven Execution JRE, so there is either automatic (derived from m-enforcer-p or falling back to project build JRE or workspace JRE) or manual override per launch configuration.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/MavenProjectPreferencePage.java
```diff
@@ -98,21 +134,151 @@ protected Control createContents(Composite parent) {
...
+
+    createDefaultJREControls(composite);
+    createAlternateJREControls(composite);
+  }
+
+  private void createDefaultJREControls(Composite comp) {
+    defaultJRERadioButton = new Button(comp, SWT.RADIO);
+    defaultJRERadioButton.setText(getDefaultJreName());
+    defaultJRERadioButton.setLayoutData(new GridData(SWT.FILL, SWT.BEGINNING, false, false, 3, 1));
+    defaultJRERadioButton.addSelectionListener(new SelectionAdapter() {
```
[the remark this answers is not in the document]
kwin: code removed

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/MavenProjectPreferencePage.java
```diff
@@ -98,21 +134,151 @@ protected Control createContents(Composite parent) {
...
+
+    alternateJREsCombo = new Combo(comp, SWT.DROP_DOWN | SWT.READ_ONLY);
+    alternateJREsCombo.setLayoutData(new GridData(GridData.FILL_HORIZONTAL));
+    alternateJREsCombo.setItems(vmInstalls.stream().map(IVMInstall::getName).toArray(String[]::new));
+    alternateJREsCombo.setVisibleItemCount(Math.min(vmInstalls.size(), 20));
+    alternateJREsCombo.setEnabled(false);
+
+    Button installedJREsButton = new Button(comp, SWT.PUSH);
+    installedJREsButton.setText(Messages.MavenProjectPreferencePage_btnInstalledJREs);
+    installedJREsButton.addListener(SWT.Selection, new Listener() {
```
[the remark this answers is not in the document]
kwin: code removed

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/MavenProjectPreferencePage.java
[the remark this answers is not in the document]
kwin: code removed

org.eclipse.m2e.core/META-INF/MANIFEST.MF
kwin: no longer relevant and therefore changes reverted

org.eclipse.m2e.core/src/org/eclipse/m2e/core/MavenPlugin.java
```diff
@@ -68,6 +69,9 @@ public static IProjectConversionManager getProjectConversionManager() {
     return MavenPluginActivator.getDefault().getProjectConversionManager();
   }
 
+  public static IJREManager getJREManager() {
+    return MavenPluginActivator.getDefault().getJREManager();
+  }
```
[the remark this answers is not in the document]
kwin: code removed

RELEASE_NOTES.md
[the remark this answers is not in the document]
kwin: I rephrased this now, please check.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/IProjectConfiguration.java
```diff
@@ -62,4 +64,13 @@ private static List<String> parseProfiles(String profilesAsText, boolean status)
...
 
+  /**
+   * Gets the required Java version for executing Maven. If {@code null} is returned the project's build JRE or (if not
+   * available) the workspace default JRE should be used. Derived from the
+   * <a href="https://maven.apache.org/enforcer/enforcer-rules/requireJavaVersion.html">{@code maven-enforcer-plugin}'s
+   * {@code requireJavaVersion} rule</a>.
+   * 
+   * @return a version spec compliant with {@link VersionRange} spec
+   */
+  String getRequiredJavaVersion();
```
HannesWell, line 75: Do you have a need for that as part of M2E's public API?
Actually I would consider this as internal.

In case there is a real need for that, I would rename it to `getExecutionJavaVersion()` or similar.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/ProjectConfigurationManager.java
```diff
@@ -500,9 +508,62 @@ private void updateProjectConfiguration(ProjectConfigurationRequest request, IPr
...
 
+    // must happen after calling the project configurators (as it checks the set Java project JRE)
+    configurationRequiredJavaVersion(mavenProjectFacade, monitor);
     log.debug("Updated project configuration for {} in {} ms.", mavenProjectFacade, System.currentTimeMillis() - start); //$NON-NLS-1$
   }
 
+  /**
+   * Configure Java version required for Maven execution based on maven-enforcer-plugin configuration
+   */
+  private void configurationRequiredJavaVersion(IMavenProjectFacade facade, IProgressMonitor monitor) {
```
HannesWell, line 519: I still would prefer to move that to the `MavenLaunchDelegate`.
This would save the logic to persist that value and unless there is a strong need to make that available in the API (see my other comment), I don't think this is necessary here. Reading the value is not very costly.

HannesWell (commented):
I could not yet look into the details but from a quick overview my main points I'd like to discuss are the new API and where the enforcer-plugin config is read. Please see the comments below.

kwin:
> > So this is no longer relevant?
> 
> @HannesWell convinced me that for this purpose it is not required and could be simplified to these steps:
> 
> 1. `maven-compiler-plugin`+`maven.enforcer-plugin` settings are evaluated an stored as raw values
> 2. on launching (if not a custom JRE is selected by the user) these raw values are mapped to the ones configured in JDT
> 
> this should reduce the scope of this PR to the minimum and allows it to be merged soon.

You guys really giving me a hard time. I did exactly that in my last update.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/IProjectConfiguration.java
laeubi: Please don't use the `IProjectConfig` for this! This class is used to group project that have the same resolver context (e.g. selected properties and profiles) but not a general purpose place to store project specific settings, beside that these are workspace specific and cannot be shared.
This should be stored in the preferences (e.g. `org.eclipse.m2e.core.prefs` or even `org.eclipse.m2e.launching.prefs`).

laeubi (changes requested):

laeubi:
I think we probably should split this into three parts (and PRs):

1. Add a way to read/store the `RequiredJavaExecutionVersion`, but don't change the tests / enforcer parsing / launching / UI ...
2. change the enforcer parsing to update this `RequiredJavaExecutionVersion`  on project updates
3. Change the launching part to take this into account, probably UI / test adjustments

That way it would be possible to be more focused and we can probbaly already merge some parts of this than trying to get this done in one big chunck.

kwin:
> I think we probably should split this into three parts (and PRs):

Sorry, not willing to invest that much time here also for me these all belong to the same feature.

> Add a way to read/store the RequiredJavaExecutionVersion

As long as you guys have not aligned on this, that won't make any sense. @HannesWell is against persisting it.

HannesWell:
> 
> You guys really giving me a hard time. I did exactly that in my last update.

Sorry for that, but I can tell you that it is also not fun, if you know that you do that to somebody.

To make your time not even harder, I added the changes how I would do it to another branch:
https://github.com/HannesWell/m2e-core/tree/pr-1137
Basically this moves read of the m-enforcer-plugin from the `ProjectConfigurationManager` to the  `MavenLaunchDelegate` and reads the config on every launch, without persisting anything.
This also makes the change smaller and less complicated. The current state is definitely not perfect and requires some fixes and streamlining but I hope it makes my suggestion clearer.
If we agree on proceeding like this, I can do the fixes/streamlining and you Konard can then just cherry-pick this and squash this into this PR.

> > As long as you guys have not aligned on this, that won't make any sense. @HannesWell is against persisting it.
> 
> As said, I think the issue is just to broad and thus things are getting mixed up, the concern was just to compute the enforced java version as part of the (default) project configuration. That is different from the topic of supporting different runtime JVM and that must be persisted somewhere and enforcer might not be the only source (e.g. I really liked the idea of having a user configuration UI for that).

Actually with the launch configs we already have a place where to persist a specific runtime JVM if the default is not sufficient. With this change (in the way it should IMHO be done) 'just' the default becomes smarter.
There still could be cases where one needs/wants another JVM than the default and creates new launch configs frequently. Then a project preference could be convenient because the JVM does not have to be adjusted in each launch. But those cases should become much more seldom with this change, so I'm in doubt if there is great demand for an extra preference this.

HannesWell (approved):
@kwin and @laeubi
I have now worked out a solution based on what I have described in previous comments, which I believe satisfies all needs. It only considers the enforcer-plugin to select the JRE to run a launched maven build and it does that, without storing anything the projects resolver config, without any new API.

Additionally this approach can read the enforcer config, if if the base-dir/pom-dir points to a project, which is not in the workspace. That is IMHO a quite common scenario, e.g. if one doesn't import the root of a git repo, which often also is the root for a maven build.

Please have a look at it and let me know what you think.
If you are fine with it, it should be submitted ASAP, so that I can release M2E this evening, just in time for 2023-03 RC1.

laeubi:
> Additionally this approach can read the enforcer config, if if the base-dir/pom-dir points to a project, which is not in the workspace. That is IMHO a quite common scenario, e.g. if one doesn't import the root of a git repo, which often also is the root for a maven build.

I must confess I don't understand this part, a configuration should be inherited by its child so it doesn't matter if the parent is in the workspace or not?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
```diff
@@ -437,51 +439,60 @@ public void setMojoExecutionMapping(Map<MojoExecutionKey, List<IPluginExecutionM
...
       IProgressMonitor monitor) {
     try {
-      return createExecutionContext().execute(mavenProject, (ctx, mon) -> {
+      return context.execute(mavenProject, (ctx, mon) -> {
         IMavenToolbox toolbox = IMavenToolbox.of(ctx);
         return toolbox.calculateExecutionPlan(tasks, setup);
       }, monitor);
     } catch(CoreException e) {
-      manager.getMarkerManager().addErrorMarkers(pom, IMavenConstants.MARKER_POM_LOADING_ID, e);
+      exceptionHandler.accept(e);
```
laeubi, line 495: This part looks suspicious (also in the original) why not just throw the exception and let it handle at the caller site?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
```diff
@@ -628,7 +654,7 @@ private MavenProject tryGetMavenProject() {
       return manager.getMavenProject(this, null);
     } catch(CoreException ex) {
       //TODO can we handle this more graceful? E.g. one error condition is that project dependencies can not be resolved, maybe we can try to load the project without dependencies then?
-      manager.getMarkerManager().addErrorMarkers(pom, IMavenConstants.MARKER_POM_LOADING_ID, ex);
+      logException(ex);
```
laeubi, line 657: same here, better throw the exception to the caller instead and handel it there

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
```diff
@@ -638,6 +664,10 @@ public IComponentLookup getComponentLookup() {
     return manager.getContainerManager().getComponentLookup(getConfiguration().getMultiModuleProjectDirectory());
   }
 
+  private void logException(CoreException e) {
```
laeubi, line 667: this does not log the exception it adds an error marker

HannesWell:
> > Additionally this approach can read the enforcer config, if if the base-dir/pom-dir points to a project, which is not in the workspace. That is IMHO a quite common scenario, e.g. if one doesn't import the root of a git repo, which often also is the root for a maven build.
> 
> I must confess I don't understand this part, a configuration should be inherited by its child so it doesn't matter if the parent is in the workspace or not?

That's right. But which project in the workspace should I check if I just have to path the root of a git repo? Iterate over all projects in the WS and check if its within the given path? But then the config could be overridden in a child-project. Maybe different in different childs. Then it's the question which should I take?
Furthermore it can be that one want's to build a project that is not in the workspace at. At least I do that sometimes, for example debugging tests is much simpler when I start the build within Eclipse, than on the CLI.

Considering all those corner cases, it is simpler to just load the pom from the given path and read its enforcer config. It is not the nicest approach, but I think it works in the most robust way.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
HannesWell: This method is called by `IMavenProjectFacade.calculateExecutionPlan/setupExecutionPlan()`, which don't permit CoreExceptions in the method signature.
So without breaking the API we cannot propagate the original exception. It would only possible when wrapping it in a RuntimeException.
But I think that would not help much here. It should have been thrown before, but changing it now is not ideal.
But we should note that for a M2E 3.0 :)

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
HannesWell: Same as above. Called by `IMavenProjectFacade.createExecutionContext()`.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
HannesWell: Same as above. Called by `IMavenProjectFacade.createExecutionContext()`.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
HannesWell: That's right. Changed it to addErrorMarker().

laeubi:
> That's right. But which project in the workspace should I check if I just have to path the root of a git repo? 

What do you mean by "root of the git repo"? Actually can only execute a `pom.xml` and then you know what project this resides on and this should be then the MavenProject to query (what might inherit from a pom further down its configuration) ... so what execution we are talking about as m2e always ever only executes one pom, and the childs should not be relevant here, if they define different settings for java I'm not sure if this is something we should consider or is even valid?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
laeubi: then wrapping it in a runtime exception (or "handle") it there, this is simply bad style and we should avoid that so pushing the handling as far as possible to the top.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/MavenProjectFacade.java
laeubi: One way would be to return an "ExceptionalExecutionContext" that throws the original exception on any call of its methods (that allow CoreException already), this is also done with the Plexuscontainer as well.

HannesWell:
> > That's right. But which project in the workspace should I check if I just have to path the root of a git repo?
> 
> What do you mean by "root of the git repo"? Actually can only execute a `pom.xml` and then you know what project this resides on and this should be then the MavenProject to query (what might inherit from a pom further down its configuration) ... so what execution we are talking about as m2e always ever only executes one pom, and the childs should not be relevant here, if they define different settings for java I'm not sure if this is something we should consider or is even valid?

As we have discussed in private chat, I have now removed the part to load the enforcer-config from a maven-project that is not in the workspace and use only the existing methods on the `IMavenProjectFacade`.
The reasoning is that we first want to have a general loading of IMavenProjectFacades from a File/Path regardless of if it is in the workspace or not (so something like `IMavenProjectFacade.of(File)`). Then later we can add use that to load the MavenConfig for a Maven launch.
For now this further simplifies this change, but still fixes mainly intended cases, where one selects Run As > Maven ... on a project in the workspace.

HannesWell:
The Jenkins build is green and the GH-workflows fail due to Tycho's current incompatibility with Maven 3.9 (looks like not all runners use Maven 3.9, some use 3.8.7).

So this is ready for submission.

Btw. since this was also an issue discussed in this PR:
Since version 5 Mockito can also mock static methods (and IIRC also final methods) out of the box, which I used here and made testing much easier.

laeubi:
Finally :+1:

PR #1140, opened 2022-12-17
Do not wrap source bundles

laeubi:
Currently m2e wraps source into new artifacts having performance and space implications and can be confusing because users expect the file to be on a different location.

Due to the recent changes in PDE this is no longer required an we can remove this workaround.

See https://github.com/eclipse-pde/eclipse.pde/pull/425

This will require a new PDE release ...

### DIFF
--- a/org.eclipse.m2e.pde.target/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.pde.target/META-INF/MANIFEST.MF
@@ -6,7 +6,7 @@ Bundle-Version: 2.0.300.qualifier
 Automatic-Module-Name: org.eclipse.m2e.pde.target
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Require-Bundle: org.eclipse.core.runtime;bundle-version="3.19.0",
- org.eclipse.pde.core;bundle-version="3.14.0",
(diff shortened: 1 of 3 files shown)


HannesWell (commented):
I think that's a good change, and think that is actually to not have sources as bundles since they are not meant to be installed in a OSGi Runtime. And better performance and less code is also beneficial. 

However I wonder if it will then still work to reference the sources-(not bundles any-more) in a Feature? In general I would recommend to not include sources in a Feature, but there might be cases where it is needed. And since it used to be possible I think there are people that do it.

It might work to add a corresponding element manually in the text of the source-tab of the Feature-Editor. But that will then show warnings in the IDE. Tycho probably can still handle this since it is not changed. 
That is probably a more general issue in PDE?

laeubi:
If thats important to anyone this needs to be fixed at PDE, the target platform still promotes this as a `TargetBundle` so PDE has all it needs to know to use this, that is the part here:

```
fInfo.setSymbolicName(sourceTarget.getSymbolicName() + ".source");
fInfo.setVersion(sourceTarget.getVersion());
```

akurtakov:
Is there a plan for this to be finished (if it's still needed at all?)?

PR #1151, opened 2022-12-21
Remove unused dependencies from m2e.core and squash Service-Components

HannesWell:
@laeubi what do you think about squashing the Service-Components? The list became quite long in the meantime.


laeubi:
I don't see any benefit as the list is automatically generated while having a manual pattern there is no guarantee that anything is ever validated and/or updated.

HannesWell:
> I don't see any benefit as the list is automatically generated while having a manual pattern there is no guarantee that anything is ever validated and/or updated.

I have to admit I didn't check if PDE/Tycho then skips the validation if a manual pattern is used. But then yes, lets keep that part as it is.

HannesWell:
Strange that it did not cause a compiler error in the IDE when I removed `com.google.common.base`, but shows up as NoClassDefFoundError immediately when launching m2e.

PR #1157, opened 2022-12-28
Show Maven artifacts via list instead of tabs for the dependency editor.

ptziegler:
The current implementation can only show ~5 items at a time. All remaining dependencies have to be selected via a combo box. For projects with several dozen dependencies, it is often faster to edit the XML file directly, rather than using this wizard.

The new implementation keeps all dependencies in a list. Clicking on one of its elements shows more detailed information in a separate page, like GAV, classifier or type.

Further improvements contain the deletion of multiple elements at a time, undo/redo operations, keybindings, as well as a noticeable performance optimizations.

**Open Topics**
~~I'd like add some proper tests, but I think this editor is beyond what can be easily tested via JUnit (primarily the interaction between the individual composites). Have there been considerations to add something like RCPTT to the test suite?~~

~~What's the indented workflow for using images? I'm currently importing them from org.eclipse.ui via ISharedImages. But I've also seen that we have copied some of them directly into the workspace.~~


ptziegler:
Before:
before
After:
after

laeubi:
@ptziegler if you really like to go with the table approach, would it not be more suitable to go this path all way down and have an editable table with five rows? This would then also allow having sorting by clicking the header by group/artifact/version/.... wdyt?

ptziegler:
> would it not be more suitable to go this path all way down and have an editable table with five rows?

I assume you mean columns and not rows? It should make the list entries (GAV+Type) a lot more readable. Switching from a ListViewer to a TableViewer shouldn't be difficult, so I'll give it a go and see how it looks.

laeubi:
|Group Id  | Artifact Id  | Version  | Classifier  | Type |
|---|---|---|---|---|
| a |b  |c  |  |jar  |
| x | y | z |  | jar |
|...  |  |  |  |  |

And then have the cells being editable items: https://www.vogella.com/tutorials/EclipseJFaceTable/article.html#column-editing-support

ptziegler:
@laeubi I think you're right. Having everything in a simple table looks a lot cleaner.

And being able to sort by column can be quite convenient, I think :)

image

laeubi:
This looks really promising, I just think we now have a lot of free "gray area", do you maybe like to try out:

- A variant where the buttons are located on the right side (you might want to use the new `BorderLayout` of SWT for that) from top to bottom?
- Or if you think its better on the to adding some text (maybe with shortcut) that describes the button e.g "Add more dependencies", "delete selected", "undo", "redo" ...

Beside that, maybe a "update to latest version" button would be useful, we already have this functionality in the editor, but it seems a good place to have it here as well!

HannesWell:
That's a very nice improvement. 👍🏽

ptziegler:
I've moved the buttons to the right side and also added an Update button, which looks up the latest versions of the selected artifacts. Though I ended up dropping the idea of using both images and text.

For some reason, they kept being centered and therefore none of the images lined up properly. Which didn't look very good. From what I found, this seems like a very old limitation in SWT ([bug 172850](https://bugs.eclipse.org/bugs/show_bug.cgi?id=172850)).

On the bright side, the new layout looks similar to the Target editor, which is probably a good thing.
image

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/ClipboardUtils.java
```diff
@@ -0,0 +1,70 @@
...
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.m2e.pde.target.MavenTargetDependency;
+import org.eclipse.m2e.pde.ui.target.editor.ClipboardParser;
+import org.eclipse.m2e.pde.ui.target.editor.MavenTargetLocationWizard;
+import org.eclipse.m2e.pde.ui.target.editor.Messages;
+import org.eclipse.swt.dnd.Clipboard;
+import org.eclipse.swt.dnd.TextTransfer;
+import org.eclipse.swt.widgets.Display;
+
+public abstract class ClipboardUtils {
```
laeubi, line 28: Can this static method be moved to `ClipboardParser` instead of having a util class?

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/Constants.java
```diff
@@ -0,0 +1,57 @@
...
+import org.eclipse.core.commands.operations.IUndoContext;
+import org.eclipse.core.commands.operations.UndoContext;
+
+public abstract class Constants {
+
+	/**
+	 * Undo context used by m2e for all internal modifications of the Maven
+	 * dependency table (e.g. the initialization).
+	 */
+	public static IUndoContext SYSTEM_CONTEXT = new UndoContext();
```
laeubi, line 24: If it is used  in the dependency table only, maybe these are better suited to be places there?

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/MavenAccessor.java
```diff
@@ -0,0 +1,73 @@
...
+ * <ul>
+ * <li>group id</li>
+ * <li>artifact id</li>
+ * <li>version</li>
+ * <li>classifier</li>
+ * <li>type</li>
+ * </ul>
+ * </p>
+ */
+public class MavenAccessor implements Function<Object, String>, BiConsumer<Object, Object> {
```
laeubi, line 33: Object seems not really they type to use here but `MavenTargetDependency` / `String`?

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/viewer/MavenTextEditingSupport.java
```diff
@@ -0,0 +1,61 @@
...
+ * <li>group id</li>
+ * <li>artifact id</li>
+ * <li>version</li>
+ * <li>classifier</li>
+ * </ul>
+ */
+public final class MavenTextEditingSupport extends EditingSupport {
+	private final MavenAccessor accessor;
+
+	public MavenTextEditingSupport(ColumnViewer viewer, MavenAccessor accessor) {
```
laeubi, line 36: Probably it would be better instead of having `MavenAccessor`  to pass the getter/setter here to remove an indirection?

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/viewer/MavenTextEditingSupport.java
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	@Override
+	protected boolean canEdit(Object element) {
+		return element instanceof MavenTargetDependency;
+	}
+
+	@Override
+	protected Object getValue(Object element) {
+		return accessor.apply(element);
```
laeubi, line 53: here would be a better place to cast the object than in the `MavenAccessor`

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/viewer/MavenTextEditingSupport.java
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	@Override
+	protected void setValue(Object element, Object value) {
+		accessor.accept(element, value);
```
laeubi, line 58: here would be a better place to cast the object than in the `MavenAccessor`

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/viewer/MavenTypeEditingSupport.java
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	@Override
+	protected boolean canEdit(Object element) {
+		return element instanceof MavenTargetDependency;
+	}
+
+	@Override
+	protected Object getValue(Object element) {
+		String value = accessor.apply(element);
```
laeubi, line 59: Do we really need to operate on the index? What about the user is entering a custom type (what is allowed)

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/viewer/TableColumnSortListener.java
```diff
@@ -0,0 +1,45 @@
...
+	private final TableViewer viewer;
+	private int sortDirection;
+
+	public TableColumnSortListener(TableViewer viewer) {
+		this.viewer = viewer;
+	}
+
+	@Override
+	public void widgetSelected(SelectionEvent e) {
+		sortDirection = sortDirection != SWT.UP ? SWT.UP : SWT.DOWN;
```
laeubi, line 39: sortDirection would probably better be a boolean and then map it to the integer afterwards? Using `!=` here makes this line looking more complex as it is.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/provider/MavenTargetColumnLabelProvider.java
```diff
@@ -0,0 +1,30 @@
...
+public class MavenTargetColumnLabelProvider extends ColumnLabelProvider {
+	private Function<Object, String> getter;
+
+	public MavenTargetColumnLabelProvider(Function<Object, String> getter) {
+		this.getter = getter;
+	}
+
+	@Override
+	public String getText(Object element) {
+		return getter.apply((Object) element);
```
laeubi, line 28: Cast `Object` to `Object`?

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/provider/MavenTargetColumnLabelProvider.java
```diff
@@ -0,0 +1,30 @@
...
+package org.eclipse.m2e.pde.ui.target.provider;
+
+import java.util.function.Function;
+
+import org.eclipse.jface.viewers.ColumnLabelProvider;
+
+public class MavenTargetColumnLabelProvider extends ColumnLabelProvider {
+	private Function<Object, String> getter;
+
+	public MavenTargetColumnLabelProvider(Function<Object, String> getter) {
```
laeubi, line 22: I think this should better be a parmeterized class.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/viewer/MavenTypeEditingSupport.java
```diff
@@ -0,0 +1,68 @@
...
+import org.eclipse.jface.viewers.EditingSupport;
+import org.eclipse.m2e.pde.target.MavenTargetDependency;
+import org.eclipse.m2e.pde.ui.target.editor.internal.Constants;
+import org.eclipse.m2e.pde.ui.target.editor.internal.MavenAccessor;
+import org.eclipse.swt.widgets.Composite;
+
+/**
+ * <p>
+ * This class provides the ability to edit the type of a Maven dependency via a
+ * combo box. Supported types are:
```
laeubi, line 29: This is wrong, the combo-box should be editable one so the user can type in any string/type these are only some common types we want to offer the user a choice without typing it in manually.

laeubi (commented):
I think the UI looks very good now some remarks on the code:

- There are a lot of whitespace/formating only changes, it would be good to reduce these to make review easier (e.g. only formating changed lines)
- Some comments on the code itself below
- You are using `PropertyChangeListeners` to update the UI, probably using `IObservables` and `ISideEffectFactory` would be more convenient than handling this manually.
- You could add a not (and your last screenshot) to the RELEASE_NOTES

ptziegler:
> There are a lot of whitespace/formating only changes, it would be good to reduce these to make review easier (e.g. only formating changed lines)

Yeah... the auto-formatter is both a blessing and a curse. I think it only affected a single class. My apologies.

But I think I've resolved all of the issues. I've split up the UI from the "model" via the ISideEffect and IObservable API and so far everything seems to be still working.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/viewer/MavenTextEditingSupport.java
ptziegler: Having this indirection was the initial idea. In the end, I removed this class completely in favor of directly passing on the getter/setter.

ptziegler:
Hm... I ran an RCPTT test earlier today and it seems like it's getting stuck in an infinite loop when accessing the IObservable containing the IOperationHistory instance.
I think I have to find a better way to update the enablement of the Undo/Redo button.

laeubi:
When I recently used RCPTT it fails to work with the target editor ... if you was able to get it working it would be great to share the RCPTT test!

Does this loop also happen in a regular eclipse? If not it might be good to report a bug to RCPTT ...

ptziegler:
> When I recently used RCPTT it fails to work with the target editor ... if you was able to get it working it would be great to share the RCPTT test!

I already planned on doing that once I'm happy with the test case.😅 But I think this should be done via a separate pull request. Because I need to include the RCPTT runner in the build job. Which isn't really difficult with the maven-rcptt-plugin, but unrelated to the dependency editor.

> Does this loop also happen in a regular eclipse? If not it might be good to report a bug to RCPTT ...

I'm not exactly sure... The UI is responsive in both the regular Eclipse and in RCPTT. So I assume the update of the IObservable is done in a background task. And because the runner is waiting for the Eclipse to become "ready", it gets stuck indefinitely.

It's entirely possible that the same happens during the normal operation, it just isn't noticable. Then again... I ran into some sporadic OutOfMemory/Heap Space errors during the past few days where I couldn't really track down the root cause. There is a good chance that those are related, if the update function is called indefinitely while the editor is open.

laeubi:
I could try to take a look at the code a bit closer today ... but in general you can try to set a breakpoint in the update function to see if it triggers again and again. Sadly there are some operations in the observable framework of eclipse that are not very optimal and one can easily create infinite updates for some collection operations. e.g. if you clear a collection (that already is empty) it is still considered a change...

ptziegler:
> but in general you can try to set a breakpoint in the update function to see if it triggers again and again

At least for the RCPTT tests, that's what I did. And also how I figured out why it got stuck. I'll test it again this evening with the regular Eclipse.

> there are some operations in the observable framework of eclipse that are not very optimal and one can easily create infinite updates for some collection operations

I don't think it's an issue with the Observable framework, but rather my custom implementation.

My idea was that the Undo/Redo buttons should be updated, whenever a new operation is added or removed from the IOperationHistory instance.

However, the DefaultOperationHistory I'm using (which is sadly the only implementation) stores the operation in normal lists instead of IObservableLists. Meaning the side-effect mechanism doesn't work here.

I wrote the OperationHistoryObservableValue class to work around this limitation but I guess it doesn't work quite as well as I had hoped.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/operation/OperationHistoryObservableValue.java
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	@Override
+	public Object getValueType() {
+		return null;
+	}
+
+	@Override
+	protected T doGetValue() {
+		fireEvent(new ChangeEvent(this));
```
laeubi, line 45: this seems not correct here why is getting a value firing a change event?

laeubi:
> At least for the RCPTT tests, that's what I did. And also how I figured out why it got stuck. I'll test it again this evening with the regular Eclipse.

Can you say where it get infinite loop?

> I wrote the OperationHistoryObservableValue class to work around this limitation but I guess it doesn't work quite as well as I had hoped.

Usually with such types it is the best to have a facade that manages the access to the underlying object that extends `AbstractObservable`. Then you has "observable-getters" (where you call `ObservableTracker.getterCalled(this);` before delegate to the original object), and "observable-setters" (where you call the delegate and then `fireChange()`), then you can use this object inside an `ISideEffect` context without any issue. As mentioned before, it is always good to check if there is an actual change to prevent unnecessary  events but thats already an optimization.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/operation/OperationHistoryObservableValue.java
ptziegler: That's what I meant... It's a quick & dirty solution to make sure the buttons are updated, whenever something is done to the IOperationHistory instance.

The Undo button, for example, checks its enablement state by calling IOperationHistory.canUndo(IUndoContext). 
But because this call doesn't access any IObservables, the side-effect for the button is never executed again. My "fix" was to create a custom dummy IObservableValue wrapping the IOperationHistory.  which notifies this side-effect, whenever accessed.

ptziegler:
> Can you say where it get infinite loop?

Not now, but I'll get back to you once I'm back home. I assume it's caused by the `fireEvent(new ChangeEvent(this))` you mentioned earlier.

> Usually with such types it is the best to have a facade that manages the access to the underlying object that extends AbstractObservable

Thanks for the suggestion! I'll try that out later today.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/operation/OperationHistoryObservableValue.java
laeubi: You should not call the `IOperationHistory` directly but for example `OperationHistoryFacade#canUndo` (like outlined below) that then delegates to the `IOperationHistory`

ptziegler:
Your recommendation was spot-on @laeubi.
The infinite loop was caused by one of the side-effect functions calling canUndo(), which then in return fired a new change event over and over again. It also happened with the normal Eclipse and went away, once I switch to a proper facade.
Thanks again for the suggestion :)
I think that's it for the editor.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/part/DependencyEditorViewerPart.java
```diff
@@ -0,0 +1,214 @@
...
+		tableColumn.setEditingSupport(new MavenTypeEditingSupport(viewer, model));
+	}
+
+	private TableViewerColumn newGenericColumn(int weight, Function<MavenTargetDependency, String> getter) {
+		TableViewerColumn tableColumn = new TableViewerColumn(viewer, SWT.NONE);
+		tableColumn.getColumn().addSelectionListener(new TableColumnSortListener(viewer, getter));
+		tableColumn.setLabelProvider(new MavenTargetColumnLabelProvider<>(getter));
+		getLayout().setColumnData(tableColumn.getColumn(), new ColumnWeightData(weight, false));
+		return tableColumn;
+	}
```
HannesWell, line 116: All classes in the new package org.eclipse.m2e.pde.ui.target.editor.internal.viewer could be inlined with the following change. This would save many lines by removing boilerplate code and would IMHO make it simpler to understand because one does not have to navigate through multiple classes.

You can added the classes documentation as documention of the corresponding methods
```suggestion
	private void createTableColumns() {
		Function<MavenTargetDependency, String> getter = MavenTargetDependency::getGroupId;
		BiConsumer<MavenTargetDependency, String> setter = MavenTargetDependency::setGroupId;
		TableViewerColumn tableColumn = newGenericColumn(250, getter);
		tableColumn.getColumn().setText(Messages.MavenTargetDependencyEditor_1);
		tableColumn.setEditingSupport(createMavenTextEditingSupport(getter, setter));

		getter = MavenTargetDependency::getArtifactId;
		setter = MavenTargetDependency::setArtifactId;
		tableColumn = newGenericColumn(250, getter);
		tableColumn.getColumn().setText(Messages.MavenTargetDependencyEditor_2);
		tableColumn.setEditingSupport(createMavenTextEditingSupport(getter, setter));

		getter = MavenTargetDependency::getVersion;
		setter = MavenTargetDependency::setVersion;
		tableColumn = newGenericColumn(250, getter);
		tableColumn.getColumn().setText(Messages.MavenTargetDependencyEditor_3);
		tableColumn.setEditingSupport(createMavenTextEditingSupport(getter, setter));

		getter = MavenTargetDependency::getClassifier;
		setter = MavenTargetDependency::setClassifier;
		tableColumn = newGenericColumn(125, getter);
		tableColumn.getColumn().setText(Messages.MavenTargetDependencyEditor_4);
		tableColumn.setEditingSupport(createMavenTextEditingSupport(getter, setter));

		getter = MavenTargetDependency::getType;
		tableColumn = newGenericColumn(125, getter);
		tableColumn.getColumn().setText(Messages.MavenTargetDependencyEditor_5);
		tableColumn.setEditingSupport(createMavenTypeEditingSupport());
	}

	private EditingSupport createMavenTextEditingSupport(Function<MavenTargetDependency, String> getter,
			BiConsumer<MavenTargetDependency, String> setter) {
		return createMavenEditingSupport(getter, setter, TextCellEditor::new);
	}

	private static final String[] DEFAULT_TYPES = new String[] { "jar", "bundle", "pom" };

	@SuppressWarnings("restriction")
	private EditingSupport createMavenTypeEditingSupport() {
		return createMavenEditingSupport(MavenTargetDependency::getType, MavenTargetDependency::setType, c -> {
			var editor = new org.eclipse.m2e.core.ui.internal.components.TextComboBoxCellEditor(c, SWT.NONE);
			editor.setItems(DEFAULT_TYPES);
			return editor;
		});
	}

	private EditingSupport createMavenEditingSupport(Function<MavenTargetDependency, String> getter,
			BiConsumer<MavenTargetDependency, String> setter, Function<Composite, CellEditor> cellEditorFactory) {
		return new EditingSupport(viewer) {

			@Override
			protected boolean canEdit(Object element) {
				return element instanceof MavenTargetDependency;
			}

			@Override
			protected Object getValue(Object element) {
				return getter.apply((MavenTargetDependency) element);
			}

			@Override
			protected void setValue(Object element, Object value) {
				MavenTargetDependency dependency = (MavenTargetDependency) element;
				String oldValue = getter.apply(dependency);
				String newValue = (String) value;

				// The table column needs to be redrawn. Otherwise we still see the old value...
				BiConsumer<MavenTargetDependency, String> realSetter = (k, v) -> {
					setter.accept(k, v);
					getViewer().update(element, null);
				};

				model.modify(dependency, realSetter, oldValue, newValue);
			}

			@Override
			protected CellEditor getCellEditor(Object element) {
				return cellEditorFactory.apply((Composite) getViewer().getControl());
			}
		};
	}

	private TableViewerColumn newGenericColumn(int weight, Function<MavenTargetDependency, String> getter) {
		TableViewerColumn tableColumn = new TableViewerColumn(viewer, SWT.NONE);
		tableColumn.getColumn().addSelectionListener(SelectionListener.widgetSelectedAdapter(e -> {
			final boolean sortAscending;
			if (viewer.getTable().getSortColumn() != e.widget) {
				// When we don't currently sort by this column, default to ascending order
				sortAscending = true;
			} else {
				// Switch between ascending & descending order
				sortAscending = viewer.getTable().getSortDirection() != SWT.DOWN;
			}
			viewer.getTable().setSortDirection(sortAscending ? SWT.DOWN : SWT.UP);
			viewer.getTable().setSortColumn((TableColumn) e.widget);

			viewer.setComparator(new ViewerComparator() {
				@Override
				public int compare(Viewer viewer, Object e1, Object e2) {
					String left = getter.apply((MavenTargetDependency) e1);
					String right = getter.apply((MavenTargetDependency) e2);
					return sortAscending ? left.compareTo(right) : right.compareTo(left);
				}
			});
			viewer.refresh();
		}));
		tableColumn.setLabelProvider(new MavenTargetColumnLabelProvider<>(getter));
		getLayout().setColumnData(tableColumn.getColumn(), new ColumnWeightData(weight, false));
		return tableColumn;
	}
```

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/part/DependencyEditorViewerPart.java
```diff
@@ -0,0 +1,214 @@
...
+				model.remove();
+			}
+		}));
+	}
+
+	/**
+	 * Updates the input and selection of this viewer with the entries from the
+	 * model. Does nothing if the entries are already up-to-date.
+	 */
+	private void onInputChange() {
```
HannesWell, line 161: This method and the following `onSelectionChange()` seem to be simple enough to be inlined, which would replace the method-reference by a lambda-block.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/operation/ModifyPropertyOperation.java
```diff
@@ -0,0 +1,59 @@
...
+import org.eclipse.core.runtime.Status;
+
+/**
+ * This class updates the property of a single element within the model. This
+ * change operation <b>does not</> trigger side-effects. This behavior is
+ * required when e.g. editing a cell in the table viewer. The data model should
+ * be updated with the new value but the input and selection should not be
+ * updated explicitly, as this is already handled by JFace.
+ */
+public class ModifyPropertyOperation extends AbstractOperation {
```
HannesWell, line 31: This class is also only used once and could be replaced by a corresponding anonymous class at the caller.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/operation/ModifyModelOperation.java
```diff
@@ -0,0 +1,66 @@
...
+import org.eclipse.core.runtime.Status;
+import org.eclipse.m2e.pde.target.MavenTargetDependency;
+import org.eclipse.m2e.pde.ui.target.editor.internal.model.TargetDependencyModel;
+
+/**
+ * This class updates both the list of all dependencies of the underlying target
+ * location, as well as the list of all currently selected elements. This change
+ * operation triggers all registered side-effects over those attributes.
+ */
+public class ModifyModelOperation extends AbstractOperation {
```
HannesWell, line 31: This class is also only used once and could be replaced by a corresponding anonymous class at the caller.

HannesWell (commented):
The result of this change looks very nice.

But I have the impression that at multiple places the code could be more compact. While it is good to separate concerns, distribution code over more classes makes it also harder to understand because one has to navigate more classes. Of course this has to be weighted but I think in this case multiple classes can be inlined and replaced by lambdas or anonymous classes, without making the other class too large. I made some suggestions in the comments below.

Since this contribution is currently greater than 1000 lines I wonder if we still have to create a Contribution Questionary to the IP team?

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/operation/ModifyPropertyOperation.java
laeubi: If it helps to keep the code more focused I think a "extra" class is fine, but it could probably made package private to show it is only used here.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/operation/ModifyModelOperation.java
laeubi: I'm not sure if inlining the class makes it more clear, but it coud be package private

laeubi (approved):
Even though there might be some improvements e.g. in the update handling I think the overall change is fine, I will open a CQ just t be sure.

@ptziegler if you like just proceed with @HannesWell to improve some code places and you might want to take a look at the `WidgetProperties` and the `ViewerProperties` class, they allow to create observable values e.g. for a selection that the can be used inside side-effects to synchronize state between model/ui in a littel more convenient ways than registering listeners manually.

laeubi:
The [handbock](https://www.eclipse.org/projects/handbook/#pmi-commands-cq) says:
> Click Committer Tools › Intellectual Property › Create a Contribution Questionnaire;

But I can't find it on the project page https://projects.eclipse.org/projects/technology.m2e 

@HannesWell any idea?

laeubi:
@HannesWell the reported p2 baseline problem is it related to:
- https://github.com/eclipse-m2e/m2e-core/pull/1168 ?

HannesWell:
> @HannesWell the reported p2 baseline problem is it related to:
> 
>     * [[Build] Move args to maven.cfg and use p2.baselineMode=failCommon in CI and fix extraction of M2E_VERSION for multi-digit versions #1168](https://github.com/eclipse-m2e/m2e-core/pull/1168) ?

Don't think so. It is probably because this branch is behind the master and the baseline replace replaced the m2e-sdk feature with the one from the master. I just updated it and that should fix that issue.



> The [handbock](https://www.eclipse.org/projects/handbook/#pmi-commands-cq) says:
> 
> > Click Committer Tools › Intellectual Property › Create a Contribution Questionnaire;
> 
> But I can't find it on the project page https://projects.eclipse.org/projects/technology.m2e
> 
> @HannesWell any idea?

Can't find it either. Maybe the need for a CQ was revoked with the recent IP policy update. But I didn't find that mail in a quick-search on the mailing lists.

ptziegler:
I applied your suggestions and inlined some of the trivial methods. For the selection listener, I don't see a benefit of switching to the side-effect framework, as the following code doesn't really look that much cleaner...
```
ISideEffectFactory factory = ...
IObservableValue<Boolean> button = WidgetProperties.buttonSelection().observe(...);
factory.create(() -> {
  if (button.getValue()) {
    doStuff()
  }
};
```

For the Undo/Redo operation, I have moved most of the logic from the model and the IUndoableOperation classes into the facade. You can't _undo_ a model, but you can undo the _history_ of model. So having the code for this in a separate class makes more sense to me.

laeubi:
Usually one uses a pattern like this:
```
button.setValue(model.isSelected)
factory.create(button::getValue, model::setSelected);
```

This automatically updates the model when the button changes and init it with the current model value. if you model is observable and you want a two way bindung you can use:
```
factory.create(model::isSelected, button::setValue);
factory.create(button::getValue, model::setSelected);
```

but this is more a bit micro-optimization here just in case :-)

laeubi:
> > This automatically updates the model when the button changes and init it with the current model value
> 
> But I don't think this pattern really works in this case, as clicking the "Add" button, for example, causes a model change. So just by running this command in the factory, I suddenly add a dependency to the model, even though I only created the button, not pressed it. :-)

Alright I probably getting confused here, so a a listener seems more suitable then.

> I'm probably completely oblivious as to how to merge the changes from m2e-core-tests back into m2e-core.
> Is there any documentation regarding the proper procedure? It's part of the pull-request but it just doesn't feel right.

I think the rcptt part should we a separate PR

ptziegler:
> the m2e-core-tests repo is just for "historic" reasons a seperate one, but we try to have the test next to the m2e repo instead of a seperate one when doing new stuff

Ah, so I simply misunderstood you. I kept wondering how I was supposed to get the commit into m2e-core when it wasn't merged into m2e-core-tests.

It's just 2 commits, so it should be easy enough to migrate.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/part/AbstractDependencyEditorPart.java
```diff
@@ -0,0 +1,35 @@
...
+ *      Patrick Ziegler - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.m2e.pde.ui.target.editor.internal.part;
+
+import org.eclipse.m2e.pde.ui.target.editor.internal.model.TargetDependencyModel;
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Layout;
+
+public abstract class AbstractDependencyEditorPart {
```
HannesWell, line 20: This super-class is only used to store fields that both sub-classes have. AFAICT it is never used as type and its only method is only used once. Therefore I'm not sure this is really necessary.

HannesWell (commented):
Its better, but I have a few more suggestions. :)

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/part/DependencyEditorViewerPart.java
```diff
@@ -0,0 +1,213 @@
...
+		});
+
+		for (TableColumn tableColumn : viewer.getTable().getColumns()) {
+			tableColumn.pack();
+		}
+	}
+
+	private TableViewerColumn newColumn(int weight, String text, Function<MavenTargetDependency, String> getter) {
+		TableViewerColumn tableColumn = new TableViewerColumn(viewer, SWT.NONE);
+		tableColumn.getColumn().addSelectionListener(new TableColumnSortListener(viewer, getter));
```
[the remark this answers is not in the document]
ptziegler: I would go even one step further and use a Comparator over the MavenTargetDependency, instead of comparing the strings directly.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/part/AbstractDependencyEditorPart.java
ptziegler: That one was bugging me as well. I initially created the class, because I need to expose the parent composite, in order to make sure that the table viewer is only as large as the height of all buttons combined.
However, I achieved the same effect by moving this information to the model, meaning this class can go for good. :)

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/part/DependencyEditorViewerPart.java
```diff
@@ -0,0 +1,267 @@
...
+				viewer.setComparator(new ViewerComparator() {
+					@Override
+					public int compare(Viewer viewer, Object e1, Object e2) {
+						return comparator.compare((MavenTargetDependency) e1, (MavenTargetDependency) e2);
+					}
+				});
+				viewer.refresh();
+			}
+		};
+	}
```
HannesWell, line 245: Instead of a anonymous class you could simply use `SelectionListener.widgetSelectedAdapter()`
```suggestion
	private SelectionListener newSortAdapter(Function<MavenTargetDependency, String> getter) {
		return SelectionListener.widgetSelectedAdapter(e -> {
			Comparator<MavenTargetDependency> comparator;

			// When we don't currently sort by this column, default to ascending order
			if (viewer.getTable().getSortColumn() != e.widget || viewer.getTable().getSortDirection() == SWT.UP) {
				viewer.getTable().setSortDirection(SWT.DOWN);
				comparator = Comparator.comparing(getter);
			} else {
				viewer.getTable().setSortDirection(SWT.UP);
				comparator = Comparator.comparing(getter).reversed();
			}

			viewer.getTable().setSortColumn((TableColumn) e.widget);
			viewer.setComparator(new ViewerComparator() {
				@Override
				public int compare(Viewer viewer, Object e1, Object e2) {
					return comparator.compare((MavenTargetDependency) e1, (MavenTargetDependency) e2);
				}
			});
			viewer.refresh();
		});
	}
```

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/part/DependencyEditorMenuPart.java
```diff
@@ -0,0 +1,94 @@
...
+	 * button is clicked, a model change has to be performed.
+	 * 
+	 * @param button      One of the buttons of the menu part.
+	 * @param onSelection The method that is called when the button is clicked.
+	 */
+	private void addSelectionChangedListener(Button button, Runnable onSelection) {
+		button.addSelectionListener(SelectionListener.widgetSelectedAdapter(event -> {
+			onSelection.run();
+		}));
+	}
```
HannesWell, line 93: The fields are only used in the `createContent()` method and can therefore be inlined in that method.
Furthermore, setting the selection-listener can also be done in the `newButton() `method.
```suggestion
public class DependencyEditorMenuPart {
	private final Composite composite;
	private final TargetDependencyModel model;

	public DependencyEditorMenuPart(Composite parent, TargetDependencyModel model) {
		this.composite = new Composite(parent, SWT.NONE);
		this.composite.setLayout(new GridLayout());
		this.composite.setLayoutData(new BorderData(SWT.RIGHT));
		this.model = model;
		createContent();
	}

	private void createContent() {
		ISideEffectFactory factory = WidgetSideEffects.createFactory(composite);

		newButton(Messages.MavenTargetDependencyEditor_7, Messages.MavenTargetDependencyEditor_12, model::add);

		Button removeButton = newButton(Messages.MavenTargetDependencyEditor_8, Messages.MavenTargetDependencyEditor_13,
				model::remove);
		factory.create(() -> removeButton.setEnabled(model.canRemove()));

		Button updateButton = newButton(Messages.MavenTargetDependencyEditor_9, Messages.MavenTargetDependencyEditor_14,
				model::update);
		factory.create(() -> updateButton.setEnabled(model.canUpdate()));

		Button undoButton = newButton(Messages.MavenTargetDependencyEditor_10, null, model.getHistory()::undo);
		factory.create(() -> undoButton.setEnabled(model.getHistory().canUndo()));

		Button redoButton = newButton(Messages.MavenTargetDependencyEditor_11, null, model.getHistory()::redo);
		factory.create(() -> redoButton.setEnabled(model.getHistory().canRedo()));

		// The table viewer should have the same height as all buttons combined
		model.setHeightHint(composite.computeSize(SWT.DEFAULT, SWT.DEFAULT).y);
	}

	/**
	 * Creates a new push-button with the given text, tool-tip and selection-change
	 * listener. Whenever the button is clicked, a model change has to be performed.
	 * 
	 * @param text        the text displayed on the button
	 * @param tooltip     the tooltip displayed for the button
	 * @param onSelection the method that is called when the button is clicked.
	 */
	private Button newButton(String text, String tooltip, Runnable onSelection) {
		Button button = new Button(composite, SWT.PUSH);
		button.setLayoutData(new GridData(SWT.FILL, SWT.CENTER, false, false));
		button.setText(text);
		button.setToolTipText(tooltip);
		button.addSelectionListener(SelectionListener.widgetSelectedAdapter(event -> onSelection.run()));
		return button;
	}
```

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/part/AbstractDependencyEditorPart.java
HannesWell: Sounds good. 👍🏽

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/part/DependencyEditorViewerPart.java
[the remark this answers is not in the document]
HannesWell: Even better. :)

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/part/DependencyEditorViewerPart.java
```diff
@@ -0,0 +1,261 @@
...
+		tableColumn.setEditingSupport(newEditingSupport(getter, setter, TextCellEditor::new));
+
+		getter = MavenTargetDependency::getVersion;
+		setter = MavenTargetDependency::setVersion;
+		tableColumn = newColumn(250, Messages.MavenTargetDependencyEditor_3, getter);
+		tableColumn.setEditingSupport(newEditingSupport(getter, setter, TextCellEditor::new));
+
+		getter = MavenTargetDependency::getClassifier;
+		setter = MavenTargetDependency::setClassifier;
+		tableColumn = newColumn(125, Messages.MavenTargetDependencyEditor_1, getter);
```
HannesWell, line 117:
```suggestion
		tableColumn = newColumn(125, Messages.MavenTargetDependencyEditor_4, getter);
```

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/part/DependencyEditorViewerPart.java
```diff
@@ -0,0 +1,261 @@
...
+		tableColumn.setEditingSupport(newEditingSupport(getter, setter, TextCellEditor::new));
+
+		getter = MavenTargetDependency::getType;
+		tableColumn = newColumn(125, Messages.MavenTargetDependencyEditor_1, getter);
```
HannesWell, line 121:
```suggestion
		tableColumn = newColumn(125, Messages.MavenTargetDependencyEditor_5, getter);
```

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/messages.properties
```diff
@@ -13,6 +13,14 @@ MavenTargetDependencyEditor_3=Version
 MavenTargetDependencyEditor_4=Classifier
 MavenTargetDependencyEditor_5=Type
 MavenTargetDependencyEditor_6=New Dependency
```
HannesWell, line 15:
```suggestion
```
Looks like this is not used anymore.

HannesWell (changes requested):
It is becoming better and better. :) Still I have a few minor more comments on the code and some labels are incorrect.

To the technical part (sorry for coming to this so late).
Empty or incomplete table rows/entries should not be persisted because they lead to immediate resolution errors when the editor is closed. Either the editor simply ignores them or should not be able to close. Ignoring empty rows, I think would be ok, for incomplete rows it would probably be better to highlight them (maybe the background color?) to make the user aware that something is missing, ideally with a corresponding hint.
Actually it would also nice to check if the specified artifact can be resolved, but we can leave that for a follow up PR, if you want.

And instead of sorting the rows by a specific column, in most cases it is probably more interesting to sort them first by group, then artifact and then version (you could simply chain corresponding `Comparators`). This way entries of the same group respectively artifacts are grouped together.
This is also the order in which they are persisted in the target-file and I think it would be nice if one can restore that display order from a per column ordering.
Furthermore it would be nice if newly added or modified rows are automatically change their position according to the current sorting.

HannesWell:
Furthermore eventually this PR should be submitted as one commit. if you want you can squash all commits into one giving it a meaning full header and message body that covers the result (no need to capture each step in that) or we will squash-and-merge commit it and try to extract a good message.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/messages.properties
HannesWell: The corresponding entry in the `messages.properties` should be removed too.

laeubi:
> Empty or incomplete table rows/entries should not be persisted 

Empty rows should be ignored, but I won't put much requirements on incomplete ones as we did not enforced this before and as the user can edit the xml anyways it does not make much sense to be over pendantic here.

> Actually it would also nice to check if the specified artifact can be resolved, 

I don't think this is a good idea as it will result in massive overhead and might produce a lot of garbage in the local repository, resolving is actually part of the target resolution and not a concern of the editor.

ptziegler:
> Empty or incomplete table rows/entries should not be persisted because they lead to immediate resolution errors when the editor is closed. Either the editor simply ignores them or should not be able to close. Ignoring empty rows, I think would be ok, for incomplete rows it would probably be better to highlight them (maybe the background color?) to make the user aware that something is missing, ideally with a corresponding hint.

Just "ignoring" them feels wrong. And I think it can lead to unexpected behavior when you, for example, simply forgot to add the version. Or in a case where we would check whether the artifact exists, the user made a typo or used the wrong type. I'm leaning towards disabling the "Finish" button + Error message, as you suggested.

> Actually it would also nice to check if the specified artifact can be resolved, but we can leave that for a follow up PR, if you want.

Depending on how complicated this is, I can probably do it as part of the validation. Though I'm a little bit concerned about the performance implications, if I talk to Maven Central, every time I add/edit a Maven artifact. But that can probably only be seen when it's actually implemented.

> And instead of sorting the rows by a specific column, in most cases it is probably more interesting to sort them first by group, then artifact and then version (you could simply chain corresponding Comparators). This way entries of the same group respectively artifacts are grouped together.
This is also the order in which they are persisted in the target-file and I think it would be nice if one can restore that display order from a per column ordering.
Furthermore it would be nice if newly added or modified rows are automatically change their position according to the current sorting.

So should it then still be possible to sort by Artifact Id/Version or only by Group Id? In that case I would simply sort by MavenTargetDependency#getKey(), which I think is what's used when the target location is saved.

> Furthermore eventually this PR should be submitted as one commit.

Absolutely. I'll probably use my first commit message with a little bit more detail. But at least while developing, I like to have an overview over the individual changes.

HannesWell:
> > Empty or incomplete table rows/entries should not be persisted because they lead to immediate resolution errors when the editor is closed. Either the editor simply ignores them or should not be able to close. Ignoring empty rows, I think would be ok, for incomplete rows it would probably be better to highlight them (maybe the background color?) to make the user aware that something is missing, ideally with a corresponding hint.
> 
> Just "ignoring" them feels wrong. And I think it can lead to unexpected behavior when you, for example, simply forgot to add the version. Or in a case where we would check whether the artifact exists, the user made a typo or used the wrong type. I'm leaning towards disabling the "Finish" button + Error message, as you suggested.

Yes, the ignore part was from my initial version of the message, where I just covered empty row. As Christoph said, I think it is fine to ignore completely empty ones, but for incomplete ones an error+ disabled finish is better.

> Empty rows should be ignored, but I won't put much requirements on incomplete ones as we did not enforced this before and as the user can edit the xml anyways it does not make much sense to be over pendantic here.

Even if that did not exist why not improve that now? IMHO that would fit quite well and should not be too much work.
Or do you think one creates only partial entries in the editor and then completes them in the text directly? I don't think we have to support that since, one can then just create the entry completely by hand.

> > Actually it would also nice to check if the specified artifact can be resolved,
> 
> I don't think this is a good idea as it will result in massive overhead and might produce a lot of garbage in the local repository, resolving is actually part of the target resolution and not a concern of the editor.

Yes a full resolution would indeed be out of scope. Do you know if it is possible to only check if a artifact exists in the local or any of the available remote repositories, without resolving/downloading it? That would at least harm the performance much less than a full resolution with a potential download.
If yes at least a warning could be helpful for users ('forbidding' it in the UI could be too much).
But as said before we can leave that part for a subsequent change and now only check for completeness of all required data.



> > And instead of sorting the rows by a specific column, in most cases it is probably more interesting to sort them first by group, then artifact and then version (you could simply chain corresponding Comparators). This way entries of the same group respectively artifacts are grouped together.
> > This is also the order in which they are persisted in the target-file and I think it would be nice if one can restore that display order from a per column ordering.
> > Furthermore it would be nice if newly added or modified rows are automatically change their position according to the current sorting.
> 
> So should it then still be possible to sort by Artifact Id/Version or only by Group Id? 

That's a good question. I mean the sorting in general has no real meaning and it just changes the displayed order in this editor. So in most cases it is probably irrelevant. I can think it could be useful if you for example search for a certain artifact and don't know the group (or similar). So I would not object if we keep the per column sorting, but I also don't think it is critical to have. So I leave that decision to you :)

> In that case I would simply sort by MavenTargetDependency#getKey(), which I think is what's used when the target location is saved.

If that isn't a string that's fine 👍🏽 If it is a string we should check if the result is as desired with the separator.


> > Furthermore eventually this PR should be submitted as one commit.
> 
> Absolutely. I'll probably use my first commit message with a little bit more detail. But at least while developing, I like to have an overview over the individual changes.

Sure that's fine and reasonable. Just wanted to make you aware. :)

laeubi:
> Yes a full resolution would indeed be out of scope. Do you know if it is possible to only check if a artifact exists in the local or any of the available remote repositories, without resolving/downloading it?

You can try issuing a version request: 
https://github.com/apache/maven-resolver/blob/master/maven-resolver-demos/maven-resolver-demo-snippets/src/main/java/org/apache/maven/resolver/examples/FindAvailableVersions.java

But this just keep in mind that this ca open a can of worms, just assume you specify additional repositories, these must then re-resolve all items (as now it might be there), then what happens if it does not work because of wrong maven settings so we need to probably offer a link to open/change the maven settings...

What could work quite out of the box would be to have a `validate` button in the dialog, then construct a MavenTargetLocation from as if one would return it from the editor and then perform a resolve() on this location, then one can fetch the status and if there are problems show these to the user.

But lets finish this before ;-)

ptziegler:
> But this just keep in mind that this ca open a can of worms, just assume you specify additional repositories, these must then re-resolve all items (as now it might be there), then what happens if it does not work because of wrong maven settings so we need to probably offer a link to open/change the maven settings...

That's not the only issue. The table is initialized with the original target location. Which also means it would use the original repositories. If the dependency can't be found on e.g. Maven Central, you would then have to construct some meta target location containing both the original and newly added (or removed) repositories, pass it to the do the whole thing again and who knows what else...
In the end, I don't think it's really worth all the trouble. Just checking for non-empty entries is probably more than sufficient. The rest can be done by the editor.

> What could work quite out of the box would be to have a validate button in the dialog, then construct a MavenTargetLocation from as if one would return it from the editor and then perform a resolve() on this location, then one can fetch the status and if there are problems show these to the user.

That sounds a lot like what the target editor is doing...

ptziegler:
I've made some improvements based on your suggestions.

- A simple validation checks, whether all required field are set. If not, an error message is shown and the "Finish" button is disabled. I've decided to also consider "empty" dependencies, simply to avoid an inconsistent behavior. For example,  creating a blank dependency uses "jar" as default packaging. Is this now an empty dependency? What if I remove the type? What if the user specifically sets it to "jar"? In the end, removing entries is very easy, so we might as well require it.
- In case the table is sorted, new items are inserted into the correct positions, rather than at the end of the list.
- Sorting by a column now also considers the remaining columns. The issue is that if one sorts by e.g. group id, then all dependencies with the same group id are sorted, for all intents and purposes, at random. The key is used to create a proper ordering. If one wants to order the dependencies in the same order they are stored in the XML file, you simply have to sort by group id.
- The composite containing the buttons has become so simple, that I decided to merge it into the table composite. As well as some general restructuring.

image

laeubi:
@ptziegler looks great and thanks for all the effort, just a really minor remark:
The button composite has a little spacing on the top, it would be good to align in with the top of the table:
grafik

ptziegler:
> The button composite has a little spacing on the top, it would be good to align in with the top of the table

Simple enough. I just had to set the margin to 0. It didn't really bother me, given that both the top and bottom had the same margin, but I don't mind either way.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/TargetDependencyModel.java
```diff
@@ -0,0 +1,339 @@
...
+	public void remove() {
+		if (!canRemove()) {
+			LOGGER.error("The remove operation is currently not possible! Ignore...");
+			return;
+		}
+
+		List<MavenTargetDependency> oldCurrentSelection = getCurrentSelection();
+		List<MavenTargetDependency> oldTargetDependencies = getTargetDependencies();
+
+		List<MavenTargetDependency> newCurrentSelection = new ArrayList<>();
```
HannesWell, line 187:
```suggestion
		List<MavenTargetDependency> newCurrentSelection;
```
Assigned value is not used.
If you want you could consider to initialize the variable with an empty list, reverse the condition of the if-block and remove the then empty else block.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/TargetDependencyModel.java
```diff
@@ -0,0 +1,339 @@
...
+			}
+			if (dependency.getArtifactId() == null || dependency.getArtifactId().isBlank()) {
+				return false;
+			}
+			if (dependency.getVersion() == null || dependency.getVersion().isBlank()) {
+				return false;
+			}
+		}
+
+		return !currentSelection.isEmpty();
```
HannesWell, line 289:
```suggestion
		Stream<String> allFields = currentSelection.stream().mapMulti((dependency, downStream) -> {
			downStream.accept(dependency.getGroupId());
			downStream.accept(dependency.getArtifactId());
			downStream.accept(dependency.getVersion());
		});
		return allFields.noneMatch(StringUtils::isBlank) && !currentSelection.isEmpty();
```

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/TargetDependencyModel.java
```diff
@@ -0,0 +1,339 @@
...
+			copy.add(dependency.copy());
+		}
+
+		return new ArrayList<>(copy);
+	}
+
+	public static boolean hasErrors(MavenTargetDependency dependency) {
+		return dependency.getGroupId().isBlank() || dependency.getArtifactId().isBlank()
+				|| dependency.getVersion().isBlank() || dependency.getType().isBlank();
+	}
```
HannesWell, line 338:
```suggestion
	public void check() {
		Stream<String> allFields = targetDependencies.stream().mapMulti((dependency, downStream) -> {
			downStream.accept(dependency.getGroupId());
			downStream.accept(dependency.getArtifactId());
			downStream.accept(dependency.getVersion());
			downStream.accept(dependency.getType());
		});
		hasErrors.setValue(allFields.anyMatch(StringUtils::isBlank));
	}

	private static List<MavenTargetDependency> deepClone(List<MavenTargetDependency> dependencies) {
		List<MavenTargetDependency> copy = new ArrayList<>();

		for (MavenTargetDependency dependency : dependencies) {
			copy.add(dependency.copy());
		}

		return new ArrayList<>(copy);
	}
```

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/TargetDependencyModel.java
```diff
@@ -0,0 +1,339 @@
...
+ * Contributors:
+ *      Patrick Ziegler - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.m2e.pde.ui.target.editor.internal;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.function.BiConsumer;
```
HannesWell, line 19: With the changes below the  imports have to be adjusted.
```suggestion
import java.util.function.BiConsumer;
import java.util.stream.Stream;

import org.apache.maven.shared.utils.StringUtils;
```

HannesWell (approved):
Looks great now! Your suggestions are perfectly fine.

Just a few minor suggestion below, but they are totally optional.

I also noticed, when playing around with it a bit that the minimum sizes of the columns as well as of the entire editor could be improved.
The column width, at least up until a certain extend, could be fitted to the content.
And it should not be possible to make the entire dialog smaller than the (radio) buttons in the lower part.
But I'm fine with doing that later.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/TargetDependencyModel.java
ptziegler: That looks a lot cleaner. Thanks for the feedback!

ptziegler:
> I also noticed, when playing around with it a bit that the minimum sizes of the columns as well as of the entire editor could be improved.
The column width, at least up until a certain extend, could be fitted to the content.

Table widths are always a funny thing to deal with...  How it currently works is that if you open the editor for the first time, the widths are adjusted so that none of the entries are obstructed. Though this might still happen, if you create new elements. I've adjusted it so that the column width is also adjusted, when the input is modified.

> And it should not be possible to make the entire dialog smaller than the (radio) buttons in the lower part.
But I'm fine with doing that later.

There I would go all the way. I calculate the preferred size of the shell when opening the editor and set that as the minimum size. You shouldn't be able to make the shell so small that you can't modify the target definition anymore.

laeubi:
I think we should just set a minimum size for the dialog, and then probably use the table column layout? That reminds me of that I always wanted to contribute a more advanced layout for table columns than whats currently available in SWT...

ptziegler:
> I think we should just set a minimum size for the dialog, and then probably use the table column layout?

But the wizard dialog is created by the target editor, so we don't have direct access to it. You also don't have a minimum size in e.g. the p2 wizard. Perhaps this is something that should be brought up to the PDE team?

laeubi:
One should be able to call IWizard#getContainer().getShell() and might resize the shell if it is too small when there is a request to show the page, but I have not tried that out.

laeubi:
> Perhaps this is something that should be brought up to the PDE team?

I think this would better be JFace/Platform UI, it would be good if a IWizardPage would have a getMinimumPageSize and getMaximumPageSize (the default can just use getControl().computeSize(-1,-1))

ptziegler:
> I think this would better be JFace/Platform UI, it would be good if a IWizardPage would have a getMinimumPageSize and getMaximumPageSize (the default can just use getControl().computeSize(-1,-1))

I've created https://github.com/eclipse-platform/eclipse.platform.ui/issues/516. It's probably for the best to postpone this topic until there's a response from the Platform UI team.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/internal/DependencyTable.java
```diff
@@ -0,0 +1,328 @@
...
+
+		getter = MavenTargetDependency::getClassifier;
+		setter = MavenTargetDependency::setClassifier;
+		tableColumn = newColumn(layout, Messages.MavenTargetDependencyEditor_4, getter, false);
+		tableColumn.setEditingSupport(newEditingSupport(getter, setter, TextCellEditor::new));
+
+		getter = MavenTargetDependency::getType;
+		tableColumn = newColumn(layout, Messages.MavenTargetDependencyEditor_5, getter, true);
+		tableColumn.setEditingSupport(newTypeEditingSupport());
+		tableColumn.getColumn();
```
HannesWell, line 122: I think this can be removed.
```suggestion
```

HannesWell (approved):
> Table widths are always a funny thing to deal with... 

UI is a funny thing and tables are especially funny, yes. :D


> > I think this would better be JFace/Platform UI, it would be good if a IWizardPage would have a getMinimumPageSize and getMaximumPageSize (the default can just use getControl().computeSize(-1,-1))
> 
> I've created [eclipse-platform/eclipse.platform.ui#516](https://github.com/eclipse-platform/eclipse.platform.ui/issues/516). It's probably for the best to postpone this topic until there's a response from the Platform UI team.

Yes, lets delay that.

I'm fine with this change as it is right now.

@laeubi are you as well?
@mickaelistria can you tell if we still have to create CQs for contributions over 1000LOC? Because I at least know that the term CQ has been deprecated, but I don't know if we have to create something else instead?

HannesWell:
> @mickaelistria can you tell if we still have to create CQs for contributions over 1000LOC? Because I at least know that the term CQ has been deprecated, but I don't know if we have to create something else instead?

I checked the [Eclipse projects handbook](https://www.eclipse.org/projects/handbook/#ip-project-content) and reviews for contributions over 1000LOC of non-Committers are still necessary.
Created https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/6372. Once that has succeeded, I plan to merge this PR.

mickaelistria:
I see a few hundreds lines that are "just" moving code or extracting methods, so from IP perspective, those lines are not really counting. If we take the new production, the new content for which IP maters, we're under 1000 lines. So I think we could skip the CQ and this would still be fine.

HannesWell:
> I see a few hundreds lines that are "just" moving code or extracting methods, so from IP perspective, those lines are not really counting. If we take the new production, the new content for which IP maters, we're under 1000 lines. So I think we could skip the CQ and this would still be fine.

Alright. Thanks for that clarification. Additionally each new file contains 12 lines of copyright/license header and do the imports count as well? However, then lets merge this. I will close the IP-issue for that reason.

@ptziegler thanks a lot for this great contribution.

PR #1159, opened 2022-12-29
Show Maven LifecycleMappings in view for active POM-Editor

HannesWell:
Follow up on #1155 (@Bananeweizen FYI) to show the lifecycle mappings of a Maven project when the POM-Editor is the active selection.
Until today I did not know I have to select the project in the Package/Project Explorer tree in order to populate the view and maybe others will have the same problem. Showing the lifecycle mappings for a active Pom-Editor seems suitable for me.

Nevertheless this probably should be improved to only populate the view if a pom-file is openend in an editor and not on every file.

### DIFF
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/views/MavenLifecycleMappingsView.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/views/MavenLifecycleMappingsView.java
@@ -13,14 +13,17 @@
 
 package org.eclipse.m2e.core.ui.internal.views;
 
+import java.util.Optional;
+
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.runtime.Adapters;
 import org.eclipse.jface.viewers.ISelection;
 import org.eclipse.jface.viewers.IStructuredSelection;
 import org.eclipse.swt.widgets.Composite;
-import org.eclipse.ui.ISelectionListener;
+import org.eclipse.ui.IEditorPart;
 import org.eclipse.ui.IViewSite;
-import org.eclipse.ui.IWorkbenchPart;
+import org.eclipse.ui.IWorkbenchPage;
+import org.eclipse.ui.IWorkbenchPartSite;
 import org.eclipse.ui.PartInitException;
 import org.eclipse.ui.part.ViewPart;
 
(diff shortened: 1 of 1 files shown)


org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/views/MavenLifecycleMappingsView.java
```diff
@@ -75,6 +65,10 @@ private void handleSelectionChanged(ISelection selection) {
     Object element;
     if(selection instanceof IStructuredSelection structuredSelection) {
       element = structuredSelection.getFirstElement();
+    } else if(selection instanceof ITextSelection) { // A P
+      element = Optional.ofNullable(PlatformUI.getWorkbench().getActiveWorkbenchWindow())
+          .map(IWorkbenchWindow::getActivePage).map(IWorkbenchPage::getActiveEditor).map(IEditorPart::getEditorInput)
```
Bananeweizen, line 70: I'm in favor of checking in JDT formatter settings to "never join wrapped lines". That would allow writing it as
```
stream
.map
.map
.orElse
```
Opinions?

Bananeweizen:
It's not necessary to restrict this to POM editors only, IMO. The view will not do anything as long as the old and new projects are the same. Therefore switching around between editors is cheap and the view still shows the correct content.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/views/MavenLifecycleMappingsView.java
HannesWell: Personally I don't think it would always be good. Sometimes it is but sometimes is actually better if the lines are joined. And I think it would be bad if one has to join them manually.
What I usually do if I want such multi-line stream-pipelines is to write a (usually empty) comment behind the last segment to prevent the formatter from joining the line.
That is indeed not idea, but never joining lines isn't IMHO either.

HannesWell:
> It's not necessary to restrict this to POM editors only, IMO. The view will not do anything as long as the old and new projects are the same. Therefore switching around between editors is cheap and the view still shows the correct content.

Indeed. Thanks for pointing that out. 
I now changed it to just set the project of the active editor if the selection is not structured.
@laeubi you wrote that initially. Any remarks?

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/views/MavenLifecycleMappingsView.java
laeubi: I must confess that I don't get what the editor has to do with that. Also should not the view already has a reference to the Workbench/Window?

HannesWell: > Also should not the view already has a reference to the Workbench/Window?

Should be done now.

laeubi (approved):
Looks good.

Issue #1150, opened 2022-12-20
Resources handling is broken when there is a .mvn folder


tmortagne:
When there is a .mvn folder (it does not even need to contain anything) at the root of the imported module or multi modules tree, resources files are not copied to `target/classes` anymore.

To reproduce:
* download and extract 
[project-with-resources.zip](https://github.com/eclipse-m2e/m2e-core/files/10269407/project-with-resources.zip)
* import it as Maven project

-> among other things, a `target/classes` is generated, but it does not contain the expected resource file named `file.txt`

* delete the `.mvn` folder
* right click on the project -> `Maven` -> `Update Project...` -> `OK`

-> the file `target/classes/file.txt` appear

I did not have this problem in Eclipse 2022-06 (I had other problems with 2022-09 like https://github.com/eclipse-m2e/m2e-core/issues/1005, so I did not really use it much).

Eclipse version: 2022-12
Maven Integration for Eclipse version: 2.1.2.20221130-2239

HannesWell:
Thanks for the detailed report. I can reproduce the issue with the current master.
It even sufficient when you just clean the project (and auto-build it) after having deleted the `.mvn` folder. Adding the .mvn folder back prevents the copying again.
I'm about to investigate it.

HannesWell:
Was able to record the following exception that is unfortunately swallowed at the moment (I think that should change and at least end up in the Error-log if not even causing a warning or error):
```
org.apache.maven.plugin.PluginExecutionException: Execution default-resources of goal org.apache.maven.plugins:maven-resources-plugin:2.6:resources failed: Unable to load the mojo 'resources' (or one of its required components) from the plugin 'org.apache.maven.plugins:maven-resources-plugin:2.6'
	at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:156)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.executeMojo(MavenExecutionContext.java:331)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.lambda$0(MavenExecutionContext.java:290)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.executeBare(MavenExecutionContext.java:364)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.execute(MavenExecutionContext.java:274)
[...]

PR #1160, opened 2022-12-30
Propagate Mojo-lookup errors and create test-case for #1150

HannesWell:
This adds a test case to reproduce #1150. It succeeds if you rename the `.mvn` folder of the project used in the test.

The reason for the failure is that in `MavenExecutionContext.executeMojo()` the looked up `BuildPluginManager` failed to look-up the the mojo for the `maven-resources-plugin:resources` goal.
The cause is probably in the `IComponentLookup` created for that multi-module project, since without a `.mvn`-folder a un-rooted look-up is created that succeeds. My first guess is that some settings are missing, but on my first look I didn't find anything.

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
@@ -317,7 +317,8 @@ public class MavenExecutionContext implements IMavenExecutionContext {
     }
   }
 
-  private static void executeMojo(MavenSession session, MojoExecution execution, IComponentLookup lookup) {
+  private static void executeMojo(MavenSession session, MojoExecution execution, IComponentLookup lookup)
+      throws CoreException {
     Map<MavenProject, Set<Artifact>> artifacts = new HashMap<>();
     Map<MavenProject, MavenProjectMutableState> snapshots = new HashMap<>();
     for(MavenProject project : session.getProjects()) {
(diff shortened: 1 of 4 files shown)


HannesWell:
That reminds me that I wanted to create a little guide about 'How to contribute a test-case'.

HannesWell:
From the tests it looks like we would have seen that earlier if the mojo execution errors were not swallowed.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
```diff
@@ -331,6 +332,7 @@ private static void executeMojo(MavenSession session, MojoExecution execution, I
       lookup.lookup(BuildPluginManager.class).executeMojo(session, execution);
     } catch(Exception ex) {
       session.getResult().addException(ex);
+      throw new CoreException(Status.error("Failed to execute mojo", ex));
```
laeubi, line 335: Thats not the place to handle this, the caller has to check the result and act accordingly (e.g. m2e has helper classes to create error markers from results).

laeubi:
> The cause is probably in the `IComponentLookup` created for that multi-module project, since without a `.mvn`-folder a un-rooted look-up is created that succeeds. My first guess is that some settings are missing, but on my first look I didn't find anything.

I don't think so. This is most probably caused by m2e not using the right context fro creation of the `MojoExecution` as mentioned in other PRs because it is invalid to use the global context (`MavenImpl`). I'll try to see if I find where this originates.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
laeubi: I checked this more deeply in m2e an it seems this actually is already handled at a higher level, even though in some places we return the result. It seems strange here to have both (throwing an exception and adding it to the result). So I think simply throwing an exception would be enough

laeubi:
merged as part of
- https://github.com/eclipse-m2e/m2e-core/pull/1161

PR #1161, opened 2022-12-31
Fix wrong realm used when executing mojos

laeubi:
Currently the MojoSetup is cached what do not seem working fine with different containers. This now changes the way to fetch a fresh configuration that takes the current context into account.

Based on 
- https://github.com/eclipse-m2e/m2e-core/pull/1160
This fixes 
- https://github.com/eclipse-m2e/m2e-core/issues/1150


laeubi:
As this fixes the previous failures and did not show any ne ones I think this is good to be merged.

HannesWell:
> As this fixes the previous failures and did not show any ne ones I think this is good to be merged.

Sorry I didn't have the time yet to review/undestand this in detail. But yes since all tests pass it is probably not too bad and 

> Currently the MojoSetup is cached what do not seem working fine with different containers. This now changes the way to fetch a fresh configuration that takes the current context into account.
> 

Can you explain where this is cached? In M2E or Maven itself?

Since this is a non obvious behavior, I think it would be good to have some comments in the code explaining why execution is cloned?!

PR #1167, opened 2023-01-02
Add Project Wizard Shortcuts

laeubi:
Currently important parts of m2e functionality are hidden in sub-menus or sub-groups. This makes it hard for users to discover them and it requires to much clicks to use them.

This adds two new short-cuts:
- For File > New > Maven Project
- For Navigator > New > Maven Module

### DIFF
--- a/org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF
@@ -41,7 +41,8 @@ Require-Bundle: org.eclipse.m2e.core;bundle-version="[2.0.0,3.0.0)",
  org.eclipse.ui.forms;bundle-version="3.4.1",
  org.eclipse.jface.text,
  org.eclipse.core.filebuffers,
- org.eclipse.ui
+ org.eclipse.ui,
+ org.eclipse.ui.navigator
 Import-Package: org.eclipse.compare.rangedifferencer,
  org.eclipse.ltk.core.refactoring,
  org.slf4j;version="[1.7.0,3.0.0)"
--- a/org.eclipse.m2e.core.ui/plugin.xml
+++ b/org.eclipse.m2e.core.ui/plugin.xml
@@ -730,4 +730,20 @@
 			</activeWhen>
 		</projectConfigurator>
 	</extension>
+	 <extension
+	       point="org.eclipse.ui.navigator.navigatorContent">
+	    <commonWizard
+           menuGroupId="org.eclipse.m2e"
+           type="new"
+           wizardId="org.eclipse.m2e.core.wizards.Maven2ModuleWizard">
+	       <enablement>
+				<or>
+					<and>
+						<instanceof value="org.eclipse.core.resources.IProject" />
+						<test property="org.eclipse.core.resources.projectNature" value="org.eclipse.m2e.core.maven2Nature" />
+					</and>
+				</or>
+			</enablement>
+	    </commonWizard>
+	 </extension>
 </plugin>
--- a/org.eclipse.m2e.jdt.ui/plugin.xml
+++ b/org.eclipse.m2e.jdt.ui/plugin.xml
@@ -344,4 +344,12 @@
             name="M2Eclipse">
       </quickFixProcessor>
    </extension>   
+	<extension point="org.eclipse.ui.perspectiveExtensions">
+	   <perspectiveExtension
+	         targetID="org.eclipse.jdt.ui.JavaPerspective">
+	      <newWizardShortcut
+	            id="org.eclipse.m2e.core.wizards.Maven2ProjectWizard">
+	      </newWizardShortcut>
+	   </perspectiveExtension>
+	</extension>
 </fragment>
--- a/org.eclipse.m2e.pde.ui/plugin.xml
+++ b/org.eclipse.m2e.pde.ui/plugin.xml
@@ -15,4 +15,12 @@
        </locationProvider>
        
     </extension>
+	<extension point="org.eclipse.ui.perspectiveExtensions">
+	   <perspectiveExtension
+	         targetID="org.eclipse.pde.ui.PDEPerspective">
+	      <newWizardShortcut
+	            id="org.eclipse.m2e.core.wizards.Maven2ProjectWizard">
+	      </newWizardShortcut>
+	   </perspectiveExtension>
+	</extension>
 </plugin>


mickaelistria:
Keep in mind that at the same time, some people (maybe even yourself a few months ago ;) complain that the menus are too crowded. So it's not all black and white. This change will cause some integrators to be annoyed because they do not want Maven as top-level element in their UI and and the proposal forcing it for all perspectives.
Do you have actual feedback from users who have failed at creating Maven projects?

The part about navigator content seems correct as there is a filter on the context;
But for the global menu, this is IMO an anti-pattern to force this element in all perspectives.

laeubi:
I think maven is an important enough topic for developers to make it more visible to the user (at least no one ever complains that IntelliJ has more visible maven support). 

I also have opened a discussion here about having a "maven perspective" (or even EPP):
- https://github.com/eclipse-m2e/m2e-core/discussions/1016

but it has not got much attention / agreement, so actually we can not really narrow this much (maybe one can add it to the java/plugin perspective only), so the global context seems valid here?

mickaelistria:
I extrapolate from the absence of positive answer to "Do you have actual feedback from users who have failed at creating Maven projects?" that you are most likely trying to fix a problem that doesn't exist...

> I think maven is an important enough topic for developers to make it more visible to the user

Not all perspectives are dedicated to development. Some products may include m2e to get some Maven support just to have it but don't need to have it as more visible than Gradle or other forms of projects.
If I'm making an IDE dedicated to Minecraft modding for teenagers, I would probably include m2e to get build and dep management, but I really wouldn't like to show kids a "Maven Project" as top-level elements. This proposal goes against that and as a result make m2e less nice to includedin RCP apps.

> at least no one ever complains that IntelliJ has more visible maven support

IntelliJ is not really used an an RCP Platform like Eclipse Platform. About no-one builds their own IJ-based IDE, so it's the kind of problem that IJ is actually not even aware of, and thus not worth being taken as example.

laeubi:
> I extrapolate from the absence of positive answer to "Do you have actual feedback from users who have failed at creating Maven projects?" that you are most likely trying to fix a problem that doesn't exist...

Do you have any actual feedback from users that don't like to have maven more visible? ;-)

At least I'm a user of m2e and I would like to have better maven support, and also at EclipseCon2022 poor user-experience / support of Eclipse was complained so I have no hard "facts" and I think no one else really has.

> Not all perspectives are dedicated to development. 

Why are they included in an "Integrated **Development** Editor" (IDE) then? And why we do offer "New Project" at all then if it is completely useless?

> Some products may include m2e to get some Maven support just to have it but don't need to have it as more visible than Gradle or other forms of projects.

Having Gradel projects being created more easily when Gradle support is installed would be fair enough (for me) if I never need that I would just not install/use gradle support...

> If I'm making an IDE dedicated to Minecraft modding for teenagers, I would probably include m2e to get build and dep management

In such a case, one won't simply not include the maven UI bundle --> no problem

> IntelliJ is not really used an an RCP Platform like Eclipse Platform. About no-one builds their own IJ-based IDE, so it's the kind of problem that IJ is actually not even aware of, and thus not worth being taken as example.

An RCP application is not affected by this, because m2e is most likely never useful for a custom RCP application and even can only work with e3 legacy mode that already pulls in a lot of "unwanted" stuff.

mickaelistria:
> Do you have any actual feedback from users that don't like to have maven more visible? ;-)

I've spent 15 years developing Eclipse plugins and hearing users that the find that the menus are too crowded about once every 2 week without interruption.

> Having Gradel projects being created more easily when Gradle support is installed would be fair enough (for me) if I never need that I would just not install/use gradle support...

So why not having all new Project Wizards in the menu, for all perspective then?

> In such a case, one won't simply not include the maven UI bundle --> no problem

One may need m2e UI from time to time without making it super-visible everywhere.

> An RCP application is not affected by this, because m2e is most likely never useful for a custom RCP application and even can only work with e3 legacy mode that already pulls in a lot of "unwanted" stuff.

That is plain wrong. I've developed multiple RCP applications that were domain-specific studios and workbench; and one cool thing with Eclipse Platform is that it allows enough customization without too much PITA to hide the "technical" items and allow to focus on the domain. I imagine an application such as Bonita which is about drawing boxes and arrows but which includes m2e UI for some occasional development tasks wouldn't really enjoy to see it's whole domain-specific UX damaged by an extra m2e entry they cannot remove.
Giving such domain-specific apps choice between "forcing a UI element everywhere" vs "not installing Maven support" is hurting Eclipse as a Platform.

laeubi:
> I've spent 15 years developing Eclipse plugins and hearing users that the find that the menus are too crowded about once every 2 week without interruption.

Then all those people should simply participate and make adjustments, sharing ideas and so on... Should mean at least one contribution from a new user every two weeks that do not sounds that bad. If not that's exactly this kind of "most of others" arguments I'm hearing everywhere when trying to make their own arguments more strong, theres even a variation of "this will benefit many users" ... In the end no one measures this and all these "others" seem to be a silent-invisible-mass.

> So why not having all new Project Wizards in the menu, for all perspective then?

If there is a perspective and one can filter some out because they are not important than this is fine, but as you already argued perspectives are not what "everyone" wants, but in general this would be useful, we currently force the user to do too many click,s and to "know" to many things where important parts are hidden.

A very good (or actually bad...) is the `Java` versus `Plugin-Devlopment` perspective that actually share so much in common and have that much overlap but still differed in some aspects and shows that the "filter on perspective" is actually only working if you have really distinct tasks to solve that never overlap.

So for an empty workspace, showing *all* options for new projects would actually suitable, after that I might hide *some* (e.g. not used in the past two month, not a single project of that kind in the workspace, whatever measure) ...

> One may need m2e UI from time to time without making it super-visible everywhere.

So should we optimize m2e for cases where people will only want "some" m2e and make life harder for those users that actually want to use m2e? This does not sound correct.

> Giving such domain-specific apps choice between "forcing a UI element everywhere" vs "not installing Maven support" is hurting Eclipse as a Platform.

Then those "domain specific-app" should contribute to m2e, making it more configurable, e.g. one can easily extract such extensions into dedicated plugins, I'm just not willing to put the burden and effort to m2e for those hypothetical use-cases without any benefit.

mickaelistria:
> So should we optimize m2e for cases where people will only want "some" m2e and make life harder for those users that actually want to use m2e?

I think we should optimize m2e to make most consumers -of all kinds- efficient with it. I've demonstrated that there are cases where this would be perceived as a severe UI regressions (eg someone doing C++ as main activity wouldn't particularly enjoy seeing more "Maven" everywere); while I don't think you've demonstrated that this proposal is highly beneficial to balance the negative impact in the wider world.

> Then those "domain specific-app" should contribute to m2e, making it more configurable, 

Platform is already very configurable by default. And it's easy for any application to add top-level wizard shortcuts with a perspective extension.
It's what you're proposing here that is actually breaking configurability. It's by no mean a problem of who's contributing or not, it's really that the suggested approach is per-se a customization anti-pattern; there is no-one else to blame about that.

There have been many people contributing to Eclipse before you, who have shared inisght and ideas and reached some conclusions and implemented it. Before changing something assuming no-one did it ever right, it's often useful to think about why things are a way instead of another, what were the criteria to decide, what would be the impact of changing it on the wider ecosystem.

The right place to host such "applicative" configuration is the product definition: you could instead try to contribute thist shortcut to Java-centric EPP packages directly. So m2e is still good to use for anyone without drawback, and "Eclipse IDE for Java developer" gives more room to Maven.

Let's just face the truth about a versatile Platform such as Eclipse: you cannot know what people want and don't want in the general context; you need the user to refine the context to make good proposal. That's why perspective extensions are supposed to target contextual perspective.

laeubi:
I have now moved the shortcut (!) to the `m2e.jdt.ui` and `m2e.pde.ui` limiting it to the main perspective of these, that is the fine grainiest choice m2e currently offers.

mickaelistria (approved):
I think it's fine that way.

PR #1177, opened 2023-01-06
Fix #1176: Strip log messages

### LINKED ISSUE #1176: Eclipse log entries with empty lines
Just had this in my Tycho IDE:
grafik

m2e creates log entries that start and end with multiple empty lines. Since the text probably comes from Maven itself, we should always trim whatever gets logged.

The message was (note the empty lines):
```


Some dependencies of Maven Plugins are expected to be in provided scope.
Please make sure that dependencies listed below declared in POM
have set '<scope>provided</scope>' as well.

The following dependencies are in wrong scope:
 * org.apache.maven:maven-model-builder:jar:3.0:compile


```

Bananeweizen:


HannesWell (approved):
Looks good. Thanks!

PR #1178, opened 2023-01-06
[Build] Only sign artifacts and inject GPG-keys on master-branch build

HannesWell:


HannesWell:
@merks I finally were able to change the M2E build to not sign artifacts and inject keys for verificiation builds as you suggested at the EclipseConm, without duplicating too much. 
This way it is should not be possible anymore to either create artifacts signed with the Eclipse certificate or to extra the M2E GPG keyring+password from malicious PRs.

merks:
Cool.  🚀

PR #1180, opened 2023-01-06
Execute RCPTT tests as part of the integration tests

ptziegler:
In relation to #1157, I think it makes sense to test complex UI features via RCPTT, rather than JUnit. Simply because there is no elegant way to describe e.g. mouse clicks or validate widget properties. It also provides another way one can contribute test cases to reproduce bugs in a controlled environment.

The structure of the test project mimics the structure of the m2e project. So tests related to e.g. `org.eclipse.m2e.pde.ui` are located in a folder with the same name. 

The tests can be executed from the command line via `mvn clean verify`

If you want to run them manually, those steps have to be followed:

- Assemble the product you want to test via `mvn clean verify -D skipTests`. The zip file is located under products/m2e-ide/target/products.
- Download and install the IDE from the official [website](https://www.eclipse.org/rcptt/download/)
- Launch RCPTT and import the `org.eclipse.m2e.tests.rcptt` project into your workspace
- Unpack and add the assembled m2e product to your applications
- Right-click on the test case you want to execute and click "Run as -> Test Cases"

There is also an Eclipse IDE integration for RCPTT, but I have never tested how good/bad it works.

From experience, it is usually more efficient to have a few large test cases, instead of many small ones. This is because preparing the test environment usually takes a few seconds. So a lot of time is wasted if the tests themselves are shorter than the preparation.

You also don't write the tests from hand. Rather, you use the record functionality to track your user input in the started application and then format and restructure the recorded statements in order to become more readable.

I've attach an initial test case to verify the functionality of the table-based dependency editor.

### DIFF
--- a/org.eclipse.m2e.rcptt.tests/.project
+++ b/org.eclipse.m2e.rcptt.tests/.project
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.eclipse.m2e.rcptt.tests</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.m2e.core.maven2Builder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.rcptt.core.builder.q7Builder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.m2e.core.maven2Nature</nature>
+		<nature>org.eclipse.rcptt.core.rcpttnature</nature>
+	</natures>
+</projectDescription>
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,6 +77,23 @@ Some tests are in a separate repository which is referenced as a Git submodule i
 * Open the project modules you want to work on (right-click > Open project) and their dependencies; approve m2e connectors installation if prompted
 * Happy coding!
 
+### ⌨️ Setting up the RCPTT Test Environment
+
+* Download and install the RCPTT IDE from [here](https://www.eclipse.org/rcptt/download/).
+
+* Create an executable application .
+    * Right-click on [m2e-ide.product](products/m2e-ide/m2e-ide.product), *Export... > Plug-in Development > Eclipse product*
+    * Execute `mvn clean verify -D skipTests` in the command line. The application is located under `products/m2e-ide/target/products`.
+ 
+ * Launch the RCPTT IDE and import [org.eclipse.m2e.rcptt.tests](org.eclipse.m2e.rcptt.tests) into your workspace.
+ 
+ * Add m2e to your applications.
+ 
+ * Right-click on the test case you want to execute, *Run As... > Test Cases*. Select the m2e application.
+ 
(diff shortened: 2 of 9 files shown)


laeubi:
@ptziegler is the test-failure expected? And could you add the information about the tests to the https://github.com/eclipse-m2e/m2e-core/blob/master/CONTRIBUTING.md as part of this change?

laeubi:
> The test is for the _new_ dependency editor. The master is still using the _old_ dependency editor.
> I can move this test to the other branch, if you prefer that.

Just let us finish the other PR and proceed with this one then once it is merged!

ptziegler:
I think there's a slight timing issue when opening the dependency editor...

RCPTT selects the target location by its exact name. But depending on whether the Maven dependencies have been resolved, this is either "Maven Central" or "Maven Central 38 plug-ins available". I can work around this issue, but I don't see a clean way to actually fix this. :/

At least according to the error log the job was still running when RCPTT tried to open the editor:
```
swt.player
      ui.job.collector.jobs
            job:Resolving "target-platform" target definition
                job.class=org.eclipse.pde.internal.ui.editor.targetdefinition.TargetEditor$TargetChangedListener$1
```

In any case, the test results are looking good and should hopefully be stable.

laeubi:
Should we open an issue at RCPTT to provide some kind of pattern matching here?

ptziegler:
> Should we open an issue at RCPTT to provide some kind of pattern matching here?

I might've spoken too early. "Maven Central.*" is a valid path. So I guess regular expressions already work. :D

> Or use some kind of loop for a while until desired String is there?

That would also work, but I don't see a major benefit compared to a try-catch block.

laeubi:
@ptziegler I fear this PR got lost somehow, can you rebase this on the current master (and probably squash changes)?
@HannesWell would you be fine with that?

ptziegler:
> and probably squash changes

I'd like to keep the two commits separate if possible. One contains the integration of RCPTT, the other one the actual test, so I'd rather treat them as isolated changes.

laeubi (approved):
I'm fine with that and hope we can even have more RCPTT tests in the future!

There is even a feature in RCPTT to record a video of the tests, so probably one day we might publish a M2Eclipse Video for a release  :-D

ptziegler:
> UTF-8 BOM plus xml decl of ISO-8859-1 is incompatible

The POM file is not encoded as UTF-8?

laeubi:
You can ignore this this is a maven 3.9 bug :-\

org.eclipse.m2e.rcptt.tests/.project
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.eclipse.m2e.tests.rcptt</name>
```
HannesWell, line 3: The project Name should Match the folder/Module Name.
```suggestion
	<name>org.eclipse.m2e.rcptt.tests</name>
```

HannesWell (approved):
I have no experience with RCPTT at all. But more Tests are always good. 
And when all pass I'm fine with it. :)

org.eclipse.m2e.rcptt.tests/.project
ptziegler: Whoops. Done and dusted. :)
I hope the test is at least somewhat understandable... Looking at the test case via a text editor is a lot more difficult than via the IDE.

laeubi:
@ptziegler thanks for enhancing the test support!

PR #1185, opened 2023-01-08
Mark jobs in progress view with m2 icon

Bananeweizen:
Looks like this at runtime: 
grafik

### DIFF
--- a/org.eclipse.m2e.binaryproject/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.binaryproject/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.m2e.binaryproject;singleton:=true
-Bundle-Version: 2.1.0.qualifier
+Bundle-Version: 2.1.100.qualifier
 Bundle-Vendor: Eclipse.org - m2e
 Bundle-Name: M2E Binary Project Core
 Require-Bundle: org.eclipse.m2e.core;bundle-version="[2.0.0,3.0.0)",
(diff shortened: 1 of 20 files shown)


laeubi:
Maybe it would make sense to have a public `MavenJob` class instead of implmenting the same over and over again in the different job impls?

Aslo the contant might better be placed in an existing public API e.g. `IMaven` ?

HannesWell:
> Maybe it would make sense to have a public `MavenJob` class instead of implmenting the same over and over again in the different job impls?
> 

Sounds reasonable.

We could even check, if it is useful to execute such job's run (optionally) in a Maven-ExecutionContext?

Bananeweizen:
> Maybe it would make sense to have a public `MavenJob` class instead of implmenting the same over and over again in the different job impls?

That's one example of where I'm not sure if coupling all kinds of different implementations just for reusing this one trivial method is really worth it. Especially since those jobs derive from multiple super type classes, so we would already have at least 3 different abstract classes to be created. We have a similar implementation in egit, and also there we agreed that redeclaring that method is more simple in the long run. Admittedly, there we also use more job families to display more icons (e.g. checkout jobs have a different icon than pull, push etc.). Not sure if multiple icons would be useful here.

> Aslo the contant might better be placed in an existing public API e.g. `IMaven` ?

I disagree about IMaven. Interfaces defining constants (instead of classes defining the same constants) is already an antipattern by itself: https://en.wikipedia.org/wiki/Constant_interface. Just adding more of more unrelated constants to the same class/interface is another antipattern. The JobFamily class is perfectly self-contained, and has no relation to anything in IMaven.

Regarding public/internal I wasn't sure what the exact policy is here. It looks like everything in m2e.some.plugin/foo.internal.bar can be used freely in all other m2e bundles. There are lots of warnings for that kind of access in the problems view without my change. Is that understanding wrong? If so, I would just move the class to a non-internal package instead.

HannesWell (approved):
To me this looks good and I don't think we should make it part of the public API, unless there is a real need. But at the moment I don't see any.

HannesWell:
@laeubi any further remarks?

laeubi:
If its fine for you I'm fine with that :-)

HannesWell:
> If its fine for you I'm fine with that :-)

I have not yet looked into it in depth, but in the GH-UI this change looks good and I assume that Michael has tested it. :)

PR #1186, opened 2023-01-08
Have a name for all modules

Bananeweizen:
Avoids the mixture between FQNs and readable names in the Maven summary.


HannesWell (approved):
Looks good, thanks.

PR #1146, opened 2022-12-18
Update the dependencies to newer minor versions

merks:
Fixes https://github.com/eclipse-m2e/m2e-core/issues/1145

PR #1197, opened 2023-01-13
Add timeout when querying search.maven.org

mickaelistria:


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenArtifactIdentifier.java
```diff
@@ -17,18 +17,25 @@
...
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.net.http.HttpResponse.BodyHandlers;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
+<<<<<<< HEAD:org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenArtifactIdentifier.java
```
laeubi, line 30: seems some merge conflict markers made it into your PR

target-platform/target-platform.target
```diff
@@ -82,7 +82,7 @@
 				<dependency>
 					<groupId>com.google.code.gson</groupId>
 					<artifactId>gson</artifactId>
-					<version>2.9.1</version>
+					<version>2.10.1</version>
```
HannesWell, line 85: Please keep the dependency as it is. Due to lsp4j's package imports, that causes TP resolution errors in the workspace:
https://github.com/eclipse-m2e/m2e-core/pull/1146#issuecomment-1362187746

I don't know why it doesn't do that in the CI.

mickaelistria: I don't think this issue is an issue of m2e and that the resolution error in the TP is something m2e should be blamed for. There is nothing that prevents gson 2.9.1 and 2.10.1 to coexist in the same TP. If LSP4J requires 2.9.1, then consumers of LSP4J must provide 2.9.1; without preventing m2e to move to newer version.

HannesWell: Then please add both versions, because only with 2.10 I can currently not launch an IDE in my m2e workspace.
The update for lsp4j happens in https://github.com/eclipse/lsp4j/pull/689.

mickaelistria: I believe both versions are supposed to be part of the resolved TP: 2.10.1 explicitly changed here, and 2.9.1 as a transitive dependency of LSP4E. Can you check your resolved TP contains both? Then can you check your launch config contains both?

HannesWell: > I believe both versions are supposed to be part of the resolved TP: 2.10.1 explicitly changed here, and 2.9.1 as a transitive dependency of LSP4E. Can you check your resolved TP contains both? Then can you check your launch config contains both?

Indeed the TP contains both. Nevertheless PDE seems to fail to wire the lsp4j to the old gson-bundle. gson isn't a singleton, so this look like a bug in PDE to me.

Issue #1196, opened 2023-01-12
IDE freeze, stack seems related to m2e PDE support


mickaelistria:
My IDE is frozen; IIRC it happened while opening the org.eclipse.jdt.core project in my workspace.
The stack trace seems to say that the freeze may be caused by a long running operation running in UI Thread. I'm reporting it here as I see m2e involved, but maybe it's a more general PDE issue.

```
main" #1 [154073] prio=6 os_prio=0 cpu=6365860.94ms elapsed=60035.56s tid=0x00007fb71c0280e0 nid=154073 runnable  [0x00007fb720ffb000]
   java.lang.Thread.State: RUNNABLE
	at sun.nio.ch.SocketDispatcher.read0(java.base@19.0.1/Native Method)
	at sun.nio.ch.SocketDispatcher.read(java.base@19.0.1/SocketDispatcher.java:47)
	at sun.nio.ch.NioSocketImpl.tryRead(java.base@19.0.1/NioSocketImpl.java:251)
	at sun.nio.ch.NioSocketImpl.implRead(java.base@19.0.1/NioSocketImpl.java:302)
	at sun.nio.ch.NioSocketImpl.read(java.base@19.0.1/NioSocketImpl.java:340)
	at sun.nio.ch.NioSocketImpl$1.read(java.base@19.0.1/NioSocketImpl.java:789)
	at java.net.Socket$SocketInputStream.read(java.base@19.0.1/Socket.java:1025)
	at sun.security.ssl.SSLSocketInputRecord.read(java.base@19.0.1/SSLSocketInputRecord.java:477)
	at sun.security.ssl.SSLSocketInputRecord.readHeader(java.base@19.0.1/SSLSocketInputRecord.java:471)
	at sun.security.ssl.SSLSocketInputRecord.bytesInCompletePacket(java.base@19.0.1/SSLSocketInputRecord.java:70)
	at sun.security.ssl.SSLSocketImpl.readApplicationRecord(java.base@19.0.1/SSLSocketImpl.java:1465)
	at sun.security.ssl.SSLSocketImpl$AppInputStream.read(java.base@19.0.1/SSLSocketImpl.java:1069)
	at java.io.BufferedInputStream.fill(java.base@19.0.1/BufferedInputStream.java:255)
	at java.io.BufferedInputStream.read1(java.base@19.0.1/BufferedInputStream.java:310)
[...]

PR #1198, opened 2023-01-13
Disable search in Maven-Central for PDE-SourceLookup

HannesWell:
Fixes https://github.com/eclipse-m2e/m2e-core/issues/1196

There could be a preference, as suggested in https://github.com/eclipse-m2e/m2e-core/issues/1196#issuecomment-1381565091, which is off by default. But as elaborated in https://github.com/eclipse-m2e/m2e-core/issues/1196#issuecomment-1381557334 the chances for a hit in the file-hash based search in central are probably low, when the pom.properties are absent.

### DIFF
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/messages.properties
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/messages.properties
@@ -237,6 +237,7 @@ MavenPreferencePage_autoUpdateProjectConfiguration=Automatically update Maven pr
 MavenPreferencePage_updateProjectRequired_title=Update project required
 MavenPreferencePage_warnIncompleteMapping=Hide warning for incomplete mapping
 MavenPreferencePage_changingPreferencesRequiresProjectUpdate=Changing the checksum policy requires updating the Maven projects for the changes to take effect. Do you want to update projects now?
+MavenPreferencePage_queryCentralToIdentifyArtifacts=Use Maven-Central to identify unknown artifacts (e.g. when locating sources)
 MavenProjectPreferencePage_btnResolve=Resolve dependencies from &Workspace projects
 MavenProjectPreferencePage_dialog_message=Maven settings have changed. Do you want to update project configuration?
 MavenProjectPreferencePage_dialog_title=Maven Settings
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenArtifactIdentifier.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenArtifactIdentifier.java
@@ -43,6 +43,7 @@ import org.eclipse.core.runtime.ILog;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.preferences.InstanceScope;
 
 import org.apache.maven.artifact.Artifact;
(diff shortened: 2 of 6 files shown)


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenArtifactIdentifier.java
```diff
@@ -58,13 +58,17 @@ private MavenArtifactIdentifier() { // static use only
   private static final ILog LOG = Platform.getLog(MavenArtifactIdentifier.class);
 
   public static Collection<ArtifactKey> identify(File classesLocation) {
+    return identify(classesLocation, true);
+  }
+
+  public static Collection<ArtifactKey> identify(File classesLocation, boolean searchCentral) {
```
laeubi, line 64: I don't thing such a parameter is useful here.
It should either be disabled for all cases or controlled by a preference, so callers don't have to individually decide what to pass here.

HannesWell: But the callers have different use-cases. Some call it only on demand (i.e. the 'old' callers) and call it in advance to get the result for 'everything'. 
So I think it makes sense to distinguish here. If you think it is better the flag can also be renamed to 'fullSearch' or to an enum.

laeubi: The use case is always the same, and the caller should not need to know about the internals (e.g. if/that maven central might be queried), so I still think it would be better in general for a user to have the choice to query central at all, e.g. if I'm behind a company proxy or a like.

HannesWell: Hmm, ok. Introduced a preference for that.

HannesWell:
Do you others have better ideas for the label (and consequently the preference id).
Furthermore actually, from the topic it should be located in the preferences of the Source-Lookup UI, but it is now in core.
But actually we could still show that at the former location but store it in the m2.core node? Or simply completely store and display it in the m2e.sourcelookup node, but query that in core.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/messages.properties
```diff
@@ -237,6 +237,7 @@ MavenPreferencePage_autoUpdateProjectConfiguration=Automatically update Maven pr
 MavenPreferencePage_updateProjectRequired_title=Update project required
 MavenPreferencePage_warnIncompleteMapping=Hide warning for incomplete mapping
 MavenPreferencePage_changingPreferencesRequiresProjectUpdate=Changing the checksum policy requires updating the Maven projects for the changes to take effect. Do you want to update projects now?
+MavenPreferencePage_queryCentralToIdentifyArtifacts=Query Maven-Central when identifying artifacts (used in Source-Lookups as last resort)
```
laeubi, line 240:
```suggestion
MavenPreferencePage_queryCentralToIdentifyArtifacts=Use Maven-Central to identify unknown artifacts (e.g. when locating sources)
```

laeubi (approved):
Looks good, I think that is fine to have it in core, this is quite basic and the identifier resides in core as well so it can be used anywhere in m2e

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/messages.properties
HannesWell: Done.

PR #1203, opened 2023-01-14
Update 'm2e-core-tests' submodule

HannesWell:


HannesWell:
@laeubi this probably/hopefully fixes the test-failures in the GH workflow since https://github.com/eclipse-m2e/m2e-core/pull/1181
IIRC in contrast to Jenkins the GH-workflow is not configured to check-out the latest submodule-master commit.

laeubi:
> @laeubi this probably/hopefully fixes the test-failures in the GH workflow since #1181 IIRC in contrast to Jenkins the GH-workflow is not configured to check-out the latest submodule-master commit.

Can we probably change that? https://github.com/actions/checkout#usage

HannesWell:
> > @laeubi this probably/hopefully fixes the test-failures in the GH workflow since #1181 IIRC in contrast to Jenkins the GH-workflow is not configured to check-out the latest submodule-master commit.
> 
> Can we probably change that? https://github.com/actions/checkout#usage

I don't see an option for that. Furthermore I actually think this way it is better because it forces you properly update the test-submodule references (which is just another element to stage in the m2e-core repo) and therefore ensures that m2e-core and m2e-core-tests are kept in sync. One just have to do that one addition D&D in the EGit staging view.

PR #1204, opened 2023-01-14
Permanently cache Central-Search results for Artifact-Identification

HannesWell:
As suggested in #1196 this PR introduces a cache to avoid querying Maven-Central for the same artifact multiple times.

The cache stores the artifact's jar-file path and maps it to the artifact-keys returned by the Maven-Central search.

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenArtifactIdentifier.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenArtifactIdentifier.java
@@ -17,6 +17,7 @@ package org.eclipse.m2e.core.internal;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
 import java.net.URI;
 import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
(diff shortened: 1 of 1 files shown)


mickaelistria:
Files are mutable, for example if using some not-versioned file name which would suddenly use a newer version without the file path/name to change. So instead of storing a cache using File as key, I would suggest to store in cache the sha1->artifactId mapping, which is immutable and reusable. It also allows that several files with different path/names but the same sha1 (so same content) would only generate a single request to search.maven.org.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/MavenArtifactIdentifier.java
```diff
@@ -102,13 +115,65 @@ private static Set<ArtifactKey> identifyCentralSearch(Path file) {
...
-      }).collect(Collectors.toSet());
+      }).collect(Collectors.toUnmodifiableSet());
     } catch(IOException | InterruptedException e) {
       LOG.log(Status.error("Failed to identify file by its hash using search.maven.org: " + file));
       return Set.of();
     }
   }
 
+  private static class ArtifactIdCache { // encapsulated in an inner-class to have it loaded only if searching Central is enabled
+    static final Map<Path, Set<ArtifactKey>> IDS = loadCachedIds();
```
mickaelistria, line 126: As mentioned, the key must be something immutable for safer caching.

mickaelistria (changes requested):

HannesWell:
> Files are mutable, for example if using some not-versioned file name which would suddenly use a newer version without the file path/name to change. So instead of storing a cache using File as key, I would suggest to store in cache the sha1->artifactId mapping, which is immutable and reusable. It also allows that several files with different path/names but the same sha1 (so same content) would only generate a single request to search.maven.org.

That's all right. I wouldn't mind to check the same file at different location multiple times, but missing a changed file at the same location would be bad. Nevertheless caching by Path has the advantage that the file-content doesn't have to be read and therefore is probably significantly faster.
I think it would be possible to maintain another path-based cache that remembers the last-modified time of a file, is build at runtime and is queried first. With that an unchanged file is only hashed once. But because I'm still think that a successful look-up in Maven-Central is very unlikely (see https://github.com/eclipse-m2e/m2e-core/issues/1196#issuecomment-1381557334), I don't want to spent more time on this right now and leave that optimization to somebody that is interested (if there is anybody).

mickaelistria (approved):

PR #1206, opened 2023-01-14
Fix discouraged access warnings for own code

Bananeweizen:
Have all m2e project accesses be silenced by exporting the related packages via x-friends. That's safe since all that code is changed and released together.

Disable discouraged access warnings in test projects. Tests can be adapted the moment that referenced class vanishes, there is no need to require stable APIs only in tests.

Finally remove no longer needed SuppressWarnings.


org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF
```diff
@@ -6,17 +6,23 @@ Bundle-RequiredExecutionEnvironment: JavaSE-17
...
-Export-Package: org.eclipse.m2e.core.ui.internal;x-friends:="org.eclipse.m2e.editor",
+Export-Package: org.eclipse.m2e.core.ui.internal;x-friends:="org.eclipse.m2e.editor,org.eclipse.m2e.scm",
  org.eclipse.m2e.core.ui.internal.actions;
   x-friends:="org.eclipse.m2e.editor,
    org.eclipse.m2e.editor.xml,
    org.eclipse.m2e.scm,
-   org.eclipse.m2e.jdt.ui",
+   org.eclipse.m2e.jdt.ui,
+   org.eclipse.m2e.tests,
+   org.eclipse.m2e.editor.tests",
```
HannesWell, line 16: For me it is a bit strange, to mark 'test-plugins' as x-friends. Can't you instead disbale discouraged access-warnings in the preferences of the tests, just like you did in the apt.tests?

Applies to other 'production' plugins too.

org.eclipse.m2e.tests.common/META-INF/MANIFEST.MF
```diff
@@ -21,7 +21,13 @@ Bundle-RequiredExecutionEnvironment: JavaSE-17
 Bundle-Vendor: Eclipse.org - m2e
 MavenArtifact-GroupId: org.eclipse.m2e
 MavenArtifact-ArtifactId: org.eclipse.m2e.tests.common
-Export-Package: org.eclipse.m2e.tests.common;x-friends:="org.eclipse.m2e.tests,org.eclipse.m2e.core.tests,org.eclipse.m2e.jdt.tests,org.eclipse.m2e.core.ui.tests,org.eclipse.m2e.wtp.tests"
+Export-Package: org.eclipse.m2e.tests.common;
+  x-friends:="org.eclipse.m2e.tests,
+   org.eclipse.m2e.core.tests,
```
HannesWell, line 26: Since this plug-in itself is only for tests, I think it would be ok to define other test-plugins as friends. But if you disable discouraged access warnings there that is probably not necessary.

Bananeweizen: Okay, will disable it in all test bundles, and then apply only the remaining fixes in non test bundles.

HannesWell:
@Bananeweizen do you want to complete this one?

akurtakov:
What is the status here?

PR #1207, opened 2023-01-15
Fix broken baseline-version config

Bananeweizen:


HannesWell (approved):
Thanks!
Now clue how that slipped in and I'm a bit surprised that this did not cause errors during XML file parsing

PR #1208, opened 2023-01-15
Don't use maven batch model locally

Bananeweizen:
Since some versions the batch mode parameter disables colored output. Therefore it should only be used on CI systems nowadays. Instead use "-ntp" to suppress all the download messages.


HannesWell (changes requested):
Batch-Mode is not only about disabling transfer-progress and output-colors, it is also about disabling the 'interactive-mode' which suppresses request for user-input and instead uses a reasonable default then.
Although we should not require that in the m2e build (currently), I think we should still use that option in the CLI. So if you don't want to have -B in local builds (what I can understand), we should add it in the CI. 
Therefore can you then also readd the `-B` option to the Maven-commands in the `Jenkinsfile` and `.github/workflows/maven.yml`.

laeubi:
> Since some versions the batch mode parameter disables colored output.

If you just want to *enable* color you can specify `-Dstyle.color=always`

Bananeweizen:
Thanks for all the hints. I've re-added the argument in all CI system related files. I've also taken liberty to use the long versions in the config, since that might be more easy to read for some people like myself.

Independent of this change, wouldn't it be also useful to remove the "--show-version" argument and instead have a maven-enforcer rule for the maven version configured?

HannesWell (approved):
Looks good! 
And thanks for the other white-space clean-ups. Did you do them by hand or do you had a tool for that? AFAICT the Jenkins-Editor Plugin in the Eclipse Marketplace sadly does not format the Jenkinsfile.

> Thanks for all the hints. I've re-added the argument in all CI system related files. I've also taken liberty to use the long versions in the config, since that might be more easy to read for some people like myself.

That's fine 👍🏽 

> 
> Independent of this change, wouldn't it be also useful to remove the "--show-version" argument and instead have a maven-enforcer rule for the maven version configured?

I'm not aware of a certain requirement about the Maven-version for the M2E-build (probably something not too old). Personally I just try to use the latest possible Maven.
For me the maven version is just informal, so I don't think an enforcer rule is necessary.
Or do you see a specific need for that?

Bananeweizen:
I often run with "show whitespace", that's how I noticed. I'm wondering if a "mixed whitespace detector plugin" would be something useful to start. :)
Also opening such files with VS instead of Eclipse can sometimes be helpful for minor formatting or whitespace improvements.

Regarding enforcer: I know that Tycho has some minimum requirements and would enforce those. In former times I often used hard coded minimum versions when there were known incompabilities between Maven and Eclipse/Tycho, but that hasn't been the case since a long time. I just thought someone wanted the version output as means to check the maven version on CI (which is what I have seen before in other projects 🙅🏻 ).

HannesWell:
Yes from Show whitespace I notice such mixings as well but a automated solution would be more convenient.

Frankly, at the moment I don't see a urgent need to use the enforcer-plugin(maybe if we use Maven-4 one day) but at the same time I would not object to use it. So please feel free to create a follow-up If you want to have that.
This change is good to go as it is.

Issue #1214, opened 2023-01-21
Build plugin execution leaking classes into the Eclipse JVM


odrotbohm:
m2e executes build plugins both during project updates and also during incremental compilation of sources of a project. A sample project available [here](https://github.com/odrotbohm/m2e-bytebuddy-memory) shows how the execution of e.g. the Bytebuddy Maven Plugin causes ~800 classes being added to the Eclipse JVM for each full project update and a low to mid two-digit value for each compilation.

I spoke to @raphw (author of the Bytebuddy Maven Plugin) and he couldn't think of any obvious problem within the plugin itself. I was wondering if this symptom could be related to how m2e handles the instances of the plugins. It seemed to reuse the instances between incremental compilation, but I am totally just observing this from the outside.

I'd be willing to help to diagnose this further if you have any pointers which code is actually creating and maintaining the build plugin instances, which object instances to look out for in a heap dump etc.

laeubi:
@odrotbohm have you checked (e.g. with JProvfler or VisualVM) that the classes are retained (e.g. after a GC?) if yes, generally the next step is to look at the GC_ root that is preventing the classes from being garbage collected.

Beside this it could help if the maven plugin participates in the Incremental build e.g. with BuildContext:
https://www.eclipse.org/m2e/documentation/m2e-making-maven-plugins-compat.html#buildcontext

odrotbohm:
I'm exploring the application using YourKit but have only seen the forest for the trees so far. Yes, the classes are retained after GCs. So, the number of loaded classes grows unbounded and ultimately results in significant off-heap memory consumption.

Yes, the plugin has already supported incremental builds (I actually contributed a draft of that) for a while. The individual compilations don't have a drastic effect on the number of loaded classes. However, a full project update sometimes adds up to a 1000 of them. 😕

laeubi:
I think then you should look at the classloader of the retained class and check what instance holds this so it can't be garbage collected.

Yourkit seem to have also an option to show GC roots: https://www.yourkit.com/docs/dotnet-profiler/2022.9/help/gc_roots.jsp.

odrotbohm:
The classloader of those classes is a `ClassRealm` and the GC Roots path for one of them looks as follows:

<img width="1452" alt="Screenshot 2023-01-21 at 16 57 13" src="https://user-images.githubusercontent.com/128577/213875423-49179bdc-a653-4788-9f2f-465cf0cf26ef.png">

[...]

PR #1218, opened 2023-01-25
Bump m2e.feature to 2.2.0 and include gson into the m2e repository

HannesWell:
The version was already bumped in commit b8a7fb34d8808b06b2a9a4b1f43a09f656f0b41c, which was reverted shortly later. But the contained changes where applied again so we should bump that version again. Furthermore significant changes have been made in the core of m2e, which justify a minor version bump.

Include the newly used version of com.google.gson into the m2e repository since nobody else seems to provide it in a p2-repo at the moment.

Both helps in https://github.com/eclipse-m2e/m2e-core/issues/1214#issuecomment-1404120935.

And simplify LifecycleManagerDisposer.

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/LifecycleManagerDisposer.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/LifecycleManagerDisposer.java
@@ -61,13 +61,7 @@ public class LifecycleManagerDisposer implements ClassWorldListener {
     lifecyclesMapField.setAccessible(true);
     Object lifecycles = lifecyclesMapField.get(manager);
     if(lifecycles instanceof Map<?, ?> map) {
-      for(Object key : map.keySet().toArray()) {
-        if(key instanceof Class<?> clazz) {
-          if(clazz.getClassLoader() == realm) {
-            map.remove(key);
-          }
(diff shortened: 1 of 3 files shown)


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/LifecycleManagerDisposer.java
```diff
@@ -61,13 +61,7 @@ private void disposeOutdatedEntries(ClassRealm realm, LifecycleManager manager)
...
     Object lifecycles = lifecyclesMapField.get(manager);
     if(lifecycles instanceof Map<?, ?> map) {
-      for(Object key : map.keySet().toArray()) {
-        if(key instanceof Class<?> clazz) {
-          if(clazz.getClassLoader() == realm) {
-            map.remove(key);
-          }
-        }
-      }
+      map.keySet().removeIf(key -> key instanceof Class<?> clazz && clazz.getClassLoader() == realm);
```
HannesWell, line 64: @laeubi that's what I had in mind in https://github.com/eclipse-m2e/m2e-core/pull/1217#discussion_r1084948482. :)

HannesWell:
@odrotbohm this should be available in the snapshots repo in about ~20/30min.

PR #1221, opened 2023-01-28
NPE in MavenSession.getAllProjects()

### LINKED ISSUE #1220: NPE in MavenSession.getAllProjects()
A related issue - https://github.com/redhat-developer/vscode-java/issues/2889#issuecomment-1399351217

This issue is caused by https://github.com/eclipse-m2e/m2e-core/commit/5640b34fceef67bf110e5bfd23244daf9921be10#diff-796a271293c65b5cac9c04a4dccddff344e43fd10f474a5cd2992c2333325704R340

snjeza:
Fixes #1220 

Signed-off-by: Snjezana Peco <snjezana.peco@redhat.com>

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
@@ -359,6 +359,7 @@ public class MavenExecutionContext implements IMavenExecutionContext {
         .setTransferListener(new ArtifactTransferListenerAdapter(monitor));
     final MavenProject origProject = mavenSession.getCurrentProject();
     final List<MavenProject> origProjects = mavenSession.getProjects();
+    final List<MavenProject> origAllProjects = mavenSession.getAllProjects();
     final ClassLoader origTCCL = Thread.currentThread().getContextClassLoader();
     try {
       if(project == null && projectSupplier != null) {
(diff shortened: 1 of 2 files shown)


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
```diff
@@ -366,15 +366,19 @@ private <V> V executeBare(MavenProject project, ICallable<V> callable, IProgress
...
       }
       return callable.call(this, IProgressMonitor.nullSafe(monitor));
     } finally {
       Thread.currentThread().setContextClassLoader(origTCCL);
       repositorySession.setTransferListener(origTransferListener);
       if(project != null) {
         mavenSession.setCurrentProject(origProject);
-        mavenSession.setProjects(origProjects != null ? origProjects : Collections.<MavenProject> emptyList());
+        List<MavenProject> projects = origProjects != null ? origProjects : Collections.<MavenProject> emptyList();
+        mavenSession.setProjects(projects);
```
laeubi, line 380: "projects" and "all projects" are not always the same!
So if you want to reset it both values must be saved before hands.

In general for plugins calling getAllProjects is bad, the method is marked with:
`/** This is a provisional method and may be removed */`
and it should actually only be used in the setup phase.

snjeza: > "projects" and "all projects" are not always the same!
So if you want to reset it both values must be saved before hands.

Fixed.

> In general for plugins calling getAllProjects is bad, the method is marked with:
/** This is a provisional method and may be removed */
and it should actually only be used in the setup phase.

maven-quarkus-plugin works from the mvn command line. See https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/DefaultMaven.java#L656

laeubi: "works" does not mean it is good style, I don't know why `maven-quarkus-plugin` is using `getAllProjects`, but the value is not the same as `getProjects` and should only be used in very rare cases (probably never).

`getProject` return all projects that are selected for a reactor build and `getAllProjects` will contain all possible projects that are discovered through modules, so a plugin using `getAllProjects` will access projects that are probably never build.

laeubi (approved):
Looks good to me.

PR #1232, opened 2023-02-03
Initialize the dependency table viewer with SWT.FULL_SELECTION.

ptziegler:
This is required for tables with more than one column, where each column should be editable. If you want to edit a cell, you currently have to first select the row you want to edit, by clicking on the item in the first column. In order to then edit a different cell, you then have to first select the new column, before the cell becomes editable.

This issue only seems to exist on Windows machines.

I noticed this issue after trying out the new editor on a Windows machine. Why this doesn't happen with GTK is anyones guess...

From the ViewerColumn documentation:
> Users setting up an editable TreeViewer or TableViewer with more than 1 column **have**
> to pass the SWT.FULL_SELECTION style bit when creating the viewer


HannesWell (approved):
Thank you for this follow-up!

PR #1247, opened 2023-02-13
Fix reading of custom Archetypes + optimize layout of preference page

laeubi:


laeubi:
Currently failing due to:
- https://github.com/eclipse-tycho/tycho/pull/2089

in combination with:
- https://github.com/eclipse-m2e/m2e-core/issues/1248

PR #1253, opened 2023-02-15
Create test-case to test consideration of properties in maven.config

HannesWell:
Add test-case initially created for https://github.com/eclipse-m2e/m2e-core/pull/852#issuecomment-1424658922 so that it can be used to check alternative solutions like https://github.com/eclipse-m2e/m2e-core/pull/1242.

### DIFF
--- a/org.eclipse.m2e.core.tests/resources/projects/mavenConfig/bundle/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.core.tests/resources/projects/mavenConfig/bundle/META-INF/MANIFEST.MF
@@ -1,0 +1,7 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Bundle
+Bundle-SymbolicName: my.bundle3
+Bundle-Version: 1.0.0.qualifier
+Automatic-Module-Name: my.bundle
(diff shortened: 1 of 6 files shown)


laeubi:
Lets see how far we got with the latest changes!

org.eclipse.m2e.core.tests/resources/projects/mavenConfig/bundle/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,7 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Bundle
+Bundle-SymbolicName: my.bundle3
```
HannesWell, line 4: When I change the BSN to `my.bundle` the tests fail when being executed with the other tests of the class, but when executed as only case it passes.

The reason is that when calling `projectRegistry.getProjectFacade(pom);` in `ProjectRegistryManager.createExecutionContext(IProjectRegistry state, IFile pom, IProjectConfiguration resolverConfiguration)` an old facade from another test-case is found that has a different multi-module root, which at that point does not exist anymore.
Looks like the clean-up of the registry is not done (at all or just too late?).

HannesWell:
Awesome, build no 214 finally succeeded. 🎉

Issue #1262, opened 2023-02-20
Allow Test to be skipped when pom specifies maven.test.skip=true


RoiSoleil:
Allow Test to be skipped when pom specifies maven.test.skip=true in M2E

rnesselrath:
We have a very specific configuration where we call the maven-compiler-plugin two times. First in the normale compile phase and second in . Reason is that we generate additional source code in between that requires the already compiled test classes from the first plugin call. Additionally we haven the option to skip the second plugin call. To sum up we have the following compiler plugin definition in our pom:

`
  <plugin>
        <artifactId>maven-compiler-plugin</artifactId>
        <version>3.8.1</version>
        <executions>
          <execution>
            <id>default-testCompile</id>
            <phase>test-compile</phase>
            <goals>
              <goal>testCompile</goal>
            </goals>
          </execution>
          <execution>
            <id>compile-generated-specs</id>
            <phase>process-test-classes</phase>
            <goals>
              <goal>testCompile</goal>
            </goals>
            <configuration>
              <skip>true</skip>
            </configuration>
          </execution>
        </executions>
`
The skip in the second execution block leads to the problem that test sources are not added to the classpath anymore.
I think changing one line would fix this issue.

 isTestCompilationSkipped = isTestCompilationSkipped ||  Boolean.TRUE
                .equals(maven.getMojoParameterValue(mavenProject, compile, "skip", Boolean.class, monitor));

PR #1263, opened 2023-02-20
Allow Test to be skipped when pom specifies maven.test.skip=true

RoiSoleil:
Fix : https://github.com/eclipse-m2e/m2e-core/issues/1262

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -16,6 +16,13 @@ The dependencies editor has been adapted to show all artifacts within a target l
 
 ![grafik](https://user-images.githubusercontent.com/70652978/212153011-160fa96a-1c06-4092-9b89-fcd7a3c2859e.png)
 
+
+### Skip on Tests
+
+Before, the property "<maven.test.skip>true</maven.test.skip>" and the "skip" configurations on "maven-compiler-plugin" and "maven-resources-plugin" were not taken into account by M2E.
+Now, they are well taken into account by M2E and this allows, depending on the need (especially compilation time), to either not compile tests or not copy test resources.
+The folders thus ignored no longer appear in the Package Explorer as "Java" folders but as standard folders.
+
 ## 2.1.0
 
 * 📅 Release Date: November 24th 2022
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -17,11 +17,14 @@ The dependencies editor has been adapted to show all artifacts within a target l
 ![grafik](https://user-images.githubusercontent.com/70652978/212153011-160fa96a-1c06-4092-9b89-fcd7a3c2859e.png)
 
 
-### Skip on Tests
+### Ignore Test Sources and Resources compilation if `maven.test.skip=true` is specified
 
-Before, the property "<maven.test.skip>true</maven.test.skip>" and the "skip" configurations on "maven-compiler-plugin" and "maven-resources-plugin" were not taken into account by M2E.
-Now, they are well taken into account by M2E and this allows, depending on the need (especially compilation time), to either not compile tests or not copy test resources.
-The folders thus ignored no longer appear in the Package Explorer as "Java" folders but as standard folders.
+The property "<maven.test.skip>true</maven.test.skip>" and the "skip" property in configurations of the "maven-compiler-plugin" and "maven-resources-plugin" are now taken into
+account by M2E in the sense that, if enabled, M2E ignores the corresponding folder, which no longer appear in the Package Explorer as "Java" folders but as standard folders.
+This allows, depending on the need (especially compilation time), to either not compile tests or not copy test resources.
+
+In general it is not recomended to use the mentioned properties but to use `-DskipTests` instead:
+https://maven.apache.org/surefire/maven-surefire-plugin/examples/skipping-tests.html
 
 ## 2.1.0
 
--- a/org.eclipse.m2e.jdt.tests/src/org/eclipse/m2e/jdt/tests/JavaConfigurationTest.java
+++ b/org.eclipse.m2e.jdt.tests/src/org/eclipse/m2e/jdt/tests/JavaConfigurationTest.java
@@ -10,28 +10,18 @@
 
 package org.eclipse.m2e.jdt.tests;
 
-import static org.junit.Assert.assertArrayEquals;
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.*;
 
-import java.io.ByteArrayInputStream;
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
+import java.io.*;
+import java.util.Arrays;
 
-import org.eclipse.core.resources.IFile;
-import org.eclipse.core.resources.IMarker;
-import org.eclipse.core.resources.IProject;
-import org.eclipse.core.resources.IResource;
-import org.eclipse.core.resources.IncrementalProjectBuilder;
-import org.eclipse.core.runtime.CoreException;
-import org.eclipse.core.runtime.FileLocator;
-import org.eclipse.jdt.core.IJavaProject;
-import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.core.resources.*;
+import org.eclipse.core.runtime.*;
+import org.eclipse.jdt.core.*;
 import org.eclipse.m2e.core.MavenPlugin;
 import org.eclipse.m2e.core.internal.preferences.MavenConfigurationImpl;
 import org.eclipse.m2e.tests.common.AbstractMavenProjectTestCase;
 import org.junit.Test;
-
 
 @SuppressWarnings("restriction")
 public class JavaConfigurationTest extends AbstractMavenProjectTestCase {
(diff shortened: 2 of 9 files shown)


laeubi:
@RoiSoleil can you probably elaborate a bit more about the context (and probably also adding a testcase to verify this)?

RoiSoleil:
> @RoiSoleil can you probably elaborate a bit more about the context (and probably also adding a testcase to verify this)?

When debugging a very big project, compiling tests can take a lot of time. Being able to disable this can save time to debug more rapidly. I will add a test for my code.

RoiSoleil:
@laeubi is that ok for you ?

laeubi (approved):
This looks fine and all tests succeed, maybe @HannesWell or @mickaelistria  like to review also.

HannesWell:
> This looks fine and all tests succeed, maybe @HannesWell or @mickaelistria like to review also.

Yes, I would ike to review this, probably tomorrow.

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/MavenClasspathHelpers.java
```diff
@@ -80,4 +80,9 @@ public static boolean hasTestFlagDisabled(MavenProject mavenProject) {
     return mavenProject != null
         && Boolean.valueOf(mavenProject.getProperties().getProperty("m2e.disableTestClasspathFlag", "false"));
   }
+
+  public static boolean isTestSkipped(MavenProject mavenProject) {
+    return mavenProject != null
+        && Boolean.valueOf(mavenProject.getProperties().getProperty("maven.test.skip", "false"));
```
HannesWell, line 86: This only checks if the `maven.test.skip` property is defined in the properties section of the current project.
It does not check if that property is set in the parent or is configured in the [maven-compiler-plugin's `testCompile` goal](https://maven.apache.org/plugins/maven-compiler-plugin/testCompile-mojo.html#skip).

According to https://maven.apache.org/plugins-archives/maven-surefire-plugin-2.12.4/examples/skipping-test.html:
`maven.test.skip is honored by Surefire, Failsafe and the Compiler Plugin.`

I'm not yet sure if it is useful to check [maven-surefire-plugin:skip](https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#skip) too, because I don't expect the test goal to trigger the compiler-plugin again.
But I haven't checked that.

Therefore I suggest to delete this method and instead check in the caller if that configuration property is set.
You could do that by creating a new flag `boolean isTestCompilationSkipped = false;` in `AbstractJavaProjectConfigurator.addProjectSourceFolders()` just before the loop that iterates over all `executions` and only considers `executions` for which `isTestCompileExecution` returns true.
Within that loop you could simply set the flag to the configured value with
`isTestCompilationSkipped = maven.getMojoParameterValue(mavenProject, compile, "skip", Boolean.class, monitor);`

This should also cover the cases mentioned above.

HannesWell (changes requested):
Sorry for the delayed reply.

In general I think the change makes sense, but I think the way how it is implemented should be more general (which would make it simpler).
Please see the comments below.

org.eclipse.m2e.jdt.tests/src/org/eclipse/m2e/jdt/tests/JavaConfigurationTest.java
```diff
@@ -57,18 +58,32 @@ public void testFileChangeUpdatesJDTSettings() throws CoreException, IOException
...
-				FileLocator.toFileURL(getClass().getResource("/projects/compilerWarnings/pom.xml")).getFile());
-		waitForJobsToComplete();
-		IProject project = importProject(pomFileFS.getAbsolutePath());
-		IFile file = project.getFile("src/main/java/A.java");
-		project.build(IncrementalProjectBuilder.FULL_BUILD, null);
-		waitForJobsToComplete();
-		IMarker[] findMarkers = file.findMarkers(IMarker.PROBLEM, true, IResource.DEPTH_ZERO);
-		assertArrayEquals(new IMarker[0], findMarkers);
-	}
+  @Test
```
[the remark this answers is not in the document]
RoiSoleil: I think it should be done in a more global way.

org.eclipse.m2e.jdt.tests/src/org/eclipse/m2e/jdt/tests/JavaConfigurationTest.java
```diff
@@ -57,18 +58,32 @@ public void testFileChangeUpdatesJDTSettings() throws CoreException, IOException
...
+    ((MavenConfigurationImpl) MavenPlugin.getMavenConfiguration()).setAutomaticallyUpdateConfiguration(true);
+    setAutoBuilding(true);
+    File pomFileFS = new File(
+        FileLocator.toFileURL(getClass().getResource("/projects/skipTest/pom.xml")).getFile());
+    waitForJobsToComplete();
+    IProject project = importProject(pomFileFS.getAbsolutePath());
+    waitForJobsToComplete();
+    IJavaProject jproject = JavaCore.create(project);
+    assertTrue(Arrays.stream(jproject.getRawClasspath())
+                     .noneMatch(cp -> cp.isTest()));
```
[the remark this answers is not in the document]
RoiSoleil: Done

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/MavenClasspathHelpers.java
RoiSoleil: Done

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
```diff
@@ -352,6 +355,8 @@ protected void addProjectSourceFolders(IClasspathDescriptor classpath, Map<Strin
           } catch(CoreException ex) {
             log.error("Failed to determine compiler test exclusions, assuming defaults", ex);
           }
+          isTestCompilationSkipped = "true" //$NON-NLS-1$
+              .equals(maven.getMojoParameterValue(mavenProject, compile, "skip", String.class, monitor)); //$NON-NLS-1$
```
HannesWell, line 359:
```suggestion
          isTestCompilationSkipped = Boolean.TRUE
              .equals(maven.getMojoParameterValue(mavenProject, compile, "skip", Boolean.class, monitor)); //$NON-NLS-1$
```
The skip field of the compiler-plugin is of type `boolean`:
https://github.com/apache/maven-compiler-plugin/blob/b225b1c29b1ee2fc2b178d2f2f1b920afe1d9a75/src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java#L67-L68

Therefore I think it is better to let maven handle the value conversion.

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
```diff
@@ -363,6 +368,8 @@ protected void addProjectSourceFolders(IClasspathDescriptor classpath, Map<Strin
       for(MojoExecution resources : projectFacade.getMojoExecutions(RESOURCES_PLUGIN_GROUP_ID,
           RESOURCES_PLUGIN_ARTIFACT_ID, mon.newChild(1), GOAL_TESTRESOURCES)) {
         testResourcesEncoding = maven.getMojoParameterValue(mavenProject, resources, "encoding", String.class, monitor); //$NON-NLS-1$
+        isTestResourcesSkipped = "true" //$NON-NLS-1$
+            .equals(maven.getMojoParameterValue(mavenProject, resources, "skip", String.class, monitor)); //$NON-NLS-1$
```
HannesWell, line 372:
```suggestion
        isTestResourcesSkipped = Boolean.TRUE
            .equals(maven.getMojoParameterValue(mavenProject, resources, "skip", Boolean.class, monitor)); //$NON-NLS-1$
```

The skip field of the resources-plugin is of type `boolean`:
https://github.com/apache/maven-resources-plugin/blob/6fce82eb63d03d7efdc993b2e222952507039b28/src/main/java/org/apache/maven/plugins/resources/TestResourcesMojo.java#L60-L61

Therefore I think it is better to let maven handle the value conversion.

HannesWell (commented):
Thanks for the update. This looks very good now, I especially appreciate the addition of even more test cases.

Just one remark regarding the conversion of boolean values below.
If that is applied this is IMO ready.

HannesWell:
Besides the code change, I think this should be mentioned in [M2E's RELEASE_NOTES](https://github.com/eclipse-m2e/m2e-core/blob/master/RELEASE_NOTES.md).
I think the change is reasonable (usually `skipTests` should be used to skip the test execution and they should still be compiled), but I think it could be unexpected for users and therefore it should be mentioned.
@RoiSoleil do you want to add such entry?

RoiSoleil:
@HannesWell it should be ok now ;)

RELEASE_NOTES.md
```diff
@@ -16,6 +16,13 @@ The dependencies editor has been adapted to show all artifacts within a target l
 
 ![grafik](https://user-images.githubusercontent.com/70652978/212153011-160fa96a-1c06-4092-9b89-fcd7a3c2859e.png)
 
+
+### Skip on Tests
+
+Before, the property "<maven.test.skip>true</maven.test.skip>" and the "skip" configurations on "maven-compiler-plugin" and "maven-resources-plugin" were not taken into account by M2E.
+Now, they are well taken into account by M2E and this allows, depending on the need (especially compilation time), to either not compile tests or not copy test resources.
+The folders thus ignored no longer appear in the Package Explorer as "Java" folders but as standard folders.
```
HannesWell, line 24:
```suggestion
### Ignore Test Sources and Resources compilation if `maven.test.skip=true` is specified

The property "<maven.test.skip>true</maven.test.skip>" and the "skip" property in configurations of the "maven-compiler-plugin" and "maven-resources-plugin" are now taken into
account by M2E in the sense that, if enabled, M2E ignores the corresponding folder, which no longer appear in the Package Explorer as "Java" folders but as standard folders.
This allows, depending on the need (especially compilation time), to either not compile tests or not copy test resources.

In general it is not recomended to use the mentioned properties but to use `-DskipTests` instead:
https://maven.apache.org/surefire/maven-surefire-plugin/examples/skipping-tests.html
```

HannesWell (approved):
Looks good now. 👍🏽

I have rephrased the Release-Notes entry a bit and will commit this with your change.

RoiSoleil:
Thank you, it's nice to see eclipse moving forward ;)

PR #1274, opened 2023-02-27
[TP] Revert to gson 2.9.1

HannesWell:
Currently gson 2.10.1 causes trouble in the SimRel. To not further contribute one more possible source of trouble, revert to gson 2.9.


HannesWell:
@laeubi do you have any idea, why Tycho always fails on macos?
Maybe @kwin can check if he can reproduce it locally?

laeubi:
The mac runner uses maven 3.9.0 so we would need at laest Tycho 3.0.3 in the extensions.xml

Issue #1275, opened 2023-02-28
m2e build is triggered by chnage in target folder


laeubi:
I have a multi-module project where random builds are triggered, and even can result in endless build cycles.

According to the Maven Workspace Build View, this is triggered by a change in the target folder:

![grafik](https://user-images.githubusercontent.com/1331477/221766070-7b9c27d9-2a17-4442-bf79-64caf32662c1.png)

One observation is that, weather or not this should trigger a change, even the parent project build is triggered by its child change.

I then open a terminal and touch a random file in the target folder a build is triggered.

laeubi:
To narrow this down I tried to remove the parent project from the workspace but still any change in the `target` folder triggers a build. The build previously has had some source folders that point to a folder inside target, but configuring these to a different location still triggers the build.

laeubi:
I now have checked this with a simple project with only one class and no items configured.
When I compile the project once and then touch the resulting jar in the target folder, each touch results in a workspace build.

mickaelistria:
Is the target folder marked as derived in project metadata? I think it used to be the criterion in order to decide whether a rebuild is necessary or not.

laeubi:
Yes, but actually this does not make any difference. Also it is possible that resources are derived (eg. generated) but still relevant for build input ... as far as I can see currently there is no filtering at all, even if one changes a property of a folder a build is triggered, but I'm currently working on a patch.

laeubi:
@mickaelistria I'm playing around a bit and found that m2e is not getting the change in the .java file but a change in the .class file! That's a bit surprising but maybe the reason why the target folder can't be excluded actually as otherwise m2e won't get a trigger at all ... I'll try to debug this a bit more in detail.

HannesWell:
I encountered that problem as well which resulted sometimes in endless builds for me, because the change in the target-folder done by a executed mojo triggered a new build.
But believed one reason was that the Eclipse-resources/builder did not know that the executed mojo did that change (e.g. because it didn't use the build-context). So maybe a refresh of the project after the m2e-builder had completed (wherever the build-context is read) could help here as well?

laeubi:
I think both should be combined, because even if someone else outside eclipse modifies a file in the target this should normally not trigger a build, I'll try to provide a PR for this and then we can check how it works out, there are just too many (legal) ways a mojo can write some stuff into the target folder that do not require a build, the target holds the result of the build so if the result triggers a build again, something is definitely wrong ...

[...]

PR #1276, opened 2023-02-28
Only trigger (incremental) builds if there are relevant changes

laeubi:
Currently any file change in a project can trigger a rebuild of that project, even if it is from the build-output or from child modules. This can produces "endless" loops of the build and produces a lot of build rush if automatic refresh is activated for the workspace or resources are refreshed on access. On the other hand, without autorefresh some changes are never detected.

This changes the following:
- check if a delta only contains irrelevant deltas and then do not trigger an incremental maven build
- refresh the whole project first to discover new files generated
- if auto refresh is enabled, do not manually refresh but use the native refresh events

This can considerably reduce the number of rebuilds happening.

Fix https://github.com/eclipse-m2e/m2e-core/issues/1275


HannesWell (commented):
This change looks good 👍🏽
And with that at least on my Windows computer my m2e workspace does not go wild anymore and then builds forever until I turn off native-hooks to detect File-system changes.

Great improvement.

PR #1277, opened 2023-02-28
Update to Tycho 3.0.3 (for Maven 3.9.x support)

laeubi:
FYI @HannesWell


laeubi:
The new error is caused by:
- https://github.com/mojohaus/build-helper-maven-plugin/issues/168

so we need to wait for maven 3.9.1 ... but this can still be merged as it does not make things worse.

PR #1280, opened 2023-03-01
Icon_Add_Wizard for m2e

Roncla:
Contributed by Octaviz Studios


laeubi:
@Roncla many thanks, build is not required because only resource changes.

PR #1282, opened 2023-03-01
Proposal for m2e Header Graphics and m2e Logo

Roncla:
Proposal for m2e Header Graphics and m2e Logo

### Variant 1
m2e_Header_orange-blue

### Variant 2
MtoEclipse_Header_red-eclipsePurple

### Variant 1
m2e_Logo_orange-blue

### Variant 2
m2e_Logo_red_eclipsePurple


HannesWell (commented):
I like the proposals. Thank you for this.

What do you others think about fading from the Maven-red from the left/letter m to the blue-purple from the usual eclipse icon (like here https://www.eclipse.org/downloads/assets/public/images/logo-eclipse.png)?

Besides that, could you provide an alternative logo where M2E is written in capital letters?

Roncla:
I have prepared a variant using capital letters.
Imho I do not like this very much since under formal aspects it does appear more pleasing in the non-capital version. In addition the writing of m2e seems to be the common way of using the acronym.

HannesWell:
> I have prepared a variant using capital letters. Imho I do not like this very much since under formal aspects it does appear more pleasing in the non-capital version. In addition the writing of m2e seems to be the common way of using the acronym.

Yes that indeed does not look so well. So I'm fine with staying at the lower case logo.

> What do you others think about fading from the Maven-red from the left/letter m to the blue-purple from the usual eclipse icon (like here https://www.eclipse.org/downloads/assets/public/images/logo-eclipse.png)?

@laeubi what do you think about adjusting the colors? I searched a little bit and got the impression that the red m2 was only used in m2e and is not the 'official' maven color. And searching at their website I didn't find any, only the multi-colored two feathers in their logo at https://maven.apache.org .Maybe would just assume the red tone in the middle of the featers and fade to the eclipse blue described above?

@Roncla would it be possible to create such variants?

assets/m2e_Header_Grafik.png
tivervac, line 1: Quick two cents here, the grey horizontal lines behind the grey `First-class...` text makes the text harder to read than expected. Maybe moving the text a bit might help. If not, what about adding a stoke/outline/shadow to that text?

tivervac, line 1: The white outline on the bottom left of the `2` in `m2eclipse` also looks a bit weird to me. Might be worth having a look at a version without it.

image

Roncla: This lines serves as a divider between the two elements. I provide a version without it here:
image

My favorite is the version with the white line.

Roncla:
Here we go with another variant of the m2e logo. I changed colors according to the mentioned suggestion above.
Hope this meets your expectations.

image

HannesWell:
> Here we go with another variant of the m2e logo. I changed colors according to the mentioned suggestion above.
> Hope this meets your expectations.

Thanks. It is a bit colorful, but I like it. :)
Coincidentally the color-gradient somehow looks a bit similar to the one of the feathers in the Maven logo: https://maven.apache.org/
But I think that fits.

What do the others think?

tivervac:
> What do the others think?

I prefer the red to blue gradient. It is indeed similar to the original maven feather

HannesWell:
I added the new variant to the initial comment (unfortunately I didn't manage quickly to make them the same size).
@eclipse-m2e/technology-m2e-committers what do you think about the proposed two logo variants?
Which color do you prefer? With or without white shadow?

laeubi:
I wonder how this will look on the header graphics background, personally I find the first variant a bit better because I don't see we are required to follow any foreign colour scheme and the red feels a bit aggressive here.

kwin:
To me the first variant looks better too. The second one has too many colors in my opinion.

HannesWell:
> I wonder how this will look on the header graphics background, personally I find the first variant a bit better because I don't see we are required to follow any foreign colour scheme and the red feels a bit aggressive here.

No we are definitely no required to follow any foreign color schema, nevertheless personally I like the idea to make an association through the color.

> Here we go with another variant of the m2e logo. I changed colors according to the mentioned suggestion above. Hope this meets your expectations.
> 

> image

Would it be possible to provide the red-blue variant with the withe shadows, like the orange-blue one was initially?
Furthermore would it be possible to use the purple-ish blue from the Eclipse IDE logo provided at https://www.eclipse.org/org/artwork/ ?
Eclipse2014_RGB

HannesWell:
And regarding the banner: I tried to apply the red-blue color schema with the purple-ish blue of the Eclipse logo on the banner with MS paint (so of course this is bad quality 😅):

grafik

@Roncla can you give a recommendation if and how to integrate the red-blue schema into the banner?
The orange shadow around the `eclipse` is actually an artifact of the fill-function of Paint, however maybe it would be a way to distinguish the text from the background and would somehow reflect the orange shadow around the Eclipse logo (earth-)ball.
On the other hand this would have many colors and could eventually like a website in the 90s.
Personally I like the background and would be fine to leave that as it is.

Alternatively the left part of the banner (`m2eclipse`) could stay orange as proposed initially.


What are the thoughts of the others about the banner?

HannesWell:
@Roncla thanks for the updated banner. I like it. :)

Could you please be so kind and provide an svg only of the logo as it is on the right part of the banner?
It would be great if you could provide it once as it is (with white 'shadows') and once without shadows (as you have already provided for previous variants with different colors).

With that I think we can make the final decisions.

Thank you.

HannesWell:
One week passed and no new opinions were raised, so we have a 2:2 tie.
In order to come to a conclusion I would like to continue with variant 2. I talked to @laeubi and although it is not his favorite, he is fine with that.

@Roncla as a final action, can you please provide the banner in variant 2 as as an svg and (electronically) sign the Eclipse Contributor agreement (ECA): https://www.eclipse.org/legal/ecafaq.php
Otherwise we are not allowed to accept this contribution.

laeubi:
Final PR is here with passing ECA:
- https://github.com/eclipse-m2e/m2e-core/pull/1477

PR #1285, opened 2023-03-02
Slightly adjust the build files

laeubi:
Local artifacts should be ignored, test failures should *not* be ignored instead adjust the action to always upload the test results and let the build just fail at the end.

### DIFF
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -2,5 +2,6 @@
 --errors
 --update-snapshots
 --no-transfer-progress
--Dmaven.test.error.ignore=true -Dmaven.test.failure.ignore=true
+--fail-at-end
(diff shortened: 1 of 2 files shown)


.mvn/maven.config
```diff
@@ -2,5 +2,6 @@
 --errors
 --update-snapshots
 --no-transfer-progress
--Dmaven.test.error.ignore=true -Dmaven.test.failure.ignore=true
```
HannesWell, line 5: The properties maven.test.error/failure.ignore should be passed in Jenkins then in order to have simpler distinctions from build failures. Maybe for the GH Workflow too?

laeubi: No, the worksflows are configured in a way that they always run the post actions.

PR #1286, opened 2023-03-02
Clarify 2.2 release notes

tivervac:
Thanks for another great release, I especially like the Updated Dependency Editor


HannesWell (approved):
Thank you for your contribution and we're glad you like the release.
Yes the new dependency editor is very nice and handy.

Issue #1293, opened 2023-03-03
Launch configuration Executing references non-existing project


laeubi:
Launching under windows a pom.xml either from the context menu or with creating a new launch config results in 

> Launch configuration Executing clean in D__DevFiles_Eclipse_Tools_IndentGuide_net.certiv.tools.indentguide 
> references non-existing project net.certiv.tools.indentguide.

see originally posted by @grosenberg in https://github.com/eclipse-m2e/m2e-core/discussions/1290

laeubi:
FYI @HannesWell Reproducer is:

- Start a dev-workspace from the IDE
- import the project form m2e "m2e-core/m2e-maven-runtime"
- right click on the pom.xml and choose Run As > Maven clean

![grafik](https://user-images.githubusercontent.com/1331477/222782005-e7154964-da2f-4345-9c00-c76abd7becfc.png)

laeubi:
The Stacktrace for the error is:
```

org.eclipse.core.runtime.CoreException: Launch configuration Executing install in_m2e-core_m2e-maven-runtime references non-existing project m2e-maven-runtime.
	at org.eclipse.jdt.launching.JavaRuntime.abort(JavaRuntime.java:1861)
	at org.eclipse.jdt.launching.JavaRuntime.getJavaProject(JavaRuntime.java:1736)
	at org.eclipse.jdt.launching.JavaRuntime.computeVMInstall(JavaRuntime.java:1775)
	at org.eclipse.jdt.launching.AbstractJavaLaunchConfigurationDelegate.getVMInstall(AbstractJavaLaunchConfigurationDelegate.java:129)
	at org.eclipse.m2e.internal.launch.MavenLaunchDelegate.getVMInstall(MavenLaunchDelegate.java:240)
	at org.eclipse.jdt.launching.AbstractJavaLaunchConfigurationDelegate.getBootpathExt(AbstractJavaLaunchConfigurationDelegate.java:363)
	at org.eclipse.jdt.launching.AbstractJavaLaunchConfigurationDelegate.getVMSpecificAttributesMap(AbstractJavaLaunchConfigurationDelegate.java:679)
	at org.eclipse.jdt.launching.JavaLaunchDelegate.getVMRunnerConfiguration(JavaLaunchDelegate.java:92)
	at org.eclipse.jdt.launching.JavaLaunchDelegate.launch(JavaLaunchDelegate.java:160)
[...]

PR #1294, opened 2023-03-03
[Launching] Only set project-name if IProject is existing IJavaProject

HannesWell:
Fixes https://github.com/eclipse-m2e/m2e-core/issues/1293

@laeubi do you want to verify it fixes the issue too?

### DIFF
--- a/org.eclipse.m2e.launching/src/org/eclipse/m2e/internal/launch/MavenLaunchDelegate.java
+++ b/org.eclipse.m2e.launching/src/org/eclipse/m2e/internal/launch/MavenLaunchDelegate.java
@@ -44,6 +44,7 @@ import org.eclipse.core.runtime.preferences.IPreferencesService;
 import org.eclipse.debug.core.ILaunch;
 import org.eclipse.debug.core.ILaunchConfiguration;
 import org.eclipse.debug.core.ILaunchConfigurationWorkingCopy;
+import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.launching.IJavaLaunchConfigurationConstants;
 import org.eclipse.jdt.launching.IVMInstall;
 import org.eclipse.jdt.launching.IVMInstall2;
@@ -228,7 +229,8 @@ public class MavenLaunchDelegate extends JavaLaunchDelegate implements MavenLaun
         }
       }
     }
-    Optional<IProject> project = getContainer(pomDirectory).map(IContainer::getProject);
+    Optional<IProject> project = getContainer(pomDirectory).map(IContainer::getProject)
+        .filter(p -> JavaCore.create(p).exists());
     if(project.isPresent()) {
       // Set the project name so that super.getVMInstall() called below, can find the JDT-Compiler JDK 
       ILaunchConfigurationWorkingCopy workingCopy = configuration.getWorkingCopy();
--- a/org.eclipse.m2e.feature/feature.xml
+++ b/org.eclipse.m2e.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.m2e.feature"
       label="%featureName"
-      version="2.2.0.qualifier"
+      version="2.2.1.qualifier"
       provider-name="%providerName"
       plugin="org.eclipse.m2e.core"
       license-feature="org.eclipse.license"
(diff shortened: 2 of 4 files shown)


laeubi:
I trust you :+1:

org.eclipse.m2e.launching/src/org/eclipse/m2e/internal/launch/MavenLaunchDelegate.java
```diff
@@ -228,14 +229,15 @@ public IVMInstall getVMInstall(ILaunchConfiguration configuration) throws CoreEx
...
+    Optional<IProject> project = getContainer(pomDirectory).map(IContainer::getProject)
+        .filter(p -> JavaCore.create(p).exists());
     if(project.isPresent()) {
       // Set the project name so that super.getVMInstall() called below, can find the JDT-Compiler JDK 
       ILaunchConfigurationWorkingCopy workingCopy = configuration.getWorkingCopy();
       workingCopy.setAttribute(IJavaLaunchConfigurationConstants.ATTR_PROJECT_NAME, project.get().getName());
       configuration = workingCopy;
     }
-    return super.getVMInstall(configuration);
+    return JavaRuntime.computeVMInstall(configuration);
```
kwin, line 240: why this change?

HannesWell: For less indirection and therefore better readability/understandability of the code.
Or do you know a reason that speaks against it?

laeubi: **Just for a future enhancement:**
Maybe we even should not use `JavaRuntime.computeVMInstall(configuration);` but mimic whats done there, e.g. for a maven launch, one could go up in the folder hierarchy and check if the folder maps a project in the workspace and simply use the closest parent project.

Beside that I only wonder why we need to set the project at all, should this not be done automatically somehow before, or was always the eclipse default jre used?

laeubi (approved):

org.eclipse.m2e.launching/src/org/eclipse/m2e/internal/launch/MavenLaunchDelegate.java
HannesWell: Agree with the first part. And since there is the possibility for clean ups in general I reverted this part again for now and made the change minimal, to reduce the chance for new bugs.
Will provide another PR for a later release that makes that more clear in general.

> Beside that I only wonder why we need to set the project at all, should this not be done automatically somehow before, or was always the eclipse default jre used?

I don't think this is set automatically since a launch-config need not to be connected to a project per-se.

laeubi: So where is the project then used? If only internally by m2e, one might specify a custom (m2e scoped) property ...

HannesWell: It is used by `JavaRuntime.computeVMInstall`, which is from JDT.

laeubi: Yes, but actually do we really *want* the (compile) project VM to execute the maven build? e.g. most of my builds currently user Java 11 but Tycho requires Java 17 (what is the default VM), so for the future I think we should use the Workspace default VM as default and the enforcer one if it is specified.

I think this would here also be the place where one might want to select a different project and/or Workspace VM for launching maven builds.

So the order should be:

1. Enforcer VM
2. Project configured m2e VM (not implemented yet --> needs a project setting)
3. Workspace default m2e VM (not implemented yet --> needs a workspace preference)
4. Default VM of the Workspace.

HannesWell: We should also consider that you can specify in the workspace preferences to which VM installation an EE is bound and IIRC that is also considers.

If that is not sufficient it could be done as suggested, but I assume if the JVM really matters one can set it in the launch config and re-use. At least in my use cases the VM to that should run the build often matches the sources Java-level.

Therefore personally I don't find it important to have extra workspace and project preferences for that. But if somebody want's to implement it and provide a PR, I'm fine with it too. :)

laeubi: EE has no meaning for maven.

HannesWell: Yes, but isn't a EE extract by the JavaRuntime class when it determines the compiler-level of the project?

laeubi: Not really, if you go to the settings you can choose a **JRE** based on the EE, so if that changes you don't need to change the config, but you can also choose an explicit one (by ID)

PR #1295, opened 2023-03-03
Fix spelling errors in user-visible error messages

robstryker:


mickaelistria:
@robstryker can you please sign the ECA, or make sure you're using the same email address between your commit and your Eclipse account?
@laeubi See, the baseline hint here is incorrect, it recommends +0.0.100 while m2e would be happy with +0.0.1.

HannesWell:
> @laeubi See, the baseline hint here is incorrect, it recommends +0.0.100 while m2e would be happy with +0.0.1.

I believe that's simply the default of the baseline plugin? Since we are not providing maintenance releases it is not necessary, but on the other hand I think it does not harm to do +100 and we would be prepared, just in case we do it some day.

laeubi:
@mickaelistria can you explain what is particular wrong in this _suggestion_? I just used this as it is very common for eclipse (platform), but of course every change (including +1) is acceptable.

@HannesWell currently it is not configurable but I'll just add a config option now, in the end it doesn't matter much, the idea behind larger increments is just that if you maintain different branches you can increment there by smaller values if you want prepare a new artifact, but thats more a deployment issue as currently there is no easy way to only deploy a single artifact.

laeubi:
@mickaelistria now +1 is _suggested_ for m2e ...

laeubi:
@robstryker thanks for looking into this and providing a patch,

robstryker:
It was a small thing, I know, but every contribution helps! Glad to participate!

PR #1299, opened 2023-03-04
Remove unactionable warnings when a plugin is ignored

### LINKED ISSUE #1298: Remove unactionable warnings when a plugin is ignored by m2e
### Discussed in https://github.com/eclipse-m2e/m2e-core/discussions/1229

Especially @kwin and @laeubi agreed on this proposal to avoid those useless warnings:

> I think the `message` part inside the `ignore` lets m2e emit that warning. I don't think this warning is useful but rather confusing therefore I propose to remove all message lines from
> 
> https://github.com/eclipse-m2e/m2e-core/blob/e3794059b1250729872daecf9c1ff0b35fc75daf/org.eclipse.m2e.core/lifecycle-mapping-metadata.xml#L131

<div type='discussions-op-text'>

<sup>Originally posted by **vtintillier** February  2, 2023</sup>
Hello,

I am not sure if the behavior I see is intended or not.

`maven-enforcer-plugin` removed their `META-INF/m2e/lifecycle-mapping-metadata.xml` file in their last 3.2.1 release (https://github.com/apache/maven-enforcer/pull/192). As a result, we see this warning in all our poms:

```
maven-enforcer-plugin (goal "enforce") is ignored by m2e
```

image

It seems this messages originate from this code:

https://github.com/eclipse-m2e/m2e-core/blob/e3794059b1250729872daecf9c1ff0b35fc75daf/org.eclipse.m2e.core/lifecycle-mapping-metadata.xml#L120-L134

So my question is: is this expected that goals ignored in `m2e-core` own `lifecycle-mapping-metadata.xml` appear as warnings in the editor, whereas goals ignored in maven plugins own `lifecycle-mapping-metadata.xml` do not?
Also is there a way to disable those warnings?

Thanks
Vivien</div>

vtintillier:
Fixes #1298

Those messages appear as warnings in the editor and in the problems view, but there is nothing the user can do to fix them. So we just remove them.

### DIFF
--- a/org.eclipse.m2e.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.m2e.core;singleton:=true
-Bundle-Version: 2.0.6.qualifier
+Bundle-Version: 2.0.7.qualifier
 Bundle-Activator: org.eclipse.m2e.core.internal.MavenPluginActivator
 Bundle-Vendor: %Bundle-Vendor
 Bundle-Localization: plugin
--- a/org.eclipse.m2e.core/lifecycle-mapping-metadata.xml
+++ b/org.eclipse.m2e.core/lifecycle-mapping-metadata.xml
@@ -59,9 +59,7 @@
         <versionRange>[3.4,)</versionRange>
       </pluginExecutionFilter>
       <action>
-        <ignore>
-          <message>maven-dependency-plugin (some goals) are ignored by m2e.</message>
-        </ignore>
+        <ignore />
       </action>
     </pluginExecution>
     <pluginExecution>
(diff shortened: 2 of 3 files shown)


laeubi:
@vtintillier you need to increase the last version part of the `org.eclipse.m2e.core` manifest by +1

kwin (approved):
LGTM, thanks.

laeubi:
@vtintillier It seems you need to adjust the pom.xml version according (@HannesWell I though we already did go full pom less do we really require a pom.xml for m2e.core?)

> Unqualified OSGi version 2.0.7.qualifier must match unqualified Maven version 2.0.6-SNAPSHOT

@kwin if the Jenkinsbuild is green we can merge this, the GH action ones are currently broken because of Maven 3.9.x issues ... I'm working on this.

vtintillier:
Thank you @laeubi, I did not notice the Jenkins build, I only saw the two github actions that failed for other PRs as well 👍

vtintillier:
@laeubi the failed build suggested to bump the feature from 2.2.0 to 2.2.100, but I only bumped to 2.2.1, as it seems previous bumps were +1 and not +100.

laeubi:
@vtintillier thanks for your patience, technical you can choose what ever you feel best, but the feature must also be incremented, usually we choose the next release version that would be 2.3.0 in this case but any other increment would also work here.

HannesWell:
> (@HannesWell I though we already did go full pom less do we really require a pom.xml for m2e.core?)

There were some technical issues that required a pom for m2e.core again. But I can check if I can remove that pom again.

@vtintillier I just submitted another PR that already contains the version bumps for the two features.
You maybe want to rebase your PR on the current master an squash all commits into one (eventually we will squash-merge it anways.

laeubi:
@vtintillier thanks for bringing this up and all your patience  and of course for the fix!

PR #1300, opened 2023-03-05
Add the banner image to the lifcycle mapping page

laeubi:
This adds the banner image for the LifecycleMappingPage to make it look more nice!

This is how it looks under Linux:
grafik

@kwin can you provide an example screenshot with mac os?
@HannesWell can you provide an example screenshot with windows?


HannesWell:
That's how it looks on Windows

grafik

PR #1309, opened 2023-03-09
Provide existing launch configs as shortcuts

laeubi:


laeubi:
Thanks for trying it out, I couldn't really find anything how this is supposed to work, even though it is implemented by some platform items, I can try to improve the platform code if you think it should work as you have described here...

laeubi:
Some finding: Hovering the arrow of the "Run" buttons in the toolbar calls that method to compute a tooltip ....

laeubi:
And a waring to people like me that always forget about that: Never put a breakpoint into menu code :-1:

laeubi:
I now created a PR to make the items available here:
- https://github.com/eclipse-platform/eclipse.platform.debug/pull/122

If this is merged, we can finally implement this in a way that for older eclipse it shows a dialog and for newer eclipse shows the actual shortcuts.

laeubi:
This now looks like this:
grafik

laeubi:
I created a followup here but from m2e side this is done now:
- https://github.com/eclipse-platform/eclipse.platform.debug/pull/164

mickaelistria:
Thanks!

Issue #1328, opened 2023-03-23
Fix and improve m2e "Add ... dependency" quick fix


howlger:
Fix and improve the m2e _"Add ... dependency"_ quick fix that has been introduced by the pull request https://github.com/eclipse-m2e/m2e-core/pull/1048 as follows:

#### (A) Avoid false positives and when adding a dependency also add the required import statement

Example to reproduce: with [these two Maven projects](https://github.com/howlger/Eclipse-IDE-improvements-videos/tree/60d4386a58aa261dc05dbd99d7853be24dff629c/sample_code) and the following sample class  in `src/test/java` of another, unrelated Maven project:
```java
import foo.List;

class SampleTest {

    @Test
//  ^^^^^
//  expected: "Add org.junit.jupiter:junit-jupiter:5.9.2 as dependency"
//  actual: no m2e quick fix
    void sample() {
        
        ModelgenMojo m= null;
//      ^^^^^^^^^^^^
//      expected and actual: "Add com.example:modelgen-mojo:1.0-SNAPSHOT as dependency"
//      correctly adds Maven dependency, but adding import statement would also be nice

        List l= null;
//      ^^^^
//      expected: no m2e quick fix, since no dependency contains "foo.List" (qualified name known by import statement)
//      actual: "Add com.example:Eclipse_2023-03_Improvements:0.0.1-SNAPSHOT as dependency"

        BiMap<String, String> p= null;
//      ^^^^^
//      expected: no m2e quick fix, since BiMap is contained in the system library only, not in a known Maven dependency
//      actual: "Add com.example:modelgen-mojo:1.0-SNAPSHOT as dependency"
        
    }

}
```

#### (B) Make sure this quick fix does not cause performance issue

See https://github.com/eclipse-m2e/m2e-core/issues/879#issuecomment-1478251753 (unclear how this can be reproduced).

#### (C) Show qualified name and offer more options where to search

Maybe by a _"Add Maven dependency containing 'List'..."_ quick fix rather than one or more _"Add ... dependency"_ quick fixes that opens a separate dialog which also provides an option whether to search only in the local Maven repository or also search.maven.org.

#### (D) Provide this or a similar quick fix on import statements that cannot be resolved 

See also:
- https://github.com/eclipse-m2e/m2e-core/issues/623
- https://github.com/eclipse-m2e/m2e-core/issues/879#issuecomment-1470177817

Issue #879, opened 2022-08-26
Quickfix to add a maven dependecy automatically to the target if a missing type is found


laeubi:
Most people have a lot of stuff already in their local maven repository, and we can use this rich meta-data for the following purpose:

1. Index all artifacts and their classes
2. If a users uses a Type found in the local repository we could suggest to add the item to the target, reload the target platform, and add an appropriate import

HannesWell:
In general a good proposal.

>     1. Index all artifacts and their classes

But creating and maintaining the index sounds like a very expensive operation and we don't get notified about all changes in the local maven repo.
Also the question is if that index is saved somewhere (potentially large) or recreated in memory everytime the IDE is started (potentially long running).

The good news is that [search.maven.org](https://search.maven.org/) already maintains such index and provides access to it via their REST API:
https://central.sonatype.org/search/rest-api-guide/

<html><body>


https://search.maven.org/solrsearch/select?q=c:junit&rows=20&wt=json | Mimics searching by classname in Advanced Search.  Returns a list of artifacts, down to the specific version, containing  the class.
-- | --

</body>
</html>

lukaseder:
I suspect this feature has shipped already, in Eclipse 2023-03, for example? I personally don't really like it. I hardly ever need to add new dependencies, and if I do, I want to do that manually, and carefully. Yet, every time I quick fix, I get tons of these m2e options that I never select:

![image](https://user-images.githubusercontent.com/734593/225352778-147dfa9a-dd49-41ee-bfd4-5657d54cea97.png)


It would be great if I could deactivate this addition. See also: https://stackoverflow.com/q/75746202/521799

laeubi:
@lukaseder you could provide a patch for a configuration option to disable this if you like.

howlger:
To support different ways of working out of the box, there are also the following options:
- Propose only _Add Maven dependency containing 'List'..._ that opens a separate dialog
- Move the proposal(s) to the bottom

laeubi:
I think the best would be if there is a check if it could be fixed by a simple import (what is represented as first choice here) because then most likely one don't want to add new dependencies.

Weather or not one "like it" or not, if it is more or less important (should be more on top or more on bottom) also depends a bit on the user ... so if one like a configuration option seems both easy to archive and most flexible.

[...]

PR #1329, opened 2023-03-23
Disable "Add ... dependency" quick fix due to bad user feedback

howlger:
The m2e _"Add ... dependency"_ quick fix seems to have some issues that should be fixed before re-enabling it: see issue https://github.com/eclipse-m2e/m2e-core/issues/1328

Feedback from dissatisfied users:
- https://github.com/eclipse-m2e/m2e-core/issues/879#issuecomment-1470177817
- https://github.com/eclipse-m2e/m2e-core/issues/879#issuecomment-1478251753
- https://stackoverflow.com/q/75817981/6505250

### DIFF
--- a/org.eclipse.m2e.jdt.ui/plugin.xml
+++ b/org.eclipse.m2e.jdt.ui/plugin.xml
@@ -336,6 +336,7 @@
             label="java">
       </keyword>
    </extension>
+   <!--
    <extension
          point="org.eclipse.jdt.ui.quickFixProcessors">
       <quickFixProcessor
@@ -343,7 +344,8 @@
             id="org.eclipse.m2e.jdt.ui.quickFixProcessor1"
             name="M2Eclipse">
       </quickFixProcessor>
-   </extension>   
+   </extension>
+   -->
 	<extension point="org.eclipse.ui.perspectiveExtensions">
 	   <perspectiveExtension
 	         targetID="org.eclipse.jdt.ui.JavaPerspective">


laeubi (changes requested):
Completely removing the feature is not a solution. Please add a m2e preference to enable / disable the feature via the UI.

howlger:
> Completely removing the feature is not a solution. Please add a m2e preference to enable / disable the feature via the UI.

I really like the idea of this feature. I don't want to remove it, nor do I want a preference to disable it (a preference wouldn't change that some users would be unsatisfied in the first place, and they would need to know that and where to disable it). In my view the problem is that there are currently false positive proposals. Even if almost all proposals are provided correctly, a few false positive proposals will result in a poor user experience. Performance might also be a blocker.

I opened issue https://github.com/eclipse-m2e/m2e-core/issues/1328 to fix and improve the feature which I guess will not be that easy. Maybe it would be better to provide a quick fix first on import statements that can't be resolved since the qualified name is known. I'm not a Maven power user, so I don't know what is the best and makes sense here. I was hoping to hear from @lukaseder and @kjaranso about what would be helpful to them, rather to hear what they don't like.

It would be great if the false positives will be fixed already in the next release. Otherwise, if more time is needed, I would appreciate if this feature would be disabled until the false positives are fixed.

lukaseder:
> I was hoping to hear from @lukaseder and @kjaranso about what would be helpful to them, rather to hear what they don't like.

Well, I would definitely love to be able to control what is displayed in quick fixes, just like in the Windows Explorer context menu. Every app/plugin/etc. has interesting ideas to contribute to a central menu, but users have very different tastes. In any case, this wouldn't be m2e specific, and perhaps a bit too broad, so I get what you're really looking for here.

What would be helpful to me:

1. I think the m2e suggestion should mostly be the lowest priority in a quick fix popup simply based on the fact that it is unlikely the most popular choice. Here's an example:

image

Sure, *maybe* I want to add a dependency, but very likely, I just want to create the type, too. There are some type names that are *very* likely from a dependency, like e.g. `Flux` from reactor. But many others aren't.

Also, how many times per day do I?

- Import stuff (all the time)
- Create types (often, but less often)
- Add dependencies (almost never, and if I do, I edit the pom.xml manually, because in my case, it's not trivial to decide where to add the dependency (e.g. in the project itself, or as a transitive dependency of the project, etc.))

2. Even if it was an occasionally good choice to add the dependency, perhaps the problem (when I first asked about this on Stack Overflow) was the overwhelming amount of options that distract me from what I *really* want to do. E.g. here's what IntelliJ does. It simply lists "Add Maven dependency..." (as a last option, even for `Flux`!):

image

I don't know if this is possible in Eclipse (a quick fix that opens up a new wizard / popup, rather than executing the action directly), but for this particular purpose, I think it would be better.

laeubi:
A QuickFix can do literally anything, so it is just a matter of implementation :-)

Sadly there only a quite vague "priority" one can use to control the position of a quickfix and also no way like for autocomplete to have distinct categories (e.g Java, E4, SWT fixes).

kjaranso:
Well, as I had mentioned before, my biggest issue was the performance. It was literally taking it over 2 minutes to return after trying to get a quick fix, the entire while eclipse was unresponsive. That's not acceptable for me. I'm not sure if that can be solved by caching everything ahead of time or using some other search mechanism. The amount of time it took varied depending on which import it was looking for, too. I'm not sure what was causing some to take longer than others, though. The Builder lombok annotation was the one that took the longest for me.

I've been using the quick fix how it's been working for years, so I'm never expecting to have it find something I don't have included in my dependencies already. It does seem like it could be convenient to have it do that for me, but at least half the time the dependency won't be found by this new scheme since it won't be in my project already. Maybe if when you try to do the quick fix and it doesn't find something, it could ask if you want to search for a dependency in your project? That way the user can decide? I don't know if you could configure it so that is optional. You could have 3 modes:
1. Always look for imports from outside project dependencies
2. Never look for imports from outside project dependencies
3. Prompt to look for imports from project dependencies if no current project dependencies are found

Does that sound reasonable? I don't know how difficult it is to implement things with this plugin, so change it if that makes more sense to you.

That being said, if you're able to figure out the performance issue such that it is never slow giving a response, then maybe those options aren't needed.

laeubi:
I'm closing this in favor of 
- https://github.com/eclipse-m2e/m2e-core/pull/1409

where this feature can be disabled on demand.

PR #1330, opened 2023-03-23
Update 'm2e-core-tests' submodule to remove embedded test artifacts

HannesWell:
Integrate changes of sub-module in:
- https://github.com/tesla/m2e-core-tests/pull/200.

### DIFF
--- a/org.eclipse.m2e.tests.common/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.tests.common/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: M2E Testing Helpers
 Bundle-SymbolicName: org.eclipse.m2e.tests.common;singleton:=true
-Bundle-Version: 2.0.3.qualifier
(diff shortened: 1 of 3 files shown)


org.eclipse.m2e.tests.common/src/org/eclipse/m2e/tests/common/HttpServer.java
```diff
@@ -657,17 +657,26 @@ public void handle(String target, Request baseRequest, HttpServletRequest reques
       for(String contextRoot : resourceDirs.keySet()) {
         String path = URIUtil.decodePath(trimContextRoot(uri, contextRoot));
         if(path != null) {
+
+          if(HttpMethod.CONNECT.asString().equals(request.getMethod())) {
```
HannesWell, line 661: @laeubi can you tell if this is the right way to handle a CONNECT request (for example to `https://repo1.maven.org/maven2/`) in this local proxy server?

HannesWell: Never mind, I just restored the few embedded artifacts for this case and reverted the change.

PR #1343, opened 2023-04-11
Use LemMinX-Maven 0.7.2

vrubezhny:


HannesWell:
The `testOpenChildThenParentResolvesParent` test case often fails since the upgrade to Maven 3.9.1. Locally it could not pre reproduced yet.
Maybe you want to give it a try although the failure is likely unrelated.

vrubezhny:
@HannesWell Yes I think so. I tried to reproduce it locally, but I wasn't able. Also, I see some other not related PRs with the same failing test, so I think the `testOpenChildThenParentResolvesParent` test failure is not really related to this one.

HannesWell (approved):
> I tried to reproduce it locally, but I wasn't able.

Too bad.

> Also, I see some other not related PRs with the same failing test, so I think the `testOpenChildThenParentResolvesParent` test failure is not really related to this one.

Yes absolutely.

mickaelistria:
Thanks guys for the patch, review and discussion!

PR #1344, opened 2023-04-11
[Build] Replace deprecated GabrielBB/xvfb-action by coactions/setup-xvfb

HannesWell:


HannesWell:
That's interesting. Do you know if there is then an equivalent GH-action like [setup-xvfb ](https://github.com/coactions/setup-xvfb) and a Jenkins Plugin like [xvfb](https://www.jenkins.io/doc/pipeline/steps/xvfb/)?

mickaelistria:
Back then, I searched and found no GitHub actions; but 4 lines of bash for mutter is IMO more interesting than 1 line of GH action for Xvfb.

laeubi:
> Back then, I searched and found no GitHub actions; but 4 lines of bash for mutter is IMO more interesting than 1 line of GH action for Xvfb.

I don't think so, please take into account that the action do much more than "just starting xvfb", e.g. it detects if you run on windows and turn it there into a noop. So if you think mutter is superior it might be good to report it to the people of that action so they can consider using mutter as an alternative.

mickaelistria:
> much more than "just starting xvfb", e.g. it detects if you run on windows and turn it there into a noop

That's not really much more, adding a `if: ${{ runner.os == 'Linux' }}` should do te trick.

>  So if you think mutter is superior

It's not just me, it's the consensus for all mainstream recent linux distributions who've stopped using X by default.

> it might be good to report it to the people of that action so they can consider using mutter as an alternative.

I don't really care about the action, I'm pretty happy with the steps described above which give more control.

laeubi:
> That's not really much more, adding a if: ${{ runner.os == 'Linux' }} should do te trick.

If the trick is to not execute anything at all on other OS (what about mac?) ... because this will now make your whole command a noop, not only skip install/start of mutter...

HannesWell:
> > Back then, I searched and found no GitHub actions; but 4 lines of bash for mutter is IMO more interesting than 1 line of GH action for Xvfb.
> 
> I don't think so, please take into account that the action do much more than "just starting xvfb", e.g. it detects if you run on windows and turn it there into a noop. So if you think mutter is superior it might be good to report it to the people of that action so they can consider using mutter as an alternative.

Created an issue: https://github.com/coactions/setup-xvfb/issues/16
Please share your opinion there as I cannot really say much about the topic.

PR #1345, opened 2023-04-11
[Setup] Simplify working set definition

HannesWell:


setup/m2e.setup
```diff
@@ -168,44 +169,39 @@
...
-      <description>The dynamic working sets for ${scope.project.label}</description>
-    </setupTask>
+      xsi:type="setup.workingsets:WorkingSetTask">
+    <workingSet
+        name="m2e-core"
+        id="m2e.core">
+      <predicate
+          xsi:type="predicates:AndPredicate">
+        <operand
+            xsi:type="predicates:RepositoryPredicate"
```
HannesWell, line 179: @merks it looks like the `RepositoryPredicate` ignores sub-modules.
While this is correct, it would be sometimes convenient to be able to match projects from submodules too.
Do you think that would be possible?

merks:
I don't understand sub-modules very well.  The implementation in org.eclipse.oomph.predicates.impl.RepositoryPredicateImpl.getRepoDirAbsolutePath(IProject) does not rely on any Git dependencies and org.eclipse.oomph.predicates.impl.RepositoryPredicateImpl.matches(IResource) would need some adjusting.  Maybe the first method should get a list of containing repos.  But that maybe one doesn't always want to consider matches from other modules.  Probably I need to see an example and use case; I'm traveling all next week...

HannesWell:
> I don't understand sub-modules very well. The implementation in org.eclipse.oomph.predicates.impl.RepositoryPredicateImpl.getRepoDirAbsolutePath(IProject) does not rely on any Git dependencies and org.eclipse.oomph.predicates.impl.RepositoryPredicateImpl.matches(IResource) would need some adjusting. Maybe the first method should get a list of containing repos. 

Thanks for the pointers. I had a quick look but didn't find a quick solution either since I've not yet worked with E/JGit. I just noticed that `GitProjectData` has a method `hasInnerRepositories()` that can tell if a repo has either `submodules` or contains a nested git repo, but I didn't find a way to access that.

> But that maybe one doesn't always want to consider matches from other modules. 

Yes considering sub-modules should be optional, probably opt-in.

> Probably I need to see an example and use case; I'm traveling all next week...

Thanks! I can create a follow up on this PR to show how I would like to use it. But it is not urgent. This is working now. :)

Btw. is there already a plan when to migrate Oomph to GitHub?

merks:
If all goes according to my plans, I will have/make time to migrate Oomph to github after the 2023-06 release winds down and before 2023-09 winds up...  I'll probably migrate the incubator first to get started if there appears to be some free time in my schedule...

I wonder if a submodule is always a direct child of the "parent" git repository such that we could just look one level in.  I wonder if submodules can have submodules?

merks:
If we added "includeNestedRepositories", and we look at the two repository locations, x and y, that come back and check (using Path.startsWith) whether y.startWith(x) to determine if the x repository path contains the y repository path.  That would work if one chooses as the prototype a project from the root repository...

HannesWell:
> If all goes according to my plans, I will have/make time to migrate Oomph to github after the 2023-06 release winds down and before 2023-09 winds up... I'll probably migrate the incubator first to get started if there appears to be some free time in my schedule...
> 
Great. Looking forward to it. :)

> I wonder if a submodule is always a direct child of the "parent" git repository such that we could just look one level in. 

According to https://git-scm.com/book/en/v2/Git-Tools-Submodules
`By default, submodules will add the subproject into a directory named the same as the repository, in this case “DbConnector”. You can add a different path at the end of the command if you want it to go elsewhere.`

From that I assume that in general a submodule can be located anywhere. But I would assume that it has to be somewhere inside the parent repo.
If one uses submodules there is a `.gitmodules` in the repository-root (I would assume always), that lists all submodules and there paths. For m2e it looks like:
```
[submodule "m2e-core-tests"]
	path = m2e-core-tests
	url = https://github.com/tesla/m2e-core-tests.git
	branch = master
```
But I assume that there is some API for that in EGit so that information can probably be obtained from some object somehow referenced from the `GitProvider`.

...

After some search I found that there is the `SubmoduleWalk` class in JGit that provides an API to iterate over submodules.
I'm currently working this out and will provide a Gerrit.



> I wonder if submodules can have submodules?

In the link mentioned above there is the following section, I think yes.

`There is another way to do this which is a little simpler, however. If you pass --recurse-submodules to the git clone command, it will automatically initialize and update each submodule in the repository, including nested submodules if any of the submodules in the repository have submodules themselves.`

HannesWell:
Created https://bugs.eclipse.org/bugs/show_bug.cgi?id=581814 and https://git.eclipse.org/r/c/oomph/org.eclipse.oomph/+/201339.

PR #1348, opened 2023-04-14
Update mockito to 5.3.0 and logback-classic to 1.2.12

merks:


merks:
I see this doesn't work.  I'll leave this in your capable hands, because I don't have more time.

Note that in the direct-from-maven repository, that missing logback dependencies is satisfied like this:

image

I guess the mockito thing will be fine and I should not have lumped them together with this one.

target-platform/target-platform.target
```diff
@@ -66,7 +66,7 @@
 				<dependency>
 					<groupId>ch.qos.logback</groupId>
 					<artifactId>logback-classic</artifactId>
-					<version>1.2.11</version>
+					<version>1.4.6</version>
```
HannesWell, line 69:
```suggestion
					<version>1.2.12</version>
```
Since logback 1.3 slf4j-2 is required, but we cannot use that yet in the Eclipse platform since mina-sshd is not compatible.
I have implemented a fix in https://github.com/apache/mina-sshd/pull/336 but it has not been released.

See also https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/981.

I'll update to logback 1.4 as soon as the new mina-sshd release is available, but until that we have to stay with 1.2 :/
The compiler error is then easy to fix :)

HannesWell (changes requested):
Thanks Ed!
Unfortunately we cannot use logback 1.3/1.4 yet.

HannesWell (approved):
Looks good now and the test-failure is unrelated.

Issue #1353, opened 2023-04-17
build-helper-maven-plugin:add-resource goal behaves differently


laeubi:
Consider the following pom:

```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>
  <groupId>test</groupId>
  <artifactId>buildhelper.test</artifactId>
  <version>0.0.1-SNAPSHOT</version>
  <build>
    <plugins>
      <plugin>
        <groupId>org.codehaus.mojo</groupId>
        <artifactId>build-helper-maven-plugin</artifactId>
        <version>3.3.0</version>
        <executions>
          <execution>
            <id>add-source</id>
            <phase>generate-sources</phase>
            <goals>
              <goal>add-source</goal>
            </goals>
            <configuration>
              <sources>
                <source>target/generated-sources</source>
              </sources>
            </configuration>
          </execution>
          <execution>
            <id>add-resource</id>
            <phase>generate-resources</phase>
            <goals>
              <goal>add-resource</goal>
            </goals>
            <configuration>
              <resources>
                <resource>
                  <directory>target/generated-resources</directory>
                </resource>
              </resources>
            </configuration>
          </execution>
        </executions>
      </plugin>
    </plugins>
  </build>
</project>
```

This will result in the following `.classpath` entry to be added:
```
<classpathentry kind="src" output="target/classes" path="target/generated-sources">
  <attributes>
	  <attribute name="optional" value="true"/>
	  <attribute name="maven.pomderived" value="true"/>
  </attributes>
</classpathentry>
```

now if the folder is created as part of the build it appears in the project.

In contrast the second execution (that adds a resource folder) is not added (because the folder does not exits) and also creating it does not make it available, one first need to create the folder, then run "Update Project" that result in the following entry:

```
<classpathentry excluding="**" kind="src" output="target/classes" path="target/generated-resources">
  <attributes>
	  <attribute name="maven.pomderived" value="true"/>
  </attributes>
</classpathentry>
```

as one can see it is missing the  `<attribute name="optional" value="true"/>` value, therefore removing the folder afterwards (e.g. as part of `mvn clean`) then one gets an error
[...]

PR #1354, opened 2023-04-17
Always add resource folders with optional attribute

laeubi:
Resource folders should be handled like a source folder that is always added with optional=true (even if the folder is not present yet) so they appear in the project as soon as they are created.

Fix https://github.com/eclipse-m2e/m2e-core/issues/1353

### DIFF
--- a/org.eclipse.m2e.jdt/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.jdt/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.m2e.jdt;singleton:=true
-Bundle-Version: 2.0.6.qualifier
+Bundle-Version: 2.3.0.qualifier
 Bundle-Localization: plugin
 Export-Package: org.eclipse.m2e.jdt,
  org.eclipse.m2e.jdt.internal;x-friends:="org.eclipse.m2e.jdt.ui",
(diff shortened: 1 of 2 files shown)


HannesWell (approved):
Sounds reasonable, but this creates new test-failures, that have to be fixed.

laeubi:
As a first step I now have split a change that allows to perform an more weak assert, this should hopefully not fail any test but is required to prepare test adjustments:
- https://github.com/eclipse-m2e/m2e-core/pull/1355

laeubi:
Alright this already lowered the failure count but still things left... will check that.

laeubi:
There is one test left that do not fails because of the classpath order:

`org.eclipse.m2e.tests.BuildPathManagerTest.test431080_flatDirectoryLayout()`

this one is a bit tricky, because currently the resourcefolder is silently skipped (as it does not exits), no it is added leading to
grafik

configuring this to the basedir (like the others as well) seems make it being ignored by **maven**
grafik

so it seems such a "flat" layout is actually not very well supported (I think @fbricon have already mentioned this), also importing the project leads to very strange behavior, in contrast to having the project root a new folder is created that is the full path of the project on disk (what is empty) like here:
grafik

So for me it seems the test is already broken by design and we probably should `@Ignore` it (+create an issue about supporting flat layouts) unless we have a good idea how to support such flat layouts, from a first though about this it seems to me that in such a case we should simply just have one source folder that points to the project root and skip all additional source and resource folders ... but thats something I defiantly not plan to implement in this PR...

HannesWell:
> 
> So for me it seems the test is already broken by design and we probably should `@Ignore` it (+create an issue about supporting flat layouts) unless we have a good idea how to support such flat layouts, from a first though about this it seems to me that in such a case we should simply just have one source folder that points to the project root and skip all additional source and resource folders ... but thats something I defiantly not plan to implement in this PR...

Sounds good 👍🏽

PR #1355, opened 2023-04-18
Relax the ClasspathHelpers to provide a minimal mapping to the cp

laeubi:
Currently the ClasspathHelpers#assertClasspath is quite strict and requires two things:

1) the patterns are exactly the same size as the classpath entries 2) the patterns occur in the same order as the classpath entries

This is quite fragile in terms of changes in the classpath e.g. if additional items are there.

This realx the method in the following way:
- it is only checked that each pattern can be matched
- order is no longer relevant
- return a map that contains the mapping


HannesWell (approved):
Have not tried this out, but the change looks reasonable in the Webinterface.
As long as all tests pass like before, I'm fine with this.

laeubi:
We seem to have one flapping? test in Github but it also has failed on master, so nothing "new", then lets go with this and give it a try!

PR #1358, opened 2023-04-20
Include the classifier when creating an ArtifactKey from a org.apachemaven.artifact.Artifact

TheHound:
Include the classifier when creating an ArtifactKey from a org.apachemaven.artifact.Artifact to avoid issues with same dependency but different classifier & scope combinations.

### DIFF
--- a/org.eclipse.m2e.core.tests/resources/projects/classifier/pom.xml
+++ b/org.eclipse.m2e.core.tests/resources/projects/classifier/pom.xml
@@ -1,0 +1,25 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>com.example</groupId>
+    <artifactId>test</artifactId>
+    <version>0.0.1-SNAPSHOT</version>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.example</groupId>
+            <artifactId>dependency-1</artifactId>
+            <version>0.0.1-SNAPSHOT</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.example</groupId>
+            <artifactId>dependency-1</artifactId>
+            <version>0.0.1-SNAPSHOT</version>
+            <classifier>classifier</classifier>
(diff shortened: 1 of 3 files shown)


laeubi:
@TheHound can you explain the specific problem this solves? Could it probably even be demonstrated as a junit test (e.g. in [org.eclipse.m2e.core.tests](https://github.com/eclipse-m2e/m2e-core/tree/master/org.eclipse.m2e.core.tests)?

TheHound:
Have updated the commit with a test case demonstrating the expected change in artifact portable strings

laeubi (approved):
Looks good to me but as its is a quite central change will request a review from some other commiters.

HannesWell (approved):
This looks like a reasonable change for me.

@mickaelistria what do you think?

HannesWell:
@mickaelistria, @akurtakov what do you think about this?
We plan do perform a release next week and I think it would be good to have this included.

mickaelistria (approved):
I don't see a reason against this change; unless test do show some related error.

laeubi:
It seems a versionbump is required, beside this I think it should be included in this release. FYI @fbricon

fbricon:
I *think* the classifier was not considered initially due to workspace project resolution. You need to ensure a workspace project A depending on workspace B and B:tests won't break for instance. 
https://github.com/tesla/m2e-core-tests/blob/0652381e7da3c6dfa5bb5e864e215da786bb2efd/org.eclipse.m2e.tests/src/org/eclipse/m2e/tests/BuildPathManagerTest.java#L740-L764 might be relevant

HannesWell:
> I _think_ the classifier was not considered initially due to workspace project resolution. You need to ensure a workspace project A depending on workspace B and B:tests won't break for instance. https://github.com/tesla/m2e-core-tests/blob/0652381e7da3c6dfa5bb5e864e215da786bb2efd/org.eclipse.m2e.tests/src/org/eclipse/m2e/tests/BuildPathManagerTest.java#L740-L764 might be relevant

The test passes in the CI. So should we still consider this or better have another solution?

fbricon:
I haven't tested that change. If you're confident it's safe, then go ahead with it.

HannesWell:
> I haven't tested that change. If you're confident it's safe, then go ahead with it.

Me neither and since I and nobody else considered the case you mentioned I'm not that confident.
Nevertheless in order to make progress here. but in a safer way I suggest to merge this after the upcoming release is out (and we are sure we don't contribute a follow up to Sim-Rel). Then we have some time to test the change via snapshots and get more confidence for the next release.

HannesWell:
Let's see how this works IRL.
Thanks @TheHound!

PR #1359, opened 2023-04-21
Use LemMinX-Maven 0.8.1 + Lemminx 0.25.0 + WWD 1.1.0

vrubezhny:


org.eclipse.m2e.repository/category.xml
```diff
@@ -27,7 +27,7 @@
    <bundle id="com.google.gson"/>
    <bundle id="com.google.gson.source"/>
    <category-def name="m2e" label="Maven Integration for Eclipse"/>
-   <repository-reference location="https://download.eclipse.org/wildwebdeveloper/releases/1.0.3/" enabled="true" />
+   <repository-reference location="https://download.eclipse.org/wildwebdeveloper/snapshots/" enabled="true" />
```
HannesWell, line 30: Can you tell if there is a wwd release planned soon?
Because we can't release m2e if we have snapshot dependencies.

vrubezhny: I'd like to see a m2e-core release (2.2.2?)  that includes #1343, so we can have a full set of Lemminx 0.24.0, Lemminx-maven 0.7.2 and WWD 1.0.5 - a last fully compatible versions before switching to Lemminx 0.25.0 (which is not API-compatible).

This PR is only a draft as of yet, made to prove that the API incompatibilities brought by switching to Lemminx 0.25.0 are fixed. So, I'm not going to suggest switching to using snapshots - I'll switch it back to using releases as soon as the new versions of WWD (planned 1.1.0) and probably Lemminx-Maven (current 0.8,0 or most likely planned 0.8.1) released.

Thanks

vrubezhny: @HannesWell Do you know if m2e-core 2.2.2 is going to be released anyhow soon?

vrubezhny: @HannesWell As I wrote above I'd like to see a full set of Lemminx 0.24 compatible products: LemMinX-Maven 0.7.2 + WWD 1.0.5 + m2e-core 2.2.2 (or 2.3.0 or whatever you name it) . Imho, these versions will be good to have in 2023-06 M2. 

Currently Lemminx 0.25 is available, but this release is not API-compatible with the listed versions, so I'm preparing the newer compatible versions of LemMinX-Maven (0.8.x) and WWD (1.1.0) as well as a draft PR  for m2e-core - I'm not ready to release yet, but probably I'll need to release LemMinX-Maven and WWD soon, in hope of inclusion of these to next (after m2e-core 2.2.2) and thus having again a full set of compatible versions based on Lemminx 0.25.

That's my plan. But if the next release of m2e-core is not going to happen soon, then I'd skip getting the set of Lemminx-0.24 based versions and go directly to releasing and including the lemminx 0.25 compatible ones. 

Sorry for a long read.

HannesWell: > @HannesWell As I wrote above I'd like to see a full set of Lemminx 0.24 compatible products: LemMinX-Maven 0.7.2 + WWD 1.0.5 + m2e-core 2.2.2 (or 2.3.0 or whatever you name it) . Imho, these versions will be good to have in 2023-06 M2.

Do you have a greater need for that or would it be just nice to have?
If you really need a release with the deps you mentioned, I can create a release, but if it is just nice to have for you I would like to save the extra work and do the next release in time for the 2023-06 RC1/2 in about a month.

For m2e we are happy to just consume the latest available releases of LemMinX(-Maven) or WWD as soon as they are available, regardless if they break the API or not. If there is an API breakage it would be nice if you provide a PR with the necessary changes (if there are any) and just lower bounds of the imported-packages/required-bundles. :)

HannesWell (approved):
Thanks for the update, it looks good.

> UPD: Well, it looks like now it takes 2-3 times longer for Error markers to appear on a resource - in my local environment it was taking about 25-35 seconds while there was only 10 seconds timeout value previously used in assert. I've changed the timeout value to 60 seconds which fixed the `EditorTest.testOpenChildThenParentResolvesParent`.

Thanks for fixing the test. I only wonder if the reason for the delayed marker occurrence is a hint for a performance regression?

vrubezhny:
> Thanks for the update, it looks good.

Thanks for merging!

> Thanks for fixing the test. I only wonder if the reason for the delayed marker occurrence is a hint for a performance regression?

Honestly. I don't know what the reason is. Some time ago I heard something on some "slowness"-like issues with maven 3.9.x, but I really cannot be 100% sure that this one is anyhow related to those problems. It can also be some eclipse changes or lemminx/lemminx-maven ones (as lemminx-maven makes more POM processing than previous versions).

Issue #1360, opened 2023-04-24
Security Best Practices: CycloneDX Maven Plugin for generating SBOMs


ghost:
Hi,

As part of the ongoing effort to improve the security of the project from the Security Team of the Eclipse Foundation, we are adding the [CycloneDX Maven Plugin](https://github.com/CycloneDX/cyclonedx-maven-plugin) to the project's POM file.

This plugin will help to generate a Software Bill of Materials (SBOM), which will provide greater visibility into the project's dependencies and enhance the overall security.

We welcome any feedback or questions about this update, so please don't hesitate to reach out if there are any concerns or if there's anything we can help with. 

Thanks.

Kind Regards,
Francisco Perez

laeubi:
The plugin is known to not work well for Tycho / Eclipse projects and the maintainer are not planning to add support here, so I don't expect much value here for m2e or any Eclipse Project unless major partsa re moved to plain maven dependencies:
- https://github.com/CycloneDX/cyclonedx-maven-plugin/issues/137

fperezel:
Hi @laeubi ,

Thanks, for the feed back, I am having a look into the issue pointed out as well as I have already updated the PR providing the outcome requested for the last commit.

I feel that regardless the cyclonedx may not suit for Tycho / Eclipse for this project provides a helpful information about all dependencies from security point of view, what do you think?

Kind Regards,

laeubi:
I think cyclone should at least support `pkg:p2` as a type or otherwise this might give a false impression that m2e uses a lot less dependencies as it is actually the case.

Another question would be who is supposed to read that data? As m2e is mostly used inside Eclipse adding that data (should it become part of the jar?) seems a bit superfluous if there is no plan for Eclipse/P2 to read that data and do anything useful with it, and if used by maven these tools might have no idea about the  `pkg:p2` type...

waynebeaton:
Try adding `-Dtycho.target.eager=true`.

fperezel:
Hi,

Closed the PR https://github.com/eclipse-m2e/m2e-core/pull/1361 due to currently we don't have a solution for tycho based builds. I am wondering if you prefer to keep this ticket open until we have a solution. Do you prefer to leave opened?

Thanks,

PR #1361, opened 2023-04-24
Applying Security Best Practices: SBOM plugin (#1360)

fperezel:
Adding the [CycloneDX Maven Plugin](https://github.com/CycloneDX/cyclonedx-maven-plugin) to the project's POM file.

resolve: #1360

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -157,6 +157,35 @@
 					</environments>
 				</configuration>
 			</plugin>
+    		<plugin>
+    		    <groupId>org.cyclonedx</groupId>
+    		    <artifactId>cyclonedx-maven-plugin</artifactId>
+				<version>2.7.0</version>
+    		    <configuration>
+    		        <projectType>library</projectType>
+    		        <schemaVersion>1.4</schemaVersion>
+    		        <includeBomSerialNumber>true</includeBomSerialNumber>
+    		        <includeCompileScope>true</includeCompileScope>
+    		        <includeProvidedScope>true</includeProvidedScope>
+    		        <includeRuntimeScope>true</includeRuntimeScope>
+    		        <includeSystemScope>true</includeSystemScope>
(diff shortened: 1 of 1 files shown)


laeubi (changes requested):
Please provide example output and describe how this helps here as cyclone does not support Eclipse/Tycho build projects see:

- https://github.com/CycloneDX/cyclonedx-maven-plugin/issues/137

fperezel:
> Please provide example output and describe how this helps here as cyclone does not support Eclipse/Tycho build projects see:
> 
> * [Cyclonedx sbom for tycho dependencies CycloneDX/cyclonedx-maven-plugin#137](https://github.com/CycloneDX/cyclonedx-maven-plugin/issues/137)

Hi,

First at all, thanks for the feedback

Finally, please find attached a zip file with the output of local test related to the latest update, 
[bom.zip](https://github.com/eclipse-m2e/m2e-core/files/11310476/bom.zip)

Thanks,

laeubi:
@fperezel I'll try to take a look into this, it would be good to explain who is using this information, e.g. is it expected to be included as a resource in the final jar? If yes we might need to increase versions for all modules to make the change effective.

fperezel:
> @fperezel I'll try to take a look into this, it would be good to explain who is using this information, e.g. is it expected to be included as a resource in the final jar? If yes we might need to increase versions for all modules to make the change effective.

Hi, answering first question about "who is using this information". I may say any consumer of the software according all recommendation and regulations related to BOM files

Regarding if it should be included as resource in the final jar. From my point of view, I do recommend to include BOM files in each final jar in order to any consumer of any of jar can check it.

Finally, I have pushed a final pom version file where CycloneDX goal has been update makeBom after tested out this option. So that could the latest pom file version and it is ready for reviewing for a possible merge.

Cheers

PR #1379, opened 2023-05-19
[Maven-Target] Simplify PDE target code and make activator obsolete

HannesWell:
Instead of clearing stale entries from the cache when the application exists, clean the cache on first access.

As discussed we I'm currently also working on Tests for Maven-Targets. They should be submitted before this PR in order to ensure that this does not break anything.


org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/BNDInstructions.java
```diff
@@ -13,63 +13,46 @@
...
-		}
+	private static Properties asProperties(String instructions) {
 		Properties properties = new Properties();
-		try (reader) {
-			properties.load(reader);
-		} catch (IOException e) {
-			throw new RuntimeException("conversion to properties failed", e);
+		try {
+			properties.load(new StringReader(instructions));
+		} catch (IOException e) { // Cannot happen
```
laeubi, line 50: if it cannot happen then an AssertionError should be thrown!

org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/CacheManager.java
```diff
@@ -129,55 +128,49 @@ public MavenTargetBundle getTargetBundle(Artifact artifact, BNDInstructions bndI
...
+			if (bundle == null) {
+				throw new IllegalStateException(CacheManager.class.getSimpleName() + " not loaded from a bundle");
+			}
+			baseDir = bundle.getDataFile("").toPath();
+			// clear all locations older than 14 days, this can be improved by
+			// 1) watch for changes in the workspace -> if target is deleted/removed from
+			// workspace we can clear the cache
+			// 2) we can add a preference page where the user can force clearing the cache
+			// or set the cache days
+			ForkJoinPool.commonPool().execute(() -> clearFilesOlderThan(baseDir, Duration.ofDays(14)));
```
laeubi, line 171: better use a dedicated thread here instead of the common pool

org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/BNDInstructions.java
HannesWell: Done.

org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/CacheManager.java
HannesWell: What would be the benefit?

laeubi: Usually the common pool assumes that no blocking happens (e.g. wait for I/O) and your work is completely CPU bound, otherwise you might degrade the performance of other items using the common pool, see for example: 
https://www.jrebel.com/blog/parallel-java-streams

So for one-shot I/O performing it is generally better to just fire up a separate thread.

HannesWell: Understand, makes sense. Thanks for the elaboration.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/ClipboardParser.java
```diff
@@ -70,16 +57,15 @@ private void parseElement(Element element) {
...
 	}
 
 	private String getTextFor(String element, Element doc, String defaultValue) {
 		NodeList nl = doc.getElementsByTagName(element);
 		Node item = nl.item(0);
 		if (item != null) {
-			String v = Objects.requireNonNullElse(item.getTextContent(), defaultValue);
-			if (!v.isBlank()) {
+			String v = item.getTextContent();
+			if (v != null && !v.isBlank()) {
```
laeubi, line 68: this do no longer return the defaultvalue when the text content is missing...

laeubi (commented):
Some files seem have only the year changed in the header but nothing else.

org.eclipse.m2e.pde.ui/src/org/eclipse/m2e/pde/ui/target/editor/ClipboardParser.java
HannesWell: Isn't that happening below in line 72?

HannesWell:
> Some files seem have only the year changed in the header but nothing else.

Don't know how this happened, anyway I reverted those year-only changes.

PR #1382, opened 2023-05-20
Add Tests for o.e.m2e.pde.targets

HannesWell:


org.eclipse.m2e.pde.target.tests/src/org/eclipse/m2e/pde/target/tests/MavenFeatureTest.java
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	@Test
+	public void testPomFeatureGeneration() {
+		// TODO Auto-generated method stub
+	}
+
+	@Test
+	public void testFeatureArtifact() {
+		// TODO Auto-generated method stub
```
HannesWell, line 87: @laeubi do you now a Eclipse Feature available as artifact from Maven-Central?

laeubi: You can search for "feature" in central and look for items ending with ".feature" and having a jar, here is an example of an very old one:

https://central.sonatype.com/artifact/org.drools/org.drools.eclipse.task.feature/5.4.0.CR1/versions

PR #1386, opened 2023-05-22
[Maven-Target] Fix detection of changed BND-Instructions and move wrapped artifact cache to workspace

HannesWell:
This consist of two changes:

1. Fix detection of changed BND-Instructions
2. Move cache for OSGi-wrapped Maven artifacts to workspace
The mentioned cache is bound to a workspace resource and therefore it makes sense to locate in the workspace's scope. This also has the advantage that the cache is deleted automatically if the workspace is deleted.

@laeubi any objections against moving the cache?

### DIFF
--- a/org.eclipse.m2e.pde.target.tests/src/org/eclipse/m2e/pde/target/tests/OSGiMetadataGenerationTest.java
+++ b/org.eclipse.m2e.pde.target.tests/src/org/eclipse/m2e/pde/target/tests/OSGiMetadataGenerationTest.java
@@ -183,7 +183,47 @@ public class OSGiMetadataGenerationTest extends AbstractMavenTargetTest {
 		assertNull(sourceAttributes.getValue(Constants.EXPORT_PACKAGE));
 		assertNull(sourceAttributes.getValue(Constants.REQUIRE_BUNDLE));
 		assertNull(sourceAttributes.getValue(Constants.DYNAMICIMPORT_PACKAGE));
+	}
 
+	@Test
+	public void testNonOSGiArtifact_missingArtifactGenerate_changedCustomInstructions() throws Exception {
+		String targetXML = """
+				<location includeDependencyDepth="none" includeDependencyScopes="" includeSource="true" missingManifest="generate" type="Maven">
+					<dependencies>
+						<dependency>
+							<groupId>com.google.errorprone</groupId>
+							<artifactId>error_prone_annotations</artifactId>
+							<version>2.18.0</version>
+							<type>jar</type>
+						</dependency>
+					</dependencies>
+					<instructions><![CDATA[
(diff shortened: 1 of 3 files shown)


laeubi:
That means that if you switch workspace you can't reuse the wrapped items what will duplicate a lot of contents, I'm currently working on a way to store the items in the maven repository what seems a more valid location anyways.

laeubi:
> You already get the duplicate since within the Cache-basedir the wrapped artifacts are scoped/sorted into directories derived from the handle's memento hash:

You can use the same target in different workspaces :-)

> Generally this sound good, but wouldn't that make it much harder to evict outdated wrappings?

The cache is simply never evicted (as it is the case with a regular maven repo), but one can delete the maven local repo if desired to reset everything.

HannesWell:
> > You already get the duplicate since within the Cache-basedir the wrapped artifacts are scoped/sorted into directories derived from the handle's memento hash:
> 
> You can use the same target in different workspaces :-)
> 
That's right.

> > Generally this sound good, but wouldn't that make it much harder to evict outdated wrappings?
> 
> The cache is simply never evicted (as it is the case with a regular maven repo), but one can delete the maven local repo if desired to reset everything.

While this makes sense if the instructions are stable, one can really litter the .m2-cache on the computer if trying out many different instructions on a large artifact.
Therefore I think there should be some sort of clean up. Maybe not as general as before.
Maybe a good compromise would be to only clean up other variants of the same artifact (with that GAV), when checking for a wrapping of it. This way it would at least be cleaned up occasionally.

Since with https://github.com/eclipse-m2e/m2e-core/pull/1387 the considerations of changed instructions is already fixed and the cache location is already moved I'm closing this one.
The new test case comes with https://github.com/eclipse-m2e/m2e-core/pull/1391.

laeubi:
> While this makes sense if the instructions are stable, one can really litter the .m2-cache on the computer if trying out many different instructions on a large artifact.

Wrapping in targets is not meant as a general purpose tool to experiment with BND instructions, and it is not meant to wrap big artifacts in general, adding metadata in the first place is always preferable.

> Therefore I think there should be some sort of clean up. Maybe not as general as before.

Maven do not do any cleanups so I don't think we should invest anything here, if people are concerned they can still cleanup their m2 repo by any suitable tool.

> Maybe a good compromise would be to only clean up other variants of the same artifact (with that GAV), when checking for a wrapping of it. This way it would at least be cleaned up occasionally.

Technically one can never know if an artifact might be required later on (e.g. you might have one build that wraps it with an instruction A while another uses B), therefore any heuristic might fail sooner or later. As wrapping is a very costly operation caching should be as aggressive as possible, especially given that wrapping should be rare and most artifacts are quite small.

PR #1391, opened 2023-05-22
Maven target clean ups and test about consideration of changed BND-instructions

HannesWell:
This PR adds the test-case initially created for https://github.com/eclipse-m2e/m2e-core/pull/1386 about consideration of changed BND-instructions.
Furthermore it cleans up the code added in https://github.com/eclipse-m2e/m2e-core/pull/1387 a bit and fixes two deprection warnings.

### DIFF
--- a/org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/shared/WrappedBundle.java
+++ b/org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/shared/WrappedBundle.java
@@ -23,12 +23,12 @@ import aQute.bnd.osgi.Jar;
 
 public final class WrappedBundle {
 
-	private DependencyNode node;
-	private List<WrappedBundle> depends;
-	private String instructionsKey;
-	private Path file;
-	private Jar jar;
-	private List<ProcessingMessage> messages;
+	private final DependencyNode node;
+	private final List<WrappedBundle> depends;
+	private final String instructionsKey;
+	private final Path file;
+	private final Jar jar;
+	private final List<ProcessingMessage> messages;
 
 	WrappedBundle(DependencyNode node, List<WrappedBundle> depends, String key, Path file, Jar jar,
 			List<ProcessingMessage> messages) {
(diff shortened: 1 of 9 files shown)


org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/shared/WrappedBundle.java
```diff
@@ -14,72 +14,22 @@
...
 	/**
 	 * @return the messages that where produced
 	 */
 	public Stream<ProcessingMessage> messages() {
-		return Stream.concat(messages.stream(), depends.stream().flatMap(dep -> dep.messages()));
-	}
-
-	@Override
-	public int hashCode() {
-		return Objects.hash(instructionsKey, node);
```
HannesWell, line 67: As far as I can tell this class (now record) is not used as key in any map.
Therefore I think it is not necessary to override hashCode/equals.

laeubi: The idea was/is that code can use this to decide if two items are equal without having to know the internal state, even though it is currently not used (at m2e) in such a way.

org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/shared/WrappedBundle.java
```diff
@@ -14,72 +14,22 @@
...
-		return jar;
-	}
-
-	/**
-	 * @return the file where the wrappes bundle is located
-	 */
-	public Path getFile() {
-		return file;
-	}
+public record WrappedBundle(DependencyNode node, List<WrappedBundle> depends, String instructionsKey,
```
laeubi, line 23: This should not be a record as it is not a pure container, this makes all its members public to the outside world what is not desired.

org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/shared/WrappedBundle.java
HannesWell: Acknowledged.

PR #1392, opened 2023-05-22
Do not embed guice in the runtime bundle

laeubi:
google guice is already a bundle, it should therefore not be required to embed it into the m2e.runtime, especially as we then export a package again.

This will reduce the size of m2e runtime, allows reuse of exiting guice bundles and prevents us from the need to bump the bundle less often in case of updates.

### DIFF
--- a/m2e-maven-runtime/org.eclipse.m2e.maven.runtime/pom.xml
+++ b/m2e-maven-runtime/org.eclipse.m2e.maven.runtime/pom.xml
@@ -19,7 +19,7 @@
 	</parent>
 
 	<artifactId>org.eclipse.m2e.maven.runtime</artifactId>
-	<version>3.9.100-SNAPSHOT</version>
+	<version>3.9.200-SNAPSHOT</version>
 	<packaging>jar</packaging>
 
 	<name>M2E Embedded Maven Runtime (includes Incubating components)</name>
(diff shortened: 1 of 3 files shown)


m2e-maven-runtime/org.eclipse.m2e.maven.runtime/pom.xml
```diff
@@ -19,7 +19,7 @@
 	</parent>
 
 	<artifactId>org.eclipse.m2e.maven.runtime</artifactId>
-	<version>3.9.100-SNAPSHOT</version>
+	<version>3.9.200-SNAPSHOT</version>
```
HannesWell, line 22:
```suggestion
	<version>3.9.101-SNAPSHOT</version>
```
This version is meant to be aligned with the embedded Maven version, which is 3.9.1 at the moment.
Since we had a version bump in this release cycle already this is not necessary.

HannesWell (commented):
Since guice is part of the Maven distribution and its version therefore controlled with it, I'm not convinced this is a good idea.
Especially because with Guice 7 the javax-annotations are not supported anymore in favor of jakarta-annotations and I don't know if Guice 6 will receive updates and when Maven Plugins will move away from javax-annotations (I suspect it will take some time).
And for Guice-5 we have the difficulties with wrapped artifacts...

But what we should do is removing the export of Guice.

PR #1393, opened 2023-05-22
[Maven-Target] Exclude sources of excluded artifacts too

HannesWell:
Currently sources of excluded artifacts are still contained in the target.

### DIFF
--- a/org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/MavenTargetLocation.java
+++ b/org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/MavenTargetLocation.java
@@ -360,12 +360,12 @@ public class MavenTargetLocation extends AbstractBundleContainer {
 			if (includeSource) {
 				try {
 					List<ArtifactRepository> repositories = getAvailableArtifactRepositories(maven);
-					Artifact resolve = RepositoryUtils.toArtifact(
+					Artifact sourceArtifact = RepositoryUtils.toArtifact(
 							maven.resolve(artifact.getGroupId(), artifact.getArtifactId(), artifact.getBaseVersion(),
 									artifact.getExtension(), "sources", repositories, new NullProgressMonitor()));
-					MavenSourceBundle sourceBundle = new MavenSourceBundle(bundle.getBundleInfo(), resolve,
+					MavenSourceBundle sourceBundle = new MavenSourceBundle(bundle.getBundleInfo(), sourceArtifact,
 							cacheManager);
-					targetBundles.addBundle(resolve, sourceBundle);
+					targetBundles.addBundle(sourceArtifact, sourceBundle);
 					targetBundles.addSourceBundle(artifact, sourceBundle);
 				} catch (Exception e) {
 					// Source not available / usable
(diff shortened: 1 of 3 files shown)


org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/MavenTargetLocation.java
```diff
@@ -208,9 +208,7 @@ private void generateFeature(TargetBundles bundles, boolean source) throws CoreE
 		} else {
 			bundleFilter = Predicate.not(bundleFilter);
 		}
-		Iterable<TargetBundle> featurePlugins = bundles.bundles() //
-				.filter(e -> !isExcluded(e.getKey()) && !isIgnored(e.getKey()))//
```
HannesWell, line 212: ignored artifacts are not added to TargetBundles.bundles so we don't have to check if they are ignored.

laeubi:
This change seem to do more than than excluding source bundles, we must always add all items to the (internal) state and only exclude them from the target state (therefore the filtering on getBundles / getFeature) because otherwhise the user can't see the items in the UI to (un) exclude them.

org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/MavenTargetLocation.java
```diff
@@ -334,6 +332,9 @@ private File unpackFeatureFile(Artifact artifact, CacheManager cacheManager) {
 
 	private void addBundleForArtifact(Artifact artifact, CacheManager cacheManager, IMaven maven,
 			TargetBundles targetBundles, IProgressMonitor monitor) {
+		if (isExcluded(artifact)) {
```
laeubi, line 335: Excluded items must be part of the `TargetBundles` as we want to still show them as ignored in the UI!

HannesWell:
> This change seem to do more than than excluding source bundles, we must always add all items to the (internal) state and only exclude them from the target state (therefore the filtering on getBundles / getFeature) because otherwhise the user can't see the items in the UI to (un) exclude them.

For the UI the DependencyNode tree seems to be used and even with this change the Editor works as expected, at least for
```
<location includeDependencyDepth="infinite" includeDependencyScopes="compile,runtime" includeSource="true" missingManifest="ignore" type="Maven">
	<dependencies>
		<dependency>
			<groupId>com.google.inject</groupId>
			<artifactId>guice</artifactId>
			<version>7.0.0</version>
			<type>jar</type>
		</dependency>
	</dependencies>
	<exclude>com.google.code.findbugs:jsr305:3.0.1</exclude>
</location>
```

grafik

laeubi:
As said the target bundle is meant to hold the whole state, if one wants to filter out things it should be filtered. e.g. there is `getMavenTargetBundle(Artifact)` that should return the target bundle regardless if it is ignored or not, also `getMavenTargetBundle(MavenTargetDependency dependency)` depends on this and so on.

If you want to filter out the sources of ignored items, then `isIgnored(...)` should be enhanced instead or even better `TargetBundles.bundles.ignoredArtifacts` should also contain the source artifact.

Issue #939, opened 2022-09-26
Automatically add configured toolchains


laeubi:
Currently one needs to configure JVMs in maven toolchains.xml and eclipse independently.

It would be good if m2e adds configured toolchains automatically to the list of JDT ones, at best these are just "readonly" items with a type = "Maven Toolchain":

![grafik](https://user-images.githubusercontent.com/1331477/192197949-a38e3d7d-7fb3-4fcb-acd1-0dc317eef92d.png)

The extension point to use would be `org.eclipse.jdt.launching.vmTypes`
Maven Modello reader is `org.apache.maven.toolchain.building.ToolchainsBuilder` and the actual items are XppDom items, so maybe it would be usefull to simply use a DomParser directly instead.

`org.eclipse.jdt.internal.debug.ui.jres.InstalledJREsBlock.VMLabelProvider` could be enhanced to support custom Label/Images

According to `org.eclipse.jdt.internal.debug.ui.jres.InstalledJREsBlock.enableButtons()` contributed VMs cant be removed, but copied and edited, we should check if there is something needed to update VMs in the toolchains or disable such features.

fbricon:
When it comes to automatic JRE management, [JRE Discovery ](https://github.com/sidespin/jre-discovery)does the job for me. I thought about adding a toolchains configurator, but since toolchains.xml already points to managed JDKs, I didn't feel like moving forward. But it'd be doable.
 
https://fbricon.github.io/posts/Eclipse-automatically-detect-managed-JREs/

laeubi:
That looks great, I still think it would be good to have toolchains.xml support because it also covers non managed setups and seems a natural addition to m2e and would work cross-platform without any effort, do not requires configuration and so on.

As you seem to have a lot of experience in that area, would you mind to possibly rethink adding this to m2e even though it won't give much more value in managed setups?

fbricon:
Feel free to open a PR to add toolchain support in JRE Discovery.
You can still fork or somehow duplicate the discovery functionality into m2e, but that wouldn't be my choice.

mickaelistria:
+1 to have m2e automatically adding toolchains as installed JRE in the IDE (when no installed JRE already points to the toolchain location), that is a natural addition.

mickaelistria:
> It would be good if m2e adds configured toolchains automatically to the list of JDT ones, at best these are just "readonly" items with a type = "Maven Toolchain":

[...]

PR #1404, opened 2023-05-30
Add JVM according to toolchains

mickaelistria:
This is current not functional as toolchains is not set in the MavenExecutionRequest.
This would require all consumers of MavenExecutionRequestPopulator in m2e to also invoke the `populateToolchains` method.

Or alternatively, we can rewrite all this code to bypass the execution request and just read the toolchains.xml file to retrieve the model and use it directly.

See https://github.com/eclipse-m2e/m2e-core/issues/939

### DIFF
--- a/org.eclipse.m2e.feature/feature.xml
+++ b/org.eclipse.m2e.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.m2e.feature"
       label="%featureName"
-      version="2.3.0.qualifier"
+      version="2.3.1.qualifier"
       provider-name="%providerName"
       plugin="org.eclipse.m2e.core"
       license-feature="org.eclipse.license"
--- a/org.eclipse.m2e.jdt/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.jdt/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.m2e.jdt;singleton:=true
-Bundle-Version: 2.3.0.qualifier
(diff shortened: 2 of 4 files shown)


mickaelistria:
I need some piece of advice here. What is best and affordable? Populate toolchains in m2e core, or just reading the toolchains model from file in the JDT integration?

laeubi:
I think `m2e.jdt` is the right place here for the code, but don't think one really needs an execution here because the best we can do is using the global tool-chains file (that might be made configurable like it is done with global/user settings in `m2e.core`).

So if we assume there is `IMavenSettings#getToolchainsFile()` backed by the preferences i would simply create a `new org.apache.maven.toolchain.io.DefaultToolchainsReader()` and read the file directly.

mickaelistria:
Ack, I will rework this patch according to your feedback.

HannesWell:
> the best we can do is using the global tool-chains file (that might be made configurable like it is done with global/user settings in m2e.core).

I think we can also consider a toolchain.xml configured in the `.mvn/maven.config` of the current context as we already started to do with properties from that file (but IIRC it is not yet fully functional).

mickaelistria:
> I think we can also consider a toolchain.xml configured in the .mvn/maven.config of the current context

As far as I know, at that stage, we don't have have a current context, we're not in the scope of a particular project or build.

laeubi:
> > I think we can also consider a toolchain.xml configured in the .mvn/maven.config of the current context
> 
> As far as I know, at that stage, we don't have have a current context, we're not in the scope of a particular project or build.

Correct, this should not (and cannot be) project specific.

mickaelistria:
I reported https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/248 which is IMO the root cause of the failing tests. I'll try to implement a workaround in this proposal in the meantime.

mickaelistria:
Those are the same failures with GitHub actions on master, let's merge.

PR #1406, opened 2023-06-02
Limit LSP4E dependency to version <= 0.23 due to the API--breaking ch…

vrubezhny:
…anges

### DIFF
--- a/org.eclipse.m2e.editor.lemminx/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.editor.lemminx/META-INF/MANIFEST.MF
@@ -2,14 +2,14 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: M2E Maven POM File Editor using Wild Web Developer, LemMinX and Maven LS extension (requires Incubating components)
 Bundle-SymbolicName: org.eclipse.m2e.editor.lemminx;singleton:=true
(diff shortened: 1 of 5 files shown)


laeubi:
why not update m2e to the latest changes and raise the **lower** version?

vrubezhny:
> why not update m2e to the latest changes and raise the **lower** version?

When a user having 2.3.0 installed updates to the latest lsp4e snapshot (which is allowed due to no restriction exists) - he/she gets broken lemminx editor.  

Imho, better is to restrict m3e 2.3.x  to `o.e.lsp4e` v.<=0.17 (which all should be API compatible) and then start updating m2e 2.4.x to the latest changes rising the lower version as you suggest.

vrubezhny:
@akurtakov @laeubi OK. I got your point. So, this PR is to be canceled. 

Sticking to providing the changes need to fix API incompatibility and limiting `lower` `o.e.lsp4e` version to be `0.17`.

laeubi:
Just from OSGi POV it would be much better if LSP4j would version the packages using [semantic-versioning](https://docs.osgi.org/whitepaper/semantic-versioning/) and m2e would import-package the used API with consumer-range, that way there is much less problems in that area.

Depending on how lsp4j is build it would also good to use the new tycho-basline plugin to ensure API compatibility on the package level.

vrubezhny:
Closing this PR in favor of #1407

PR #1407, opened 2023-06-02
Use LSP4E v.0.23.0 and adapt to comming API-breaking changes

vrubezhny:


vrubezhny:
Note that, as discussed in #1406 , it's  not possible to correctly set the lower limit for LSP4E plugin dependency in m2e-core because we're on 0.23 release of LSP4E, not a snapshot, so we can't reference a not released version

laeubi:
@vrubezhny can you explain what is the purpose of this change if there is nothing released yet?
If the code change is already in an released version it should be fine to use that version as a lower bound.

vrubezhny:
@laeubi The `org.eclipse.lsp4e.LanguageServers` API which is used now in `o.e.m2e.editor.lemminx` was introduced since  `o.e.lsp4e` v.0.16.0 , so I've updated the lover limit for this dependency to be "0.16.0" instead of "0.13.7". 

Limiting to "0.17.0" is not possible and not needed off course and came from WWD which depends on LSP4E master (installed from a snapshot).

laeubi:
> The org.eclipse.lsp4e.LanguageServers API which is used now in o.e.m2e.editor.lemminx was introduced since o.e.lsp4e v.0.16.0 , so I've updated the lower limit for this dependency to be "0.16.0" instead of "0.13.7".

This sound reasonable and is the way to go I think so from m2e POV everything should be safe for future versions.

laeubi (approved):
Looks sane from my side but @mickaelistria should review as well.

mickaelistria:
Thanks @vrubezhny for the fix and @laeubi for the review (here and on previous patch). I think the current patch is by far the best approach and this seems correct (and build and tests are still working). So let's merge.

PR #1411, opened 2023-06-05
Consume dependencies of maven archetype from the target platform

laeubi:
Currently we use a special packed "uber-bundle" that contains everything as library entries. This makes the installation and building quite complicated as we need to take special care regarding making the source available. Also upgrading individual items is not possible without version bumps, on the other hand the m2e-target support already offers all we need here and can even generate a feature out of the selected items.

This changes the structure in the following way:
- mark the m2e archetype bundle with extensible-api header
- add all dependencies of maven archetype as fragment to the target
- remove the embedded jars


laeubi:
Jenkins build failure seems unrelated.

HannesWell (commented):
Since the m2e.archetype common is now empty and only serves as Fragment host
I think we should better find another host and delete it.
Maybe those artifacts also work well with a non-flat classpath? Then we could just use them as regular bundels?

laeubi:
The problem is that this will pull in a lot of other dependencies (e.g velocity) or we need to very specifically craft the headers and review the results, so my intend here was to mainly get rid of the embedded jars but retain current behavior.
Currently only `m2e.ui` seem using the classes, but we can think about making `org.eclipse.m2e.archetype.common` a regular bundle and move some classes there (e.g. `IArchetype` and alike)

org.eclipse.m2e.archetype.common/META-INF/p2.inf
```diff
@@ -0,0 +1,14 @@
+requires.0.namespace = osgi.bundle
```
HannesWell, line 1: Can't we add generic provided capabilities to the fragments and require that in the manifest instead of using p2-instructions?

org.eclipse.m2e.archetype.common/.classpath
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-17">
+		<attributes>
+			<attribute name="module" value="true"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry kind="src" path="src"/>
```
HannesWell, line 9: It looks like this Plugin currently does not have a source folder and therefore it is not created in the IDE when one checks out this project, so I assume this will lead to an error on a clean workspace?

HannesWell (commented):
> The problem is that this will pull in a lot of other dependencies (e.g velocity) or we need to very specifically craft the headers and review the results, so my intend here was to mainly get rid of the embedded jars but retain current behavior. 

That's right, but we do that for the aggregated Plugin that embedded the jars already:

https://github.com/eclipse-m2e/m2e-core/blob/ea3b6cf122d75d4773ee0be445c24f1827ecd8ba/m2e-maven-runtime/org.eclipse.m2e.archetype.common/pom.xml#L144-L148

My guess is that those instructions can be simply copied to the TP and BND will select what's suitable for those bundles and it should work like now.

> Currently only `m2e.ui` seem using the classes, but we can think about making `org.eclipse.m2e.archetype.common` a regular bundle and move some classes there (e.g. `IArchetype` and alike)

If we can make the archetype-artifacts plain bundles I would just leave those classes in m2e.ui until one needs them somewhere else.

laeubi:
> My guess is that those instructions can be simply copied to the TP and BND will select what's suitable for those bundles and it should work like now.

I have tried that first, but this currently only works because we have a flat classpath and nothing is really imported.

> Can't we add generic provided capabilities to the fragments and require that in the manifest instead of using p2-instructions?

I don't see that we gain anything from that beside making things more complex, as we would need one capability per fragment, that is only fulfilled by the fragment itself while we actually only want that if the base is installed it also installs all fragments.

laeubi:
Lets see how this works out, we can change / improve it later on anyways :-)

PR #1418, opened 2023-06-12
Add latest release update-site of m2eclipse

laeubi:
With m2eclipse we try to fix issues found as soon as possible and to regular releases also between the simrel-cycle, but many users can only benefit from that after a next quarterly release as they often just use the eclipse defaults, even though we ask users whenever possible to add the m2eclipse update sites, it seems more valuable to have these added by default so users get notified also between releases about new updates and we get a faster feedback-cycle.

As discussed here: https://github.com/eclipse-packaging/packages/pull/33


laeubi:
P2 will inform the user if it tries to install in incompatible version, so updates are not applied automatically. So nothing will "break" just because the repository is available.

If you look here: https://github.com/eclipse-m2e/m2e-core#-installation the only option currently is `snapshot` or `latest` and I really don't see that we support any older update versions anyways.

If this is a real concern, we can provide a `https://download.eclipse.org/technology/m2e/releases/2.x/latest/` but as said I currently don't see we will support others than latest anyways.

HannesWell:
> P2 will inform the user if it tries to install in incompatible version, so updates are not applied automatically. So nothing will "break" just because the repository is available.
> 
That's right, but it requires that everybody uses  version-ranges as recommend (but not always done).


> If you look here: https://github.com/eclipse-m2e/m2e-core#-installation the only option currently is `snapshot` or `latest` and I really don't see that we support any older update versions anyways.
> 

That's right, but one can use a repo of e specific version.
Yes we don't support other version than latest, but if one just wants to use a specific version, he/she gets the latest repo added, which has to be removed again if not wanted. And that removal is unfortunately cumbersome.

> If this is a real concern, we can provide a `https://download.eclipse.org/technology/m2e/releases/2.x/latest/` but as said I currently don't see we will support others than latest anyways.

I would be in favour of that, ideally having a structured composite repo. But we probably need better support for that in Tycho. At the moment it is quite cumbersome to build.

laeubi:
> That's right, but it requires that everybody uses version-ranges as recommend (but not always done).

But that's not the task of m2e to support such bad practice.

> That's right, but one can use a repo of e specific version.

Theoretically yes, but I suspect not much people are doing that anyways as we do not promote these, and old repos where deleted/moved in the past as well, so actually there is only latest.

> Yes we don't support other version than latest, but if one just wants to use a specific version, he/she gets the latest repo added, which has to be removed again if not wanted. And that removal is unfortunately cumbersome.

If a user don't want it it can be removed or disabled, I think that's acceptable.

> I would be in favour of that, ideally having a structured composite repo. But we probably need better support for that in Tycho. At the moment it is quite cumbersome to build.

Tycho has nothing to do with deployments, and one don't really needs to build a composite, the question is more if there is someone dedicated to perform all this steps for something a bit hypothetical case with only very small benefits.

PR #1434, opened 2023-06-15
Require Maven 3.9.0 or newer due to usage of SyncContextFactory only

kwin:
introduced there

Compare with discussion at
https://github.com/eclipse-m2e/m2e-core/discussions/1433


laeubi (approved):
good idea

HannesWell (approved):
Yes this totally makes sense for the reason you mentioned in the discussion.
Thanks!

PR #1435, opened 2023-06-15
Support launch without pom.xml

### LINKED ISSUE #1427: Regression in launching without a project location
We have code that runs `archetype:crawl` which doesn't require a project, and we set the ATTR_POM_DIR to

```java
workingCopy.setAttribute(MavenLaunchConstants.ATTR_POM_DIR, "${workspace_loc}");
```
This has been this way for probably 5 years or more with no issues. Upon running this launch code in the latest Eclipse 4.28, the following error appears:
```
java.lang.IllegalArgumentException: Path must include project and resource name: /pom.xml
	at org.eclipse.core.runtime.Assert.isLegal(Assert.java:66)
	at org.eclipse.core.internal.resources.Workspace.newResource(Workspace.java:2164)
	at org.eclipse.core.internal.resources.Container.getFile(Container.java:196)
	at org.eclipse.m2e.internal.launch.MavenLaunchDelegate.readEnforcedJavaVersion(MavenLaunchDelegate.java:248)
	at org.eclipse.m2e.internal.launch.MavenLaunchDelegate.getVMInstall(MavenLaunchDelegate.java:224)
	at org.eclipse.jdt.launching.AbstractJavaLaunchConfigurationDelegate.getBootpathExt(AbstractJavaLaunchConfigurationDelegate.java:363)
	at org.eclipse.jdt.launching.AbstractJavaLaunchConfigurationDelegate.getVMSpecificAttributesMap(AbstractJavaLaunchConfigurationDelegate.java:679)
	at org.eclipse.jdt.launching.JavaLaunchDelegate.getVMRunnerConfiguration(JavaLaunchDelegate.java:92)
	at org.eclipse.jdt.launching.JavaLaunchDelegate.launch(JavaLaunchDelegate.java:160)
	at org.eclipse.m2e.internal.launch.MavenLaunchDelegate.launch(MavenLaunchDelegate.java:133)
```

Even if you launch using the UI, setting "project location" to `${workspace_loc}`, the launch dialog does not tell you this isn't allowed and you are able to launch, getting the same error in the Error Log immediately. You can't even properly edit the configuration in this state, every action you will get a popup error message.

image

This is a regression caused by https://github.com/eclipse-m2e/m2e-core/commit/32dc6407c919bb8eb8217df4a3ccddff18ab1ed4, and m2e no longer supports running such goals without a project (even though Maven does).

Please provide a temporary workaround in the meantime.

## System Information

Eclipse Platform
Version: 2023-06 (4.28)
Build id: I20230605-0440

Eclipse.org - m2e	M2E Maven Integration for Eclipse Core	2.3.0.20230523-2033	org.eclipse.m2e.feature
Eclipse.org - m2e	M2E Maven Integration for Eclipse Core	2.2.100.20230523-1752	org.eclipse.m2e.pde.feature

kwin:
This closes #1427

### DIFF
--- a/org.eclipse.m2e.launching/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.launching/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.m2e.launching;singleton:=true
-Bundle-Version: 2.0.500.qualifier
+Bundle-Version: 2.0.600.qualifier
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.core.runtime,
  org.eclipse.core.variables,
--- a/org.eclipse.m2e.launching/src/org/eclipse/m2e/internal/launch/MavenLaunchDelegate.java
+++ b/org.eclipse.m2e.launching/src/org/eclipse/m2e/internal/launch/MavenLaunchDelegate.java
@@ -35,6 +35,7 @@ import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.ILog;
+import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Path;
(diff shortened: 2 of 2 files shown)


org.eclipse.m2e.launching/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.m2e.launching;singleton:=true
-Bundle-Version: 2.0.500.qualifier
+Bundle-Version: 2.0.501.qualifier
```
HannesWell, line 5:
```suggestion
Bundle-Version: 2.0.600.qualifier
```
By default the service/micro version should be bump by `+100`, just in case we have to do a in between maintenance release later (we have not yet done it, but since that is cheap why not be prepared, just in case).

org.eclipse.m2e.launching/src/org/eclipse/m2e/internal/launch/MavenLaunchDelegate.java
```diff
@@ -245,13 +246,15 @@ public static String readEnforcedJavaVersion(File pomDirectory, IProgressMonitor
...
+        IPath pomPath = Path.fromOSString(IMavenConstants.POM_FILE_NAME);
+        if(container.get().exists(pomPath)) {
+          IFile pomFile = container.get().getFile(pomPath);
+          IMavenProjectFacade mavenProject = projectManager.create(pomFile, true, new NullProgressMonitor());
+          if(mavenProject != null) {
+            return readEnforcedVersion(mavenProject, monitor);
+          }
         }
       }
-      //TODO: handle the case if the pomDirectory points to a project not in the project. Then load the bare project.
```
HannesWell, line 254: I think the comment should stay, since it still holds and covers another case than the one here.
Although the text has a mistake.

HannesWell (commented):
Makes sense.
Thanks for the fix!

HannesWell (approved):

PR #1440, opened 2023-06-21
Add a dedicated feature to install the maven runtime

laeubi:


mickaelistria:
Why a feature? It's totally possible to directly install a bundle and to show it in the installation dialog. A single-bundle feature doesn't bring much there.

laeubi:
> Why a feature? It's totally possible to directly install a bundle and to show it in the installation dialog.

1. It probably won't be a single bundle in the future
2. feature have more rich metadata to present to the user 
3. as far as I know only features are displayed under "installed software"

mickaelistria:
For 1, I can't tell, but at te moment it's not the case
For 2, you're right.
For 3, you're wrong.

laeubi:
> I can't tell, but at te moment it's not the case

It is on my todo-list, so whats the real issue with having a feature for this instead of referencing a bundle? I see no real benefit of that nor any real drawback in using a feature right now.

laeubi:
Superseded by https://github.com/eclipse-m2e/m2e-core/pull/1443

PR #1445, opened 2023-06-22
Retain Maven 3.8 compatibility

laeubi:
m2e can perfectly build with maven 3.8 runtime so we should retain this as long as possible to have a way for people to switch to older runtimes but using newer m2eclipse.

The only issue is that we need to use a deprecated class here but unless it is really removed from maven we can simply use that without an issue.


HannesWell:
Sounds not too hard to achieve.
As you said a profile could solve this. In this case the inclusion of the maven.runtime bundle as well as activating the baseline compare could be moved to a profile that is active by default and disabled if a certain property is activated.

Issue #1446, opened 2023-06-23
Error without toolchains


laeubi:
@mickaelistria It seems now m2e prints an error each time if one does not has a `toolchains.xml` can you take a look?

```
java.io.FileNotFoundException: C:\Users\xxxxx\.m2\toolchains.xml (Das System kann die angegebene Datei nicht finden)
        at java.base/java.io.FileInputStream.open0(Native Method)
        at java.base/java.io.FileInputStream.open(FileInputStream.java:216)
        at java.base/java.io.FileInputStream.<init>(FileInputStream.java:157)
        at org.codehaus.plexus.util.xml.XmlReader.<init>(XmlReader.java:128)
        at org.codehaus.plexus.util.xml.XmlStreamReader.<init>(XmlStreamReader.java:67)
        at org.codehaus.plexus.util.ReaderFactory.newXmlReader(ReaderFactory.java:123)
        at org.apache.maven.toolchain.io.DefaultToolchainsReader.read(DefaultToolchainsReader.java:55)
        at org.eclipse.m2e.jdt.LookupJDKToolchainsJob.run(LookupJDKToolchainsJob.java:50)
        at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

PR #1447, opened 2023-06-23
Do not fail on non-existing toolchains file

mickaelistria:
+ process global toolchains file as well as user one

Fixes https://github.com/eclipse-m2e/m2e-core/issues/1446


mickaelistria:
@laeubi can you please verify it on your end too?

laeubi (approved):
Looks good!

PR #1448, opened 2023-06-23
Revert "Move to use the platform license check"

HannesWell:
Now that https://github.com/eclipse/dash-licenses/pull/247 is merged I think we can use the original workflow again.

This reverts commit 8c6d5719b8a4a2b4261c0d8c96fec6e087b26726.


HannesWell:
I'm now a committer in the dash-licenses project and can now fix issues easily, so I think it is time to switch back.

PR #1449, opened 2023-06-25
Use p2.inf to generate installable maven runtimes and add release note

laeubi:
This adds two new units that can be used to install additional runtimes into m2eclipse and a release note that explains how to use this.

### DIFF
--- a/org.eclipse.m2e.repository/p2.inf
+++ b/org.eclipse.m2e.repository/p2.inf
@@ -1,0 +1,48 @@
+# Define IU for Maven 3.8
+units.0.id = org.eclipse.m2e.maven.runtime.38.iu
+units.0.version = 1.0.0
+units.0.properties.1.name=org.eclipse.equinox.p2.name
+units.0.properties.1.value=Maven 3.8.x
+units.0.provides.0.namespace=org.eclipse.equinox.p2.iu
+units.0.provides.0.name=org.eclipse.m2e.maven.runtime.38.iu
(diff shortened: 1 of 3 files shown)


org.eclipse.m2e.repository/p2.inf
```diff
@@ -0,0 +1,48 @@
...
+units.0.provides.0.version=1.0.0
+units.0.requires.0.namespace = osgi.bundle
+units.0.requires.0.name = org.eclipse.m2e.maven.runtime
+units.0.requires.0.range = [3.8.0,3.9.0)
+units.0.requires.0.greedy = true
+units.0.requires.0.optional = false
+units.0.update.id = org.eclipse.m2e.maven.runtime.38.iu
+units.0.update.range = [1.0.0,1.0.0]
+
+# Define IU for Maven 3.8
```
HannesWell, line 17:
```suggestion
# Define IU for Maven 3.9
```

HannesWell (approved):
I have not understand this in depth, but what I see/understand sounds reasonable to me.

For clarification, this adds two IUs to the m2e p2-repo, where each requires the m2e.maven.runtime in a specific minor version range, so that one can install that artificial units that pull in the actual bundles?

laeubi:
> For clarification, this adds two IUs to the m2e p2-repo, where each requires the m2e.maven.runtime in a specific minor version range, so that one can install that artificial units that pull in the actual bundles?

yes.

PR #1450, opened 2023-06-25
Simplify tycho-surefire-plugin configuration

HannesWell:
Use the properties defined/read by the tycho-surefire-plugin directly instead of assigning their value to the corresponding properties.


pom.xml
```diff
@@ -185,19 +185,6 @@
...
 				</plugin>
-				<plugin>
-					<groupId>org.eclipse.tycho</groupId>
-					<artifactId>tycho-surefire-plugin</artifactId>
-					<version>${tycho-version}</version>
-					<configuration>
-						<useUIHarness>${tycho.surefire.useUIHarness}</useUIHarness>
-						<useUIThread>${tycho.surefire.useUIThread}</useUIThread>
-						<includes>
-							<include>**/*Test.java</include>
```
HannesWell, line 196: Looks like with the default Pattern we have now an extra test, but it is failing and has to be fixed.

PR #1452, opened 2023-06-26
Remove Maven Build shortcut

laeubi:
Now we show the launch configs in the menu already, the Run As > Maven Build can be removed as it is obsolete now:

- if no config is there it is equivalent to the dialog variant
- if there is a config it is already shown as menu item

### DIFF
--- a/org.eclipse.m2e.launching/plugin.xml
+++ b/org.eclipse.m2e.launching/plugin.xml
@@ -46,29 +46,6 @@
   </extension>
 
    <extension point="org.eclipse.debug.ui.launchShortcuts">
-      <shortcut id="org.eclipse.m2e.core.pomFileAction"
-                class="org.eclipse.m2e.actions.ExecutePomAction"
-                icon="icons/m2.gif"
-                label="%m2.popup.pomFile.label"
-                modes="run,debug">
-         <contextualLaunch>
-           <contextLabel label="%m2.popup.pomFile.label" mode="run"/>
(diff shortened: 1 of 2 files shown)


laeubi:
@mickaelistria would you mind asking Redhat to assign an unused resource pack to m2eclipse? I noticed that Jenkins is currently way behind because it lacking executors when we have multiple PRs pending.

laeubi:
@HannesWell @mickaelistria @fbricon any concerns?

fbricon:
Why can't we build the PRs on Github Actions only? The only reason why we're using jenkins is so that we can sign the artifacts on the Eclipse infra and push CI builds to the snapshots area. PRs don't need any of that.

mickaelistria:
We need to detect errors ASAP, including the ones related to signing.

fbricon:
> We need to detect errors ASAP, including the ones related to signing.

No signing is happening on PR builds. Only master and version branches get signed: 
https://github.com/eclipse-m2e/m2e-core/blob/1cf7ac239dcd8c54d99225d2db8e88db7015b59d/Jenkinsfile#L46-L48

laeubi:
I think in the past it was valuable to have some different system and setup to test, and assume the Jenkins build as the ultimate check as if it fails there *after the fact* its quite annoying, so unless we can get rid of Jenkins CI completely I think it should stay for PRs as well.

mickaelistria (approved):

claudioweiler:
Please, revert this commit, see #1618
Also, this commit is incomplete as shortcut binding still exists, but doing nothing.

Issue #117, opened 2021-03-30
Bug 528399 - Download sources by default


mickaelistria:
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=528399

laeubi:
@mickaelistria do you plan to work on this?

mickaelistria:
Submitted https://github.com/eclipse-m2e/m2e-core/pull/1456

PR #1456, opened 2023-06-26
Download sources by default

mickaelistria:
Fixes https://github.com/eclipse-m2e/m2e-core/issues/117


laeubi:
@mickaelistria it seems we need to update the submodule so the test changes are picked up by github!

PR #1459, opened 2023-06-27
Test what causing the test failure

laeubi:
- add message changes


laeubi:
It seems the new dependency causing the test-failure that's really strange...

laeubi:
@mickaelistria any idea why the test is sensitive to the imports in the manifest?

mickaelistria:
It may be some activation ordering or something of that sort. For instance with VM detection in JDT, some tests have started to fail because it induced a change in preference initialization order.

HannesWell:
@laeubi is this still relevant?

laeubi:
Yes this is still relevant unless someone can explain why we get a test failure for a completely unrelated change.

Adding `org.eclipse.core.variables` to the manifest let the `test org.eclipse.m2e.jdt.tests.RunTest#testRunTwice` fail for no obvious reason.

akurtakov:
What are the failures seen? I don't seem to be able to find them here.

laeubi:
It seems even github deletes run logs after such long time, I rebased and resolved the conflict so lets see if it still fails.

akurtakov:
I see only the regular failures as on master thus closing this one.

PR #1463, opened 2023-06-29
Fix bad casting causing all entries to be unknown

kysmith-csg:
A change made by https://github.com/eclipse-m2e/m2e-core/commit/6eb17dbe07be05d3f237cf8288b070fd3731f1ff#diff-28403ac299e4b874aaf674d927c7a0fa60fe79ea6fbfd07a5b827e51d84684e6L446 causes the Lifecycle Mappings preferences to show everything is "unknown" because `IPluginExecutionMetadata` and `LifecycleMappingMetadataSource` do not share a common base class and thus could never be `instanceof`. 

Current behavior as of 2.2.0:
Screenshot 2023-06-29 at 15 02 21

Fix:
Screenshot 2023-06-29 at 15 04 20

### DIFF
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/LifecycleMappingsViewer.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/LifecycleMappingsViewer.java
@@ -85,6 +85,37 @@ import org.eclipse.m2e.core.ui.internal.Messages;
 
 @SuppressWarnings("restriction")
 public class LifecycleMappingsViewer {
+
+  /**
+   * Uninteresting lifecycle phase
+   */
+  private static final String UNINTERESTING = "uninteresting"; //$NON-NLS-1$
+
+  /**
+   * Maven plugin lifecycle source
+   */
+  private static final String MAVEN_PLUGIN = "maven-plugin"; //$NON-NLS-1$
+
+  /**
+   * Pom lifecycle source
+   */
+  private static final String POM = "pom"; //$NON-NLS-1$
+
+  /**
+   * Eclipse extension lifecycle source
+   */
+  private static final String EXTENSION = "extension"; //$NON-NLS-1$
+
+  /**
+   * Default execution source
+   */
+  private static final String DEFAULT = "default"; //$NON-NLS-1$
+
+  /**
+   * Unknown lifecycle source
+   */
+  private static final String UNKNOWN = "unknown"; //$NON-NLS-1$
+
   /*package*/TreeViewer mappingsTreeViewer;
 
   /*package*/boolean showPhases = false;
@@ -102,7 +133,7 @@ public class LifecycleMappingsViewer {
   private IProject project;
 
   void updateMappingsTreeViewer() {
-    if(mappingsTreeViewer == null) {
+    if(mappingsTreeViewer == null || mappingsTreeViewer.getControl().isDisposed()) {
       return;
     }
     mappingsTreeViewer.refresh();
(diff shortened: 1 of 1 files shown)


HannesWell:
@laeubi can you please have a look?

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/LifecycleMappingsViewer.java
```diff
@@ -444,7 +444,8 @@ String getSourcelabel(MojoExecutionKey execution, List<IPluginExecutionMetadata>
     LinkedHashSet<String> sources = new LinkedHashSet<>();
     if(mappings != null && !mappings.isEmpty()) {
       for(IPluginExecutionMetadata mapping : mappings) {
-        if(mapping instanceof LifecycleMappingMetadataSource metadata) {
+        LifecycleMappingMetadataSource metadata = ((PluginExecutionMetadata) mapping).getSource();
```
laeubi, line 447: Either use `IPluginExecutionMetadata` (without a cast) or add a check that mapping is actually of type `PluginExecutionMetadata` ...

laeubi (changes requested):

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/LifecycleMappingsViewer.java
```diff
@@ -444,52 +475,56 @@ String getSourcelabel(MojoExecutionKey execution, List<IPluginExecutionMetadata>
...
+              sources.add(getSourceLabel(artifact, detailed));
+            } else if(source instanceof MavenProject mavenProject) {
+              sources.add(getSourceLabel(mavenProject, detailed));
+            } else if(source instanceof Bundle bundle) {
+              sources.add(getSourceLabel(bundle, detailed));
+            } else {
+              sources.add(UNKNOWN); //$NON-NLS-1$
+            }
           }
         } else if(mapping instanceof DefaultPluginExecutionMetadata) {
```
laeubi, line 496: This path must be outside the `if(mapping instanceof PluginExecutionMetadata metadata) {` block.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/LifecycleMappingsViewer.java
```diff
@@ -444,52 +475,56 @@ String getSourcelabel(MojoExecutionKey execution, List<IPluginExecutionMetadata>
...
-          sources.add("default"); //$NON-NLS-1$
+          sources.add(DEFAULT); //$NON-NLS-1$
         } else {
```
laeubi, line 498: this one as well, so we have a chain like 

```
if(mapping instanceof PluginExecutionMetadata metadata) {
 ...
} else if(mapping instanceof DefaultPluginExecutionMetadata) {
 ...
} else {
  sources.add(UNKNOWN):
}
```

laeubi (changes requested):

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/LifecycleMappingsViewer.java
kjsmita6: It is already.
```java
if (mapping instanceof PluginExecutionMetadata) {
    if (source == null) {
        // unknown
    } else {
        // source label or unknown
    }
} else if (mapping instanceof DefaultPluginExecutionMetadata) {
    // default
} else {
    // unknown
}

Comment (kysmith-csg) [reply org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/LifecycleMappingsViewer.java:496]: Accidentally replied on the wrong account, but this is how it is.

```java
if (mapping instanceof PluginExecutionMetadata) {
    if (source == null) {
        // unknown
    } else {
        // source label or unknown
    }
} else if (mapping instanceof DefaultPluginExecutionMetadata) {
    // default
} else {
    // unknown
}
```

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/LifecycleMappingsViewer.java
kysmith-csg: See above, this is exactly how it is.

laeubi (approved):

PR #1465, opened 2023-07-04
Register the maven protocol handler as a DS component

laeubi:
FYI @merks this will make the protocol available as soon as possible even if m2e is currently not activated by other means.


laeubi:
This currently references `IMaven` as a service to circumvent:

- https://github.com/eclipse-equinox/equinox/pull/290

if that is fixed, we should switch back to using `IMaven maven = MavenPlugin.getMaven();` to allow the component being activated without any prerequisites.

laeubi:
@mickaelistria in commit 0fc95231a540e9c25baaa7e46d324e6832938629 you reverted registering this with ECF do you remember what was the problem back there?

mickaelistria:
@laeubi yes, it was failing because there was a circular dependency at exec between ECF (loading m2e extension) and m2e (loading ECF)  and this was causing failures to startup m2e. But with DS, since the service are more lazily resolved, it could be fine.

laeubi:
I have tried searching the code but can't find that m2e is using ECF anywhere do you have a pointer where the dependency to ECF originates?

mickaelistria:
The dependency was probably originating from the addition of the protocol handler itself. I don't remember the details, but I remember that I thought that something lazier could probably solve it.

laeubi:
Alright, I'll maybe try to debug this, actually the extension registry should also allow to use it lazy so maybe something to improve for ECF...

laeubi:
@mickaelistria I have now debugged this a bit and it seems all ECF does is actually register the extension point as a service, so this seems redundant here them, but it could be more lazy anyways I'll open a PR.

laeubi:
I have now created
- https://github.com/eclipse/ecf/pull/52

laeubi:
I also discovered one problem with ECF and custom handlers, created
- https://github.com/eclipse-equinox/p2/pull/298

PR #1479, opened 2023-07-18
Build with Tycho 4.0.0

HannesWell:


HannesWell:
@laeubi the tycho 4 snapshot is not available anymore, so I switched to the release.
When you think that Tycho 5 is stable enough to be tested I'm fine with switching to the Tycho 5 snapshot.

laeubi:
As long as we not need any snapshot features I think the release version is fine.

PR #1491, opened 2023-07-30
[Setup] Remove disable P2-director and set id for worksing sets

HannesWell:
The connectors defined in the disabled P2-director are not compatible with m2e and are unlikely to become compatible in the future (project seems abandoned) and are actually not necessary anymore.


HannesWell:
Setup is not build -> No need to wait for the result.

PR #1492, opened 2023-07-30
Move o.e.m2e.maven.runtime to m2e's git-repo root

HannesWell:
Since `o.e.m2e.maven.runtime` is the only remaining module in m2e-maven-runtime there is no need anymore to keep `m2e-maven-runtime` as its parent project and the general build configuration separated.

@laeubi any objection?


HannesWell:
Somehow the feature's build-qualifier does not take the version and qualifier change of o.e.m2e.maven.runtime into account:

```
[INFO] --- tycho-packaging:4.0.0:build-qualifier-aggregator (default-build-qualifier-aggregator) @ org.eclipse.m2e.feature ---
[INFO] The project's OSGi version is 2.4.100.20230705-1302
```

And consequently the feature is replaced with the baseline from the snapshot.

laeubi:
> @laeubi any objection?

If you think it helps managing m2e build why not :-)

laeubi:
> Somehow the feature's build-qualifier does not take the version and qualifier change of o.e.m2e.maven.runtime into account:

You need to bump the version manually, the qualifier update will likely miss non Tycho projects.

HannesWell:
> > Somehow the feature's build-qualifier does not take the version and qualifier change of o.e.m2e.maven.runtime into account:
> 
> You need to bump the version manually, the qualifier update will likely miss non Tycho projects.

Was this always required? IIRC in the past, this was not necessary. But this was likely in the time, when the m2e.maven.runtime participated as regular Eclipse Plugin in the build.

Do you think it would be feasible to also consider non Eclipse Plugin Projects when computing the qualifier?

laeubi:
The qualifier computation is quite fragile already and requires a special naming of artifacts/bundles so I'm not sure it is worth to invest more time as necessary but if you like you can debug the case and check why it is not considered.

HannesWell:
> The qualifier computation is quite fragile already and requires a special naming of artifacts/bundles so I'm not sure it is worth to invest more time as necessary but if you like you can debug the case and check why it is not considered.

Understand. Then lets keep this as task for the future. Since we will encounter this problem on every maven update, I think the chances are not too low, that one is annoyed enough bis this in the future.^^

PR #1493, opened 2023-07-30
Build with Tycho 4.0.1

HannesWell:


.mvn/extensions.xml
```diff
@@ -3,6 +3,6 @@
   <extension>
     <groupId>org.eclipse.tycho</groupId>
     <artifactId>tycho-build</artifactId>
-    <version>4.0.0</version>
+    <version>4.0.1</version>
```
laeubi, line 6: should we probably externalize the version into maven.config?

HannesWell: This would be nice, in fact I already started this a while ago in https://github.com/eclipse-m2e/m2e-core/pull/848, but IIRC there is something missing.
I'll have a look at that.

laeubi: Probabbly just change the extraction of tycho version as a minimal step ;-)

HannesWell: Lets continue the discussion about that in #848 and merge this for now.

PR #1494, opened 2023-07-30
Make m2e.core.ui host of archetype-fragments and delete archetype.common

HannesWell:
The Plugin o.e.m2e.archetype.common is empty and only serves as host for the maven-archetype fragment-bundles. Since they are only used by m2e.core.ui, we can simply make that the host of those fragments and save one project in m2e.

@laeubi what do you think?


laeubi:
I don't think this is really worth because it require special treatment of the m2.ui bundle.
If you think we really should save bundles, one can make the archtype common the host bundle (by adding an extra bnd mapping for it to the TP) and the other ones fragments of that bundle but be aware that this is actually a major change for the feature.

HannesWell:
> one can make the archtype common the host bundle (by adding an extra bnd mapping for it to the TP) and the other ones fragments of that bundle 

Did that and also re-used the jdom2 bundle provided by maven-bnd Orbit, so that we don't have to define an own one.

> but be aware that this is actually a major change for the feature.

I wouldn't consider this a braeking change in thise case since the bundles contained in this PDE are more internal and the same feature (in a functional sense is still provided). But I worked around this, simply by keep the generated bundle name as `org.eclipse.m2e.archetype.common`.

org.eclipse.m2e.feature/feature.xml
```diff
@@ -26,7 +26,33 @@
...
+         fragment="true"
+         unpack="false"/>
+
+   <plugin
+         id="org.eclipse.m2e.archetype.maven-artifact-transfer"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         fragment="true"
+         unpack="false"/>
```
HannesWell, line 55: @laeubi do you see any reason that speaks against including the archetype bundles in this feature?
I would prefer it over the p2.inf file since it is less magic.

laeubi: The p2.inf ensures that all fragments are always installed together.

HannesWell: > The p2.inf ensures that all fragments are always installed together (especially for cases like tests).

This can also be achieved by corresponding Import-Packages and is less magic and even less to type.

laeubi: No this can't work because a fragment is always optional, so a consumer would need to import a package from each fragment even if it is never used.

HannesWell: > No this can't work because a fragment is always optional, so a consumer would need to import a package from each fragment even if it is never used.

That's right, a fragment is optional but m2e.core.ui already needs to import a package from each of the archetype artifacts (if one does not use require-bundle). The only import that would not be necessary from a class-loading/compiling point of view is the maven-transfer. But I think one not so necessary import is IMHO ok. And just to save that I wouldn't add the p2.inf.

HannesWell:
@laeubi this change is IMHO now ready, do you have any objects?

target-platform/target-platform.target
```diff
@@ -5,12 +5,14 @@
 	<locations>
 		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
 			<repository location="https://download.eclipse.org/eclipse/updates/4.28/"/><!--Keep in sync with repo-ref in org.eclipse.m2e.repository/category.xml-->
+			<repository location="https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/2023-09/"/>
```
laeubi, line 8: Even though from the XSD it is supported this it is not really supported by the PDE UI so I think we better use a separate location here.

target-platform/target-platform.target
```diff
@@ -106,17 +108,32 @@
...
+					<artifactId>archetype-common</artifactId>
 					<version>3.2.1</version>
 					<type>jar</type>
 				</dependency>
+			</dependencies>
+			<instructions><![CDATA[
+Bundle-Name:           M2Eclipse supplement fragment for ${mvnGroupId}:${mvnArtifactId}:${mvnVersion}
+version:               ${version_cleanup;${mvnVersion}}
+Bundle-SymbolicName:   org.eclipse.m2e.archetype.common
+Bundle-Version:        ${version}.1
```
laeubi, line 124: this will result in a `3.2.1.1` version is this really desired?

target-platform/target-platform.target
```diff
@@ -106,17 +108,32 @@
...
 				<dependency>
 					<groupId>org.apache.maven.archetype</groupId>
-					<artifactId>archetype-catalog</artifactId>
+					<artifactId>archetype-common</artifactId>
 					<version>3.2.1</version>
 					<type>jar</type>
 				</dependency>
+			</dependencies>
+			<instructions><![CDATA[
+Bundle-Name:           M2Eclipse supplement fragment for ${mvnGroupId}:${mvnArtifactId}:${mvnVersion}
```
laeubi, line 121: This is not a fragment.

target-platform/target-platform.target
HannesWell: Yes the UI does not handle that so nice. But that's effectivly how those IU-Locations are resolved.
Any way, I changed it as you suggested.

target-platform/target-platform.target
HannesWell: Yes it is.
It is a (manual) qualifier bump so that those that use the snapshots don't miss the update.
When there is a new archetype version, it can be removed.

target-platform/target-platform.target
HannesWell: Removed fragment. But I think something like 
`M2Eclipse Archetype supplement based on ${mvnGroupId}:${mvnArtifactId}:${mvnVersion}`
would be better here and below.

target-platform/target-platform.target
laeubi: 1 is really a bad qualifier, why not do like in maven-runtime and use 

`${version}01` what will result in 3.2.101

target-platform/target-platform.target
HannesWell: That is the artifactId. That's fine, but I would also like to mention in the name too that this bundle is wrapped, so I think simply `M2Eclipse's ${mvnArtifactId}` is good.

PR #1496, opened 2023-08-01
Abstract the delta handling from the resource API

laeubi:
Currently m2e is directly wired to the resource delta API, while this seems good in the context of eclipse for m2e this imposes to restrictive API as it is all noimplement/extend interfaces and offer much more than m2e actually using.

This aims to implement a thin wrapper for further improvement of delta detection especially when it comes to resource see for example https://github.com/eclipse-m2e/m2e-core/issues/1302

This will not fix anything but is an important first step to ensure things are still working as before.

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/builder/IIncrementalBuildFramework.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/builder/IIncrementalBuildFramework.java
@@ -17,7 +17,6 @@ import java.io.File;
 import java.util.Set;
 
 import org.eclipse.core.resources.IProject;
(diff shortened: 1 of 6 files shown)


laeubi:
FYI @martinlippert if you can check out the patch and see if it introduces any new problem it would be great. This is a very first step to improve the situation, we need to be able to further detect changes that are not seen by the resource API.

martinlippert:
> FYI @martinlippert if you can check out the patch and see if it introduces any new problem it would be great. This is a very first step to improve the situation, we need to be able to further detect changes that are not seen by the resource API.

@BoykoAlex 👆Maybe you can give this a try?

laeubi:
Tests are green (except the usual macos) so lets give it a try so we can have additional PRs based on this.

PR #1498, opened 2023-08-02
Record the build timestamp and assume every file changed that is newer

laeubi:
Currently we only analyze the current delta but this can have several subtile problems:

- when a file is modified outside of eclipse it might not be recognized by eclipse immediately
- when files are modified outside the "interesting folders" the delta maybe not includes this
- if a build fails the next delta will not necessarily include all files of the previous delta because the error might be in a java source file and a resource file was changed

This adds a project scoped timestamp to record when the last successful build happens and assumes every file changed that has a timestamp larger than the last build, this still does not handle deletes properly but improves the situation already for existing files. It also leads to probably more files queried/assumed changed in case of failing builds or when the build timestamp is unknown but ensures that we only ever process more files than required but never less ones.

FYI @BoykoAlex @martinlippert

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/builder/MavenBuilderImpl.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/builder/MavenBuilderImpl.java
@@ -16,15 +16,18 @@ package org.eclipse.m2e.core.internal.builder;
 import static org.eclipse.core.resources.IncrementalProjectBuilder.FULL_BUILD;
 
 import java.io.File;
+import java.text.DateFormat;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Date;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/builder/MavenBuilderImpl.java
```diff
@@ -452,4 +465,64 @@ public void clean(MavenSession session, IMavenProjectFacade projectFacade,
...
+      this.project = project;
+    }
+
+    public void commit() {
+      this.lastBuild = System.currentTimeMillis();
+    }
+
+    @Override
+    public String toString() {
+      return "BuildState for " + project + " lat reccorded timestamp "
```
szarnekow, line 485: Nit: `reccorded` -> `recorded`

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/builder/MavenBuilderImpl.java
```diff
@@ -78,6 +83,8 @@ public class MavenBuilderImpl {
 
   private final List<IIncrementalBuildFramework> incrementalBuildFrameworks;
 
+  private final Map<IProject, ProjectBuildState> deltaState = new ConcurrentHashMap<>();
```
szarnekow, line 86: Did you consider storing this as a session property on the project to get automatic cleanup of the map?

laeubi: @szarnekow sounds like a good idea, do you like to propose a PR? This is currently a bit experimental but seem to solve the issues we see with missing resource copy, so improvements are always welcome!

szarnekow: I'd love to but I don't see a test for this so I'm not quite sure which problem this in fact solves and how to ensure that a refactoring doesn't break it.

Issue #1481, opened 2023-07-19
Empty maven console in 2023-06


reckart:
I have upgraded my m2e installation to 2.3.0 in Eclipse 2023-06 and now no longer get output in the Maven Console View. Relevant plugins seem to be installed.

```
M2E - Maven Integration for Eclipse	2.3.0.20230523-2033	org.eclipse.m2e.feature.feature.group	Eclipse.org - m2e
M2E - SLF4J over Logback Logging	2.1.2.20230523-2106	org.eclipse.m2e.logback.feature.feature.group	Eclipse.org - m2e
```

Is this (again) a known issue?
Is there a known fix or workaround?

Older similar issue: https://github.com/eclipse-m2e/m2e-core/issues/182

reckart:
ping?

HannesWell:
Can you also list the versions of the `slf4j.api` and `logback` bundles?

And check (e.g. in the Host OSGi console) if the `aries.spifly.bundle` and `logback` bundles are started?
But since as far as I know in EPP the start levels have not been configured yet, I assume they are not started.

reckart:
<img width="645" alt="Screenshot 2023-07-24 at 20 45 24" src="https://github.com/eclipse-m2e/m2e-core/assets/1410238/8ae18ff6-e9fd-4177-893f-75b71ba37e26">
<img width="743" alt="Screenshot 2023-07-24 at 20 45 43" src="https://github.com/eclipse-m2e/m2e-core/assets/1410238/93b21c0c-9028-4175-95a6-7360edd7c59c">

... need to install the PDE tooling for the host console ... takes a moment

reckart:
According to the OSGI console, logback seems to be resolved only and spifly does not even seem to be installed.

```
2627	reference:file:plugins/org.eclipse.m2e.logback_2.1.200.20230523-2106.jar
  RESOLVED    org.eclipse.m2e.logback_2.1.200.20230523-2106 [2627]
2544	reference:file:plugins/ch.qos.logback.classic_1.2.12.jar
  RESOLVED    ch.qos.logback.classic_1.2.12 [2544]
2545	reference:file:plugins/ch.qos.logback.core_1.2.12.jar
  RESOLVED    ch.qos.logback.core_1.2.12 [2545]
```

reckart:
[...]

PR #1512, opened 2023-08-17
Configure start behavior of logback-class through m2e.logback.feature

HannesWell:
and add 'jakarta.servlet-api' as requirement to ensure its present until Aries can handle its absense for logback:
https://github.com/apache/aries/pull/233

Fixes https://github.com/eclipse-m2e/m2e-core/issues/1481

### DIFF
--- a/org.eclipse.m2e.logback.feature/p2.inf
+++ b/org.eclipse.m2e.logback.feature/p2.inf
@@ -1,0 +1,25 @@
+#The purpose of the following advices is to configure ch.qos.logback.classic as auto-started with start-level 2 when installing this feature 
+#Create a requirement on the fragment we are creating
+requires.0.namespace=org.eclipse.equinox.p2.iu
+requires.0.name=configure.logback.classic
+requires.0.range=[$version$,$version$]
+requires.0.greedy=true
+
+#Create a IU fragment named configure.logback.classic
+units.0.id=configure.logback.classic
+units.0.version=$version$
+units.0.provides.0.namespace=org.eclipse.equinox.p2.iu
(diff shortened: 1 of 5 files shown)


org.eclipse.m2e.logback.feature/p2.inf
```diff
@@ -0,0 +1,22 @@
...
+units.0.instructions.configure=setStartLevel(startLevel:2);\
+                               markStarted(started:true);
+units.0.hostRequirements.1.namespace=osgi.bundle
+units.0.hostRequirements.1.name=ch.qos.logback.classic
+units.0.hostRequirements.1.range=[1.3,1.5]
+units.0.hostRequirements.1.greedy=false
+units.0.requires.1.namespace=osgi.bundle
+units.0.requires.1.name=ch.qos.logback.classic
+units.0.requires.1.range=[1.3,1.5]
+units.0.requires.1.greedy=false
```
HannesWell, line 22: @laeubi can you tell what I'm doing wrong here?
I want to set the `ch.qos.logback.classic` bundle as auto-started with level 2.

laeubi: If i rember correctly  you need to include the p2.inf into the build.properties.

HannesWell: Yes that was definitely missing.
What was also missing was the `installBundle(bundle:${artifact})`. Otherwise logback.classic was just removed from the `bundles.info` file.

HannesWell:
In my local testing with a local build this now worked. But I'll try again tomorrow with the IDE build in the CI.

PR #1515, opened 2023-08-19
Make sure the last modified timestamp is set correctly

laeubi:
Linux stores modification time with nano-seconds precision, but we currently set the timestamp of the cached file only in milliseconds. This can lead to the situation that the source is considered a few nanoseconds older than the cache file and therefore the cache is regenerated every time.

### DIFF
--- a/org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/shared/MavenBundleWrapper.java
+++ b/org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/shared/MavenBundleWrapper.java
@@ -136,10 +136,7 @@ public class MavenBundleWrapper {
 			Map<DependencyNode, WrappedBundle> visited = new HashMap<>();
 			WrappedBundle wrappedNode = getWrappedNode(node, instructionsLookup, visited);
 			for (WrappedBundle wrap : visited.values()) {
-				Jar jar = wrap.getJar();
-				if (jar != null) {
-					jar.close();
-				}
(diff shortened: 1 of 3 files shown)


HannesWell (approved):
LGTM

laeubi:
The test failure was that if the artifact was already wrapped once with error it might be still created (bnds principle is to never really 'fail' but only emit an error/warning), so the second time it finds the cache file and proceeds without an error.

I now choose to simply delete the result file in case of an error, a more sophisticated approach might would store the error message along with the cached file but this has some uncertainties e.g if the error was not caused by the artifact itself so this seems the safe and easy path for now as one should not optimize for the error case.

org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/shared/MavenBundleWrapper.java
```diff
@@ -136,10 +136,7 @@ public boolean visitEnter(DependencyNode n) {
 			Map<DependencyNode, WrappedBundle> visited = new HashMap<>();
 			WrappedBundle wrappedNode = getWrappedNode(node, instructionsLookup, visited);
 			for (WrappedBundle wrap : visited.values()) {
-				Jar jar = wrap.getJar();
-				if (jar != null) {
-					jar.close();
-				}
+				wrap.getJar().ifPresent(jar -> jar.close());
```
HannesWell, line 139: You could probably use a Method reference here.

HannesWell (approved):
Makes sense to me.

PR #1516, opened 2023-08-20
Add latest tm4e release (0.8) to m2e's TP directly

HannesWell:
This allows to use snakeyaml from Maven-Central which is included in tm4e's repo. Therefore we can now drop the Orbit repo, which only served as provider of the old snakeyaml verson.

@vrubezhny as you suggested in https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/pull/1252#issuecomment-1685037743. Do you see any potential problem with this?
This will also help in https://github.com/eclipse-m2e/m2e-core/pull/1494, because the latest version of tm4e can use the latest snakeyaml 2 version (or one of its dependencies, not sure which one is causing the trouble), which the release included in wildwebdeveloper can't.

### DIFF
--- a/org.eclipse.m2e.repository/category.xml
+++ b/org.eclipse.m2e.repository/category.xml
@@ -50,8 +50,9 @@
      </query>
    </iu>
    <category-def name="m2e" label="Maven Integration for Eclipse"/>
+   <repository-reference location="https://download.eclipse.org/eclipse/updates/4.28/" enabled="true" />
    <repository-reference location="https://download.eclipse.org/wildwebdeveloper/releases/1.3.0/" enabled="true" />
+   <repository-reference location="https://download.eclipse.org/tm4e/releases/0.8.0/" enabled="true" />
    <repository-reference location="https://download.eclipse.org/lsp4j/updates/releases/0.21.0/" enabled="true" />
    <repository-reference location="https://download.eclipse.org/lsp4e/releases/0.23.0/" enabled="true" />
-   <repository-reference location="https://download.eclipse.org/eclipse/updates/4.28/" enabled="true" />
 </site>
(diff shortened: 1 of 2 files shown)


vrubezhny:
> @vrubezhny as you suggested in [eclipse-wildwebdeveloper/wildwebdeveloper#1252 (comment)](https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/pull/1252#issuecomment-1685037743). Do you see any potential problem with this?

WWD do not depend anyhow on m2e-core, so I don't know if this PR  changes anything in this regard. I'll split the https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/pull/1252 to make WWD to use the same direct repository for tm4e so the latest release of tm4e and its language pack is available in WWD anyway, making it possible to switch using the language configs and grammars in the following PRs.

HannesWell:
> WWD do not depend anyhow on m2e-core, so I don't know if this PR changes anything in this regard.

Yes that's right. I was more wondering if it could be a problem if a more recent version of tm4e is used in m2e than in WWD or lsp4e/j? In my simple testing everything worked fine.

mickaelistria:
I don't see a technical concern here, but I see an organizational one: m2e will have to manage 1 dependency which is already managed at best transitively (in Wild Web Developer). I'm afraid m2e won't monitor and react to TM4E updates as well as TM4E, so in the end, m2e will miss an update of TM4E and a newer version of TM4E will be pulled transitively by Wild Web Developer, so the content in the m2e target file will be outdated and irrelevant.

HannesWell:
> I don't see a technical concern here, but I see an organizational one: m2e will have to manage 1 dependency which is already managed at best transitively (in Wild Web Developer). I'm afraid m2e won't monitor and react to TM4E updates as well as TM4E, so in the end, m2e will miss an update of TM4E and a newer version of TM4E will be pulled transitively by Wild Web Developer, so the content in the m2e target file will be outdated and irrelevant.

I hope that everyone that updates any repo in this Location will also check the others, at least I would do it respectivly ask for that in a review. And even if it missed the P2 Planner will select the latest possible artifacts any way so in worst case we have a repo defined that is actually not used.

If wwd switch to include the latest tm4e release (nnot only SimRel) in the future I'm happy to remove this again, but now I need this for the Archetype restructuring.

PR #1533, opened 2023-08-27
Change Micro-Version schema in wrapped archetype artifacts to 100-steps

HannesWell:
Before the Maven-archetype artifacts where obtained and wrapped in the TP, when they were embedded in a single 'o.e.m2e.archetype.common' bundle, said bundle had the version 3.2.104 and used 100er step as micro version where the hundreds matched the version of the original maven artifact's version.
This schema should be preserved for now to not have back-steps in versions until the next minor of Maven archetype is available.

This also caused problems in SimRel: https://git.eclipse.org/r/c/simrel/org.eclipse.simrel.build/+/203841

This is currently a draft since it requires some fine-tuning for maven-artifact-transfer.

### DIFF
--- a/org.eclipse.m2e.feature/build.properties
+++ b/org.eclipse.m2e.feature/build.properties
@@ -13,4 +13,3 @@
 bin.includes = feature.xml,\
                feature.properties,\
                p2.inf
-pom.model.property.tycho.baseline.skip = true
--- a/target-platform/target-platform.target
+++ b/target-platform/target-platform.target
@@ -123,7 +123,8 @@
 Bundle-Name:           M2Eclipse's ${mvnArtifactId}
 version:               ${version_cleanup;${mvnVersion}}
 Bundle-SymbolicName:   org.eclipse.m2e.archetype.common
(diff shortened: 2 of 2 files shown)


HannesWell:
We now we have a kind of mixed versioning schema:
- org.eclipse.m2e.archetype.common 3.2.104
- org.eclipse.m2e.archetype.catalog 3.2.1.2
- org.eclipse.m2e.archetype.descriptor 3.2.1.2
- org.eclipse.m2e.archetype.maven-artifact-transfer 0.13.1.2

`org.eclipse.m2e.archetype.common` should have a higher version than 3.2.103 (which was the version of the plugin that embedded all archetype artifacts), but we cannot have a fixed micro-version for all fragments, since maven-artifact-transfer has a version schema dependent to the archetype artifacts.

So I think this is the best we can do for now. With the next release of archetype we can at least remove the qualifiers from the fragments and with the next minor version we can even use the original version for `o.e.m2e.archetype.common`.

merks:
This seems like a good compromise.

PR #1536, opened 2023-08-30
Clarify framework stack for pom.xml editor in README

kwin:


HannesWell:
That's a good Idea.
Can you please add wwd and tm4e as well
- https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper
- https://github.com/eclipse/tm4e

kwin:
TBH I am not sure how they relate to Lemminx/LSP. Feel free to extend my PR or we do a followup PR afterwards.

HannesWell:
> TBH I am not sure how they relate to Lemminx/LSP.

@vrubezhny can you explain that?
I assume tm4e and wwd are responsible for tokenizing/editing the text, aren't they?

HannesWell:
> tm4e is used for syntax highlighting in the generic editor using textmate grammars. Wild Web Developer provides the XML edition using LemMinX. m2e adds LemMinX-Maven into Wild Web Developer's LemMinx.

Thanks for that. I updated the list with that info.

@kwin I'm fine with this as it is now. Please rebase-merge if you agree.

In general we could think about moving the editor line a little bit lower in the list, because I don't think it is m2e responsibility to first advertise the projects it depends on.

PR #1579, opened 2023-10-15
Use SPDX identifier for license

Bananeweizen:
Recommended by Maven documentation and Eclipse EMO.

PR #1478, opened 2023-07-18
Unit test for no resources in the target folder

BoykoAlex:
In support of issue #1302

PR #1587, opened 2023-10-30
ignore insignificant licence change in baseline-check

fbricon:
Change-Id: I3667aef7789dc8d94c97ee685787c54a0982b8a7
Signed-off-by: Fred Bricon <fbricon@gmail.com>

### DIFF
--- a/m2e-parent/pom.xml
+++ b/m2e-parent/pom.xml
@@ -120,6 +120,7 @@
 								<ignore>META-INF/ECLIPSE_.SF</ignore>
 								<ignore>jars/*.jar</ignore>
 								<ignore>Eclipse-SourceReferences</ignore>
+								<ignore>Bundle-License</ignore>
 							</ignores>
 							<extensions>true</extensions>
 						</configuration>


mickaelistria (changes requested):
See https://github.com/eclipse-m2e/m2e-core/pull/1478#issuecomment-1785172200 , I think we should favor other solutions.

fbricon:
I will not go over dozens of manifests and pom files to fix it

m2e-parent/pom.xml
```diff
@@ -120,6 +120,7 @@
 								<ignore>META-INF/ECLIPSE_.SF</ignore>
 								<ignore>jars/*.jar</ignore>
 								<ignore>Eclipse-SourceReferences</ignore>
```
mickaelistria, line 122: Actually, even those shouldn't be ignored as they're supposed to be stable across rebuilds. And if they're not stable, then they should be made stable.

fbricon:
CI ignores the directive anyway. Fuck it.

mickaelistria:
The failure is a consequence of 248d8fce98037e865120e58afffaf911e471d374 / https://github.com/eclipse-m2e/m2e-core/pull/1579

mickaelistria:
The proper fix is just to bump bundle version as they now have different content.

m2e-parent/pom.xml
HannesWell: Unfortunately they are not stable since they contain the ID of the commit from which the bundle was build.

mickaelistria: If configured correctly, the referenced commit is the last commit that has affected the particular module; which is stable as long as the bundle is not changed.

HannesWell: That's a point.
But I likely added this because it failed the build, but if it would work without ignoring that, it would be even better.

PR #1592, opened 2023-10-31
Exclude cli source

mickaelistria:


mickaelistria:
The `verify-repository` step is not running because build diverges between PR and master (which is IMO usually an anti-pattern), so the CI build doesn't verify this change.
I think we should just keep signing profile for PRs as well, as we can see here it's not something tat always work smoothly.

mickaelistria:
Fixed and merged as https://github.com/eclipse-m2e/m2e-core/commits/master

HannesWell:
> The `verify-repository` step is not running because build diverges between PR and master (which is IMO usually an anti-pattern), so the CI build doesn't verify this change.
> I think we should just keep signing profile for PRs as well, as we can see here it's not something tat always work smoothly.

In generell I agree that it should be similar.
But this is on purpose, because if we would sing PRs anybody could create a PR with whatever content and then could download the signed artifacts from Jeninks.

PR #1594, opened 2023-11-07
Update dependencies: Aries SPIFly, webtools, lsp4j, lsp4e, commons-cli and Guava

HannesWell:
Apache Aries SPIFly is currently on vote, thus is is only a draft.


HannesWell:
Aries SPIFly 1.3.7 is now available on Maven-Central. 🎉

Issue #1600, opened 2023-11-11
Use quotes on project name


Bananeweizen:
It might be useful to add single quotes around the project name when configuring, to make multi word names more readable:
![grafik](https://github.com/eclipse-m2e/m2e-core/assets/406876/d74651dc-bd40-4eb3-8835-26f41f8b8570)

PR #1601, opened 2023-11-11
Use quotes on project name #1600

Bananeweizen:
Also fix 2 additional properties ending in blanks. One can be removed, the other should be trimmed and the blank be added in code.

Looks like this after the change: 
grafik


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/messages.properties
```diff
@@ -110,7 +108,7 @@ ProjectConfigurationManager_task_importing2=Importing project {0}
 ProjectConfigurationManager_task_refreshing=Refreshing projects
 ProjectConfigurationManager_task_updating=Updating configuration for {0}
 ProjectConfigurationManager_task_updating_projects=Updating Maven projects
-ProjectRegistryManager_task_project=project {0}
+ProjectRegistryManager_task_project=Project ''{0}''
```
HannesWell, line 111: Are the double single quotes necessary (Sorry I'm only on my phone and can't try it out atm.)?

laeubi:
@Bananeweizen can you explain why the quotes are required?

Bananeweizen:
@laeubi an example is visible in the linked issue: #1600. project names with blanks are hard to read in the unquoted version. And quotes are used in many of the other m2e progress messages anyway.

@HannesWell The properties are used in NLS, which in turn uses MessageFormat. And that uses single quote ranges for escaped text: 
> Within a String, a pair of single quotes can be used to quote any arbitrary characters except single quotes. For example, pattern string "'{0}'" represents string "{0}", not a FormatElement. A single quote itself must be represented by doubled single quotes '' throughout a String.

https://docs.oracle.com/javase/8/docs/api/java/text/MessageFormat.html

laeubi:
I'm not sure if the quotes make things better readable, maybe we should improve the message itself to be more expressive?

Bananeweizen:
Well, I'm sure that the quotes make the message more readable. :) No need to extend the message, especially because that is only the detail progress message (the lower part of the progress bar), the "what" is in the upper part of the progress bar anyway.

HannesWell (approved):
> Well, I'm sure that the quotes make the message more readable. :) No need to extend the message, especially because that is only the detail progress message (the lower part of the progress bar), the "what" is in the upper part of the progress bar anyway.

I share this assessment.
LGTM.

HannesWell:
Thanks Michael.

PR #1603, opened 2023-11-13
Add ability to add source/resource folder outside of the project basedir using linked folder

RoiSoleil:
Add ability to add source/resource folder outside of the project basedir using linked folder


org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
```diff
@@ -496,10 +510,14 @@ private void addResourceDirs(IClasspathDescriptor classpath, IProject project, M
...
-      File resourceDirectory = new File(directory);
-      IPath relativePath = getProjectRelativePath(project, directory);
-      IResource r = project.findMember(relativePath);
-      if(r == project) {
+      File resourceDirectory = new File(resource.getDirectory());
+      String resourceCanonicalPath = resourceDirectory.getAbsolutePath();
+      try {
+        resourceCanonicalPath = resourceDirectory.getCanonicalPath();
+      } catch(IOException ex) {
+      }
```
HannesWell, line 518: I have not yet tried this out, but I'm not sure this is correct.
If `resource.getDirectory()` is a relative path (haven't checked if that can happen), then it will resolve against the running directory of the application, which will be something like the IDE location.
But here it should be resolved against the project directory, shouldn't it?

Furthermore if `resourceDirectory.getAbsolutePath()` is the fall back in case `getCanonicalPath()` throws, it should only be called in the catch block.

HannesWell (commented):
Thank you for this PR.

In general it looks reasonable.

I couldn't try it out yet, but already have some remarks, see below.

Can you please rebase this on top of the current master?

RoiSoleil:
Thx for your kind review ! I would to the changes you ask for. Nice to see that PR are welcomed ;)

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
RoiSoleil: When i checked resource.getDirectory() were only absolutePath no relativePath

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
```diff
@@ -496,10 +510,14 @@ private void addResourceDirs(IClasspathDescriptor classpath, IProject project, M
       if(directory == null) {
         continue;
       }
-      File resourceDirectory = new File(directory);
-      IPath relativePath = getProjectRelativePath(project, directory);
-      IResource r = project.findMember(relativePath);
-      if(r == project) {
+      File resourceDirectory = new File(resource.getDirectory());
```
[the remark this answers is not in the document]
RoiSoleil: Done

RoiSoleil:
@HannesWell All tests are ok now. Can you review ?

RoiSoleil:
@HannesWell Do you have time for a review ?

HannesWell:
> @HannesWell Do you have time for a review ?

Sorry I'm currently in the pre-christmas rush but will review this ASAP hopefully within the next week.

RoiSoleil:
Ok thx for the reply ;)

RoiSoleil:
@HannesWell Could you take a look ?

HannesWell:
> @HannesWell Could you take a look ?

Thank for your patience and sorry for the delay.

I had a coarse look and was wondering if the change in the m2e-tests submodule is still necessary/intended? If yes please open a PR for that in the tesla repository and resolve the conflicts.

I'll do a detailed review tomorrow, but I think many changes, if any will be required.

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
```diff
@@ -496,9 +512,14 @@ private void addResourceDirs(IClasspathDescriptor classpath, IProject project, M
         continue;
       }
       File resourceDirectory = new File(directory);
-      IPath relativePath = getProjectRelativePath(project, directory);
-      IResource r = project.findMember(relativePath);
-      if(r == project) {
```
HannesWell, line 501: This check does not work anymore, since r is now always an instance/handle for an IFolder and cannot be a project anymore.
Unfortunately my alternative I just have pushed doesn't work either since the absolute path is resolved against the project (with stripped-off drive), when `project.getFolder(projectLocation)` is called.

I have to think about another elegant/efficient solution to check if `resourceCanonicalPath` points to the project location.

HannesWell (commented):
The change looks almost good (one point below).

I have applied a few minor simplifications.

@RoiSoleil please let me know if the changes in the m2e-tests submodule are still necessary.

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
HannesWell: I didn't found a solution as simple/fast as the identity check before, so went on with the obvious one and changed it to compare the absolute resource path with the absolute project location.

In general this class could benefit from a medium clean-up where the re-compuation of the project-location could be avoided. But for now this is fine.

HannesWell (approved):
Given the tests pass, this is now fine for me.
@RoiSoleil are you fine with it as well?

RoiSoleil:
> Given the tests pass, this is now fine for me. @RoiSoleil are you fine with it as well?
I tested it on my use case and it's ok. Thx for your time by the way ;)

RoiSoleil:
@HannesWell Do you need help to solde the failing test ?

HannesWell (approved):
Finally the build is green.
Thank you @RoiSoleil for this contribution.

RoiSoleil:
@HannesWell Thank you for your time and work on this PR.

PR #1625, opened 2023-12-17
classpathentry should respect pom resource excludes/includes

qxo:
so we build project on eclipse without org.eclipse.m2e.core.maven2Builder


org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
```diff
@@ -579,14 +579,29 @@ private void addResourceDirs(IClasspathDescriptor classpath, IProject project, M
...
 
   private void addResourceFolder(IClasspathDescriptor classpath, IPath resourceFolder, IPath outputPath,
-      boolean addTestFlag) {
+      boolean addTestFlag, Resource resource) {
     log.info("Adding resource folder " + resourceFolder);
-    IClasspathEntryDescriptor descriptor = classpath.addSourceEntry(resourceFolder, outputPath, DEFAULT_INCLUSIONS,
-        new IPath[] {IPath.fromOSString("**")}, false /*optional*/);
+    IClasspathEntryDescriptor descriptor = classpath.addSourceEntry(resourceFolder, outputPath, 
+      toIPathList(resource.getIncludes(), null),
+      toIPathList(resource.getExcludes(), "**/*.java"), false /*optional*/);
```
HannesWell, line 586: Can you please explain why you changed the default exclusion pattern? In a quick search I didn't found a reference for that. Please provide a link, in case that's the Maven default exclusion.

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
```diff
@@ -579,14 +579,29 @@ private void addResourceDirs(IClasspathDescriptor classpath, IProject project, M
...
+    if (fileNames == null) {
+      return defaultPattern != null ? new IPath[] {IPath.fromOSString(defaultPattern)} : DEFAULT_INCLUSIONS;
+    }
+    final List<IPath> retList = new ArrayList<>();
+    for (final String files : fileNames) {
+      retList.add(IPath.fromOSString(files));
+    }
+    if (defaultPattern != null) {
+      retList.add(IPath.fromOSString(defaultPattern));
+    }
```
HannesWell, line 601: If a default, pattern is specified I don't think it is supposed to be added in any case.
Please provide a reference if I'm mistaken.

HannesWell (changes requested):
Hello @qxo, thank you for providing this PR.
In general this change looks reasonable, but we have to work through a few things until it can be considered for submission.

First of all, you have to sign the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ecafaq.php), otherwise we are not allowed to accept your change.

Second please add a test-case to org.eclipse.m2e.jdt.tests in order to demonstrate the missing feature and to ensure it will continue to work in the future.

About the change itself, please see the remarks below.

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
qxo: On my some project, the maven builder often not work, so I had to change the exclusion pattern.

After the change, we don't need maven builder anymore

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
qxo: toIPathList(resource.getIncludes(), null) ==> defaultPattern  is null

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
HannesWell: I'm sorry that it does not work for your projects, but M2E is a tool used by a lot of users and has to be as consistent as possible to Maven CLI invocations.
And if standard maven does not define `**/*.java` as default exclusion we won't do it in m2e.

So if you want to exclude it, you have to define it in your pom.xml. You can wrap that in a profile which is activated if the property `m2e.version` is set. M2E injects the property `m2e.version` when projects are build in the IDE, consequently the profile is not activated if build with CLI Maven.

But in general I wonder why you even add `*.java` files in to your resources folder, they are usually placed in the `src` folder.

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
HannesWell: Yes, but I meant the opposite, `defaultPattern` shall not be added at all if filenames is not null. Otherwise the defaults are added all the time, which is not the purpose of defaults.
```suggestion
```

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
qxo: ok `toIPathList(resource.getExcludes(), "**/*.java")` -> `toIPathList(resource.getExcludes(), null)`

PR #1626, opened 2023-12-19
Removes "only" in contributing section

vogella:
"m2e only accepts contributions" has been replaced with "m2e accepts contributions". Only sounds like a restriction which I think it not necessary.


mickaelistria:
Thanks

Issue #1511, opened 2023-08-16
m2e plugin sometimes "loses" resources - using snapshot 2.1.201.20230809-1423


tarioch:
In some cases, the resources are deleted from the target folder and not copied over again.

e.g. one case I encounter is, if I do any change in the pom file, all my resources disappear until I touch any of the resources or java files. 

This is the split up from the issues which are still existing in snapshot 	2.1.201.20230809-1423

See https://github.com/eclipse-m2e/m2e-core/issues/1302 for more details.

Reproduction: https://github.com/eclipse-m2e/m2e-core/pull/1478

laeubi:
@tarioch It actually should have been improved by the last m2e eclipse release, but we can only try to make it less often as the underlying cause is here:
- https://github.com/apache/maven-filtering/pull/77

but it hasn't gotten much attention from the maven people yet (and I'm not a maven committer).

laeubi:
Beside that, sponsoring is a good way [to get specific issues fixed](https://github.com/eclipse-m2e/m2e-core/issues?q=is%3Aissue+label%3Asponsored), so if this is crucial to your business and you likes to speed up the development of m2e in general a [sponsoring](https://github.com/sponsors/laeubi?frequency=one-time) would allow me to assign more time-slots to bug fixing or contact me for an individual contract for example fixing a specific issue, m2e is sadly manly drive by volunteers working on their (spare) free time so it often takes some time if no one can dedicate time or money.

tarioch:
Thanks for your response. Unfortunately even with latest version it still is still very, very bad. Is there a fundamental change why this issue happens since 2022-12? I haven't seen it before.

tarioch:
Did do some more testing, now used the plain eclipse java installation from eclipse and a very tiny project (empty pom file, one empty class and one dummy resource).

### 2022-12/m2e 2.1.2
Adding/removing a dependency to the pom.xml doesn't impact the resources, classes don't seem to get recompiled, resources are untouched

### 2023-03/m2e 2.2.1 (same behavior also in latest 2023-09)
Adding/removing a dependency to the pom.xml does seem to trigger a rebuild, afterwards the resource file is gone and the class file has a newer timestamp

Any idea what the big change in m2e 2.2 was that changed this behavior? From the release notes I saw that there was a change to execute plugins by default in incremental builds and I tried with the setting changed to ignore, but it's still the same behavior.

lihomev5:
[...]

PR #1629, opened 2023-12-20
Fix "m2e plugin sometimes 'loses' resources" (#1511)

kohlschuetter:
The Java Builder may delete files from the project output directory that need to be re-created by the m2e Maven Builder.

With commit 8e5cd4967eaca2698e8f7109461320e11a11d4d7 (a fix for #1275), any changes to the project output directory were ignored, leading to unexpected errors when running an application after modifying the project POM: resources were missing from the target classpath, leading all sorts of unexpected program behavior.

This change allows marking these changes as relevant, as long as the following conditions are met:

- The expected resource no longer exists (i.e., it was deleted by another builder, plugin or outside process)
- The modification applies to a resource in the classes or test-classes folder (i.e., outputLocation/testOutputLocation)

https://github.com/eclipse-m2e/m2e-core/issues/1511 https://github.com/eclipse-m2e/m2e-core/issues/1275

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/builder/MavenBuilderImpl.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/builder/MavenBuilderImpl.java
@@ -209,6 +209,10 @@ public class MavenBuilderImpl {
     if(project == null || buildOutputLocation == null) {
       return true;
     }
+
+    IPath outputLocation = projectFacade.getOutputLocation();
+    IPath testOutputLocation = projectFacade.getTestOutputLocation();
+
     IPath projectPath = project.getFullPath();
     List<IPath> moduleLocations = projectFacade.getMavenProjectModules().stream()
         .map(module -> projectPath.append(module)).toList();
@@ -219,7 +223,12 @@ public class MavenBuilderImpl {
         IPath fullPath = delta.getFullPath();
         if(buildOutputLocation.isPrefixOf(fullPath)) {
           //anything in the build output is not interesting for a change as it is produced by the build
-          //lets see if there are more interesting parts...
+          // ... unless a classpath resource that existed before has been deleted, possibly by another builder
+          if(!resource.exists() && ((outputLocation != null && outputLocation.isPrefixOf(fullPath))
+              || (testOutputLocation != null && testOutputLocation.isPrefixOf(fullPath)))) {
+            hasRelevantDelta.set(true);
+            return false;
+          }
           return true;
         }
         for(IPath modulePath : moduleLocations) {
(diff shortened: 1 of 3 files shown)


kohlschuetter:
CI failure: "Baseline problems found! Project version ..." can be fixed with https://github.com/eclipse-m2e/m2e-core/pull/1630

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/builder/MavenBuilderImpl.java
```diff
@@ -219,7 +223,12 @@ private boolean hasRelevantDelta(IMavenProjectFacade projectFacade, IResourceDel
         IPath fullPath = delta.getFullPath();
         if(buildOutputLocation.isPrefixOf(fullPath)) {
           //anything in the build output is not interesting for a change as it is produced by the build
-          //lets see if there are more interesting parts...
+          // ... unless a classpath resource that existed before has been deleted, possibly by another builder
+          if(!resource.exists() && ((outputLocation != null && outputLocation.isPrefixOf(fullPath))
```
laeubi, line 227: Please split this into smaller parts or maybe a method as such nested conditions are hard to read and understand, for example the frist and the second only differ in the location,

laeubi:
> CI failure: "Baseline problems found! Project version ..." can be fixed with #1630

Please include any version changes in your PR here.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/builder/MavenBuilderImpl.java
```diff
@@ -219,7 +225,11 @@ private boolean hasRelevantDelta(IMavenProjectFacade projectFacade, IResourceDel
         IPath fullPath = delta.getFullPath();
         if(buildOutputLocation.isPrefixOf(fullPath)) {
           //anything in the build output is not interesting for a change as it is produced by the build
-          //lets see if there are more interesting parts...
+          // ... unless a classpath resource that existed before has been deleted, possibly by another builder
+          if(!resource.exists() && isOutputOrTestOutput.test(fullPath)) {
```
laeubi, line 229:
```suggestion
          if(isOutputOrTestOutput.test(fullPath) && !resource.exists()) {
```
checking the prefix is probably a bit cheaper than checking all resources for existence.

laeubi (changes requested):
The change looks good in general now and it seems a valid assumption that if the output folder is gone something must be happen.

But thinking more about it I think this is more fundamentally problem even though it is fixing things for resource plugin others might behave odd, one example is the maven-bundle-plugin or bnd-maven-plugins that might also place resources there.

Because of this, please apply the following changes:

1. `hasRelevantDelta` should return an enum that has the values: `DELTA` (the current 'true' return value), `FULL` (when detecting the new output folder is gone), `IRRELEVANT` (the current 'false' return value.
2. In `setupProjectBuildContext` a value of `FULL` should result in a `null` build delta passed so we make sure a full build is executed and not an incremental one.

kohlschuetter:
@laeubi I think your recently requested changes are beyond the scope of the bug, and therefore better suited for a separate refactoring. I also feel that it's more reasonable if you implement these changes the way you envision them, it will save us both unnecessary back and forth.

PR #1634, opened 2023-12-25
Perform full build in case the output directory was deleted

laeubi:
Actually the check should also be enhanced that if the `target` folder was deleted ... currently only the `target/classes``target/test-classes` are considered as a fatal deletion.

We just need to check how this behaves on a clean build ...


HannesWell (approved):
LGTM.

Issue #391, opened 2021-11-03
Update Project removes custom classpath entries


peterjeschke:
We use external annotations in our maven project. Updating the maven project causes the .classpath file to be regenerated, removing all references to the external annotations. Is it possible to put the configuration for external annotations in the pom.xml so that the m2e plugin configures the .classpath file correctly?

sebthom:
I am having the same issue. Would be great if the custom external annotations specified would survive a maven project update.

![image](https://user-images.githubusercontent.com/426959/199519674-5bb66c9f-3035-44d9-a5ef-a8c46e6b360e.png)

laeubi:
@sebthom it seems this is not really something the usual committer/contributors encounter and thus handled with very low priority, so probably you like to provide a PR to fix that together with a test-case to verify it works?

Beside that, sponsoring is a good way [to get specific issues fixed](https://github.com/eclipse-m2e/m2e-core/issues?q=is%3Aissue+label%3Asponsored), so if this is crucial to your business and you likes to speed up the development of m2e in general a [sponsoring](https://github.com/sponsors/laeubi?frequency=one-time) would allow me to assign more time-slots to bug fixing or contact me for an individual contract for example fixing a specific issue.

PR #1635, opened 2023-12-27
 fix: updating project config removes external annotations path #391 

sebthom:


laeubi:
@sebthom can you squash your commits and add a commit message that described the current and improved state?

sebthom:
@laeubi done

sebthom:
@HannesWell @laeubi I would greatly appreciate it if this could be included in the next m2e release. 🤞

HannesWell (approved):
@sebthom thank you for this contribution, it looks good.
I just reworked it a bit to reduce code-duplication and to stream-line the creation of new classpath-entries, but the behavior should be the same (as your tests shows).


https://github.com/eclipse-m2e/m2e-core/pull/1664

sebthom:
@HannesWell Looks good to me. Thanks for reviewing/updating!

HannesWell:
> @HannesWell Looks good to me. Thanks for reviewing/updating!

Great. Thanks for the fast feedback.
All builds are green, submitting.

PR #1638, opened 2024-01-03
Add lifecycle mapping for maven-toolchains-plugin #1637

G-Ork:
For #1637

### DIFF
--- a/org.eclipse.m2e.core/lifecycle-mapping-metadata.xml
+++ b/org.eclipse.m2e.core/lifecycle-mapping-metadata.xml
@@ -199,6 +199,19 @@
         <ignore/>
       </action>
     </pluginExecution>
+    <pluginExecution>
+      <pluginExecutionFilter>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-toolchains-plugin</artifactId>
+        <versionRange>[0,)</versionRange>
+        <goals>
+          <goal>toolchain</goal>
+        </goals>
+      </pluginExecutionFilter>
+      <action>
+        <ignore />
+      </action>
+    </pluginExecution>
 
     <!-- commonly used codehaus plugins -->
     <pluginExecution>


.gitignore
```diff
@@ -4,3 +4,6 @@ bin/
 .polyglot*
 .tycho-consumer-pom.xml
 pom.tycho
+.classpath
+.settings
+.project
```
HannesWell, line 9: Please revert this addition. There files are relevant for Eclipse project.
```suggestion
```

org.eclipse.m2e.apt.core/lifecycle-mapping-metadata.xml
```diff
@@ -96,6 +96,19 @@
...
+      <pluginExecutionFilter>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-toolchains-plugin</artifactId>
+        <versionRange>[2.0,)</versionRange>
+        <goals>
+          <goal>toolchain</goal>
+        </goals>
+      </pluginExecutionFilter>
+      <action>
+        <ignore></ignore>
```
HannesWell, line 109:
```suggestion
        <ignore />
```

org.eclipse.m2e.apt.core/lifecycle-mapping-metadata.xml
```diff
@@ -96,6 +96,19 @@
...
+      </action>
+    </pluginExecution>
```
HannesWell, line 111: Since this the maven-toolchains-plugin is a very general one I think its silencing is better placed into `org.eclipse.m2e.core/lifecycle-mapping-metadata.xml`

HannesWell (changes requested):
Thank you for this PR, the general intend looks good, but the lifecycle mappings should not be added to the `m2e.apt.core` plugin (see the comments).

.gitignore
G-Ork: I will do, but they are not under version control. They should not be under version control. Just sayin'

org.eclipse.m2e.apt.core/lifecycle-mapping-metadata.xml
G-Ork: I've moved the metadata. Thank you for the hint,

.gitignore
HannesWell: They are under version control. Why should they not be under version control? M2E mainly consists out of Eclipse Plugin projects that require at least a `.project` and `.classpath` file since they cannot be derived, like for example for Maven projects. Or do you have something else in mind?

org.eclipse.m2e.core/lifecycle-mapping-metadata.xml
```diff
@@ -199,6 +199,20 @@
         <ignore/>
       </action>
     </pluginExecution>
+    <pluginExecution>
+      <pluginExecutionFilter>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-toolchains-plugin</artifactId>
+        <versionRange>[2.0,)</versionRange>
```
HannesWell, line 206: Do you have a specific reason for this version range?
Looking at M-C, there is no version 2 at all:
https://repo1.maven.org/maven2/org/apache/maven/plugins/maven-toolchains-plugin/

G-Ork: No specific version, just grobe Kelle.

.gitignore
G-Ork: .classpath is not. should be ignored. :)

HannesWell (approved):
Looks good now. Lets wait for the result of the CI.

.gitignore
HannesWell: Sorry, but no for Plugin projects `.classpath` should not be ignored.

HannesWell:
@G-Ork it looks like this causes new test-failures. Could you have a look at them?

G-Ork:
I will. Seemed to be code thats generated somehow. Can not find the test classes mentioned.

HannesWell:
> I will. Seemed to be code thats generated somehow. Can not find the test classes mentioned.

These tests live in another repository, i.e. https://github.com/tesla/m2e-core-tests/, which is included as sub-module of this repository. Unfortunately, for legal reasons we are not allowed to add these tests to this repository. :/

The contribution guide explains how to do that manually: https://github.com/eclipse-m2e/m2e-core/blob/master/CONTRIBUTING.md#%EF%B8%8F-setting-up-the-development-environment-manually
or you just use the provided Oomph setup.

Anyway, it looks you already found a hot candidate for being the cause of the failures. 🚀

G-Ork:
Is the check **Build M2Eclipse / build (macos-latest) (pull_request)**  related to [m2e-core-tests](https://github.com/tesla/m2e-core-tests/) or is this another problem? Looks like it was executed before my last fixes and not again afterwards.

HannesWell:
> Is the check **Build M2Eclipse / build (macos-latest) (pull_request)** related to [m2e-core-tests](https://github.com/tesla/m2e-core-tests/) or is this another problem? Looks like it was executed before my last fixes and not again afterwards.

This is unrelated, to your change. All tests look good now. 👍🏽

Could you please just revert the version bumps in `org.eclipse.m2e.core/META-INF/MANIFEST.MF` and `org.eclipse.m2e.core/pom.xml`. This plugin's version has already been bumped since the last release so bumping it again should not be necessary.
Then this PR is ready for submission.

G-Ork:
Thank you for the patience. I have removed the version bumps. 
I've read it but did not check it :man_facepalming:

HannesWell (approved):
> Thank you for the patience. I have removed the version bumps. I've read it but did not check it 🤦‍♂️

No problem, you are welcome.
Thanks again for this contribution.

HannesWell:
> but m2e seems to not perform this currently in `org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.createExecutionRequest(IMavenConfiguration, IComponentLookup, Settings)` this would be much more appropriate than disable the execution of the mojo.

Actually this was implement as part of the PR that added the UI:
https://github.com/eclipse-m2e/m2e-core/pull/1643/files#diff-796a271293c65b5cac9c04a4dccddff344e43fd10f474a5cd2992c2333325704

and is now performed at
https://github.com/eclipse-m2e/m2e-core/blob/53c0eafe4f20deeedde635fd21f53be63e1d544b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java#L211-L214

So should we consider reverting this? @G-Ork what's your opinion about that?

G-Ork:
Hi, i am currently preparing a PR with unit tests as asked for the change of the UI. The tests are finished and where necessary because the settings where not applied to the maven launcher. My fault. My Plan was to create the PR for the UI  and the unit tests. The PR Should be ready today or at least this weekend. 

After this i would dig into the open questions about the internal launch. This relate to the code you quoted. My opinion is that we should try to make it work. Or make it work somehow. I stumbled already over the code that configures different JVMs based on the enforcer plugin. That feature was added last year. I didn't test it yet but i guess i will help.

In my opinion this is quite hacky as the enforcer plugin is a  assertion plugin and not for configuration. I was surprised as another developer mention that. Also found inside the hidden [releasenotes](https://github.com/eclipse-m2e/m2e-core/blob/044cb60d31a621fd0dbce9e5b0997adc9da1e68e/RELEASE_NOTES.md?plain=1#L141). I would prefer a toolchain approach. I also run into this requirement this week by trying to migrate a project to java 21. Eclipse used the start JRE and i wasn't able to build with the internal build. Workaround was to use the Maven launcher and configure a 21 JDK.

PR #1639, opened 2024-01-03
Bug 571572 - UI for setting the location of the toolchain.xml files #114

G-Ork:
I've added the text field to configure the user toolchains definition.  Its just for the convenience.

Other changes:

- toolchain validation based on the maven factories.
- open file in eclipse editor just like the user settings file.
- Set the user toolchains file in MavenExecutionContext.

Issue #114, opened 2021-03-30
Bug 571572 - UI for setting the location of the toolchain.xml files


mickaelistria:
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=571572

> The location of the toolchain.xml files (user and global scopes) should be able to be set from the UI. The current "User Settings" preferences page could be the right place for it.
> 
> Then, the -gt and/or -t options would be used when invoking Maven so that the values set in the UI were passed to Maven.

laeubi:
@mickaelistria as you recently worked in that area, do you think you can add support for configuring toolchains?

mickaelistria:
Sorry, I don't plan to work on it at this moment. Should I close as I reprorted it and it got inactive for nore than 2 years?

laeubi:
We should just keep it I think it would be useful to have this as a reminder.

PR #1643, opened 2024-01-08
 Bug 571572 - UI for setting the location of the toolchain.xml files #114 

G-Ork:
I've added the text field to configure the user toolchains definition.  Its just for the convenience.

Other changes:

- toolchain validation based on the maven factories.
- open file in eclipse editor just like the user settings file.
- Set the user toolchains file in MavenExecutionContext.

Sorry i've crushed the previous PR #1639 by lack of git skills :)


org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/IMaven.java
```diff
@@ -215,6 +216,14 @@ <T> T getMojoParameterValue(MavenProject project, String parameter, Class<T> typ
 
   List<SettingsProblem> validateSettings(String settings);
 
+  /**
+   * Validates the toolchains definition
+   * 
+   * @param toolchains The path to the toolchains definition file to test.
+   * @return List of all problems. Is never <code>null</code>.
+   */
+  List<Problem> validateToolchains(String toolchains);
```
laeubi, line 225: I think we should not add new methods to `IMaven` anymore, can you addit to the `org.eclipse.m2e.core.internal.IMavenToolbox` instead?

G-Ork: Null problemo. method relocated.

HannesWell (changes requested):
@G-Ork thank you for this contribution.
It looks really good and is a good enhancement for M2E.

I have some remarks below.
In general, especially in the preferences, there is now (and even before) a lot of code-duplication. But this can also be cleaned up in a follow-up change.

HannesWell (approved):
@G-Ork I have now applied all the requested changes by myself and added a separate first commit where I did the unifications.

With that this is ready for submission.

What would be great, if you could add a test-case in a follow-up PR to verify that it works as expected and is ensured to continue to work in the future.

G-Ork:
@HannesWell Thank you for for doing the cleanup. I will care for a test the next days. 

I have also a strong wish to change the design of that page. I do not like the vertical stacked layout (label, input). I would group in global and user settings and make the layout horizontal (label -> input -> button(s)). I guess we separate this in two PRs?

HannesWell:
> I have also a strong wish to change the design of that page. I do not like the vertical stacked layout (label, input). I would group in global and user settings and make the layout horizontal (label -> input -> button(s)). I guess we separate this in two PRs?

Thanks for working on the test. Yes please split that into two separate PRs.
If you want to re-design the page please also keep in mind that the labels don't have the same length and that the length can even vary (i.e. if the specified file is not available respectively no choice is made, like for the global settings). I'm not sure about the beauty of the page if that varies, but nevertheless I'm interested about your solution. :)

G-Ork:
Here is a rough draft. I would remove the links for the editor and replace them with buttons. For sake of completeness i would also list the global toolchain. Icons for the buttons instead of the labels would also be more colorful. 

Your opinion in advance would be nice to prevent unnecessary work.

Draft-Gridlayout-ico

HannesWell:
> Your opinion in advance would be nice to prevent unnecessary work.

Looks good in general. But regarding Icons vs Text, it is quite common in the eclipse UI to use texts on buttons instead of icons. For the sake of consistency we should probably keep that.

mickaelistria:
> it is quite common in the eclipse UI to use texts on buttons instead of icons

It's used sometimes in Platform though (eg Project > Properties > Resources has icons for navigate or copy).
I think it's a decent pattern per se, but in such case where there is an icon, we need to think about accessibility and people using a screen reader: if icon doesn't show text directly, it is important to still set a text in hover so screen readers can find something to read.

G-Ork:
I do not insist on icons. So the next is just my opinion and not serious attempt to argument for icons:

The use of icons and the constants for common icons is not well documented. I guess this this the main reason for the plain designs. 

@mickaelistria I think accessibility is not the problem. The button will get tooltip.

HannesWell:
> The use of icons and the constants for common icons is not well documented. I guess this this the main reason for the plain designs.

That could be the case, but I'm not sure.

But I suggest you create a PR from your current changes and we discuss it there. :)

PR #1653, opened 2024-01-18
Avoid throwing NPE from EclipseLogger if no throwable was passed

### LINKED ISSUE #1652: EclipseLogger can throw NPE if throwable is null
```
Caused by: java.lang.NullPointerException: Cannot invoke "java.lang.Throwable.getMessage()" because "t" is null
	at org.eclipse.m2e.core.internal.embedder.EclipseLogger.warn(EclipseLogger.java:71)
	at org.apache.maven.monitor.logging.DefaultLog.warn(DefaultLog.java:72)
[snip]
```

I think the logger should be very robust and not fail on absent throwables.

szarnekow:
This closes #1652

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/EclipseLogger.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/EclipseLogger.java
@@ -40,7 +40,7 @@ class EclipseLogger implements Logger {
   @Override
   public void debug(String msg, Throwable t) {
     if(isDebugEnabled()) {
-      log.debug(msg + " " + t.getMessage(), t);
+      log.debug(toMessageString(msg, t), t);
     }
   }
 
@@ -54,7 +54,7 @@ class EclipseLogger implements Logger {
   @Override
   public void info(String msg, Throwable t) {
     if(isInfoEnabled()) {
-      log.info(msg + " " + t.getMessage(), t);
+      log.info(toMessageString(msg, t), t);
     }
   }
 
@@ -68,7 +68,7 @@ class EclipseLogger implements Logger {
   @Override
   public void warn(String msg, Throwable t) {
     if(isWarnEnabled()) {
-      log.warn(msg + " " + t.getMessage(), t);
+      log.warn(toMessageString(msg, t), t);
     }
   }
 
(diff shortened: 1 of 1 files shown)


laeubi (approved):
Looks good

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/EclipseLogger.java
```diff
@@ -143,4 +143,14 @@ public Logger getChildLogger(String name) {
...
   }
+
+  private String toMessageString(String msg, Throwable t) {
+    if(t == null || t.getMessage() == null) {
+      return msg;
+    }
+    if(msg == null) {
+      return t.getMessage();
+    }
+    return msg + " " + t.getMessage();
```
HannesWell, line 154: For log messages it is usually good practice to only pass a format string and its arguments separately so the string object creation can be avoided if the message is eventually not logged.

szarnekow: I read your words but I don't understand what I can do about it.

laeubi: As this is a 1:1 as before I think its actually fine here, all calls are already guarded by a check if the log level is enabled so it will be logged with no doubt.

szarnekow: You're probably not referring to avoiding `t.getMessage()` in case the message is produced lazily, are you?

szarnekow: Also I don't see an overload in slf4j with a format string *and* a throwable.

HannesWell: > I read your words but I don't understand what I can do about it.

I was thinking about something like https://www.slf4j.org/manual.html#typical_usage

But Christoph is right, the calls to this method are already guarded (Sorry I oversaw that at my phone).

HannesWell (approved):

szarnekow:
I don't think the test failure can be related to the proposed change, can it?

PR #1657, opened 2024-01-23
Don't set target feature version to default if it is already set

kysmith-csg:
Previously, editing a target feature location would always set the version to `1.0.0.qualifier`, even if the version was set already. This is because this `initialize()` method is called _after_ the feature details are filled in.

Now, editing an existing feature will fill in the details as shown in the target file. New features will continue to be set to `1.0.0.qualifier`.

### DIFF
--- a/org.eclipse.m2e.pde.feature/feature.xml
+++ b/org.eclipse.m2e.pde.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.m2e.pde.feature"
       label="%featureName"
-      version="2.3.0.qualifier"
+      version="2.3.1.qualifier"
       provider-name="%providerName"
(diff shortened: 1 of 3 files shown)


laeubi (approved):
looks reasonable

laeubi:
@kysmith-csg you need to bump the version of the affected bundle.

kysmith-csg:
> @kysmith-csg you need to bump the version of the affected bundle.

Done, sorry about that.

HannesWell:
@kysmith-csg is this not relevant anymore?

kysmith-csg:
@HannesWell  it is still valid. I went to Github to update my fork and I completely forgot this pull request was open still. Luckily I still had the commit locally :)

Issue #1661, opened 2024-01-30
Allow configiration of automatic project updates per project


HannesWell:
Currently it is only possible at the workspace level to configure if
- Update Maven projects on startup
- Automatically update Maven projects configuration

is enabled. 

![grafik](https://github.com/eclipse-m2e/m2e-core/assets/44067969/66c53298-a393-4cd2-91e9-dd11fc476539)

While this is useful in general it can be annoying if one has a specifically configuration project that does not match the automatic configuration. Therefore it should be possible to disable this automatic configuration per-project so that one can e.g. add the corresponding preference under version control.

PR #1664, opened 2024-02-02
Make automatic configuration updates configurable per project

HannesWell:
And degrade problems about out-dated project configuration to severity 'info', if automatic updates are disabled for a project.

grafik

Fixes https://github.com/eclipse-m2e/m2e-core/issues/1661

@laeubi what do you think?
I'm not sure adding API methods is really necessary (for now I just added it for completes/symmetry) but it is probably sufficient to have this internal.

### DIFF
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/project/MavenUpdateConfigurationChangeListener.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/project/MavenUpdateConfigurationChangeListener.java
@@ -25,6 +25,7 @@ import org.eclipse.core.resources.IResourceChangeListener;
 import org.eclipse.core.runtime.CoreException;
 
 import org.eclipse.m2e.core.MavenPlugin;
+import org.eclipse.m2e.core.internal.project.ResolverConfigurationIO;
 import org.eclipse.m2e.core.ui.internal.UpdateMavenProjectJob;
 
 
@@ -40,10 +41,10 @@ public class MavenUpdateConfigurationChangeListener implements IResourceChangeLi
 
   @Override
   public void resourceChanged(IResourceChangeEvent event) {
-    if(isDisabled()) {
+    if(isAutoConfigurationUpdateDisabled()) {
       return;
     }
-    List<IProject> outOfDateProjects = null;
+    List<IProject> outOfDateProjects;
     try {
       OutOfDateConfigurationDeltaVisitor visitor = new OutOfDateConfigurationDeltaVisitor();
       event.getDelta().accept(visitor);
@@ -52,10 +53,12 @@ public class MavenUpdateConfigurationChangeListener implements IResourceChangeLi
       LOG.error("An error occurred while checking for out-of-date configuration markers", e);
       return;
     }
+    outOfDateProjects = outOfDateProjects.stream() //
+        .filter(ResolverConfigurationIO::isAutomaticallyUpdateConfiguration).toList();
     updateProjectConfiguration(outOfDateProjects);
   }
 
-  private boolean isDisabled() {
+  public static boolean isAutoConfigurationUpdateDisabled() {
     return !MavenPlugin.getMavenConfiguration().isAutomaticallyUpdateConfiguration();
   }
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -8,6 +8,13 @@
 
 Updated the embedded Maven from version 3.9.5 to 3.9.6; [Maven 3.9.6 Release Notes](https://maven.apache.org/docs/3.9.6/release-notes.html).
 
+### New project preference for automated Maven project configuration updates
+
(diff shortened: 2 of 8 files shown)


HannesWell:
> I'm not sure adding API methods is really necessary (for now I just added it for completes/symmetry) but it is probably sufficient to have this internal.

I have now reverted the changes of the `IProjectConfiguration` interface, but could not avoid new methods in `ResolverConfiguration`. In order to prevent/discourage external usage, I annotated them with `@noreference`.

HannesWell:
> > I'm not sure adding API methods is really necessary (for now I just added it for completes/symmetry) but it is probably sufficient to have this internal.
> 
> I have now reverted the changes of the `IProjectConfiguration` interface, but could not avoid new methods in `ResolverConfiguration`. In order to prevent/discourage external usage, I annotated them with `@noreference`.

Now the change is fully internal and does not need any new methods in any API (that would need `@noreference`).

laeubi (approved):
Looks good just one thing for the UI I think it must be a radio group with the options:

- [ ] Use Workspace default (<maybe link to workspace settings>
- [ ] Update Automatically
- [ ] Never update automatically

Also from the code it currently looks like if the project has **enabled** it but it is disabled in the workspace settings no update will ever happen.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/project/MavenUpdateConfigurationChangeListener.java
```diff
@@ -40,10 +41,10 @@ public class MavenUpdateConfigurationChangeListener implements IResourceChangeLi
 
   @Override
   public void resourceChanged(IResourceChangeEvent event) {
-    if(isDisabled()) {
+    if(isAutoConfigurationUpdateDisabled()) {
```
laeubi, line 44: This must be removed or disbaled in the workspace can't be changed by a project.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/project/MavenUpdateConfigurationChangeListener.java
```diff
@@ -52,10 +53,12 @@ public void resourceChanged(IResourceChangeEvent event) {
       LOG.error("An error occurred while checking for out-of-date configuration markers", e);
       return;
     }
+    outOfDateProjects = outOfDateProjects.stream() //
+        .filter(ResolverConfigurationIO::isAutomaticallyUpdateConfiguration).toList();
     updateProjectConfiguration(outOfDateProjects);
   }
 
-  private boolean isDisabled() {
+  public static boolean isAutoConfigurationUpdateDisabled() {
```
laeubi, line 61: I would add an IProject parameter here then it could:

1. Check if there is a project config and if yes choose this one unless it is configured to use the workspace default
2. fallback to the global default

```suggestion
  public static boolean isAutoConfigurationUpdate(IProject) {
```

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/project/MavenUpdateConfigurationChangeListener.java
```diff
@@ -52,10 +53,12 @@ public void resourceChanged(IResourceChangeEvent event) {
       LOG.error("An error occurred while checking for out-of-date configuration markers", e);
       return;
     }
+    outOfDateProjects = outOfDateProjects.stream() //
+        .filter(ResolverConfigurationIO::isAutomaticallyUpdateConfiguration).toList();
```
laeubi, line 57:
```suggestion
        .filter(MavenConfigurationChangeListener::isAutoConfigurationUpdate).toList();
```

HannesWell:
> Also from the code it currently looks like if the project has **enabled** it but it is disabled in the workspace settings no update will ever happen as there is an early exit.

That's right and this was actually by design, based on the assumption that this is mainly used to disable specific projects in their preferences and that the corresponding preferences is committed in the SCM.
I assume the other way round is not very common since updates are enabled by default and if one disables them it is probably expected to have a workspace-global effect.

With that in mind I wanted to keep the preferences simple. But the current behavior is already reflected in the project preference since it is disabled, when the workspace preference is disabled:
grafik

laeubi:
I think it depends a bit, e.g. if I have a problem with many projects I might want to have this only on some projects where it works well, or only enable it step by step ... but of course we can add it also if its really a demand it is just that usually project settings take precedence over global settings so this might be a bit of a surprise.

HannesWell:
> I think it depends a bit, e.g. if I have a problem with many projects I might want to have this only on some projects where it works well, or only enable it step by step ... but of course we can add it also if its really a demand it is just that usually project settings take precedence over global settings so this might be a bit of a surprise.

In that case, you are right. But I think the current setup is sufficient for most case (I assume in the workspace it is turned on in most cases as it is default) and I would prefer to wait for further enhancements until there is a real demand for it.

Issue #700, opened 2022-05-16
Consider injecting maven-surefire/failsafe-plugin argLine into Eclipse JUnit Runner


victornoel:
I noticed that the `argLine` I specify in the maven-surefire-plugin configuraton are not taken into account by the JUnit Runner.

I'm thinking it would be relevant for m2e to inject those into Eclipse's.

This is a subject that was raised in the past here: https://stackoverflow.com/questions/12392926/maven-surefire-arguments-when-running-junit-test-via-eclipse

And today I'm facing it becuase with Java 17, it is very common to pass some more option to the JVM in order to allow some inter-module access:
```xml
			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-surefire-plugin</artifactId>
				<configuration>
					<argLine>--add-opens java.base/java.lang=ALL-UNNAMED --add-opens jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED</argLine>
				</configuration>
			</plugin>
```

chicobento:
Intellij supports this feature: https://www.jetbrains.com/help/idea/running-tests.html
I do really miss this on eclipse/m2e to improve developer experience.

mickaelistria:
@chicobento would you be willing to contribute this feature to m2e?

chicobento:
hi @mickaelistria , I'd love to contribute with this feature but unfortunately I dont think I'll have time available in the near future. Will try if I find some.

mickaelistria:
I suspect you could start by putting a breakpoint in EclipseMavenLauncher.runMaven(....) and then debug it to see the call stack, From the call stack, you'll be able to find out what is the piece of code that seems to suit best for this task.

Issue #174, opened 2021-04-20
systemPropertyVariables ignored


jeacott1:
it looks like this is an ancient issue and despite it claiming here it was transferred, I couldn't find any ref to it here.
ref https://bugs.eclipse.org/bugs/show_bug.cgi?id=388683

the maven failsafe/surefire ```<systemPropertyVariables>``` properties are ignored.

HannesWell:
The last comment on that issue unfortunately tends to be misleading. It was intend as general statement that m2e issues are moved to github.

However, as stated in comment 1 of the linked bug, executing a Test via `Run As -> JUnit Test` is not connected to the Maven configuration at all and therefore does not consider Maven configurations at the moment (and there are no plans to implement that).

As a workaround you can add the desired system-properties to the JUnit launch-configuration in the `Arguments` tab. To store and share that launch configuration, I suggest to save it as a shared file (`Common` tab) in a project and check it into your version-control-system.

garretwilson:
I would like to revisit this issue.

On [SLF4J-592](https://jira.qos.ch/browse/SLF4J-592) we're discussing the best approach for specifying a logging provider to be used only in Maven tests without interfering with the main application logging provider (if any). The idea is to configure a logging provider for tests in a parent POM with `test` scope, but have it take precedence in descendant POM tests even if the main project has the production logging provider on the classpath.

The simplest and elegant solution we're converging on so far is to simply have a system property e.g. `slf4j.provider` to indicate the logging provider to use in the case that several appear on the classpath.

Whether this works depends on whether it is possible to specify a system property for Maven tests. Both the [Maven Surefire Plugin](https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#systemPropertyVariables) and the [Maven Failsafe Plugin](https://maven.apache.org/surefire/maven-failsafe-plugin/integration-test-mojo.html#systemPropertyVariables) provide a `<systemPropertyVariables>`, with a description, "List of System properties to pass to a provider". The Surefire [Using System Properties](https://maven.apache.org/surefire/maven-surefire-plugin/examples/system-properties.html) documentation have some examples.

I've tested this in Maven 3.9.1 with Surefire.

* I specified a system property e.g. `foo` in `<systemPropertyVariables>` in a parent POM.
* I retrieved the system property in the unit test of an inheriting descendant POM, and the value was what I had set. I assume it will work in Failsafe as well for integration tests.

[...]

PR #1672, opened 2024-02-13
Surefire/Failsafe plugin configuration propagated to Junit/TestNG launch configuration

treilhes:
The following arguments are supported: <br/>
`<argLine>`, <br/>
`<environmentVariables>`, <br/>
`<systemPropertyVariables>`, <br/>
`<workingDirectory>`,<br/>
`<enableAssertions>`,<br/>

Configuration is propagated on unit test launch configuration creation and also when executing `maven > update project`

By default and if found, the plugin maven-dependency-plugin (goal: properties) is executed before updating the launch configuration to load properties.

If properties set by other plugins are used in the failsafe/surefire plugin configuration it is possible to override the default loading behaviour by providing the list plugins and goals to execute using the property `m2e.launch.configuration.prerequisites`

The expected format is as follow:

groupId1:artifactId1:goal1[,groupIdX:artifactIdX:goalX]*

Ex:
<m2e.launch.configuration.prerequisites>org.apache.maven.plugins:maven-dependency-plugin:properties,org.codehaus.mojo:properties-maven-plugin:read-project-properties</m2e.launch.configuration.prerequisites>

### DIFF
--- a/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/UnitTestSupport.java
+++ b/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/UnitTestSupport.java
@@ -1,0 +1,435 @@
+/********************************************************************************
+ * Copyright (c) 2024, 2024 Pascal Treilhes and others
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Pascal Treilhes - initial API and implementation
+ ********************************************************************************/
+
+package org.eclipse.m2e.jdt.internal;
+
+import java.io.File;
+import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.nio.file.Files;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.StringJoiner;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.debug.core.DebugPlugin;
+import org.eclipse.debug.core.ILaunchConfiguration;
+import org.eclipse.debug.core.ILaunchConfigurationType;
+import org.eclipse.debug.core.ILaunchConfigurationWorkingCopy;
+import org.eclipse.debug.core.ILaunchManager;
+import org.eclipse.jdt.launching.IJavaLaunchConfigurationConstants;
+import org.eclipse.jdt.launching.JavaRuntime;
+
+import org.apache.maven.plugin.Mojo;
+import org.apache.maven.plugin.MojoExecution;
+import org.apache.maven.project.MavenProject;
+
+import org.eclipse.m2e.core.MavenPlugin;
+import org.eclipse.m2e.core.embedder.IMaven;
+import org.eclipse.m2e.core.internal.IMavenConstants;
+import org.eclipse.m2e.core.project.IMavenProjectFacade;
+import org.eclipse.m2e.jdt.internal.launch.MavenRuntimeClasspathProvider;
+
+
+/**
+ * This class is used to support the launch configuration of JUnit and TestNG tests
+ */
+public class UnitTestSupport {
+
+  /**
+   * Feature flag to enable or disable the support
+   */
+  private static final boolean FEATURE_ENABLED = Boolean
+      .parseBoolean(System.getProperty("m2e.process.test.configuration", "true"));
+
+  private static final Logger LOG = LoggerFactory.getLogger(UnitTestSupport.class);
+
+  /**
+   * org.apache.maven.surefire.api.testset.TestListResolver method name to check if a test should run
+   */
+  private static final String SHOULD_RUN_METHOD = "shouldRun";
+
+  /**
+   * org.apache.maven.plugin.surefire.AbstractSurefireMojo method name to get the included and excluded tests
+   */
+  private static final String GET_INCLUDED_AND_EXCLUDED_TESTS_METHOD = "getIncludedAndExcludedTests";
+
+  /**
+   * Launch configuration attribute for the main class
+   */
+  private static final String LAUNCH_CONFIG_MAIN_CLASS = "org.eclipse.jdt.launching.MAIN_TYPE";
+
+  /**
+   * Launch configuration attribute for the project
+   */
+  private static final String LAUNCH_CONFIG_PROJECT = "org.eclipse.jdt.launching.PROJECT_ATTR";
+
+  /**
+   * Launch configuration attribute for the VM arguments
+   */
+  public static final String LAUNCH_CONFIG_VM_ARGUMENTS = "org.eclipse.jdt.launching.VM_ARGUMENTS";
+
+  /**
+   * Launch configuration attribute for the environment variables
+   */
+  public static final String LAUNCH_CONFIG_ENVIRONMENT_VARIABLES = "org.eclipse.debug.core.environmentVariables";
+
+  /**
+   * Launch configuration attribute for the system properties
+   */
+  private static final String LAUNCH_CONFIG_WORKING_DIRECTORY = "org.eclipse.jdt.launching.WORKING_DIRECTORY";
+
+  /**
+   * Launch configuration attribute for the working directory
+   */
+  private static final String PLUGIN_ARGLINE = "argLine";
+
+  /**
+   * surefire/failsafe mojo configuration element for the environment variables
+   */
+  private static final String PLUGIN_ENVIRONMENT_VARIABLES = "environmentVariables";
+
+  /**
+   * surefire/failsafe mojo configuration element for the system properties
+   */
+  private static final String PLUGIN_SYSPROP_VARIABLES = "systemPropertyVariables";
+
+  /**
+   * surefire/failsafe mojo configuration element for the working directory
+   */
+  private static final String PLUGIN_WORKING_DIRECTORY = "workingDirectory";
+
+  /**
+   * surefire/failsafe mojo configuration element for the enable assertions
+   */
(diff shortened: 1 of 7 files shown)


treilhes:
Solves #700  and probably #174

HannesWell:
@treilhes thank you for this contribution. In general it is interesting and definitely a good thing to support.

Unfortunately I won't have time to review this large contribution over the weekend and will then be on vacation for three weeks.
So unless another committer reviews this your patience is kindly requested.

treilhes:
@HannesWell 
Don't worry, my patience will handle this just fine. Have great holidays. thank you for having warned

treilhes:
@HannesWell, I miss you so much, I long to hear from you about your vacation (and to get a bit of review) :broken_heart:

HannesWell:
> @HannesWell, I miss you so much, I long to hear from you about your vacation (and to get a bit of review) 💔

My vacation was great, thanks for asking. I'll do my best to review this at the weekend (when I arrived back home a full ToDo-List hit immediately).

treilhes:
@HannesWell Bump ❤️

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/UnitTestSupport.java
```diff
@@ -0,0 +1,776 @@
...
+      return isTestHandled;
+    }
+
+    /**
+     * Lookup method in class and ancestors
+     * 
+     * @param instance the instance on which we search the method
+     * @return an optional method
+     */
+    private Optional<Method> findMethod(Object instance, String methodName, Class<?>... parameters) {
```
HannesWell, line 526:
```suggestion
    private Method findMethod(Object instance, String methodName, Class<?>... parameters) {
```
If you return the Method (or null) directly you can save the intermediate `orElseThrow()` in the caller and terminate the Optional chain earlier and can continue with plain references.

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/UnitTestSupport.java
```diff
@@ -0,0 +1,776 @@
+/********************************************************************************
+ * Copyright (c) 2024, 2024 ptreilhes and others
```
HannesWell, line 2: For the copy-right header you should use your full name (i.e. `surname lastname`).

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/UnitTestSupport.java
```diff
@@ -0,0 +1,776 @@
...
+    }
+
+    /**
+     * Execute the prerequisites mojos
+     * 
+     * @param configuration the configuration
+     * @param facade the maven project facade
+     * @param monitor the progress monitor
+     */
+    private void loadPrerequisites(ILaunchConfiguration configuration, IMavenProjectFacade facade,
```
HannesWell, line 438: Can't/shouldn't the list of mojos executed before the test-plugin be derived from the computed Maven lifecycle?

HannesWell (commented):
Hi @treilhes first of all a big sorry for the delay and thanks for your patience.
It's just currently a lot to do, but I still have you on my TODO list a will try my best that we can complete this before the next release (ideally with some time left).

I had a a very brief overview over this change and I find your solution to adjust the launch config very interesting. I assume this works reliably in various scenarios like creating via a short-cut or modifying an existing config? If all works as expected that's a smart solution.
I'm asking because having a mechanism for that is a longer standing issue:
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1183

Overall I didn't see any blockers and I'm glad you added extensive tests.

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/UnitTestSupport.java
treilhes: ok

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/UnitTestSupport.java
treilhes: ok

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/UnitTestSupport.java
treilhes: This one is a bit more complex as i think it can greatly impact performance. I'm afraid executing all plugins available in the maven lifecycle until this phase without knowing which ones are effectively producing properties is risky.
The only plugin provided by the maven team can safely be handled by default but about the others ....i'm not sure.
On the other hand providing the list of plugins to execute using a property feals more like a dirty hack than a clean solution.

If you have a way to know if a plugin generate some properties or a cleaner way to provide the list of plugin to execute it would be great as i plan to use the same kind of loading in another PR to be able to provide values for --patch-module directives from the compiler plugin to eclipse .classpath file in the fututre.

At this point i only can think of other ways like:
- providing a PR to maven surefire/failsafe/compiler plugin to provide the same kind of feature as the property but i'm skeptical the maven team will accept this PR more dedicated to IDE/Eclipse than maven itself
- providing some warning markers and some potential "quick fix" into the pom warning the user that plugins aren't executed until some configuration (property/project's preference) is set. (I'm completly clueless about how to do that but it would be cool but also time consuming)

What do you (or any other interested ) think about that?

treilhes: Hello @HannesWell,
After some research it seems the best way to handle this may be the lifecycle-metadata-mapping.
Currently a plugin may have the following configuration in the mapping:

ignore
execute(runOnConfiguration/runOnIncremental)
generator

So i propose to add a boolean "runOnLaunchConfiguration" or something more generic to fit the compiler case like "provideProperties" to the execute element to mark all plugins that need to be executed before setting up a launch configuration. 

About the "generator" case, a new method can be added to org.eclipse.m2e.core.project.configurator.AbstractProjectConfigurator :

`public void configureLaunchConfiguration(LaunchConfigurationRequest request, IProgressMonitor monitor) throws CoreException`

or something like

`public void loadPropertiesInContext(??, IProgressMonitor monitor) throws CoreException`

What do you think about that ?

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/UnitTestSupport.java
```diff
@@ -0,0 +1,776 @@
...
+
+  /**
+   * org.apache.maven.plugin.surefire.AbstractSurefireMojo method name to get the included and excluded tests
+   */
+  private static final String GET_INCLUDED_AND_EXCLUDED_TESTS_METHOD = "getIncludedAndExcludedTests";
+
+  /**
+   * Launch configuration attribute for the main class
+   */
+  public static final String LAUNCH_CONFIG_MAIN_CLASS = "org.eclipse.jdt.launching.MAIN_TYPE";
```
HannesWell, line 86: Please restrict the visibility as far as possible. And some constants are even unused.

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/UnitTestSupport.java
```diff
@@ -0,0 +1,776 @@
...
+      if(args.getEnvironmentVariables() != null) {
+        copy.setAttribute(LAUNCH_CONFIG_ENVIRONMENT_VARIABLES, args.getEnvironmentVariables());
+      }
+
+      copy.doSave();
+    }
+
+    private static IMavenProjectFacade getMavenProjectFacade(IJavaProject project) {
+      return MavenPlugin.getMavenProjectRegistry().getProject(project.getProject());
+    }
```
HannesWell, line 395: This can be inlined too.

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/UnitTestSupport.java
HannesWell: I think leveraging the lifecycle-mapping-metadata is the the better approach.
It would be ideal if we could re-use the existing metadata, but I assume it doesn't fit well in all cases, does it?
If new arguments are added, it should as general as possible and as specific as necessary.

But from what you have described enhancing the https://github.com/codehaus-plexus/plexus-build-api would maybe be another option?

Maybe @laeubi can help here too?

HannesWell: In order to simplify the review of this big change, I think it would be good to move the resolution of variables created by other plugins into a separate PR.
Would that be ok?

HannesWell (commented):
I made another pass over this PR and have a couple of suggestions, mainly about reducing code size.

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/UnitTestSupport.java
treilhes: ok let's continue this discussion in another PR ;)

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/UnitTestSupport.java
treilhes: restricted but still kept public some constants used in test

org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/UnitTestSupport.java
treilhes: inlined

HannesWell (approved):
@treilhes thank you for this contribution, that's really a nice extension to m2e and good work.

I'll leave the last question open for now because I would like to have this in the RC1 for the upcoming m2e release that I'm currently preparing.
We can still adjust this later.

treilhes:
Hello @HannesWell,
Replacing the MAVEN_CLASSPATH_PROVIDER existence check by a maven nature check is totaly valid.
If i remember well the classpath check come from the ModuleSupport class and it was a lazy copy/paste.
It seems this check was also replaced by a maven nature check as i can't find it anymore.

HannesWell:
Thank you for the clarification.
Do you want to provide a follow-up to remove it?

treilhes:
Hello @HannesWell,
You can find the fix in #1818

PR #1673, opened 2024-02-13
Support project local maven settings

laeubi:
Currently m2e has one global way to specify the global/user settings file, but one can also configure this through .mvn/maven.config files and while m2e already supports specification of properties and profiles there this does not work for (global)settings. This leads to the situation that user probably getting different results when running a mojo as part of eclipse or on the commandline.

This now adds the infrastructure and support for projects having a local configuration as well by augmenting the maven.config if available.

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
@@ -17,6 +17,7 @@ package org.eclipse.m2e.core.internal.embedder;
 import static org.eclipse.m2e.core.internal.M2EUtils.copyProperties;
 
 import java.io.File;
+import java.io.IOException;
 import java.util.ArrayDeque;
 import java.util.Collections;
 import java.util.Deque;
@@ -25,6 +26,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.function.Function;
@@ -73,6 +75,7 @@ import org.eclipse.m2e.core.embedder.ICallable;
 import org.eclipse.m2e.core.embedder.IComponentLookup;
 import org.eclipse.m2e.core.embedder.IMavenConfiguration;
 import org.eclipse.m2e.core.embedder.IMavenExecutionContext;
+import org.eclipse.m2e.core.embedder.MavenSettingsLocations;
 import org.eclipse.m2e.core.internal.MavenPluginActivator;
 import org.eclipse.m2e.core.internal.Messages;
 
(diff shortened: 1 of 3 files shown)


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
```diff
@@ -1103,4 +1065,82 @@ public MavenExecutionContext createExecutionContext() {
...
+          request.setUserSettingsFile(user);
+        }
+        try {
+          settings = lookup.lookup(SettingsBuilder.class).build(request).getEffectiveSettings();
+        } catch(SettingsBuildingException ex) {
+          String msg = "Could not read settings.xml, assuming default values";
+          log.error(msg, ex);
+          /*
+           * NOTE: This method provides input for various other core functions, just bailing out would make m2e highly
+           * unusuable. Instead, we fail gracefully and just ignore the broken settings, using defaults.
```
tivervac, line 1133:
```suggestion
           * unusable. Instead, we fail gracefully and just ignore the broken settings, using defaults.
```

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
```diff
@@ -1103,4 +1065,82 @@ public MavenExecutionContext createExecutionContext() {
...
+           */
+          settings = new Settings();
+        }
+        globalSettingsLength = gs;
```
tivervac, line 1137: Should we update even if we catch an exception? If we saw an exception, we have to reload anyway, don't we?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenProperties.java
```diff
@@ -73,14 +75,63 @@ public class MavenProperties {
...
+        for(String arg : new String(Files.readAllBytes(configFile.toPath())).split("\\s+")) {
+          if(!arg.isEmpty()) {
+            args.add(arg);
+          }
+        }
+        CLIManager manager = new CLIManager();
+        try {
+          commandline = manager.parse(args.toArray(String[]::new));
+        } catch(ParseException ex) {
+          throw new IOException("can't parse file " + configFile, ex);
```
tivervac, line 120:
```suggestion
          throw new IOException("Couldn't parse file " + configFile, ex);
```

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/ISettingsFiles.java
```diff
@@ -0,0 +1,39 @@
...
+
+
+/**
+ * Holds the locations of the maven global and user settings, global settings if specified are considered first but user
+ * settings might override.
+ */
+public interface ISettingsFiles {
+  /**
+   * Locates the global settings file, this takes <code>.mvn/maven.config</code> into account and otherwise returns the
+   * workspace configured one or even the maven default if no such settings are available.
```
tivervac, line 26:
```suggestion
   * Locates the global settings file, this takes <code>.mvn/maven.config</code> into account and otherwise returns the
   * one configured in the workspace or even the maven default if no such settings are available.
```

org.eclipse.m2e.core/src/org/eclipse/m2e/core/project/ISettingsFiles.java
```diff
@@ -0,0 +1,39 @@
...
+   * 
+   * @return the global settings file or <code>null</code>
+   */
+  File getGlobalSettingsFile();
+
+  /**
+   * Locates the user settings file, this takes <code>.mvn/maven.config</code> into account and otherwise returns the
+   * global configured one or even the maven default if no such settings are available.
```
tivervac, line 34:
```suggestion
   * Locates the user settings file, this takes <code>.mvn/maven.config</code> into account and otherwise returns the
   * globally configured one or even the maven default if no such settings are available.
```

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenImpl.java
laeubi: Yes the idea is unless you change the file parsing will always fail the same way (e.g. invalid content)

laeubi:
So now after some preliminary work this is finally ready and works to discover a changed local repository in the user and/or global settings!

Another issue might be toolchain files, this seems only partly supported but is out of scope of this issue.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
```diff
@@ -149,32 +152,59 @@ protected MavenExecutionRequest newExecutionRequest() throws CoreException {
...
+      }
+      MavenSettingsLocations mavenSettingsLocations = mavenArgs
+          .map(mavenCfg -> mavenCfg.getSettingsLocations(workspaceConfiguration))
+          .orElseGet(workspaceConfiguration::getSettingsLocations);
+      if(request == null) {
+        //create a fresh one ....
+        request = createExecutionRequest(workspaceConfiguration, containerLookup, mavenSettingsLocations);
+        request.setBaseDirectory(basedir);
+      } else {
+        //update existing configuration, we might have copied from an outer context here but the multi module directory is different we need to update some things then...
```
tivervac, line 171:
```suggestion
        //update existing configuration, we might have copied from an outer context here, but if the multi-module directory is different, we need to update some things...
```

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java
```diff
@@ -149,32 +152,59 @@ protected MavenExecutionRequest newExecutionRequest() throws CoreException {
...
+      }
+      request.setMultiModuleProjectDirectory(multiModuleProjectDirectory);
     }
-    request.setMultiModuleProjectDirectory(multiModuleProjectDirectory);
-
     return request;
   }
 
+  private boolean pathEquals(File file1, File file2) {
+    if(file1 != null && file2 != null) {
```
tivervac, line 187: Could they both be null? That would make them equals

laeubi: Better safe than sorry, I adjusted the code accordingly.

laeubi:
This improvement was gently sponsored by  [Sigasi](https://www.sigasi.com/).

PR #1674, opened 2024-02-14
Test removal of possibly duplicate init of local repository

laeubi:


laeubi:
I verified here that that test is either flaky or failing already without my change:
- https://github.com/eclipse-m2e/m2e-core/pull/1676

PR #1675, opened 2024-02-14
Consider the settings files location in project configuration

laeubi:
Currently the settings files are not considered when grouping projects by there project configuration.

As the settings files can have an impact they should be considered when grouping projects.

This is part of 
- https://github.com/eclipse-m2e/m2e-core/pull/1673

after further investigation I'm not sure if it is required or useful at all and if not 
- https://github.com/eclipse-m2e/m2e-core/pull/1674

is actually enough, because if the multi module basedir is the same the settings needs also be the same, the same is true as well for profiles and user properties, so the main question would be if there are cases where the settings are defined differently other than `maven.config` or m2e preferences.

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenProperties.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenProperties.java
@@ -237,6 +237,20 @@ public class MavenProperties {
     }
   }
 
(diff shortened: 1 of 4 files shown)


HannesWell:
Does this contain parts not already submitted in other PRs or can this be closed?

laeubi:
> Does this contain parts not already submitted in other PRs or can this be closed?

I rebased and updated this now.

PR #1676, opened 2024-02-14
Test JPMS fails

laeubi:


laeubi:
I also checked this test in the local IDE and it works there correctly.

treilhes:
Hello @laeubi,

I was also looking into it to solve the failing test for my PR #1672 and i did reproduce the issue and found the following:

In JpmsConfigurationTest
When updating the pom content (line 103)

`pomFileWS.setContents(new ByteArrayInputStream(argsSet1Content.getBytes()), true, false, null);`

A IResourceChangeEvent is triggered but MavenUpdateConfigurationChangeListener never receive the event so the project JreContainer isn't updated when we want to get properties from jreAttributes.

Why ? Because MavenUpdateConfigurationChangeListener isn't registered as a listener into the NotificationManager

The MavenUpdateConfigurationChangeListener  is supposed to be  regsitered in M2EUIPluginActivator.start() but in this case the start method is never called.

I was able to solve the issue by removing some lines from org.eclipse.m2e.core.ui\META-INF\MANIFEST.MF

By removing the latest 6 lines from the manifest:

```
Service-Component: OSGI-INF/org.eclipse.m2e.core.ui.internal.archetype.ArchetypeGenerator.xml,
 OSGI-INF/org.eclipse.m2e.core.ui.internal.archetype.ArchetypePlugin.xml
Bundle-ActivationPolicy: lazy
Service-Component: OSGI-INF/org.eclipse.m2e.core.ui.internal.archetype.ArchetypeGenerator.xml,
 OSGI-INF/org.eclipse.m2e.core.ui.internal.archetype.ArchetypePlugin.xml
Bundle-ActivationPolicy: lazy


```

I don't have any knowledge about osgi but it seems those directives are duplicates but it is just a guess

laeubi:
@treilhes interesting observation, do you like to open a PR that removes the duplicates, this does not looks correct it seems these are cam in by c3fa2edafe0a7f6e08fdb04fb508f767c7ac619d as an accident (maybe merge /rebase problem)

treilhes:
#1679 created, waiting for the build judgement

treilhes:
#1679 Seems good

PR #1677, opened 2024-02-14
Cache maven.config and support reading the (global) settings path

laeubi:
Currently the maven.config is read and parsed on each access, this now adds a caching stage to the MavenProperties that checks if the file actually has changed and if not just use the local copy.

This is split from:
- https://github.com/eclipse-m2e/m2e-core/pull/1673

to make the PR smaller and focus on the actual changes.


laeubi:
Test failure documented here:
- https://github.com/eclipse-m2e/m2e-core/pull/1676

PR #1680, opened 2024-02-17
Use preferences in property tester

laeubi:
The Maven property tester can hang the IDE as it query the project registry what might be a lengthy task.

To determine if a project has workspace resolution we can simply go directly to the preferences.

This is similar to what we have done previously for automatic project updates.


HannesWell (commented):
I'm not 100% certain that the  `IProjectConfiguration.isResolveWorkspaceProjects()` was the value from the preferences in all cases. In general the returned value can also be set, but as far as I can oversee it at the moment it was at least the case for all `IProjectConfiguration` in the project registry and other configs which had their value set where only used for other tasks.
So I think this is fine.

PR #1681, opened 2024-02-17
Resolve the local repository path relative to the multi module directory

laeubi:
If a user specify a relative path in the settings.xml this currently leads a bit to unspecified behavior. Java would then resolve it to the current working directory what might give desired results on the commandline (but sometimes also unexpected ones) but in the IDE it is not really controllable.

This now resolves the local repository path against the multi module directory what is the only user controllable constant directory we can offer here at the moment.


HannesWell (commented):
Sounds reasonable 👍🏽

PR #1682, opened 2024-02-17
Only consider existing artifacts for the classpath

laeubi:
Currently only if the artifact file is null it is not considered but if the file does not exits JDT complains. It even seems there was once in a time such a check but it is disabled, this is an attempt to check if we can reenable it again.


HannesWell (commented):
Looks like there are test-failures now. Don't know if they are only due to changed circumstances or are really meaning full.
It would be good to add a test for the case you described.

Sadly the commented out line goes back until the `Initial commit at Eclipse`. Would have been nice to know why this was disabled. :/

HannesWell:
> sadly some tests are written wit some numbers of problems in the expectation without looking at the content.

Yes, that's what I meant with non meaningful tests :)
One should check each failing tests if its just the number of expected CP-entries that has to be adjusted or if the failure is really revealing a potential problem.

> At least it show it has an effect .. but probably not something for the current release.

Yes, lets keep this for the next cycle.

PR #1685, opened 2024-02-18
chore: .gitignore add .vscode .settings for codespace

qxo:


.gitignore
```diff
@@ -4,3 +4,6 @@ bin/
 .polyglot*
 .tycho-consumer-pom.xml
 pom.tycho
+.settings
```
HannesWell, line 7: We should continue to track .settings files. Ignoring `.vscode` is fine.
```suggestion
```

akurtakov:
After a year without interest in applying requests from review one can assume this one will not be finished.

PR #1687, opened 2024-02-18
Maven global settings and toolchains UI cleanup

G-Ork:
I redesigned the Maven user settings and add the sections _Global_ & _User._  For booth there are now _settings_ & _toolchains_ options. 

- Added default path hints for global settings and global toolchains. They are only rendered for external maven installations. Don't forget to apply changes to see them.
- Unify settings override through goals field. All four settings could be put into the goal in short form (-gt. -gs, -t, -s).
- Throw exception during launch if path to a settings file is invalid.
- Provide unit-tests for maven launcher for all four settings
- Fix a bug in settings validation. Parser problems are not copied to result if parser do not crash. All well formed xml-files are treated as valid.
- Add XSDs for settings and toolchains as eclipse catalog contributions. 
  -  Could not tests this as i am not able to start M2E-IDE this the _WST XML_ feature.


laeubi:
> Add XSDs for settings and toolchains as eclipse catalog contributions.
> 
>     Could not tests this as i am not able to start M2E-IDE this the WST XML feature.

@mickaelistria @fbricon can you probably help here?

laeubi:
@G-Ork can you share a screenshot before/after?

G-Ork:
@laeubi Holy moly fast. Explain more what kind of screenshot you like.

laeubi:
> I redesigned the Maven user settings and add the sections Global & User. For booth there are now settings & toolchains options.

So what ever you redesigned in the UI, e.g dialog before your change, and how it looks now :-)

org.eclipse.m2e.core.tests/src/org/eclipse/m2e/core/IBiConsumer.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ *
+ * @param <T> the type of the first argument to the operation
+ * @param <U> the type of the second argument to the operation
+ * 
+ * @see Consumer
+ * @author Georg Tsakumagos
+ */
+@FunctionalInterface
+public interface IBiConsumer<T, U> {
```
laeubi, line 34:
```suggestion
public interface CoreBiConsumer<T, U> {
```

laeubi:
@G-Ork you also need to consider two other places, one is

`org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.createExecutionRequest(IMavenConfiguration, IComponentLookup, MavenSettingsLocations, File)`

where we need to set the Global/User Toolchain file initally, then in `org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.newExecutionRequest()` you probabbly need to update them if the are configured in the `MavenProperties` differently

G-Ork:
Okay i was not exact. I am able to start the M2E-IDE but the WST XML feature are not loaded. Here is the screenshot based on this PR.

**Current (released)**
Maven-Usersettings-Released

**Draft from PR**
Maven-Usersettings-Draft

org.eclipse.m2e.core.tests/src/org/eclipse/m2e/core/IBiConsumer.java
G-Ork: Okay sounds better. Same applies to the supplier.

G-Ork:
> @G-Ork you also need to consider two other places, one is
> 
> where we need to set the Global/User Toolchain file initally, then in `org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.newExecutionRequest()` you probabbly need to update them if the are configured in the `MavenProperties` differently

@laeubi From what i can see it your feature is not complete. The [.mvn/maven.config ](https://maven.apache.org/configure.html#mvn-maven-config-file) can contain much more than just [alternative settings](https://github.com/eclipse-m2e/m2e-core/blob/93f9495666d3d2f32433093d9b46aec845c1eaee/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/MavenExecutionContext.java#L164) location. 

Any chance we can refactor this to be more easy and support more options like .e.g. "-T3 -U "?
As far i can see it should work because you only care for the settings file. This part uses the configuration pointing to the settings files.

laeubi:
> From what i can see it your feature is not complete.

That's why I said you probably want to enhance it to support toolchains

> Any chance we can refactor this to be more easy and support more options like .e.g. "-T3 -U "?

There is no easy way as not all options can be supported and probably need to be supported differently. For example `-T` does not makes any sense as m2e only ever uses one build thread internally so claiming to use more might even be confusing. `-U` looks easy enough at a first glance, but this is usually controlled by the UI so when should we set it?

So I think one should really focus on things one by one when there is a demand.

HannesWell:
@G-Ork thanks for this contribution.
Since you listed many points in your intial comment I wonder if it would be possible (without too much effort) to split this into multiple PRs? This would also make it easier to review.
Furthermore for contributions of non-Committers over 1000LOC we would actually need an separate IP-clearance (usually not denied, but extra work).

I guess that the UI refactoring and the core changes can probably be splitted, similar to the schema additions.

G-Ork:
@HannesWell The high LOC includes the xsd files which are IP from maven. I think we could split XSD changes from this one into a separate PR. The changes for UI are so small i would not exclude them from the unit-tests and m2e layers.

HannesWell:
> @HannesWell The high LOC includes the xsd files which are IP from maven. I think we could split XSD changes from this one into a separate PR. The changes for UI are so small i would not exclude them from the unit-tests and m2e layers.

Sounds good. I have to admit that I first have to recall for what the XSD files are actually used.

G-Ork:
Split PR into two separate and close this:

- #1691 
- #1690

G-Ork:
Split PR into two separate and close this:

- #1691 
- #1690

PR #1690, opened 2024-02-18
Add XML schemas for Settings and Toolchains to Eclipse XML catalog.

G-Ork:
Add the open available Schemas from Apache Maven. THis should help with content assist if namespace in settings file is declared.

Source for the Schemas: https://maven.apache.org/xsd/


laeubi:
@mickaelistria can you review? DO we need a CQ for this?
What I'm wondering is if m2e is the right place for this maybe this better is placed at WWD or lemminx(-maven)?

mickaelistria:
> DO we need a CQ for this?

Yes we do need a CQ as we want to include code with a non-EPL license (and including in code-base is a bigger problem than just referencing 3rd party, or delivering 3rd party in binary)

> What I'm wondering is if m2e is the right place for this maybe this better is placed at WWD or lemminx(-maven)?

That's a very good question. and the answer I support is https://github.com/eclipse/lemminx-maven/issues/201 , but it's unfortunately not so immediately achievable.
But having the schemas in m2e can be useful as a workaround because the schemas defined in extensions will still be injected to lemminx by Wild Web Developer, so adding schemas here should make lemminx use them.

mickaelistria:
> Hmm neither WWD nor Lemminx are the right places.

LemMinX-Maven is the right place, it's an extension of lemminx dedicated to Maven files development.

> I don't get the license problems

Please read the Eclipse Development Process to know more about it.

> If ASF 2.0 is not compatible you could not bundle maven core.

Committing sources in the repository or including them in a distribution ask for different level of compatibility/tracking in the EDP.
I'm almost certain this is legal, but the proper way to do it is that we track the request via some CQ and that we get Eclipse Foundation legal people approving that. We actually had to do it as well for Maven in the past, without even committing sources in.

HannesWell:
Thank you @G-Ork for this.
I tried this out but I have to admit that I cannot see what exactly is enhanced by adding the schemata. I compared a Eclipse instance launched from my m2e workspace that has them available with my dev-eclipse (that doesn't have them yet) and when opening for example my local `toolchains.xml` I get a content-assist that is working sometimes well and sometimes not and the documentation displayed when hovering over an element. But that's the same in both cases.
The only difference I see is that in the head-line it says `xsi:schemaLocation` in my dev-eclipse and `xsi:schemaLocation` with catalog in the eclipse launched with this PR. Since these settings.xml and toolchains.xml usually have their schema defined I wonder if missing schemata are simply downloaded from the specified schema-location?
Or could you please show where this makes a difference?

> I don't get the license problems. If ASF 2.0 is not compatible you could not bundle maven core.
> If so the XSDs are the least problem.

I it's probably not a problem, but as Mickael said, we still need a clearance for that. It's like visiting another (license-) country when you have to apply for a visa. It is usually granted, but you have to provide all the details and somebody has to approve it.

laeubi:
I created https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/13350 now, as we have done this already for other schemas I think there is a purpose (maybe offline usage).

G-Ork:
@laeubi So the iplab issue is primary for the bookkeeping and the unlikely case of a veto?

@HannesWell Adding the schemas to the catalog enable the (eclipse) XML validator(s) to locate the schemas and validate them. No remote access needed. Its more than just code assist for the editor. The most projects do not include Maven toolchains or setting files. But with  #1673 it is more likely.  In m2e we have them to. Actually without namespace. 

Defining schema locations to remote resources is bad practice. Imagine how much http requests hit common schema locations. The latency piles up noticeably on large workspaces with many different schemas. Having a catalog is needed if using a secured validator as those permits download from unknown sources. XML parsers configs should also be customized to prevent billion laughs attack aka. XML Bombs and other funny things. Commercial editors like Oxygen XML Editor offer this security by default. Eclipse XML Validator has also some security options. But i am not sure what the option "Honour all XML schema locations" is about.

laeubi:
> @laeubi So the iplab issue is primary for the bookkeeping and the unlikely case of a veto?

Yes it needs a manual review so we do not include any "forbidden" ones, but as we already include some of the xsds here its more a formalism than something I expect would require deeper analysis / investigations.

HannesWell (commented):
> Defining schema locations to remote resources is bad practice. Imagine how much http requests hit common schema locations. The latency piles up noticeably on large workspaces with many different schemas. Having a catalog is needed if using a secured validator as those permits download from unknown sources.

OK makes sense to me. 👍🏽 
Then lets wait until the CQ is approved.

In the meantime can you please adjust the message of your commit to something like the PR title: `Add XML schemas for Settings and Toolchains to Eclipse XML catalog`
You can just amend your commit and force-push the amended commit to this PR's branch.

Another thing that could be tought about (but it could also be done in a separate change) is to unify the file names of the existing schema that currently don't match the name in https://maven.apache.org/xsd/.

G-Ork:
@HannesWell: I'll do, the next days.

G-Ork:
I changed the commit. Just a savety question. Some of the xsd where renamed and slightly changed. e.g. [maven-4.0.0.xsd](https://maven.apache.org/xsd/maven-4.0.0.xsd). Looks like it is not the first time. 
Do you like to update the name and the content?

Other renamed files:

- profiles-1.0.0.xsd
- settings-1.0.0.xsd

G-Ork:
@HannesWell Feedback requested...

org.eclipse.m2e.model.edit/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-Name: M2E Maven Project Model Edit
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Bundle-Vendor: Eclipse.org - m2e
-Bundle-Version: 2.0.400.qualifier
+Bundle-Version: 2.0.401.qualifier
```
HannesWell, line 5:
```suggestion
Bundle-Version: 2.0.500.qualifier
```

HannesWell (commented):
> I changed the commit. Just a safety question. Some of the XSD where renamed and slightly changed. e.g. [maven-4.0.0.xsd](https://maven.apache.org/xsd/maven-4.0.0.xsd). Looks like it is not the first time. Do you like to update the name and the content?
> 
> Other renamed files:
> 
>     * profiles-1.0.0.xsd
> 
>     * settings-1.0.0.xsd

Yes please. I think it would be good if all the schemata  embedded in m2e.model.edit match the names in https://maven.apache.org/xsd/.

org.eclipse.m2e.model.edit/META-INF/MANIFEST.MF
G-Ork: Thanks for the hint. Changed.

G-Ork:
I've changed the naming and removed the old maven 4 schema with underscore. If its not hosted by maven we should not need it.

How generic was the iplab response? 
There are some schemas that could be helpful like the assembly schemas for people who write maven assembly descriptors like me :)

I'd like to support them also. 
In a separate PR for sure ...

HannesWell:
> I've changed the naming and removed the old maven 4 schema with underscore. If its not hosted by maven we should not need it.

Thanks! That looks good.
Just one tiny thing, the org.eclipse.m2e.feature feature needs a +100 bump of its version's service segment. Then the build should be happy.

Can you also please squash all your commits into one commit with a meaningful message covering this change as a whole and rebase plus force-push it to on top of the master branch?

> 
> How generic was the iplab response? There are some schemas that could be helpful like the assembly schemas for people who write maven assembly descriptors like me :)
> 

The one created looks relatively generic for the entire content, but I'm not 100% certain so I would just create another one for a new PR. It's not a big task. :)

> I created https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/13350 now,

HannesWell (approved):
I just re-based this PR and resolved the conflicts.
This is now ready. Thanks @G-Ork for this contribution.

G-Ork:
Thank you for the merge. Also without the stash.
Sorry i was busy.

PR #1691, opened 2024-02-18
Unified UI setting for toolchains

G-Ork:
I redesigned the Maven user settings and add the sections Global & User. For booth there are now settings & toolchains options.

- Added default path hints for global settings and global toolchains. They are only rendered for external maven installations. Don't forget to apply changes to see them.
- Unify settings override through goals field. All four settings could be put into the goal in short form (-gt. -gs, -t, -s).
- Throw exception during launch if path to a settings file is invalid.
- Provide unit-tests for maven launcher for all four settings
- Fix a bug in settings validation. Parser problems are not copied to result if parser do not crash. All well formed xml-files are treated as valid.

Maven-Usersettings-Draft


akurtakov:
Is there interest if finishing this one?

laeubi:
@G-Ork can you pleas squash all commits and rebase on main?

G-Ork:
I'll do my best to squeeze the squash it into this weekend.

PR #1692, opened 2024-02-18
Update to LemMinX-Maven 0.11.1

HannesWell:
To get the latest enhancements in LemMinX-Maven for m2e:
- https://github.com/eclipse/lemminx-maven/pull/552
-  https://github.com/eclipse/lemminx-maven/pull/553

### DIFF
--- a/org.eclipse.m2e.editor.lemminx.tests/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.editor.lemminx.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: M2E Maven POM File Editor using Wild Web Developer, Lemminx and Maven LS extension Tests
 Bundle-SymbolicName: org.eclipse.m2e.editor.lemminx.tests
-Bundle-Version: 2.0.1.qualifier
+Bundle-Version: 2.0.2.qualifier
 Automatic-Module-Name: org.eclipse.m2e.editor.lemminx.tests
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Bundle-Vendor: Eclipse.org - m2e
--- a/org.eclipse.m2e.editor.lemminx.tests/src/org/eclipse/m2e/editor/lemminx/tests/EditorTest.java
+++ b/org.eclipse.m2e.editor.lemminx.tests/src/org/eclipse/m2e/editor/lemminx/tests/EditorTest.java
@@ -12,6 +12,7 @@ package org.eclipse.m2e.editor.lemminx.tests;
 
 import static org.junit.Assert.assertTrue;
 
+import java.io.IOException;
 import java.io.InputStream;
 import java.util.Arrays;
 import java.util.Set;
@@ -21,7 +22,7 @@ import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
-import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.resources.IncrementalProjectBuilder;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.jface.text.TextSelection;
@@ -36,7 +37,6 @@ import org.eclipse.swt.widgets.TableItem;
 import org.eclipse.ui.IWorkbenchPage;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.ide.IDE;
-import org.eclipse.ui.part.FileEditorInput;
 import org.eclipse.ui.tests.harness.util.DisplayHelper;
 import org.eclipse.ui.texteditor.ITextEditor;
 import org.eclipse.ui.texteditor.ITextEditorActionConstants;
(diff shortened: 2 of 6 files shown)


HannesWell:
@mickaelistria, @laeubi any idea, why the `EditorTest` now fail?
I can reproduce it locally but I understand why there are no markers published (the changes don't affect that). If I import the test pom's in a IDE launched from my m2e workspace the expected error markers show up.

laeubi:
@HannesWell are the adjustment of the test class required here due to the update? If not I would revert those as a first step.

laeubi:
One test complains:

```
Feb 18, 2024 11:51:14 PM org.eclipse.tm4e.core.registry.Registry _doLoadSingleGrammar
WARNING: No grammar source for scope [source.java]
Feb 18, 2024 11:51:14 PM org.eclipse.tm4e.core.internal.rule.RuleFactory _compilePatterns
WARNING: CANNOT find grammar for scopeName [source.java]. I am [source.xml@org.eclipse.wildwebdeveloper.xml]
Feb 18, 2024 11:51:14 PM org.eclipse.tm4e.core.internal.rule.RuleFactory _compilePatterns
WARNING: REMOVING BeginEndRule{id=79,name=source.java.embedded.xml} ENTIRELY DUE TO EMPTY PATTERNS THAT ARE MISSING
```

so maybe the test is not happy about the warnings?

laeubi:
@HannesWell I have now tried to test this and with a local m2e I see no (language server) error markers generated when I open it in a "pure" maven project, but maybe it is intentional, if I open it with `Generic Editor`, if I insert some gibberish in the file I do get error markers from language server, so in general something happens...

mickaelistria:
>  I do get error markers from language server, 

Are those real Maven errors reported by the LS, or just some XML syntax stuff?

mickaelistria:
I also support that the test shouldn't be changed with the update of lemminx-maven if possible.
Note that Lemminx-Maven doesn't care about underlying IMavenProject and so on from m2e. So the structure of the project doesn't matter. Actually, it's even better to make the test independent of m2e IMavenProject or nature stuff, so when m2e drops usage of natures or becomes more flexible with project structure, the tests would still run.

laeubi:
> > I do get error markers from language server,
> 
> Are those real Maven errors reported by the LS, or just some XML syntax stuff?

It looks like XML validation errors, I think the best would be to bring this testcase to lemminx-maven to see if it fails there to @mickaelistria can you suggest a test for testing diagnostic reports, I'll try to create one then.

laeubi:
Okay I noticed that lemminx has actually a test for it `pom-without-artifactId.xml` / `org.eclipse.lemminx.extensions.maven.participants.SimpleModelTest.testMissingArtifactIdError()` that seem to work, so somewhere now is something missing in between.

mickaelistria:
If lemminx-maven was released, it can be assumed to work.
If no Maven info is reported to m2e, it's most likely that Maven-LemMinX is not properly integrated: a dependency may be missing, or conflicting... There are options to debug the LS and enable more logging, or to even pass the necessary flags to debug it (from another IDE instance).

laeubi:
@szarnekow can you probably have a look here?
It seems m2e uses latest and greatest lsp4j, lemminx itself seem to have some issues:
- https://github.com/eclipse/lemminx/pull/1627

if I manually raise the versions in lemminx-maven the test still runs (even though some other can't work due to changed API)

mickaelistria:
According to your comment, it seems to me that the issue is more that the version of lemminx embedded in Wild Web Developer is not compatible with this release of lemminx-maven.
However, I don't see any change betweem currnet and previous release of lemminx-maven that has affected the code in a way that makes it rely on mutable APIs that might now be broken.

mickaelistria:
I've checked and my former comment is incorrect: both Wild Web Developer and LemMinX-Maven do include/require the same 0.27 version of lemminx. So it doesn't look like an inconsistency in here.
And the version of LSP4J inside m2e shouldn't really matter because the LS is running as a separate process, and that the version of LSP4J is taken from the lemminx "uber-jar". It's a different execution domain and there is no strong reason that the LSP4J in the IDE leaks into the language sever.

laeubi:
The only changes are my ones and dependency updates, but its hard to guess if WWD is/could be affected...

mickaelistria:
Does one of your change in lemminx-maven require a new version of the resolver, newer than what m2e includes and injects in lemminx together with lemminx-maven?

laeubi:
My changes should be 100% backward compatibile, the only thing I could think of is maybe maven API version I'm using the `org.apache.maven.cli.CLIManager.SET_USER_PROPERTY` over the (deprecated) `org.apache.maven.cli.CLIManager.SET_SYSTEM_PROPERTY` ... but would expect that the compiler inline this anyways an lemminx-maven ships with what it compiles for?

laeubi:
> Does one of your change in lemminx-maven require a new version of the resolver, newer than what m2e includes and injects in lemminx together with lemminx-maven?

How does m2e inject maven to leminx?!?

laeubi:
It seems to be the constant, I have no clue why it compiles / run in lemminx-maven but fails at m2e.... if I recompile with the deprecated one everything works.

mickaelistria:
> How does m2e inject maven to leminx?!?

MavenRuntimeClasspathProvider resolves the list of jars that are added to LemMinX's classpath.

> It seems to be the constant, I have no clue why it compiles / run in lemminx-maven but fails at m2e....

Maybe LemMinX-Maven tests uses a different version of Maven that what is in m2e?

laeubi:
Then it might be that m2e uses an outdated maven-embedder jar I still don't understand why the compiler is not inline this compile time constant!

HannesWell:
> @HannesWell are the adjustment of the test class required here due to the update? If not I would revert those as a first step.

They are not, but as written above they don't influence the test-failure.

> > It seems to be the constant, I have no clue why it compiles / run in lemminx-maven but fails at m2e....
> 
> Maybe LemMinX-Maven tests uses a different version of Maven that what is in m2e?

Is https://github.com/eclipse/lemminx-maven/issues/526 related?

laeubi:
I think i found it here: 
- https://github.com/eclipse-m2e/m2e-core/pull/1693

This all is really a bit unfortunate as the forked process seems to use a flat classpath and therefore did not notice if anything is missing even though it is declared on the OSGi level... also it sounds strange that m2e gives jars to the language server if it already seem to be a standalone process why does it not using its own?

mickaelistria:
> also it sounds strange that m2e gives jars to the language server if it already seem to be a standalone process why does it not using its own?

IIRC, the goal was to save some ~30MB, but if it appears that it's overall better to have lemminx-maven come with its own Maven jars despite the extra resource cost, it would be OK to do it.

HannesWell:
> I think another option would be to run the server in a (isolated) OSGi framework, that way it can simply use already installed bundles... but I first need to check how WWD start up everything.

Can't it just run in the same OSGi runtime? IIRC @mickaelistria you presented a while ago that the JDT-LS now also runs in the same process which even allows sharing objects, in case of m2e project information and caches.

mickaelistria:
> I think another option would be to run the server in a (isolated) OSGi framework, that way it can simply use already installed bundles... but I first need to check how WWD start up everything.

The idea has strong technical merits, it was already discussed but the "human resource" aspect weigh more here:
Neither LemMinX nor LemMinx-Maven are tested nor supported in an OSGi environment at the moment. Those might work, but then it would mean what we have in Eclipse land becomes a much differnet distribution from what is tested upstream. As the delta with upstream becomes bigger, it means the maintenance effort would grow as well, which is the opposite of what we want with language servers overall.
I would rather see lemminx-maven duplicating Mavan jars that starting to deal with maintenance of lemminx and lemminx-maven in an OSGi environment.

laeubi:
> I would rather see lemminx-maven duplicating Mavan jars that starting to deal with maintenance of lemminx and lemminx-maven in an OSGi environment.

It seems there is already `lemminx-maven-0.11.2-SNAPSHOT-zip-with-dependencies.zip` what is currently 11 mb.

laeubi:
I have now created:
- https://github.com/eclipse/lemminx-maven/pull/556

this would allow m2e to look at the manifest and either use OSGi resolver or P2 resolver to get know what the lemminx-maven has as a dependency.

laeubi:
Finally everything green :1st_place_medal:

PR #1698, opened 2024-02-21
Add testcase for classifier with transitive dependencies

laeubi:
Currently if a dependency is specified  the transitive items seem to be not correctly collected, at laest there is one artifact (ehcache) that has a transitive dependency that is invalid (javax.xml) and one can select a different classifier (jakarta) to use the jakarta ones but this seem not to happen at the moment.

This adds a testcase to illustrate the problem.


laeubi:
Looking further on the problem it seems that classified artifacts don't have own metadata. So what happens here is that the classified artifact is actually used, but when one enables dependency inclusion it includes the dependencies of the main artifact (what is bad in this case!).

I therefore tend to make it that way, that even if requested for a classified artifact no dependencies are included at all. if one really wants this the user needs to add both, the classified and the main artifact, or can add the pom directly here.

laeubi:
The same behavior/testcase needs to be added to Tycho.

PR #1700, opened 2024-02-26
Trace test execution directories in the remote maven execution

laeubi:
Currently running tests in maven via m2e is possible but not very convenient as one still needs to navigate to the results then open the correct file. Another pitfall is that even if one has opened the file once, the "classic" JUnit view not update the view even when the file changes afterwards.

This now adds a new process tracking of test executions directories that then can be watched on the m2e side and display the new advanced JUnit view when the run has finished.


laeubi:
This is how it currently looks like and serves as a first POC

https://github.com/eclipse-m2e/m2e-core/assets/1331477/61d8ad18-ee3b-4f89-90b2-2384d3ac1049

to make it really cool / convenient a bit more work needs to be done:

1. The eclipse test results view has some concurrency issues and if m2e is "too fast" can have some UI glitches, one can see this in the video where full the results are only show after selecting "Expand all"
2. The View currently does not support a concept of "projects" so we fake them by adding each project as a test-suite, it would be good to enhance this by the concept of projects/groups so it is more clear what belongs where.
3. Reading files *after* the run works but is not really dynamic and might better watch the directory
4. Even better it would be to impose an API at [plexus-build-api](https://github.com/codehaus-plexus/plexus-build-api/) that allows m2e to intercept the process execution and install a test execution listener, then one would get real run time information

Overall it looks like a quite cool feature to explore, so if anyone is interested in sponsoring further development or working on some of the mentioned points please let me know.

PR #1706, opened 2024-03-03
Make the remove maven listener a process in the launch

laeubi:
Currently m2e uses an additional socket connection to receive project events from the remote maven project. This is not visible to the user and requires some internal maps to maintain state.

This refactors this into a more generic event facility and make it visible as a Process in the debug view, this also has the advantage that the process itself can be used to maintain the state.


akurtakov:
Would you please rebase this if you plan to finish it?

laeubi:
Its still on my todo list...

PR #1719, opened 2024-03-17
fix: the `Maven Project Builder' duplicate issue

qxo:
uncheck it then do `maven -> Update Project...` we'll see the duplicate


akurtakov:
Do you think you can a test case so we know that it will not regress again?

PR #1731, opened 2024-04-04
Only ever return pom path for type=pom

### LINKED ISSUE #1730: Never return path to `pom.xml` for artifact files except for type = "pom"
Currently the logic in https://github.com/eclipse-m2e/m2e-core/blob/5f945212542f8aee7ca9c20dd59a3dd2dfee3d5a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/EclipseWorkspaceArtifactRepository.java#L84-L102 may return the pom.xml path for the file. This is used in https://github.com/eclipse-m2e/m2e-core/blob/5f945212542f8aee7ca9c20dd59a3dd2dfee3d5a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/EclipseWorkspaceArtifactRepository.java#L145 (an implementation of the Maven SPI [org.apache.maven.artifact.repository.ArtifactRepository](https://maven.apache.org/ref/3.9.3/maven-artifact/apidocs/org/apache/maven/artifact/repository/ArtifactRepository.html#find(org.apache.maven.artifact.Artifact)))

Although returning the path to the `pom.xml` makes sense for type being `pom` it doesn't for any other type. In that case one should rather fall back to `null`.

kwin:
This closes #1730


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/registry/EclipseWorkspaceArtifactRepository.java
```diff
@@ -77,8 +77,8 @@ protected File resolveAsEclipseProject(String groupId, String artifactId, String
 
     if(context.resolverConfiguration.isResolveWorkspaceProjects()) {
       IWorkspaceRoot root = ResourcesPlugin.getWorkspace().getRoot();
-      IPath file = pom.getLocation();
-      if(file == null) {
+      IPath pomFile = pom.getLocation();
+      if(pomFile == null) {
         return ProjectRegistryManager.toJavaIoFile(pom);
       }
       if(!POM_EXTENSION.equals(extension)) {
```
kwin, line 84: this needs further cleanup as well, as actually POM_EXTENSION is the type (not the extension), but the type is only passed from the legacy https://github.com/eclipse-m2e/m2e-core/pull/1731/files#diff-22a899eac2926341327db3736d6ed24c7f9ce9a64a154f805af15b424d6fed75L146 but not from the modern resolver SPI https://github.com/eclipse-m2e/m2e-core/pull/1731/files#diff-22a899eac2926341327db3736d6ed24c7f9ce9a64a154f805af15b424d6fed75R141

kwin: Seems that `pom` type is having extension `pom` (https://maven.apache.org/ref/3.9.6/maven-core/artifact-handlers.html) despite its file name `pom.xml`.

mickaelistria: If you look at how the pom files are published in Maven repos (eg https://repo.eclipse.org/content/repositories/tycho/org/eclipse/tycho/tycho-core/4.0.8-SNAPSHOT/ ), then you can see that they are indeed published as `artifactId-version.pom` . The `xml` is the "source" extension, but the derived artifact is a `.pom` file.

kwin:
@laeubi Any idea about the failing test cases? They seem related to the change, but I don't understand why.

kwin:
> so you need to return something existing for it to work

IMHO returning `null` here forces Maven to use the regular local Maven repository (instead of using Eclipse source artifacts). I think it is just about test cases setting wrong expectations...

laeubi:
> > so you need to return something existing for it to work
> 
> IMHO returning `null` here forces Maven to use the regular local Maven repository (instead of using Eclipse source artifacts). I think it is just about test cases setting wrong expectations...

If Workspace Artifacts are enabled then they must be resolved if possible, return null would mean that the project do not exits in workspace what is wrong.

akurtakov:
Is there plan to finish this one?

laeubi:
> Is there plan to finish this one?

As mentioned I think the approach is wrong to return `null`, as it will then mean the artifact is "unresolved".

I also miss a testcase that shows why we change it and that it actually works as expected.

PR #1688, opened 2024-02-18
Build with Tycho 4.0.6

HannesWell:

PR #1733, opened 2024-04-09
Update tycho to 4.0.8

akurtakov:


HannesWell:
All these test-failures happen also in the attempt to update to Tycho 4.0.6: https://github.com/eclipse-m2e/m2e-core/pull/1688

I haven't investigated it yet, maybe @laeubi can tell?

laeubi:
From the buildfailure it is a problem due to classfile changes (ECJ update?)

HannesWell:
> From the buildfailure it is a problem due to classfile changes (ECJ update?)

The update to Tycho 4.0.6 already failed in #1688 and IIRC the comparator errors were not the problem and the fixed ECJ version was removed in an attempt to resolve that failure, but it seems like it didn't fix it.
But lets fix the comparator errors first.

HannesWell:
> From the buildfailure it is a problem due to classfile changes (ECJ update?)

The comparator errors are resolved and now we (again) have the 500 test-failures.

laeubi:
As they all fail the same way maybe some change in P2 causing this, I would try to add a direct dependency in the test plugin to make sure its not caused by some indirect dependency missing now.

HannesWell:
> As they all fail the same way maybe some change in P2 causing this, I would try to add a direct dependency in the test plugin to make sure its not caused by some indirect dependency missing now.

The plugin `org.eclipse.m2e.tests` already requires the bundle `org.eclipse.m2e.maven.runtime`.

HannesWell:
> > > As they all fail the same way maybe some change in P2 causing this, I would try to add a direct dependency in the test plugin to make sure its not caused by some indirect dependency missing now.
> > 
> > 
> > The plugin `org.eclipse.m2e.tests` already requires the bundle `org.eclipse.m2e.maven.runtime`.
> 
> The error says that the bundle `org.eclipse.m2e.tests.common` can not find the class...

Right (I just read the first line and checked the test class). But even `org.eclipse.m2e.tests.common` already requires `org.eclipse.m2e.maven.runtime`.

laeubi:
Interestingly I get these errors locally as well with Tycho 4.0.4

laeubi:
As I can "reproduce" it [here](https://github.com/eclipse-m2e/m2e-core/pull/1770) with Tycho 4.0.4 I suspect it must be some kind of classpath ordering problem, e.g that for consistency we order the dependencies differently (either Tycho or P2)...

HannesWell:
With Tycho 4.0.9 we could finally update: https://github.com/eclipse-m2e/m2e-core/pull/1837

PR #1720, opened 2024-03-17
[TP] update to Eclipse 4.31 and latest versions of other dependencies

HannesWell:

PR #1734, opened 2024-04-09
Properly create IJavaProject in JpmsConfigurationTest

akurtakov:
The pattern used to cast java nature to java project is wrong and we have seen it used in many places but broke for some reason with 2024-03. This reduces the failing tests of
https://github.com/eclipse-m2e/m2e-core/pull/1720 .


akurtakov:
Unrelated test failure.

PR #1736, opened 2024-04-10
AbstractJavaProjectConfigurator: Fix NPE if EE has no compliance options

### LINKED ISSUE #1735: NPE in AbstractJavaProjectConfigurator
Steps to reproduce:

- install Eclipse 4.32M1 and m2e - https://download.eclipse.org/eclipse/downloads/drops4/S-4.32M1-202404061800/
- try to import https://github.com/eclipse-m2e/m2e-core/tree/master/org.eclipse.m2e.jdt.tests/projects/compilerSettings

m2e throws NPE at AbstractJavaProjectConfigurator.java, line 128, and the project isn't configured correctly.
The issue is caused by https://github.com/eclipse-equinox/equinox/issues/571

sratz:
Fixes #1735.

### DIFF
--- a/org.eclipse.m2e.jdt/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.jdt/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.m2e.jdt;singleton:=true
-Bundle-Version: 2.3.401.qualifier
+Bundle-Version: 2.3.500.qualifier
 Bundle-Localization: plugin
 Export-Package: org.eclipse.m2e.jdt,
  org.eclipse.m2e.jdt.internal;x-friends:="org.eclipse.m2e.jdt.ui",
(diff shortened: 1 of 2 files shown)


HannesWell (commented):
Thanks Sebastian for this. 
In general it looks good, but I would like to take this as an opportunity to replace var by the actual type in the changed code and to remove the `JRE` element from the `supportedEETypes` set.

andi-huber:
Thanks Sebastian, patch works for me  (Eclipse 4.32M1). You saved my day.

sratz:
> Thanks Sebastian for this. In general it looks good, but I would like to take this as an opportunity to replace var by the actual type in the changed code

Done.

> Furthermore the service version of `org.eclipse.m2e.jdt` needs a (+100) bump and probably the feature as well.

I bumped the jdt plugin and features where it is included by +100. Can you double-check?

> and to remove the `JRE` element from the `supportedEETypes` set.

Hmm, shouldn't this be done in a different change/PR to keep the bugfix minimal?

sratz:
> Thanks for the update! I fixed the accidental removal of imports which lead to compilation errors.

Thanks! Weird, I have no idea how that removed import slipped in :D

> > > and to remove the `JRE` element from the `supportedEETypes` set.
> > 
> > 
> > Hmm, shouldn't this be done in a different change/PR to keep the bugfix minimal?
> 
> That's also fine, see #1739 :)

👍

HannesWell:
> > Thanks for the update! I fixed the accidental removal of imports which lead to compilation errors.
> 
> Thanks! Weird, I have no idea how that removed import slipped in :D

That can happen. Maybe your local m2e.maven.runtime had errors and then JDT decided to remove the imports.

Anyway now it's fixed. Thanks again.

PR #1738, opened 2024-04-11
Unstable ConsoleTest

akurtakov:
Give one more minute for the maven build to complete


akurtakov:
Doesn't look like this helps. Closing.

PR #1758, opened 2024-05-17
Avoid triggering IResourceDelta.ENCODING change if encoding is already set to the good value

RoiSoleil:
Avoid triggering IResourceDelta.ENCODING change if encoding is already set to the good value

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/ProjectConfigurationManager.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/ProjectConfigurationManager.java
@@ -26,6 +26,7 @@ import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
 import java.util.regex.Matcher;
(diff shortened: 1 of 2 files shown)


org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/ProjectConfigurationManager.java
```diff
@@ -488,8 +488,10 @@ private void updateProjectConfiguration(ProjectConfigurationRequest request, IPr
     MavenProject mavenProject = request.mavenProject();
     Properties mavenProperties = mavenProject.getProperties();
     String sourceEncoding = mavenProperties.getProperty("project.build.sourceEncoding");
-    log.debug("Setting encoding for project {}: {}", project.getName(), sourceEncoding); //$NON-NLS-1$
-    project.setDefaultCharset(sourceEncoding, monitor);
+    if(project.getDefaultCharset() == null || !project.getDefaultCharset().equals(sourceEncoding)) {
```
HannesWell, line 491:
```suggestion
    if(Objects.equals(project.getDefaultCharset(), sourceEncoding)) {
```
That should also cover the case that the project's encoding is already null and no encoding is specified and would resetting it to null again. Furthermore the project's charset is then only fetched once.

If you think that's fine it should be applied to all other changes as well.

HannesWell (changes requested):
In general this looks sane, but I think the implementation can be slightly enhanced.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/project/ProjectConfigurationManager.java
RoiSoleil: Done ;)

HannesWell (approved):
Looks good, thank you. :)

thahnen:
Thanks a lot, @RoiSoleil and @HannesWell, for fixing this issue! Much appreciated :bow:

PR #1762, opened 2024-05-21
Fix build time warning about build-helper version

akurtakov:


akurtakov:
Same tests fail on master.

PR #1763, opened 2024-05-25
[Runtime] Embed Maven 3.9.7

HannesWell:


org.eclipse.m2e.maven.runtime/pom.xml
```diff
@@ -20,19 +20,26 @@
...
 
 	<name>M2E Embedded Maven Runtime (includes Incubating components)</name>
 
+	<repositories>
+		<repository>
+			<id>maven-staging</id>
+			<url>https://repository.apache.org/content/repositories/maven-2125/</url>
+		</repository>
+	</repositories>
+
```
HannesWell, line 34: This has to be reverted once Maven 3.9.7 is officially released, which should happen maybe today already.

org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/ReadonlyMavenExecutionRequest.java
```diff
@@ -522,6 +522,16 @@ public MavenExecutionRequest setUseLegacyLocalRepository(boolean useLegacyReposi
     return request.setUseLegacyLocalRepository(useLegacyRepository);
   }
 
+  // @Override // Support compilation against Maven 3.8
```
HannesWell, line 525: @laeubi in the CI all M2E bundles build against the maven 3.8.6 runtime version and therefore compilation fails if these new methods are annotated with `@Override`. I assume that's intended and you probably don't want to raise the lower bound to Maven 3.9, don't you?

HannesWell: OK this will of course cause problems in the implementation since the new methods are also called.

So we either have to raise the lower bound or use reflection here.

org.eclipse.m2e.maven.runtime/pom.xml
HannesWell: Done.

HannesWell:
Maven 3.9.7 was just released.
@laeubi I'm just waiting for your answer regarding 3.8 compatibility.

laeubi:
> @laeubi in the CI all M2E bundles build against the maven 3.8.6 runtime version and therefore compilation fails if these new methods are annotated with @Override. I assume that's intended and you probably don't want to raise the lower bound to Maven 3.9, don't you?

I think we should keep 3.8.x compatibility, I'm not sure why not marking it with `Override` should cause any issue, as I assume only the new 3.9.x maven will call the method (if used) or does m2e directly call this method?

laeubi:
I also wonder if it would not be better to implement R/O MavenExecutionRequest  as a [proxy class](https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Proxy.html) instead that simply delegates to the original for all method except those that starts with "set" in the handler?

HannesWell:
> > @laeubi in the CI all M2E bundles build against the maven 3.8.6 runtime version and therefore compilation fails if these new methods are annotated with @OverRide. I assume that's intended and you probably don't want to raise the lower bound to Maven 3.9, don't you?
> 
> I think we should keep 3.8.x compatibility, I'm not sure why not marking it with `Override` should cause any issue, as I assume only the new 3.9.x maven will call the method (if used) or does m2e directly call this method?

No m2e does not call this method. It didn't exist before Maven 3.9.7 and this PR only contains changes to make maven 3.97 work :)
I think it is also ok, to compile `ReadonlyMavenExecutionRequest` against Maven 3.9.7 and call the new `isIgnoreTransitiveRepositories()`. If at runtime there is only Maven 3.8.6 available I think a `NoSuchMethodError` would then only be thrown if one would try to call that method, which should not be the case. So I think its fine to compile against 3.9.7 instead of 3.8.6. 
But that brings the question, why does Tycho compile against 3.8.6 if I don't raise the lower bound in the manifest? In the IDE it is compiled against the latest version, but Tycho seems to use the oldest one. Can that be changed?

HannesWell:
> I also wonder if it would not be better to implement R/O MavenExecutionRequest as a [proxy class](https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Proxy.html) instead that simply delegates to the original for all method except those that starts with "set" in the handler?

Haven't thought about that, I only wonder if this could be a performance regression?
But I also don't plan to invest more time then necessary to make this work.

laeubi:
> But that brings the question, why does Tycho compile against 3.8.6 if I don't raise the lower bound in the manifest? In the IDE it is compiled against the latest version, but Tycho seems to use the oldest one. Can that be changed?

Don't know actually always the highest version should be used, but it makes sense to use the lower bound so no one can call "higher methods".

> Haven't thought about that, I only wonder if this could be a performance regression?

I don't think it is something to care about, at least it would decouple us from maven changes...

HannesWell:
> > But that brings the question, why does Tycho compile against 3.8.6 if I don't raise the lower bound in the manifest? In the IDE it is compiled against the latest version, but Tycho seems to use the oldest one. Can that be changed?
> 
> Don't know actually always the highest version should be used, but it makes sense to use the lower bound so no one can call "higher methods".

That's right, but it should be consistent with the IDE and I also assumed Tycho uses the latest version available. Maybe something special was set up when you added the Maven 3.8 artifact?

> 
> > Haven't thought about that, I only wonder if this could be a performance regression?
> 
> I don't think it is something to care about, at least it would decouple us from maven changes...

That's right. I cannot assess to what extent it would impact the performance or if it would be irrelevant.
I just think using the classic approach would be simpler to read/understand. And as written above, I think implementing the new method and delegate it to the request would also not cause problems if only Maven 3.8 is available because the method is never called.

PR #1765, opened 2024-05-25
Update commons-cli to 1.8.0

merks:


merks:
FYI, I'm experimenting with further automating and streamlining SimRel project's target platform updates.

merks:
The tests appear to be quite unstable.  I don't think any of this is related to this change.  Or?

HannesWell (approved):
> The tests appear to be quite unstable. I don't think any of this is related to this change. Or?

Unfortunately that's right.

But thank you for the update.

Issue #1766, opened 2024-05-25
M2E's Mac build fails when executing m2e.rcptt.tests


HannesWell:
The mac build of m2e fails for some time in `org.eclipse.m2e.rcptt.tests` while launching the rcptt test runtime.
From the log below it looks like some bundles are missing, but I wonder why this is only the case for mac?
We have one profile for mac, but actually I don't expect that to interact with the rcptt test?

@ptziegler you set them up, can you tell why this happens only on Mac?
@kwin IIRC you are on Mac, aren't you? Can you tell or test locally what happens?

```
[INFO] --- rcptt-maven-plugin:2.5.4:execute (default-execute) @ org.eclipse.m2e.rcptt.tests ---
[...]

PR #1767, opened 2024-05-25
Update to latest version of RCPTT runner to enable aarch64 support

ptziegler:
Fixes https://github.com/eclipse-m2e/m2e-core/issues/1766

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,7 +25,6 @@
 	<name>Maven Integration for Eclipse (root)</name>
 
 	<properties>
-		<rcptt-maven-version>2.5.4</rcptt-maven-version>
 		<tychoDefaultTestArgLine>-Xmx800m</tychoDefaultTestArgLine>
 		<tycho.testArgLine>${tychoDefaultTestArgLine}</tycho.testArgLine>
 		<surefire.timeout>900</surefire.timeout>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -25,7 +25,7 @@
 	<name>Maven Integration for Eclipse (root)</name>
 
 	<properties>
-		<rcptt-maven-version>2.5.4</rcptt-maven-version>
+		<rcptt-maven-version>2.5.5-MO</rcptt-maven-version>
```
HannesWell, line 28: There is already an M3 available.
```suggestion
		<rcptt-maven-version>2.5.5-M3</rcptt-maven-version>
```

ptziegler: Good point. Perhaps I shouldn't simply pick the first entry 😅 

image

HannesWell: `M3` isn't even listed there :O, but the repo from the pom contains it:
https://repo.eclipse.org/content/repositories/rcptt-releases

ptziegler: The tests passes with the M2. Should I add the Eclipse repo and switch to the M3 or keep the one from Maven Central?

HannesWell: Awesome. 🎉
The repo is already configured in the rcptt project pom.xml.
So we can just go to M3. AFAICT the rcptt maven plugin is not at maven-central.
But since that property is only used in one place, could you please take this occasion and just inline it into the m2e.rcptt.tests plugin?

ptziegler: Done

HannesWell (approved):
Thank you.

Can you please update the commit message, since we now have confidence that it works (we don't test aarch64 support anymore).

ptziegler:
> Thank you.
> 
> Can you please update the commit message, since we now have confidence that it works (we don't test aarch64 support anymore).

I've updated both the commit message and the PR

HannesWell (approved):
Great. Thanks again!

PR #1770, opened 2024-06-01
Extract maven-resolver-api from the embedded runtime bundle

laeubi:
Currently we embed the maven-resolver-api jar into m2e, but the API itself is already a bundle an we should consume it as such so we can:

1) update it independently
2) make it reusable by others without require the full runtime 3) cut down the size of the runtime bundle


target-platform/target-platform.target
```diff
@@ -32,15 +32,6 @@
...
 		</location>
-		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<repository location="https://download.eclipse.org/wildwebdeveloper/releases/1.3.4/"/>
-			<repository location="https://download.eclipse.org/tm4e/releases/0.11.0/"/>
-			<repository location="https://download.eclipse.org/lsp4e/releases/0.26.0/"/>
-			<repository location="https://download.eclipse.org/lsp4j/updates/releases/0.22.0/"/>
-			<repository location="https://download.eclipse.org/mylyn/updates/release/4.2.0/"/>
-			<unit id="org.eclipse.wildwebdeveloper.xml.feature.feature.group" version="0.0.0"/>
-			<unit id="org.eclipse.lsp4e" version="0.0.0"/>
-		</location>
```
HannesWell, line 43: Could you please revert the formatting changes that also remove comments.
The Target Editor unfortunately does not respect the existing element order.

HannesWell (commented):
We can try that, but I would be in favor to do it after the upcoming release for the 24-06 SimRel.
I have no clue if maven-core works that well if the resolve is not on the same flat classpath.
The bundle has indeed OSGi metadata, but I would be (positively) surprised if everything works already.

laeubi:
> I have no clue if maven-core works that well if the resolve is not on the same flat classpath.

I don't think its a big issue as its all only API, but realized that this will require to change all bundles previously requiring `maven.runtime`, I therefore changed it not to a require-bundle with reexport.

target-platform/target-platform.target
laeubi: I can do that but on the long run we either should not put comments in the target or enhance PDE to retain them in some way otherwise one is always forced to do text-editing targets

PR #1772, opened 2024-06-02
Force qualifier update of m2e.maven.runtime to resign embedded jansi

HannesWell:
Update build-qualifier (probably) because of changed Mac-signing certificates.


HannesWell:
This wasn't sufficient. I created https://github.com/eclipse-m2e/m2e-core/pull/1774 to simply ignore the entire jansi-jar for now.

PR #1788, opened 2024-07-11
Add bnd-maven-plugin integration to m2e

laeubi:
PDE now offers so called "shared components" (based on bndtools) we can leverage them to give a better user experience in m2e without needing the user to have bndtools (or pde) installed at all.

This PR is to track progress and make sure the current WIP can be integrated with the current code base.

### DIFF
--- a/org.eclipse.m2e.bnd.ui/.classpath
+++ b/org.eclipse.m2e.bnd.ui/.classpath
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-21"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry kind="src" path="src"/>
+	<classpathentry kind="output" path="bin"/>
+</classpath>
--- a/org.eclipse.m2e.bnd.ui/.project
+++ b/org.eclipse.m2e.bnd.ui/.project
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.eclipse.m2e.bnd.ui</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.jdt.core.javabuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.pde.ManifestBuilder</name>
+			<arguments>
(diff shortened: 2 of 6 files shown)


laeubi:
With the current rather rough implementation it is now possible to use the Bundle Repository View and see the (implicit) project Artifacts Bundle repository used by bnd-maven-plugins:

grafik

FYI @stbischof

stbischof:
What Features are planned in general? An which in this pr?

juergen-albert:
@laeubi great news! Have you tested this with bndtools installed? bndtools comes with an m2e connector as well. We should try to avoid that both fight each other.

laeubi:
> What Features are planned in general? An which in this pr?

There is no "masterplan" at all, for this PR I would like to get the "Bundle Repositories View" working as this is currently the most prominent (and useful) thing to know out of the box (and rather trivial to implement).

> Have you tested this with bndtools installed? bndtools comes with an m2e connector as well. We should try to avoid that both fight each other.

As bndtools never worked reliable for me its hard to guess if something is breaking or is "by design", beside that this is to actually avoid needing bndtools at all. Anyways bndtools is free to adapt or join the development in any way I just would not enforce that or invest time with that due to poor contribution experience of bnd(tool) project.

laeubi:
> If you implement something like this in the core and break the m2e support in bnd would be a bit unfair.

Every project chooses how to integrate into the eco-system, bndtools has made its decision as well but other are not meant to take that burden of design decisions.

> I'm not 100% sure what your Adapter does, so it might be totally fine.

It adapts an Eclipse project to a Bnd project so it can be used together with the [pde reusable components](https://github.com/eclipse-pde/eclipse.pde/tree/master/ui/org.eclipse.pde.bnd.ui) I tried to make bndtools able to participate but got only insult back in return and wasted a lot of time discussing things completely off-topic
- https://github.com/bndtools/bnd/pull/5885

So even if bnd is bound to "we don't understand it so it must be bad", other projects (like m2e) can still adapt to give a better **I**ntegrated **D**evelopment **E**xperience in their [IDE](https://en.wikipedia.org/wiki/IDE).

>  I guess it will show rather quick, if something breaks or not.

If you want you can checkout the PR build it with `-DskipTests` and then install the resulting updatesite in whatever Eclipse desired, m2e itself does not include bndtools so I can't test it as part of the usual m2e development.

laeubi:
I added a readme describing the basics [here](https://github.com/eclipse-pde/eclipse.pde/tree/master/ui/org.eclipse.pde.bnd.ui) now.

laeubi:
I now added a POC for a lemminx extension that adds autocompletion support to `pom.xml` files that are using an embedded `<bnd>` instructions

grafik

this is usually put into `CDATA` but sadly currently there is a bug
- https://github.com/eclipse/lemminx/issues/1694

Also fetching the text content seems to be flawed and should be fixed to make such auto completion more useful:
- https://github.com/eclipse/lemminx/issues/1695

PR #1811, opened 2024-08-24
[Maven Runtime] Reduce number of manually maintained version and update build-plugins

HannesWell:
1. Remove manually maintained version of exported packages.
The `bnd-maven-plugin` can derive the version of the `org.eclipse.aether*` and `org.eclipse.sisu*` packages perfectly fine based on the OSGi metadata that are embedded in these jars (since some time).
Furthermore the version of the exported `org.eclipse.sisu*` packages has been false before (due to the false version being used manually).

3. Exclude the dependency to `commons-codec` explicitly instead of mentioning it as provided to exclude it from the runtime bundle. Excluding it does not require a version to be defined.

### DIFF
--- a/m2e-parent/pom.xml
+++ b/m2e-parent/pom.xml
@@ -163,7 +163,7 @@
 				<plugin>
 					<groupId>org.apache.maven.plugins</groupId>
 					<artifactId>maven-clean-plugin</artifactId>
-					<version>3.2.0</version>
+					<version>3.4.0</version>
 				</plugin>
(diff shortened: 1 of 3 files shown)


HannesWell:
Verified that the build results are as expected and `org.eclipse.aether*` packages are exported with version `1.9.22` while `org.eclipse.sisu*` packages are exported with version 0.9.0.

Updating the embedded Maven is now even simpler (if no extra adjustments are necessary).

PR #1813, opened 2024-08-24
Fix failing tests for the M2E-PDE connector console printout handler

HannesWell:
And clean-up ConsoleTest.

The `ConsoleTest.testConsole_debuggerAttachmentAndLinkAlignmentAndBehavior_tychoProject` failed because the test-case used the outdated Tycho version 3.0.0, which seems to have a dependency that pulls in apache `commons-parent` version 35 that is considered as invalid by recent maven versions:

```
org\apache\commons\commons-parent\35\commons-parent-35.pom:
UTF-8 BOM plus xml decl of ISO-8859-1 is incompatible (position: START_DOCUMENT seen <?xml version="1.0" encoding="ISO-8859-1"... @1:42)
```


HannesWell:
Great. The `ConsoleTest.testConsole_debuggerAttachmentAndLinkAlignmentAndBehavior_tychoProject` is fixed, the remaining failures are for a follow-up.

PR #1816, opened 2024-08-27
Migrate RCPTT test for MavenTargetDependencyEditor to SWTBot

ptziegler:
In the long run, using RCPTT is probably causing more trouble than it's worth. It therefore makes sense to switch to a more lightweight solution like SWTBot for testing.


ptziegler:
This also has the added benefit of allowing one to execute those tests directly within the Eclipse IDE.

HannesWell:
Thank you @ptziegler for this migration, as said I have no idea about this part and trust your assessment. 
As long as the tests pass I'm happy with it.
@laeubi or somebody else do you have any special interest in this?

Patrick, could you then revert the change from https://github.com/eclipse-m2e/m2e-core/pull/1163 regarding the formats of the `tycho-p2-director-plugin`? In fact the entire `formats` section can be removed since all values are the default then and with that actually the entire section for the `tycho-p2-director-plugin` in that `pluginsManagement` element is obsolete. Could you please remove that entirely in a second commit of this PR?

target-platform/target-platform.target
```diff
@@ -45,6 +45,10 @@
 			<repository location="https://download.eclipse.org/cbi/updates/license/"/>
 			<unit id="org.eclipse.license.feature.group" version="0.0.0"/>
 		</location>
+		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
+			<repository location="https://download.eclipse.org/technology/swtbot/releases/4.2.1/"/>
```
ptziegler, line 49: A lot of the dependencies are part of the release train. Wouldn't it be cleaner to consume them via https://download.eclipse.org/releases/2024-06, rather than collecting everything over 6 locations?

ptziegler:
The instability seems to be gone. I need to make sure that the target platform has been resolved before executing the test. Otherwise the Maven target editor can't be opened reliably.

> In fact the entire formats section can be removed since all values are the default then and with that actually the entire section for the tycho-p2-director-plugin in that pluginsManagement element is obsolete

The entry itself needs to stay to specifiy the Tycho version, but the execution block is gone.

>2024-08-31T15:25:59.5306983Z [WARNING] Some problems were encountered while building the effective model for org.eclipse.m2e:m2e-ide:eclipse-repository:2.0.0
>2024-08-31T15:25:59.5310284Z [WARNING] 'build.plugins.plugin.version' for org.eclipse.tycho:tycho-p2-director-plugin is missing.
>2024-08-31T15:25:59.5311392Z [WARNING] 
>2024-08-31T15:25:59.5312391Z [WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.

akurtakov:
Would you please look into failing tests (https://ci.eclipse.org/m2e/job/m2e/job/PR-1816/20/) ? And sorry for the merge commit (I wanted to rebase but misclicked!) please rebase on master and force push.

target-platform/target-platform.target
HannesWell: This indeed seems cumbersome but is intentionally, because m2e contributes to the SimRel and the repository you mentioned is the SimRel aggregation.
And since we want to make sure that there are no self-loops in the SimRel we don't use that repository.

ptziegler:
I think the initial approach of opening the Maven editor via the target-platform editor is just not very reliable... A lot depends on when the dependencies are resolved, which is effectively arbitrary.
I now try to instantiate the Maven editor directly, which should overall be faster and more reliable. This of course has the downside that this needs to know the actual classes and is therefore less agnostic.

ptziegler:
All the SWTBot tests completed at least once now... The other failures are  very likely unrelated.

akurtakov:
Thanks for all the work. Merging.

HannesWell:
Thank you @ptziegler!

Issue #1824, opened 2024-09-04
Deferred properties in surefire argLine cause problems in 2.6.2


reckart:
Looks like the new support for `argLine` causes trouble. If one uses a deferred property in the argline like, this ends up unresolved in the JUnit run configuration.

E.g. 

```
<argLine>-Xmx512 @{argLineExtra}</argLine>
```

will create an unrunnable test run configuration that fails with

```
Error: could not open `{argLineExtra}`
```

ahoehma:
Same here :)

I have this settings in my surefire-plugin:

```
<plugin>
          <groupId>org.apache.maven.plugins</groupId>
          <artifactId>maven-surefire-plugin</artifactId>
          <version>3.5.0</version>
          <executions>
            <execution>
              <id>default-test</id>
              <phase>test</phase>
              <goals>
                <goal>test</goal>
              </goals>
              <configuration>
                <!-- @formatter:off Avoid newlines inside argLine! -->
                <!-- Sets the VM argument line used when unit tests are run. See jacoco plugin -->
                <argLine>-Xshare:off -Djdk.net.URLClassPath.disableClassPathURLCheck=true @{jacoco.surefireArgLine}</argLine>
                <!-- @formatter:on Avoid newlines inside argLine! -->
                <printSummary>true</printSummary>
                <useUnlimitedThreads>false</useUnlimitedThreads>
                <perCoreThreadCount>false</perCoreThreadCount>
                <parallelOptimized>false</parallelOptimized>
                <threadCount>${test-thread-count}</threadCount>
                <parallel>${test-parallel}</parallel>
                <forkCount>${test-fork-count}</forkCount>
                <reuseForks>${test-fork-reuse}</reuseForks>
                <rerunFailingTestsCount>${test-rerun-count}</rerunFailingTestsCount>
                <skipAfterFailureCount>${test-skip-after-failure-count}</skipAfterFailureCount>
                <parallelTestsTimeoutForcedInSeconds>400</parallelTestsTimeoutForcedInSeconds>
                <parallelTestsTimeoutInSeconds>400</parallelTestsTimeoutInSeconds>
                <properties>
                  <property>
                    <name>dataproviderthreadcount</name>
                    <value>${test-testng-dataproviderthreadcount}</value>
                  </property>
                </properties>
                <includes>
                  <include>${test-includes}</include>
                </includes>
                <reportFormat>plain</reportFormat>
                <consoleOutputReporter>
                  <disable>true</disable>
[...]

PR #1827, opened 2024-09-09
FIX: unit test configuration handles deferred var @{x} from surefire/failsafe

treilhes:
Fixes https://github.com/eclipse-m2e/m2e-core/issues/1824
Fixes https://github.com/eclipse-m2e/m2e-core/issues/1832

### DIFF
--- a/org.eclipse.m2e.jdt.tests/src/org/eclipse/m2e/jdt/tests/UnitTestLaunchConfigConfigurationTest.java
+++ b/org.eclipse.m2e.jdt.tests/src/org/eclipse/m2e/jdt/tests/UnitTestLaunchConfigConfigurationTest.java
@@ -56,26 +56,30 @@ public class UnitTestLaunchConfigConfigurationTest extends AbstractMavenProjectT
 	private static final String SUREFIRE_ARGS_SET = """
 			<configuration>
 				<argLine>
-					--argLineItem=surefireArgLineValue
+					--argLineItem=surefireArgLineValue --undefinedArgLineItem=${undefinedProperty}
 				</argLine>
 				<systemPropertyVariables>
 					<surefireProp1>surefireProp1Value</surefireProp1>
+					<surefireEmptyProp>${undefinedProperty}</surefireEmptyProp>
 				</systemPropertyVariables>
 				<environmentVariables>
 					<surefireEnvironmentVariables1>surefireEnvironmentVariables1Value</surefireEnvironmentVariables1>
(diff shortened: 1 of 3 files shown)


org.eclipse.m2e.jdt.tests/src/org/eclipse/m2e/jdt/tests/UnitTestLaunchConfigConfigurationTest.java
```diff
@@ -279,6 +279,33 @@ public void test_configuration_must_be_updated_with_failSafe_config_when_created
...
+		updateProject(project);
+		waitForJobsToComplete();
+
+		ILaunchConfiguration[] updatedConfigurations = LAUNCH_MANAGER.getLaunchConfigurations(type);
+		assertTrue(updatedConfigurations.length == 1);
+
+		ILaunchConfiguration config = updatedConfigurations[0];
+		String argLine = config.getAttribute(UnitTestSupport.LAUNCH_CONFIG_VM_ARGUMENTS, "");
+		assertTrue(argLine.contains("-javaagent")); // resolved jacoco agent
+		assertTrue(argLine.contains("@{titi.tata}")); // unresolved property, maybe removing the placeholder would be
```
reckart, line 305: I'd say yes - removing it would probably be better ;) 

Although I believe that in a CLI build, it would also not be resolved and stay as is.

treilhes: I hesitated but in the end I preferred a quick failure in the case of an undefined variable than spending time looking for where the problem comes from

@HannesWell, @laeubi, @akurtakov what are your thought on this ?

HannesWell: I agree.
First M2E should behave as much as possible as a standalone Maven execution on the CLI.
And second respectively in general a fail-fast behavior is usually better.

HannesWell (approved):
Thanks a lot for these fixes @treilhes and sorry for the late reply.

I made a few small adjustments in `UnitTestSupport.resolveDeferredVariables()` to be more precise in case of multiple occurrences of equal placeholders (although the result should be the same) and to handle a `null` value or empty properties.

I plan to submit this change this evening. Please let me know if you are not fine with any adjustments.

HannesWell:
Thanks again.

Btw. I noticed that the tests you added are skipped in the CI. Locally they passed, but it would be good to have them running. 
Could you please look into that?

PR #1844, opened 2024-09-27
Fix pgp sign

akurtakov:
Previous commit didn't make it work.


akurtakov:
Another direct merge due to no signing on PRs.

PR #1845, opened 2024-09-27
Keep only the artifacts of the last 2 builds

fredg02:
In order to save disk space, the number of builds to keep with artifacts should be limited to 2.

This change should be applied to all branches.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 pipeline {
 	options {
 		timeout(time: 45, unit: 'MINUTES')
-		buildDiscarder(logRotator(numToKeepStr:'10'))
+		buildDiscarder(logRotator(numToKeepStr:'5', artifactNumToKeepStr: 'master'.equals(env.BRANCH_NAME) ? '5' : '1' ))
 		disableConcurrentBuilds(abortPrevious: true)
 		timestamps()
 	}


Jenkinsfile
```diff
@@ -1,7 +1,7 @@
 pipeline {
 	options {
 		timeout(time: 45, unit: 'MINUTES')
-		buildDiscarder(logRotator(numToKeepStr:'10'))
+		buildDiscarder(logRotator(numToKeepStr:'10', artifactNumToKeepStr: '2'))
```
laeubi, line 4: Maybe we should simply discard old build after e.g. 30 days anyways?

fredg02: Does not hurt, but artifacts should still be discarded (and not wait for 30 days to lapse).

laeubi: Yes but even 2 artifacts are wasted space for older branches no one really looks at... we mostly retain them for historic reasons or if one really wants to back port things so I think they can be pruned completely probably if nothing changes for a long time.

fredg02: I never tested which option has priority. In a perfect world, artifacts of builds that are older than 30 days would be discarded, despite the option to keep 2 builds with artifacts. Realistically, limiting the number of builds saves more disk space for most use-cases, though.

Example:
* artifacts size: 100MB
* 10 jobs with 10 builds for 30 days => 10GB
* 10 jobs with max 2 builds with artifacts =>  2GB

This example applies to a standard CI build of a main branch where development happens for a long time. It's a different story for a branch that is only used temporarily, but is then kept around forever and never gets cleaned up. So cleaning up old branches is a best practice as well. :)

HannesWell: Does `artifactNumToKeepStr` in contrast to `numToKeepStr` mean that the artifacts for the builds 3 to 10 are discarded and only the other data of the run are kept?

I think for PR builds it's fine to even just keep the artifacts of the last build, but for the master it might be good to retain more. But I think it's sufficient to keep only 5 builds in general.

So my suggestion would be

```suggestion
		buildDiscarder(logRotator(numToKeepStr:'5', artifactNumToKeepStr: 'master'.equals(env.BRANCH_NAME) ? '5' : '1' ))
```

fredg02: > Does artifactNumToKeepStr in contrast to numToKeepStr mean that the artifacts for the builds 3 to 10 are discarded and only the other data of the run are kept?

Yes.

HannesWell (approved):
Thank you.

Just squashed and force-pushed the changes manually (the result of a Squash-And-Merge in the GH UI is not so nice).

HannesWell:
@laeubi if you're fine as well I'll submit this.

HannesWell:
Thank you @fredg02 for this.

I'll try to update open branches as much as possible, but currently the m2e JIPP seems to be down entirely. I'll check again tomorrow and if that's a permanent outage I'll create a help-desk issue.

HannesWell:
@fredg02 for the jobs that couldn't be rebased and therefore picked this up, I have now manually deleted old jobs. 
The m2e-JIPP should now have lower storage usage.

PR #1847, opened 2024-09-27
[PDE] Reduce usage of PDE internals in production code

HannesWell:
Follow-up on https://github.com/eclipse-m2e/m2e-core/pull/1835.

In general we could also consider to add API constants for such constant Strings in PDE.


laeubi:
Official API sounds good especially as encoding it fixed in code has some drawbacks as well.

PR #1852, opened 2024-10-01
Stop building products

akurtakov:
Not used by anything nor published.


akurtakov:
RCPTT tests still use , looks like I have to give up for now.

PR #1854, opened 2024-10-03
Fix syntax error in shell script for deploy

akurtakov:
Ubuntu uses dash as default shell thus explicitly call bash.


akurtakov:
Deploy script is executed only for master builds thus push directly as it would not be verified.

PR #1856, opened 2024-10-03
[Build] Remove irrelevant propertys and build in parallel at GH

HannesWell:


HannesWell:
Parallel build on Jenkins was unstable.
But running that in parallel would be even more important since Jenkins is the actual bottleneck.

PR #1858, opened 2024-10-03
Revert "Simplify pgp sign"

HannesWell:
This reverts change in pgp-signing done in the commits ca976ab480bab0d31c0af7aa3297fcf99eccb711 and
1dd211b9fba841e0c54744f6a0850bd320ee6a68


HannesWell:
The simplification was not impossible but just broken because of a typo, that was fixed in https://github.com/eclipse-m2e/m2e-core/pull/1859.
So this is obsolete.

PR #1842, opened 2024-09-27
Simplify pgp sign

akurtakov:

PR #1859, opened 2024-10-03
[Build] Fix definition of pgp-secretKeys property to fix pgp signing

HannesWell:
Follow-up on https://github.com/eclipse-m2e/m2e-core/pull/1842.


HannesWell:
The signing in a replayed build succeeded.

PR #1862, opened 2024-10-04
Stop jarsigning

akurtakov:
PGP sign is enough, simrel no longer requires jarsigning and build becomes faster and more stable without the network roadtrips caused by eclipse-jarsigner.

### DIFF
--- a/m2e-parent/pom.xml
+++ b/m2e-parent/pom.xml
@@ -199,15 +199,6 @@
 						</dependency>
 					</dependencies>
 				</plugin>
-				<plugin>
-					<groupId>org.eclipse.cbi.maven.plugins</groupId>
-					<artifactId>eclipse-jarsigner-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


HannesWell:
@laeubi are you fine with this?
I think it's worth a try.

akurtakov:
Merging in this case.

HannesWell:
Since this change, when updating to the latest snapshots of m2e I always get the message that the m2e artifacts are unsigned.
I don't know why this is the case, but it's bad and unless we can fix this quickly I contemplate to revert this for the release I plan to do this evening.

akurtakov:
I have just installed everything from the snapshots repo and didn't get warnings and looking at https://www.eclipse.org/downloads/download.php?file=/technology/m2e/snapshots/latest/artifacts.xml.xz I see pgp.signatures so please share exactly which bundles are unsigned for you.

PR #1864, opened 2024-10-04
Do not include o.e.m2e.tests.common in sdk feature

akurtakov:
This bundle is only useful for the other test bundles which are not included in the p2 repo at all.

### DIFF
--- a/org.eclipse.m2e.sdk.feature/feature.xml
+++ b/org.eclipse.m2e.sdk.feature/feature.xml
@@ -51,12 +51,4 @@
          id="org.eclipse.m2e.pde.feature.source"
          version="0.0.0"/>
 
-   <plugin
-         id="org.eclipse.m2e.tests.common"
-         version="0.0.0"/>
-
-   <plugin
-         id="org.eclipse.m2e.tests.common.source"
-         version="0.0.0"/>
-
 </feature>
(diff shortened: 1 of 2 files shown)


org.eclipse.m2e.sdk.feature/feature.xml
```diff
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.m2e.sdk.feature"
       label="%featureName"
-      version="2.7.0.qualifier"
+      version="2.7.100.qualifier"
```
HannesWell, line 5: The feature version has already been bumped this release cycle.
```suggestion
      version="2.7.0.qualifier"
```

HannesWell (changes requested):

org.eclipse.m2e.sdk.feature/feature.xml
akurtakov: I will revert the version bump and force qualifier update as the first

akurtakov:
I give up as build fails with " org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho:tycho-baseline-plugin:4.0.9:verify (baseline-check) on project org.eclipse.m2e.sdk.feature: Baseline problems found! Project version: 2.7.0.20241004-1536, baseline version: 2.6.2.20240828-1954, suggested version: 3.0.0" and there is no easy way to add filter but just a big bang skip.

laeubi:
> Baseline problems found!

As its q bit hard to find in the log, here is the baseline report summary

grafik

so removal is a breaking change for features, especially for an SDK feature that might be used inside a target file it means users are suddenly missing items.

PR #1855, opened 2024-10-03
[Build] Fix specification of bash as shell for deployment

HannesWell:
Follow-up on https://github.com/eclipse-m2e/m2e-core/pull/1854.

PR #1866, opened 2024-10-04
[Build] Remove workaround for previously different default shell

HannesWell:
On the new Ubuntu images initially the 'dash' shell was used by default and now it was changed to use 'bash' again.

So we don't have to explicitly specify the shell anymore.

See https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5061

This reverts the workarounds added via
- https://github.com/eclipse-m2e/m2e-core/pull/1854
- https://github.com/eclipse-m2e/m2e-core/pull/1855
- https://github.com/eclipse-m2e/m2e-core/pull/1857

On the other hand, we could also keep the explicit specification of the shell so that reads are aware what to expect?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,7 +25,7 @@ pipeline {
 					string(credentialsId: 'gpg-passphrase', variable: 'KEYRING_PASSPHRASE')
 				]) {
 				xvnc(useXauthority: true) {
-					sh '''#!/bin/bash -x
+					sh '''
 						mavenArgs="clean verify --batch-mode -Dmaven.test.failure.ignore=true -Dtycho.p2.baselineMode=failCommon"
 						if [[ ${BRANCH_NAME} == master ]] || [[ ${BRANCH_NAME} =~ m2e-[0-9]+\\.[0-9]+\\.x ]]; then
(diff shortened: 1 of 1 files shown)


akurtakov:
I believe it's best to keep the explicit declaration of shell for clarity.

laeubi:
It would even be more clear if we won't need any shell magic at all :-D

e.g. in Tycho I use this to get different settings for deploy/not deploy:

https://github.com/eclipse-tycho/tycho/blob/cec877685f3dca852fb157b76766c49c029beeaf/Jenkinsfile#L3-L9
https://github.com/eclipse-tycho/tycho/blob/cec877685f3dca852fb157b76766c49c029beeaf/Jenkinsfile#L37

HannesWell:
Ok, then let's just keep it.

> It would even be more clear if we won't need any shell magic at all :-D

That would be nice but due to the task that's done, I think it doesn't become overall simpler in this case if one just uses declarative pipeline and wants to avoid duplications.

And the XPath extraction would probably have to be implemented in groovy like it would be done in Java. And that wouldn't be simpler at all.

PR #1874, opened 2024-10-10
Provide improved support for long property values

merks:
- Make the MavenPropertyDialog resizeable and provide dialog settings to remember the size.
- Use a table layout for the property table of the MavenLaunchMainTab.

### DIFF
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/dialogs/MavenPropertyDialog.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/dialogs/MavenPropertyDialog.java
@@ -13,9 +13,10 @@
 
 package org.eclipse.m2e.core.ui.internal.dialogs;
 
-// import org.eclipse.debug.ui.StringVariableSelectionDialog;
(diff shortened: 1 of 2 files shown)


merks:
I literally needed to edit a value in some other editor and paste it because it just wasn't possible with the tiny limited-sized dialog:

image

The table layout  also improves the use of the existing real estate.

merks:
A more realistic value that I was editing is like this one:
```
${file_extract_matching_pattern:${platform-dependencies},UTF-8,<version.maven.tycho>([^<]+)</version.maven.tycho>,$1,3.0.1}
```
I'm experimenting with a dynamic variable where I can specify a file to be read with the given encoding, a pattern that's used to find a match in the content, a replacement string used to extract the content, and a default value if the pattern doesn't match.  It's quite a bit to specify!

laeubi (approved):
Looks good to me

HannesWell:
Thank you for this one as well.

If one want's to enhance M2E's launch dialog further I think it would be good if one could edit property keys and values directly in the table without the need of a separate dialog at all.

merks:
Yes, as well up and down buttons.   I know order doesn't matter, but it's still nice to organize.

PR #1876, opened 2024-10-11
Avoid grabbing focus when creating the launch configuration main tab

merks:
- Normally one can, in the Run Configurations dialog, use the keyboard in the left control to navigate each of the launch configurations but that does work for m2e because the configuration control grabs away the focus.


HannesWell:
Thank you Ed!

This actually bugged me as well for some time, but I never tried got the time to find the cause.

PR #1888, opened 2024-11-15
Provide auto completion for bnd instructions in maven xml documents

laeubi:
The bnd-maven and felix-bundle plugin provide a way to use bnd-instructions to build OSGi bundles. As this is a complex syntax that can not be expressed as regular maven-mojo configuration lemminx-maven can not supply any useful completions.

This adds a new lemminx-extension that provides such completions in a very basic way to support people writing such custom configuration.

Here is a little video showing the new completion support:

https://github.com/user-attachments/assets/a1c78594-d0f9-4b19-a3ef-286c9c74069a

@mickaelistria this shows how to deploy a lemminx-extension as a regular bundle inside m2e and providing additional jars to the flat-classpath using OSGi wiring API


laeubi:
This is currently quite basic and not super-smart, but will open up a whole bunch of new things to support, e.g. the next thing would be to add diagnostics to one can get errors, warnings and so on.

I therefore plan to merge this after the next m2e release for 2024-12 so we can further improve / experiment with that!

mickaelistria:
> @mickaelistria this shows how to deploy a lemminx-extension as a regular bundle inside m2e and providing additional jars to the flat-classpath using OSGi wiring API

That's really good, it does make things better. It's still not immune to libraries incompatibilities inside LemMinX process though, but that's another story.

laeubi:
Yes it makes it just a little bit easier now from a deployment perspective as one does not need to collect everything manually. For more one would need to change how lemminx itself is launched, e.g if everything would be embedded inside the bundle (as with a very minimal example [here](https://github.com/eclipse-lemminx/lemminx/pull/1703)) and only imports what is required to share the API then one could probably make it an connect-osgi-framework where each extension is its own bundle and encapsulates all requirements except the shared classes like lsp4j...

laeubi:
Has to be adapted to:
- https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/pull/1741#issuecomment-2644621538

laeubi:
I have now rebased and adjusted this, due to a bug in PDE this currently not working smooth during development in the IDE:

- https://github.com/eclipse-pde/eclipse.pde/pull/1603

PR #1900, opened 2024-11-27
Rename junit.yml to unit.yml

copyleft21:


akurtakov:
There have been no explanation what the benefit is from this one despite being requested by committer.

PR #1901, opened 2024-11-28
Perform Maven target dependency update in non-UI thread

ptziegler:
Depending on how many artifacts are updated, this might be a long-running operation and should therefore be run without blocking the application.

### DIFF
--- a/org.eclipse.m2e.pde.feature/feature.xml
+++ b/org.eclipse.m2e.pde.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.m2e.pde.feature"
       label="%featureName"
-      version="2.3.300.qualifier"
+      version="2.3.400.qualifier"
       provider-name="%providerName"
       plugin="org.eclipse.m2e.core"
       license-feature="org.eclipse.license"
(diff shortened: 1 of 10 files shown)


laeubi:
@ptziegler can you check with latest release? I Think that I should have fixed this in PDE so it performs updates in a separate thread. If not let me know I think this is better handled in the PDE editor than directly in the target location implementation.

ptziegler:
> @ptziegler can you check with latest release? I Think that I should have fixed this in PDE so it performs updates in a separate thread. If not let me know I think this is better handled in the PDE editor than directly in the target location implementation.

I've just tested it with org.eclipse.pde 3.13.3000.v20250228-0140 and org.eclipse.m2e.pde.target 2.1.1.20250223-0727 and the problem still exists. 

More specifically, the issue lies with the following button, which is used to updated all selected dependencies. Those updates are done in the UI thread and if, like in my case, 200 or so artifacts are updated at once, the IDE freezes for ~1-2 minutes.

image

Given that this is purely an interaction between the dialog and Maven Central, I'm not sure if this has anything to do with PDE editor itself.

laeubi:
> Given that this is purely an interaction between the dialog and Maven Central, I'm not sure if this has anything to do with PDE editor itself.

I see, I thought this is referencing to the update button in the target editor, so is there anything left here or can we merge this after conflicts are resolved?

ptziegler:
I've rebased everything and fixed the merge conflict. Unless there are any related test failures, that should now include everything. If anyone is interested, this is how it looks like:

image

PR #1917, opened 2025-01-28
Add workflow to automate target updates

laeubi:


pom.xml
```diff
@@ -180,6 +180,28 @@
...
 				</plugin>
+				<plugin>
+					<groupId>org.eclipse.tycho.extras</groupId>
+					<artifactId>tycho-version-bump-plugin</artifactId>
+					<version>${tycho-version}</version>
+					<executions>
+						<execution>
+							<id>update-target</id>
+							<configuration>
+								<updateEmptyVersion>false</updateEmptyVersion>
```
laeubi, line 191: I have currently disabled this as we use the `0.0.0` version in the target, the question for me would be if we can automate this if it is then maybe better to use exact versions in the target.

laeubi:
As we do not have a PAT here I use the github one to get the things going:

- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5588

PR #1919, opened 2025-01-29
Rename master > main

laeubi:


laeubi:
@HannesWell @merks can you probabbly look at the setup files if I did adjusted them correctly?

merks:
Looks ok. It will effectively disable the project setup reference in existing workspace.setup, i.e., existing installations will no longer see or use the project setup.

merks:
If you want to fix and existing installation after this is committed, you can use Navigate -> Open Setup -> Workspace, use Open in Text Editor on the context menu, change master to main, and save:

image

laeubi:
@HannesWell it seems that changing the markdown file generates a version bump, I tried to fix that for the future but the configuration looks odd, it tells it wants pom.xml changes (but why?!?) and disables all jgit ignores... confusing.

org.eclipse.m2e.editor.lemminx/pom.xml
```diff
@@ -60,7 +60,13 @@
 				<groupId>org.eclipse.tycho</groupId>
 				<artifactId>tycho-packaging-plugin</artifactId>
 				<configuration>
-					<jgit.ignore>nothing</jgit.ignore> <!-- empty (whitespace needed), don't ignore pom.xml as changes in here affect payload -->
+					<!-- overrides from parent as we don't want to ignore pom changes! -->
+					<ignores>
+						<ignore>OSGI-INF/*</ignore>
+						<ignore>META-INF/ECLIPSE_.RSA</ignore>
+						<ignore>META-INF/ECLIPSE_.SF</ignore>
```
HannesWell, line 67:
```suggestion
						<ignore>META-INF/ECLIPSE_.*</ignore>
```

HannesWell (commented):
> @HannesWell it seems that changing the markdown file generates a version bump, I tried to fix that for the future but the configuration looks odd, it tells it wants pom.xml changes (but why?!?) and disables all jgit ignores... confusing.
> 
In case of the lemminx-bundle the inclusion is of the embedded lemminx jar is driven by the pom.xml file. And in case the lemminx version is changed in this pom we want to have a an updated qualifier.
However I don't see in the parent (anymore) that pom.xml files are generally excluded. Maybe that's from the past.

laeubi:
> In case of the lemminx-bundle the inclusion is of the embedded lemminx jar is driven by the pom.xml file. 

But should not the embedded jar itself already make a difference?

org.eclipse.m2e.feature/feature.xml
```diff
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.m2e.feature"
       label="%featureName"
-      version="2.7.0.qualifier"
+      version="2.7.1.qualifier"
```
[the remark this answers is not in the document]
laeubi: m2e uses only single increments see:

https://github.com/eclipse-m2e/m2e-core/blob/main/CONTRIBUTING.md#%EF%B8%8F-version-bump

org.eclipse.m2e.editor.lemminx/pom.xml
laeubi: I think I will rvert this alltogehter and dont touch it.

laeubi:
Further refinements can better be made in separate PRs.

HannesWell:
> > In case of the lemminx-bundle the inclusion is of the embedded lemminx jar is driven by the pom.xml file.
> 
> But should not the embedded jar itself already make a difference?

The jar is not committed into the repository, it's just copied into the project by the build. Therefore a change in the jar itself doesn't change the git-timestamp.

PR #1932, opened 2025-02-20
Update org.eclipse.m2e.workspace.cli to 0.4.0

akurtakov:


HannesWell:
@akurtakov have you checked in detail if this update is really save? If I'm not mistaken this update required adaption in M2E and could break code depending on M2E and we therefore skipped this update on purpose.

akurtakov:
I haven't noticed anything in a cursory check. Tests also succeeded.

PR #1936, opened 2025-02-21
Test variables

laeubi:
I can't reopen

- https://github.com/eclipse-m2e/m2e-core/pull/1459

so this is the new one


laeubi:
I see the test is failing locally as well ... if I modify the test so I can see the UI I see the expected text is actually printed to the console.. but for some reason it does not show up in the proxy...

laeubi:
I now hopefully have found a way to make the test more stable and side-effect free:

- https://github.com/eclipse-m2e/m2e-core/pull/1943

It seems that adding `org.eclipse.debug.ui` to the required bundles enable "something" (maybe the console view or similar) that makes it impossible for the test to reliable capture its output.

Issue #286, opened 2021-07-21
move custom 'exclude' to 'dependency' element


laeubi:
Currently the m2-pde target location uses a custom `<exclude>` element to exclude a decency artifact.
With #212 we support the standard <dependency> element from the maven model what support a exclusion in the following way:

```
<dependency>
	<groupId>org.group</groupId>
	<artifactId>artifact1</artifactId>
	<exclusions>
		<exclusion>
			<groupId>bad.group</groupId>
			<artifactId>artifact3</artifactId>
		</exclusion>
	</exclusions>
</dependency>
```

to better align with common techniques know to maven-users we should handle exclusions the same way here.

vogella:
Just for documentation for people stumbling about this issue, the custom (using the non Maven syntax) element tag would look like the following:

````
<dependency>
	<groupId>org.group</groupId>
	<artifactId>artifact1</artifactId>
	<exclude>
               <groupId>bad.group</groupId>
		<artifactId>artifact3</artifactId>
         </exclude>
</dependency>

````

laeubi:
Actually there should be a button for this but this seems currently broken :-(

laeubi:
@vogella that's how it is supposed to look like:
![grafik](https://user-images.githubusercontent.com/1331477/208263554-132f07e7-7a3b-4704-ba65-07610c4ccf6b.png)
![grafik](https://user-images.githubusercontent.com/1331477/208263589-84259fac-cef3-44b5-b125-a71b391d86fd.png)

vogella:
Thanks @laeubi

laeubi:
With

- https://github.com/eclipse-m2e/m2e-core/pull/1939

this is now much easier to implement, in fact we should be able to simply pass the defined exclude to the now used resolver dependency object and it should work. I'll try to prepare a PR for this.

If we have this it would of course be nice to have some kind of UI, similar to what we have for adding roots.

PR #1939, opened 2025-02-21
Fix transitive artifact collection in m2e targets and make sharable

laeubi:
Currently m2e uses the usual maven resolver to find transitive dependencies but this works differently than what we want here. Instead we want all dependencies and their dependencies and then filter for the given scopes provided if any match.As this is a non trivial operation, it seems useful to make the code also sharable with Tycho so we ensure there is always the same result in Tycho as in m2e.

This now creates a new MavenDependencyCollector that can be shared with Tycho that traverses the dependency tree in a way we need here and collects everything along the way. This also has the advantage, that if we choose to only collect direct dependencies, not the whole tree has to be traversed as before.


lorenzodallavecchia:
Hi, sorry to revive this conversation, but I'm not sure if this is a bug or a question, and which other issue may exist for it (maybe #286).

After introducing the new `MavenDependencyCollector` in 2.8.0, I see that the "Maven" location is no longer supporting some Maven features correctly.

- The `exclusions` are completely ignored (this is the proposed standard replacement for top-level `exclude`, see #286).
- The managed dependencies (`dependencyManagement`) only take effect when used in the base POM or the same POM that declares the `dependency`. They no longer work for fixing the versions of _transitive_ dependencies.

Both of these features used to work before the introduction of the new `MavenDependencyCollector`.

@laeubi Since you say that the goal is not to use the same Maven resolver, I wonder, are these limitations intended? Or did you just mean that we should not share code with Maven, but still respect its semantics?

PR #1942, opened 2025-02-23
[TP] Update to Archetype 3.3.1 and unify archetypes OSGi metadata

HannesWell:
- Adapt to API changes in archetype-common
- Updating to Archetype 3.3. allows to remove the 'maven-artifact-transfer' dependency.


org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypePlugin.java
```diff
@@ -113,7 +114,6 @@ protected void configure() {
     final ContainerConfiguration cc = new DefaultContainerConfiguration() //
         .setClassWorld(new ClassWorld("plexus.core", ArchetypeArtifactManager.class.getClassLoader())) //$NON-NLS-1$
         .setClassPathScanning(PlexusConstants.SCANNING_INDEX) //
-        .setAutoWiring(true) //
         .setName("plexus"); //$NON-NLS-1$
     container = new DefaultPlexusContainer(cc, logginModule);
     archetypeArtifactManager = container.lookup(ArchetypeArtifactManager.class);
```
HannesWell, line 119: When trying this change locally this failed.
I assume this is a consequence of https://github.com/apache/maven-archetype/pull/206?
@laeubi any hint how to fix it?

laeubi: I'm not complete sure *what* failed, but disable autowiring does not looks like what one wants. so why should it be removed?

HannesWell: The lookup fails with no such element. 
And auto-wiring is implied with our config.

HannesWell: > Looking at the implementation autowiring is off by default and even if not I think we should make it explciti.

It's implied by the classPath-scanning mode:
https://github.com/eclipse-sisu/sisu-project/blob/8ea80a1289fe8efa8d8a95e3991c5bbf2c8f4bd6/org.eclipse.sisu.plexus/src/main/java/org/codehaus/plexus/DefaultContainerConfiguration.java#L126-L134

But we can also keep it explicit

> Was is maybe missing here is to enable .setJSR250Lifecycle(true)

At least locally, unfortunately this didn't help.

laeubi: Then "something" seems missing on the classpath for this to work :-\

laeubi:
@HannesWell I have checked this now and the problem ist that with the move to sisu now all components are tried to be looked up, and we have removed some dependencies intentionally. I'll check if we can use some bnd magic or if I can somehow intercept the loading of unwanted parts.

laeubi:
Another issue seem that at least locally I can't see that the fragment approach works her all archetype artifacts are now plain bundles.

target-platform/target-platform.target
```diff
@@ -133,60 +133,46 @@
...
 Bundle-SymbolicName:   org.eclipse.m2e.archetype.${replacestring;${mvnArtifactId};archetype-}
-# Change the following value to '${version}' once the next minor version of archetype is availble
-Bundle-Version:        ${version}.2
+Bundle-Version:        ${version}
 Bundle-Vendor: 	       Eclipse.org - m2e
-Fragment-Host:         org.eclipse.m2e.archetype.common;bundle-version="${range;[===,+);3.2.104}"
-Export-Package:        org.apache.maven.*;-noimport:=true;provider=m2e;mandatory:=provider;version="${version}";x-friends:="org.eclipse.m2e.core.ui"
-Import-Package:        !*
+Fragment-Host:         org.eclipse.m2e.archetype.common;bundle-version="${range;[===,===];${version}}"
+-removeheaders ${if;isHostBundle;Fragment-Host;}
```
laeubi, line 166:
```suggestion
-removeheaders ${if;${isHostBundle};Fragment-Host,Require-Bundle;}
```

target-platform/target-platform.target
```diff
@@ -133,60 +133,46 @@
...
+Eclipse-ExtensibleAPI: true
+Require-Bundle:        \
+                       org.eclipse.m2e.maven.runtime;bundle-version="[3.8.0,4.0.0)", \
+                       com.ibm.icu
+Import-Package:        org.jdom2*
+Export-Package:        \
+                       META-INF.plexus;-noimport:=true;x-internal:=true, \
```
laeubi, line 173:
```suggestion
                       META-INF.plexus;-noimport:=true;x-internal:=true, \
                       META-INF.sisu;-noimport:=true;x-internal:=true, \
```

laeubi:
With the proposed changes it works for me using plexus here. the main problem was that the fragments where not generated due to missing `${...}` in if clause, so it used the literal what always evaluated to "true"

laeubi:
By the way we probably should contribute to the archetype-plugin and ask if they can use constructor injection instead of field injection, then we can avoid the whole DI container thing and just create a default instance what should be enough for our case.

target-platform/target-platform.target
HannesWell: Good catch, thank you. 
That was definitely a problem!

HannesWell:
> With the proposed changes it works for me using plexus here. the main problem was that the fragments where not generated due to missing `${...}` in if clause, so it used the literal what always evaluated to "true"

Yes, then the implementation of the interface is found, however unfortunately the fields are still not injected. I have no clue why :/


> By the way we probably should contribute to the archetype-plugin and ask if they can use constructor injection instead of field injection, then we can avoid the whole DI container thing and just create a default instance what should be enough for our case.

Yes that would be great.

target-platform/target-platform.target
```diff
@@ -133,60 +133,46 @@
...
-Fragment-Host:         org.eclipse.m2e.archetype.common;bundle-version="${range;[===,+);3.2.104}"
-Export-Package:        org.apache.maven.*;-noimport:=true;provider=m2e;mandatory:=provider;version="${version}";x-friends:="org.eclipse.m2e.core.ui"
-Import-Package:        !*
+Fragment-Host:         org.eclipse.m2e.archetype.common;bundle-version="${range;[===,===];${version}}"
+-removeheaders ${if;${isHostBundle};Fragment-Host;}
+Eclipse-ExtensibleAPI: true
+Require-Bundle:        \
+                       org.eclipse.m2e.maven.runtime;bundle-version="[3.8.0,4.0.0)", \
+                       com.ibm.icu
+Import-Package:        org.jdom2*
```
laeubi, line 171:
```suggestion
Import-Package:        org.jdom2*,javax.inject
```

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypePlugin.java
```diff
@@ -114,8 +125,21 @@ protected void configure() {
         .setClassWorld(new ClassWorld("plexus.core", ArchetypeArtifactManager.class.getClassLoader())) //$NON-NLS-1$
         .setClassPathScanning(PlexusConstants.SCANNING_INDEX) //
         .setAutoWiring(true) //
+        .setJSR250Lifecycle(true) //
         .setName("plexus"); //$NON-NLS-1$
+
+    ClassSpace space = new BundleClassSpace(FrameworkUtil.getBundle(ArchetypeArtifactManager.class));
```
laeubi, line 131:
```suggestion
    ClassSpace space = new BundleClassSpace(FrameworkUtil.getBundle(ArchetypeArtifactManager.class));
     final Module repositorySystemModule = new AbstractModule() {
        @Override
        protected void configure() {
          try {
            bind(RepositorySystem.class).toInstance(MavenPluginActivator.getDefault().getRepositorySystem());
          } catch(CoreException ex) {
            ex.printStackTrace();
          }
        }
      };
```

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypePlugin.java
```diff
@@ -114,8 +125,21 @@ protected void configure() {
...
+    Injector injector = Guice.createInjector(logginModule,
+        new WireModule(new SpaceModule(space, BeanScanning.INDEX, false)));
+    ArchetypeArtifactManager mgr1 = injector.getInstance(DefaultArchetypeArtifactManager.class);
```
laeubi, line 134:
```suggestion
```

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypePlugin.java
```diff
@@ -114,8 +125,21 @@ protected void configure() {
         .setClassWorld(new ClassWorld("plexus.core", ArchetypeArtifactManager.class.getClassLoader())) //$NON-NLS-1$
         .setClassPathScanning(PlexusConstants.SCANNING_INDEX) //
         .setAutoWiring(true) //
+        .setJSR250Lifecycle(true) //
         .setName("plexus"); //$NON-NLS-1$
+
+    ClassSpace space = new BundleClassSpace(FrameworkUtil.getBundle(ArchetypeArtifactManager.class));
+    Injector injector = Guice.createInjector(logginModule,
```
laeubi, line 132:
```suggestion
    Injector injector = Guice.createInjector(logginModule, repositorySystemModule,
```

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/archetype/ArchetypePlugin.java
```diff
@@ -114,8 +125,21 @@ protected void configure() {
...
+        new WireModule(new SpaceModule(space, BeanScanning.INDEX, false)));
+    ArchetypeArtifactManager mgr1 = injector.getInstance(DefaultArchetypeArtifactManager.class);
+    try {
+      ArchetypeArtifactManager mgr2 = injector.getInstance(ArchetypeArtifactManager.class);
```
laeubi, line 136:
```suggestion
      ArchetypeArtifactManager mgr2 = injector.getInstance(ArchetypeArtifactManager.class);
      System.out.println("The manager is: " + mgr2);
      //Nothing injected so we can create them directly...
      InternalCatalogArchetypeDataSource internalCatalog = new InternalCatalogArchetypeDataSource();
      LocalCatalogArchetypeDataSource localCatalog = new LocalCatalogArchetypeDataSource();
```

laeubi (commented):
With the proposed changes I finally was able to use the injector instead of plexus and getting a proper injected instance.

Main problems here:

- `javax.inject` was not imported and so guice was not finding injectable fields / constructors
- We need a module that supplies the `RepositorySystem` required by the downloader

laeubi:
@HannesWell I have pushed my current state and it looks like everything is injected correctly here:
- https://github.com/eclipse-m2e/m2e-core/pull/1951

HannesWell:
> Main problems here:
> 
>     * `javax.inject` was not imported and so guice was not finding injectable fields / constructors
> 
>     * We need a module that supplies the `RepositorySystem` required by the downloader

Yes, thanks for finding them. Since explicitly listing the required imports seems to be very error-prone I not went with the opposite approach and excluded everything that we want to strip of. In the future we can simply inspect the generated Manifest and see what doesn't resolve and can then decide if we want to add it or exclude it.

I have also continued based on your changes and combined the two modules supplied by us and also found a way with that we can continue to use the `archetypeDataSourceMap`.

target-platform/target-platform.target
```diff
@@ -133,60 +133,54 @@
...
+                       !org.apache.commons.collections, \
+                       !org.apache.maven.shared.invoker, \
+                       org.apache.maven.*;provider=m2e, \
+                       org.eclipse.aether.*;provider=m2e, \
+                       org.codehaus.plexus.*;provider=m2e, \
+                       *
+Export-Package:        \
+                       META-INF.sisu;-noimport:=true;x-internal:=true, \
+                       org.apache.maven.archetype.*;-noimport:=true;provider=m2e;mandatory:=provider;version="${version}";x-friends:="org.eclipse.m2e.core.ui", \
+                       org.apache.maven.*;-noimport:=true;provider=m2e;mandatory:=provider;version="${version}";x-friends:="org.eclipse.m2e.core.ui"
```
laeubi, line 183: Do we really want this here? What maven packages are required to be exported beside the archetype ones?

HannesWell: No was just a left-over of the maven-artifact-transfer that was required before. Removed it.

HannesWell:
All tests except those for Windows succeed now but since I'm not sure if this a flaw in the test or an actual problem I'm postponing this to the next release cycle as I have to create the release for the 2025-03 now and have no more time to investigate this now.

PR #1951, opened 2025-03-02
Use Sisu/Guice instead of plexus container for component lookup

laeubi:


laeubi:
Included here now: 
- https://github.com/eclipse-m2e/m2e-core/pull/1942

PR #1956, opened 2025-03-12
Add maven build shortcut to plugin.xml

kufnerg:
Removing this shortcut broke the "shift-alt-x + m" shortcut to simply run a preconfigured maven build. Currently the "shift-alt-x + m" shortcut has no effect what so ever.

### DIFF
--- a/org.eclipse.m2e.launching/plugin.xml
+++ b/org.eclipse.m2e.launching/plugin.xml
@@ -46,6 +46,29 @@
   </extension>
 
    <extension point="org.eclipse.debug.ui.launchShortcuts">
+      <shortcut id="org.eclipse.m2e.core.pomFileAction"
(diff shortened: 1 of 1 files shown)


laeubi (changes requested):
@kufnerg this does not seem the right approach, this adds a new Launch Shortcut just to make a a key combination work, I would expect this to be a keybinding instead of a launch shortcut.

laeubi:
I would go that way, define a keybinding and then run the required actions:

https://www.vogella.com/tutorials/EclipseCommandsKeybindings/article.html

kufnerg:
screenshot
I don't know. I think the actions suggested in the little popup shown in eclipse after "shift-alt-x" should work out of the box without extra configuration !

PR #1958, opened 2025-03-13
Enclose system properties containing white space with quotation marks

### LINKED ISSUE #1904: Surefire configuration of properties with spaces in values constructs incorrect Eclipse configuration
First, thank you for fixing delayed property evaluation in m2e! Now we're on to the next issue ;)

We include a upstream parent pom that declares some helpful system properties for Surefire, like:

```
<pluginManagement>
...
  <plugins>
    <plugin>
      <artifactId>maven-surefire-plugin</artifactId>
      <configuration>
          <systemPropertyVariables combine.children="append">
            <sun.jnu.encoding>${project.build.sourceEncoding}</sun.jnu.encoding>
            <user.timezone>UTC</user.timezone>
            <java.awt.headless>true</java.awt.headless>
            <java.util.logging.SimpleFormatter.format>%1$tY-%1$tm-%1$td %1$tH:%1$tM:%1$tS %4$s %5$s%6$s%n</java.util.logging.SimpleFormatter.format>
         </systemPropertyVariables>
       </configuration>
     </plugin>
   </plugins>
</pluginManagement>
```

This gets faithfully written to Eclipse surefire config:
```
-ea
 -Xmx256m -Dfile.encoding=UTF-8 -Dws.default-timeout=PT5m
-Djava.awt.headless=true
-Djava.util.logging.SimpleFormatter.format=%1$tY-%1$tm-%1$td %1$tH:%1$tM:%1$tS %4$s %5$s%6$s%n
-Dsun.jnu.encoding=UTF-8
-Duser.timezone=UTC
```

Because the format has a space in it, the constructed argument list is not parsed correctly and fails the test:

```
Error: Could not find or load main class %1$tH:%1$tM:%1$tS
Caused by: java.lang.ClassNotFoundException: %1$tH:%1$tM:%1$tS
```

The configuration of properties is XML and thus clearly the value is a single argument - I don't believe adding any quoting to the POM is appropriate, since this is not shell syntax. I think m2e should handle passing property values that contain spaces correctly through to Eclipse.

Thank you and please let me know if I can provide further information.

Paranoja:
# Summary

Fixes #1904

For Surefire and Filesafe plugins with `<systemPropertyVariables />` containing properties with white space characters:
* tab (`U+0009`)
* line feed (`U+000a`)
* carriage return (`U+000d`)
* space (`U+0020`)

wrap values with double quotation marks (`"`).

# Details

I chose these four characters as an intersection of a set of characters, supported by XML, and std::isspace() functions.

Command line reference for `java` command states in [Standard options for java](https://docs.oracle.com/en/java/javase/17/docs/specs/man/java.html#standard-options-for-java)

> -Dproperty=value
> Sets a system property value. The property variable is a string with no spaces that represents the name of the property. The value variable is a string that represents the value of the property. **If value is a string with spaces**, then enclose it in quotation marks (for example -Dfoo="foo bar").

_(emphasis mine)_

that values with spaces must be enclosed, however in section [Using the JDK_JAVA_OPTIONS Launcher Environment Variable](https://docs.oracle.com/en/java/javase/17/docs/specs/man/java.html#using-the-jdk_java_options-launcher-environment-variable) it states:
> The content of the JDK_JAVA_OPTIONS environment variable is a list of arguments separated by white-space characters (as determined by isspace())

and [std::isspace()](https://en.cppreference.com/w/cpp/string/byte/isspace) provides this list:
> * space (0x20, ' ')
> * form feed (0x0c, '\f')
> * line feed (0x0a, '\n')
> * carriage return (0x0d, '\r')
> * horizontal tab (0x09, '\t')
> * vertical tab (0x0b, '\v')

While [000 \[C0 Controls and Basic Latin, C1 Controls and Latin-1 Supplement\]](https://www.w3.org/TR/xml-entity-names/000.html) section of [XML Entity Definitions for Characters (3rd Edition)](https://www.w3.org/TR/xml-entity-names/) specifies, which Unicode characters with code points between `U+00000` and `U+000FF` are valid XML entities.

Manual testing if `java` preserves said whitespace when passed to it through command line with following test program:
```java
package test;

import static java.util.stream.Collectors.joining;

public class SpaceTest {

	public static void main(String[] args) {
		System.getProperties().forEach((name, val) -> {
			if (((String) name).startsWith("prop.")) {
				System.out.append((String) name).println(':');
				System.out.println(((String) val).chars().mapToObj("%02X"::formatted).collect(joining(" ", "\t[", "]")));
			}
		});
	}

}

```
have shown following behavior on Fedora 41 and `OpenJDK Runtime Environment (Red_Hat-21.0.6.0.7-1) (build 21.0.6+7)`:
```bash
java -p ${ws}/test/bin -Dprop.whitespaces="$(printf "{\t\n\v\f\r }")" -m test/test.SpaceTest
```
and output:
```
prop.whitespaces:
        [7B 09 0A 0B 0C 0D 20 7D]
```

### DIFF
--- a/org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.m2e.core.ui;singleton:=true
-Bundle-Version: 2.4.0.qualifier
+Bundle-Version: 2.4.1.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-21
 Bundle-Name: %Bundle-Name
 Bundle-Vendor: %Bundle-Vendor
--- a/org.eclipse.m2e.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.m2e.core;singleton:=true
-Bundle-Version: 2.7.3.qualifier
+Bundle-Version: 2.7.4.qualifier
 Bundle-Activator: org.eclipse.m2e.core.internal.MavenPluginActivator
 Bundle-Vendor: %Bundle-Vendor
 Bundle-Localization: plugin
--- a/org.eclipse.m2e.editor.lemminx/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.editor.lemminx/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: M2E Maven POM File Editor using Wild Web Developer, LemMinX and Maven LS extension (requires Incubating components)
 Bundle-SymbolicName: org.eclipse.m2e.editor.lemminx;singleton:=true
-Bundle-Version: 2.0.900.qualifier
+Bundle-Version: 2.0.901.qualifier
 Automatic-Module-Name: org.eclipse.m2e.xmlls.extension
 Import-Package: javax.inject;version="[1.0.0,2.0.0)",
  org.apache.commons.cli;version="1.6.0",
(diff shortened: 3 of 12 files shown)


Paranoja:
Looks like `org.eclipse.m2e.pde.connector.tests.TychoConnectorTest` is flaky:
* [build no. 11](https://ci.eclipse.org/m2e/job/m2e/job/PR-1958/11/testReport/org.eclipse.m2e.pde.connector.tests/TychoConnectorTest/importTychoPluginWithDS/) -- failed with no apparent reason. 
* [build no. 12](https://ci.eclipse.org/m2e/job/m2e/job/PR-1958/12/testReport/org.eclipse.m2e.pde.connector.tests/TychoConnectorTest/importTychoPluginWithDS/) -- fixed by itself with no changes to the code.

Looks like the PR is finally ready to be reviewed/merged.

holly-cummins:
The Quarkus team would like to see this as well (I was about to submit the exact same PR).

stevenschlansker:
Hallelujah, it works! I confirmed latest snapshot fixes the issue. Thank you @Paranoja

Issue #1979, opened 2025-04-17
Update m2e so that it produces new versions of all "older" artifacts


merks:
Consider `org.eclipse.m2e.apt.core` as an example.

The version currently in my bundle pool is considered unsigned:
```
$jarsigner.exe  -verbose -verify plugins/org.eclipse.m2e.apt.core_2.3.0.20241001-1350.jar

           0 Thu Oct 03 10:09:56 CEST 2024 META-INF/
           0 Thu Oct 03 10:09:56 CEST 2024 org/
           0 Thu Oct 03 10:09:56 CEST 2024 org/eclipse/
           0 Thu Oct 03 10:09:56 CEST 2024 org/eclipse/m2e/
           0 Thu Oct 03 10:09:56 CEST 2024 org/eclipse/m2e/apt/
           0 Thu Oct 03 10:09:56 CEST 2024 org/eclipse/m2e/apt/internal/
           0 Thu Oct 03 10:09:56 CEST 2024 org/eclipse/m2e/apt/internal/compiler/
           0 Thu Oct 03 10:09:56 CEST 2024 org/eclipse/m2e/apt/internal/preferences/
           0 Thu Oct 03 10:09:56 CEST 2024 org/eclipse/m2e/apt/internal/processor/
           0 Thu Oct 03 10:09:56 CEST 2024 org/eclipse/m2e/apt/internal/utils/
           0 Thu Oct 03 10:09:56 CEST 2024 org/eclipse/m2e/apt/preferences/
           0 Thu Oct 03 10:09:56 CEST 2024 META-INF/maven/
           0 Thu Oct 03 10:09:56 CEST 2024 META-INF/maven/org.eclipse.m2e/
           0 Thu Oct 03 10:09:56 CEST 2024 META-INF/maven/org.eclipse.m2e/org.eclipse.m2e.apt.core/
        1189 Thu Oct 03 10:09:56 CEST 2024 META-INF/MANIFEST.MF
    ?    824 Thu Oct 03 10:09:56 CEST 2024 org/eclipse/m2e/apt/MavenJdtAptPlugin.class
    ?  16379 Thu Oct 03 10:09:56 CEST 2024 org/eclipse/m2e/apt/internal/AbstractAptConfiguratorDelegate.class
    ?   7898 Thu Oct 03 10:09:56 CEST 2024 org/eclipse/m2e/apt/internal/AbstractAptProjectConfigurator.class
[...]

PR #1980, opened 2025-04-18
Force qualifier changes to update jar-signed bundles to be PGP-signed

merks:
https://github.com/eclipse-m2e/m2e-core/issues/1979

### DIFF
--- a/org.eclipse.m2e.apt.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.apt.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.m2e.apt.core;singleton:=true
-Bundle-Version: 2.3.0.qualifier
+Bundle-Version: 2.3.100.qualifier
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.27.0,4.0.0)",
  org.eclipse.core.resources,
--- a/org.eclipse.m2e.apt.ui/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.apt.ui/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.m2e.apt.ui;singleton:=true
-Bundle-Version: 2.0.500.qualifier
+Bundle-Version: 2.0.600.qualifier
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.ui,
  org.eclipse.core.runtime,
--- a/org.eclipse.m2e.binaryproject.ui/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.binaryproject.ui/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.m2e.binaryproject.ui;singleton:=true
-Bundle-Version: 2.1.0.qualifier
+Bundle-Version: 2.1.100.qualifier
 Bundle-Vendor: Eclipse.org - m2e
 Bundle-Name: M2E Binary Project UI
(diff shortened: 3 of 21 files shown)


merks:
@akurtakov  @laeubi 

Hmmm, apparently we can't just force a qualifier change.   I'm not sure I understand why not.  What should I do instead?  Increment actual versions in the MANIFEST.MF?

mickaelistria:
>  I'm not sure I understand why not. 

If version x.y.z.qualifier was already released (is available in baseline), then any build of x.y.z.differentQualifier or x.y.z.qualifier but with different content gets rejected because an already released version of x.y.z exists.

So...

> Increment actual versions in the MANIFEST.MF

Yes, for the bundles that were not touched since last release.

> And then fight the feature version increment battle until it's done...

It's likely that most features already got a version bump for other changes, so this fight shouldn't be too long.

merks:
I see things like this:
```
11:35:28  Running org.eclipse.m2e.apt.tests.M2eAptProjectconfiguratorTest
11:50:35  [ERROR] Timeout of 900s exceeded. Process was killed
```
I know I'm seeing my Orbit builds take 3-6 times longer than normal and failing infrastructure problems (427 from dash):

image

So probably the changes are good but the build is kind of unstable right now...

HannesWell:
> I changed all the force qualifiers to be a +1 service increment as suggested on slack by @akurtakov so let's so how that goes.

Yes that should also do the job. Thank you.
The only nit-pick, I usually try to use the same +100 scheme here like in platform etc. I pushed a corresponding update to your branch.

> So probably the changes are good but the build is kind of unstable right now...

I think so as well. Due to the update we have another build and another chance for the tests to pass. If it times out again, I plan to submit this nevertheless.

merks:
I know the 100 thing for service releases, but keep in mind that there is actually no binary change at all, so there cannot really exist ever a service release between these two binary identical build version.  In fact this is kind of a service change of that nature.

mickaelistria:
> I know the 100 thing for service releases

I don't think we ever cared about +0.0.100 in m2e; going by +0.0.1 has always been fine as maintaining multiple stream has always been avoided.

> but keep in mind that there is actually no binary change at all

There are no source change, but just creating new artifacts with a newer version of Tycho and compiler can easily result in different binaries from the same source.
I think if you just want to replace the p2 metadata for an artifact that was already released, it's mostly a matter of reinvoking the p2-metadata goal after the replacement to ensure those metadata get refreshed.

merks:
No, we really need a new version number so that the unsigned content in the pools out in the wild will no longer be used.  Anything less will not work.

merks:
Looks like timeouts again.

HannesWell:
> > I know the 100 thing for service releases
> 
> I don't think we ever cared about +0.0.100 in m2e; going by +0.0.1 has always been fine as maintaining multiple stream has always been avoided.

I tried to keep us this option open in the past, although I think it's unlikely it's ever used respectively I hope we never have to use it.

> I know the 100 thing for service releases, but keep in mind that there is actually no binary change at all, so there cannot really exist ever a service release between these two binary identical build version. In fact this is kind of a service change of that nature.

That's right. But for simplicity I would still apply this pattern here, as otherwise we would have to bump the version on the next code change within this release cycle for each project.

I would be in favour of submitting this now as it is. If one strongly prefers only +1 bumps I can revert to them.

merks:
I have no opinion on the amount of the increment, so 100 is fine with me.  I only care that the problem goes away!  😁

mickaelistria:
100 is as fine as 1, so let's stick to that change.

HannesWell:
Then lets go and hope that the master build succeeds...

HannesWell:
The main-build was finally successful.

laeubi:
> > > I know the 100 thing for service releases
> > 
> > 
> > I don't think we ever cared about +0.0.100 in m2e; going by +0.0.1 has always been fine as maintaining multiple stream has always been avoided.
> 
> I tried to keep us this option open in the past, although I think it's unlikely it's ever used respectively I hope we never have to use it.

I think it does not really matter much about the actual numbers, but [the contribution guide says](https://github.com/eclipse-m2e/m2e-core/blob/main/CONTRIBUTING.md#%EF%B8%8F-version-bump):

> The delta for version bumps are:
>    +0.0.1 (next micro) for a bugfix, or an internal change that doesn't surface to APIs

And currently this is used inconsistently ... so if we want to change this, then the documentation must be updated and the baseline check should be adjusted [to suggest +100](https://tycho.eclipseprojects.io/doc/latest/tycho-baseline-plugin/verify-mojo.html#increment).

PR #1985, opened 2025-05-05
Add SVGs to org.eclipse.m2e.core.ui

Michael5601:
This PR adds SVGs for all icons in the bundle `org.eclipse.m2e.core.ui` except for the following as these do not exist as SVGs yet:

banner_lifecycleMappingPage.svg
banner_new_module.svg
banner_new_project.svg
banner_update_project.svg
m2eclipse48.svg

See also this [PR](https://github.com/eclipse-platform/eclipse.platform/pull/1797) for more information.


HannesWell:
Thanks for this PR. This particular change looks good. Out of curiosity, where did you get the icons from?
I think some the missing icons are contained in:
- https://github.com/eclipse-m2e/m2e-core/tree/main/assets

Do you plan to also contribute SVG images for the icons in?
- org.eclipse.m2e.editor
- org.eclipse.m2e.pde.ui

Some of them are probably just copies from the platform.
I also noticed that m2e still has plenty of gif icons, not only PNG. It would be great to get a replacement for them as well.

And last, but not least, @laeubi what's your opinion about keeping the old png/gif icons? For the Eclipse TLP we kept the icons for now in case some downstream plugins reference them e.g. via corresponding platform-plugin URLs etc. (as resources cannot be really hidden or somehow marked as internal).
For m2e I would be open to just remove the replaced icons immediately, without a deprecation period.
I assume m2e bundles are used less and if there are downstream consumers of the m2e icons, that cannot adapt to SVG immediately, they can still copy the old icons in their bundle.

laeubi:
Some SVGs are also here:

https://github.com/eclipse-platform/eclipse.platform.images/tree/master/org.eclipse.images/eclipse-svg/org.eclipse.m2e.core.ui/icons

beside that, m2e has to wait for the next release before we can start using SVG icons (that's why the build currently fails) but then I don't see why we need to retain old image formats.

Michael5601:
> Out of curiosity, where did you get the icons from?

I got them from the images repository. @laeubi also posted the link:
https://github.com/eclipse-platform/eclipse.platform.images/tree/master/org.eclipse.images/eclipse-svg/org.eclipse.m2e.core.ui/icons

> Do you plan to also contribute SVG images for the icons in?
> 
> * org.eclipse.m2e.editor
> * org.eclipse.m2e.pde.ui
> 

Yes I do. I maintain a [list](https://github.com/eclipse-platform/eclipse.platform.images/issues/129) of missing SVGs that I want to contribute. I will add the two bundles you mentioned to the list. 

> I also noticed that m2e still has plenty of gif icons, not only PNG. It would be great to get a replacement for them as well.

I also mentioned the amount of GIFs that is still used in Eclipse (mostly example/test bundles with low prio) in the above list. I have to check how much effort it will be to create them. Maybe I restrain this task to only creating the most important icons.

Michael5601:
> 
> @Michael5601 can you please update this PR to remove the old icons for which replacements are added with this PR?

Done. :)

HannesWell (approved):
After applying the necessary version bumps the build succeeded.

Thank you for this.

PR #2003, opened 2025-06-11
add proper module support which has been kmssing for over 8 years now,

crowlogic:
y'all are welcome

https://www.reddit.com/r/eclipse/comments/14vv5hm/maven_update_classpath_vs_modulepath/

```
Maven Update - Classpath vs Modulepath
🙋🏻‍♂️ Help Request

I have a number of Java 11 module-enabled projects using maven for build. Every time I run a maven update it moves the Maven Dependencies object from the Modulepath to the Classpath, creating linkage errors. I then have to manually edit properties of the project and move it back to Modulepath, which then fixes the errors.

The issue is referenced in the Stack Overflow post [here](https://stackoverflow.com/questions/62038751/why-and-when-is-eclipse-putting-maven-dependencies-in-the-classpath-and-not-in-t). I've tried to place a modulePath entry in the POM file as described in the answer in that article, but eclipse does not recognize that element under the configuration tree.

I tried updating the version of the maven-compiler-plugin to the latest 3.11.0, which does offer a tantalizing useModulePath entry:

<plugin>
    <version>3.11.0</version>
    <configuration>
        <release>11</release>
        <useModulePath>true</useModulePath>
    </configuration>
</plugin>

But that doesn't appear to change the behavior. The article [here](https://stackoverflow.com/questions/60473309/eclipse-maven-dependencies-moved-under-classpath-after-maven-update-project) claims that moving the entry from Classpath to Modulepath has no effect, but that is demonstrably false because it removes the third attribute line in the .classpath file:

<classpathentry kind="con" path="org.eclipse.m2e.MAVEN2_CLASSPATH_CONTAINER">
    <attributes>
        <attribute name="maven.pomderived" value="true"/>
        <attribute name="org.eclipse.jst.component.nondependency" value=""/>
        <attribute name="module" value="true"/>
    </attributes>
</classpathentry>

This approximately matches the change referenced in the first article:

<classpathentry kind="con" path="org.eclipse.m2e.MAVEN2_CLASSPATH_CONTAINER">
    <attributes>
        <attribute name="maven.pomderived" value="true"/>
        <attribute name="module" value="true"/>
    </attributes>
</classpathentry>

It also causes problems in the eclipse workspace about indirectly referenced class files - the following is an example:

The type software.amazon.awssdk.awscore.client.builder.AwsClientBuilder cannot be resolved.  It is indirectly referenced from required .class files.

If it's not that it's something else - that's just the first error I wrote down to demonstrate the error. My current eclipse version is 2022-06 (4.24.0) and current build ID is 20220609-1112.

I've been having to make the manual change after every rebuild for years - I don't remember it ever working after I moved to maven. I've just ignored it until now, but I'm trying to tie up loose ends before bringing in another developer.

I posted this a few days ago in [r/javahelp](https://www.reddit.com/r/javahelp/), but this sub might be more appropriate.
```

### DIFF
--- a/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
+++ b/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/AbstractJavaProjectConfigurator.java
@@ -28,6 +28,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
+import java.util.Arrays;
 
 import org.osgi.framework.Version;
 import org.slf4j.Logger;
@@ -41,6 +42,7 @@ import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.QualifiedName;
 import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.jdt.core.IAccessRule;
@@ -167,7 +169,7 @@ public abstract class AbstractJavaProjectConfigurator extends AbstractProjectCon
     String executionEnvironmentId = getExecutionEnvironmentId(options);
     addJREClasspathContainer(classpath, executionEnvironmentId);
 
-    addMavenClasspathContainer(classpath);
+    addMavenClasspathContainer(javaProject, classpath);
 
     addCustomClasspathEntries(javaProject, classpath);
 
@@ -259,8 +261,8 @@ public abstract class AbstractJavaProjectConfigurator extends AbstractProjectCon
         environmentId);
     return null;
   }
-
-  protected void addMavenClasspathContainer(IClasspathDescriptor classpath) {
+  
+  protected void addMavenClasspathContainer(IJavaProject javaProject, IClasspathDescriptor classpath) {
     List<IClasspathEntryDescriptor> descriptors = classpath.getEntryDescriptors();
     List<IAccessRule> accessRules = new ArrayList<>();
     boolean isExported = false;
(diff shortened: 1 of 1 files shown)


crowlogic:
remove existing m2e plugin:

```
eclipse/eclipse \
  -application org.eclipse.equinox.p2.director \
  -noSplash \
  -uninstallIU org.eclipse.m2e.feature.feature.group
```

install:

```
eclipse/eclipse \
  -application org.eclipse.equinox.p2.director \
  -noSplash \
  -repository file:///wherever/git/m2e-core/org.eclipse.m2e.repository/target/repository \
  -installIU org.eclipse.m2e.feature.feature.group
```

crowlogic:
It's functional enough for me. I've done what I can. It's too much of a
nightmare to work with this monstrous. Build system, there's not really
enough return. In it for me, so I'm just gonna leave it here. So that the
next hapless victim does some research at least finds a solution rather
than stares into the void like I did. And no, the author name was not
wrong. I said that purposely because I can

mickaelistria:
>  It's not gonna pay off more than the amount of pain

You already faced most of the pain. Approving the ECA on the other hand is a simple click to a disclaimer that basically tells that Eclipse Foundation is not individually responsible if some illegal (stolen or intentionally introducing security threats) code is merged in.
We (maintainers) can help dealing with versioning or other small things in code to merge this in at some point, but we really are not allowed to merge anything involving your contribution unless you agree on signing the ECA.

crowlogic:
after struggling far too long trying to get this baseline check crap to go away. *every single configuration flag that existed* failed to disable it, i had to remove it from the m2e-parent pom completely. 

This should probabably be noted in the CONTRIBUTION.md lest anyone else tries to work on something and gets furious and smashes the laptop to pieces because the failure of a rather long build process that only lets you know whatever you tried didnt work after 5 minutes of nearly setting the machine on fire to compile all this stuff

crowlogic:
To be clear, it's not like you are helping me unless I'm getting paid for
this. Just like I said, I have no symptom. This is just free work. I'm
offering to do if someone else can't pick up a piece of an obviously
working patch. Then so bad it stays there. It's really no skin off my back

crowlogic:
You are the one that asked me to do extra work' if I can't get a little
help at least before signing the e, c, a, then it's not even worth signing

PR #2020, opened 2025-08-11
Update 'm2e-core-tests' submodule

HannesWell:


HannesWell:
Ideally this awaits https://github.com/eclipse-m2e/m2e-core-tests/pull/1, so that we can have that updated included.

PR #2041, opened 2025-09-29
Update to Tycho 5

laeubi:


.github/workflows/maven.yml
```diff
@@ -10,8 +10,12 @@ on:
      - 'main'
 
 jobs:
+  check-versions:
+    uses: eclipse-platform/eclipse.platform.releng.aggregator/.github/workflows/checkVersions.yml@master
+    with:
+      botName: Eclipse M2Eclipse Bot
+      botMail: m2e-bot@eclipse.org
```
HannesWell: This is now in-cooperated at other places.
```suggestion
```

org.eclipse.m2e.jdt/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.m2e.jdt;singleton:=true
-Bundle-Version: 2.4.101.qualifier
+Bundle-Version: 2.4.102.qualifier
```
HannesWell: Ideally we let the version increment workflow do it's thing now?
And also use the +100 schema that's implemented in that workflow.

laeubi: This is done already by the bot and as always I can only reference to the documentation here that m2e uses +1 increments: 

https://github.com/eclipse-m2e/m2e-core/blob/main/CONTRIBUTING.md#%EF%B8%8F-version-bump

.github/workflows/maven.yml
laeubi: reverted...

org.eclipse.m2e.jdt/META-INF/MANIFEST.MF
HannesWell: > This is done already by the bot 

In fact the now used version-increment has the +100 schema implemented:

> I can only reference to the documentation here that m2e uses +1 increments:

Then lets update it:
- https://github.com/eclipse-m2e/m2e-core/pull/2055

laeubi:
Warnings are unavoidable at the moment.

PR #2048, opened 2025-10-17
Restore build shortcut

### LINKED ISSUE #1618: Maven build shortcut not working
In eclipse 2023-06 the maven build shortcut (alt+shift+x+m) is ok, mas upgrande to version 2023-09 and 2023-12 the shortcut doesnt work anymore.

kwin:
This closes #1618

### DIFF
--- a/org.eclipse.m2e.launching/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.launching/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.m2e.launching;singleton:=true
-Bundle-Version: 2.1.101.qualifier
+Bundle-Version: 2.1.102.qualifier
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.core.runtime;bundle-version="[3.27.0,4.0.0)",
  org.eclipse.core.variables,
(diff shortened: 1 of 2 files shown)


kwin:
>  is not what we want:

Well, this seem to be mostly you. 

> instead one should register a keybinding with whatever shortcut we like.

To what action? The action allowing to select a Maven execution is perfectly fine for the shortcut and desired by most customers.

laeubi:
> In that case, the old behavior gave me a nice dialog that was maneuverable with arrow key and enter to swiftly select the build i want.

Nothing prevents you from show that dialog with the proposed solution in https://github.com/eclipse-m2e/m2e-core/pull/2049 that would maybe even make it simpler. Launching just the last was just a very first example if something else work better fine with that.

So fetch the PR, adjust it as you think it best fit, create a PR and we can merge it. If it happens before the end of November it will be part of the release (and before part of the snapshots).

> I still dont relay see why this old behavior is perceived as "dirty" in the first place.

Because what we see in the poup in the first place is what you will see when you click the Maven build (+ you get the behavior from the ... one right below so it is confusing to do something else after you used it once), so it is redundant, clutters the UI, inconsistent and only a few items get Shortcuts so we want to keep that list short.

> I still feel like listing all maven build in the popup is more trouble than its worth, also since it is pretty much permanently overlooked.

This feature is not provided by m2e but the Eclipse platform and many people like that... so everyone has different workflows and like different things.

> All it did was cost me lots of time 

 Then you might want to invest some times here and implement a permanent and consistent solution... or wait unless someone else picks it up if it is not _THAT_ important to you...

> explaining to our trainees over and over again, that clicking "Maven Build" will create a new launch config every time, which is all but obvious.

Beside that the three dots consistently in Eclipse and other UIs indicate a new dialog opens, so one would hope not to need explain it to everyone again.

PR #2053, opened 2025-10-18
Use Maven 4 as the runtime for m2eclipse

laeubi:
Maven 4 will soon be released and should be able to execute Maven 3 projects as well. We should therefore upgrade m2e to use Maven 4 from now on.

For users that require Maven 3.9 line we can offer the latest 3.9.11 maven runtime as an additional install option like we already did for Maven 3.9.6


stbischof:
Whooop whooop!!!!! Thank you a lot for your work

laeubi:
We get a lot of deprecation warnings now but that is expected and we should fix those later on.
We maybe even just accept those for now to retain backward compatible for users that choose to use an older runtime.

laeubi:
@HannesWell this is currently only a draft so may contain some changes that must be cleaned up before a final version. I now have made it to compile again with Maven4 but possibly retain backwards compatible with Maven 3.9.x ...

laeubi:
I added a null check and a catch for the unsupported operation, but both seem to indicate some change in Maven 4 that we need to reconsider how m2e works today, but lets see how many tests might actually fail...

laeubi:
We are currently hit by this bug:
- https://github.com/apache/maven/issues/11552

laeubi:
Upgraded to Maven 4-rc6 now. Now we have 182 failing test (-32)

laeubi:
I fixed one issue with diverging dependencies adn now at laest a basic maven 4 test are working

https://ci.eclipse.org/m2e/job/m2e/job/PR-2053/38/testReport/org.eclipse.m2e.core/Maven4Test/

grafik

but there seems to be a lot more to do to make the old tests happy...

laeubi:
I currently think given the gap between M3+M4 that probably instead of my idea here to allow people switch runtime (what can be problematic if one has a mixture of 3+4) that instead we must have these runtimes run in parallel.
The one needs to decide *per project* what runtime must be used, so one has e.g. one maven-4 and one maven-3 runtime. This might influence workspace resolution  but seems acceptable.

This would then mean that a runtime not only offers e.g. maven classes + facades (like here) but e.g. an own implementation of `IMaven` / `IMavenProjectFacade` / `IMavenExectuion` / ... and at best all other code operates on only these interfaces.

As a very basic step this would go into this direction:

- https://github.com/eclipse-m2e/m2e-core/pull/2221
- https://github.com/eclipse-m2e/m2e-core/pull/2222

so we decouple everything from the maven-core API systematically.

laeubi:
I've proposed this work item to the [IDE-WG Development Funding Initiative](https://eclipseide.org/working-group/development-funding-initiative/) for potential funding:

- https://gitlab.eclipse.org/eclipse-wg/ide-wg/ide-wg-dev-funded-efforts/ide-wg-dev-funded-program-planning-council-top-issues/-/work_items/47

Please keep expectations in check — funding decisions and subsequent work will likely take some time. That said, anyone interested in helping move this forward is of course welcome, and funding from outside the IDE-WG program is always appreciated too. And if you'd like to have a direct say in prioritizing efforts like this one, consider [joining the IDE-WG](https://eclipseide.org/working-group/#about)!

PR #2054, opened 2025-10-18
Run version increment workflow on PRs and result publication

HannesWell:
Follow-up on
- https://github.com/eclipse-m2e/m2e-core/pull/2052


HannesWell:
My external testing makes me sufficient enough that this now works as intended.
Lets see if it really does.

PR #2055, opened 2025-10-18
Clarify and set up usage of +100 increments for service version segments

HannesWell:
The workflow for automated version increments also uses the +100 increment schema for the service version.


HannesWell:
> Why does contributing MD change require version bumps?!?

I assume it's because this also changes the parent pom, which then propagates somehow to the features?

laeubi:
Seems resonable...

> > Why does contributing MD change require version bumps?!?
> 
> I assume it's because this also changes the parent pom, which then propagates somehow to the features?

Okay that makes sense.

PR #2073, opened 2025-10-25
Restore maven build shortcut without adding a toolbar button

### LINKED ISSUE #1618: Maven build shortcut not working

Omega1001:
Reconnected the shortcut to an action that calls the previous code.
This is working without bringing the Button back, as Requested in #2048 
Also closes #1618 

Restores the previous behavior of the shortcut: (be reconnecting the old code)
If no maven launch configuration exists: Open the Create Dialog.
If exactly one configuration exists: Launch that configuration.
If more than one exists: Open a filterable/maneuverable Dialog to select which one you want.

@laeubi How's this?

### DIFF
--- a/org.eclipse.m2e.launching/plugin.xml
+++ b/org.eclipse.m2e.launching/plugin.xml
@@ -34,6 +34,11 @@
              description="%m2.shortcut.description.generate-sources"
              categoryId="org.eclipse.debug.ui.category.run"/>
   </extension>
+  
+  <extension point="org.eclipse.ui.handlers">
+    <handler commandId="org.eclipse.m2e.core.pomFileAction.run"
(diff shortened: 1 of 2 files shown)


Omega1001:
> @Omega1001 many thanks for looking into this, you need to sign the ECA so we can accept your contribution, but as all tests pass already it seems working fine apart from that!

Already did right after the push.
Github account is also linked.
Guess the update is a poll, so it should pick that up any time now.

laeubi (approved):
Looks fine to me, lets try!

PR #2080, opened 2025-11-07
Use Tycho 5.0.1-SNAPSHOT

laeubi:
As currently the target update fails


laeubi:
Only one new compiler warning (maybe due to upgrade in JDT compiler)

laeubi:
Target Update flow works again!

- https://github.com/eclipse-m2e/m2e-core/pull/1923

Issue #2076, opened 2025-10-27
Automatic installation of the development environment fails


kevloral:
When setting up the development environment using the Eclipse installer, the process fails because, while the _m2e-core_ repository uses _main_, _m2e-core-tests_ uses _master_:

> Executing startup tasks
[..]
Performing Preference /instance/org.eclipse.oomph.setup.ui/showToolBarContributions = true
Performing JRE for JavaSE-17 = F:\Java\jdk\21
[--]
Performing Git Clone https://github.com/eclipse-m2e/m2e-core.git  (main)
Cloning Git repo https://github.com/eclipse-m2e/m2e-core.git to F:\Java\eclipse\m2e-main\git\m2e-core
[..]
Configure https://github.com/eclipse-m2e/m2e-core.git
Setting autocrlf = true
Creating local branch main
Resetting hard
Cloning Git repo https://github.com/eclipse-m2e/m2e-core-tests.git to F:\Java\eclipse\m2e-main\git\m2e-core\m2e-core-tests
java.lang.Exception: java.lang.Exception: org.eclipse.jgit.api.errors.TransportException: Remote branch 'main' not found in upstream origin
[..]
Caused by: org.eclipse.jgit.api.errors.TransportException: Remote branch 'main' not found in upstream origin
  at org.eclipse.jgit.api.FetchCommand.call(FetchCommand.java:249)
  at org.eclipse.jgit.api.CloneCommand.fetch(CloneCommand.java:322)

PR #2082, opened 2025-11-10
The m2e-core-tests Git repository is now imported from master

kevloral:
Related to #2076, I have changed the _m2e.setup_ file so that the automatic installation of the development environment works again. Now, the _m2e-core-tests_ Git repository is imported separately from _master_. 

Obviously, this change should be undone once _m2e-core-tests_ is moved from _master_ to _main_.

### DIFF
--- a/setup/m2e.setup
+++ b/setup/m2e.setup
@@ -98,8 +98,7 @@
   <setupTask
       xsi:type="git:GitCloneTask"
       id="git.clone.m2e.core"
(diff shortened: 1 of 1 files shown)


kevloral:
If changing the branch is as easy as opening a ticket there, that is definitely better than modifying the m2e.setup file. I will do so and, therefore, I am closing this PR.

laeubi:
Yes it should be possible and then only needs approval, so let me know and I'll do that!

PR #2099, opened 2025-12-05
Retrieve effective instead of raw model when importing projects

### LINKED ISSUE #1948: "Import existing Maven projects" does not recognize modules leveraging interpolated properties
When executing "Import (Existing Maven Project)" in a Maven project having several modules which leverage property values with interpolation, those are not detected.

For example for a multi-module pom like

```
...
<properties>
    <subproject.abbreviation>test</subproject.abbreviation>
</properties>

<modules>
    <module>${subproject.abbreviation}-core</module>
    <module>${subproject.abbreviation}-config</module>
    <module>${subproject.abbreviation}-frontend</module>
</modules>
...
```

the modules are not detected

kwin:
This will also interpolate correctly.
Use ProjectBuilder instead of ModelProcessor to parse the POM. Encapsulate in a dedicated MavenExecutionContext.

This closes #1948

### DIFF
--- a/org.eclipse.m2e.feature/feature.xml
+++ b/org.eclipse.m2e.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.m2e.feature"
       label="%featureName"
-      version="2.10.0.qualifier"
+      version="2.11.0.qualifier"
       provider-name="%providerName"
       plugin="org.eclipse.m2e.core"
       license-feature="org.eclipse.license"
--- a/org.eclipse.m2e.scm/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.scm/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.m2e.scm;singleton:=true
-Bundle-Version: 2.1.100.qualifier
+Bundle-Version: 2.1.200.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-21
 Bundle-Name: %Bundle-Name
 Bundle-Vendor: %Bundle-Vendor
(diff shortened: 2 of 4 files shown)


laeubi:
@kwin please be aware that at some places we might need the raw model as we can not create a full maven session now. Also interpolation is maybe not always wanted. So if project import is the use-case here, I would suggest to

1. create a testcase for this particular case
2. only change it there by starting a session beforehands

kwin:
> at some places we might need the raw model as we can not create a full maven session now. Also interpolation is maybe not always wanted

Can you elaborate? What are those places? Under which circumstances you don't want interpolation?

org.eclipse.m2e.core/src/org/eclipse/m2e/core/embedder/MavenModelManager.java
```diff
@@ -106,13 +105,26 @@ public org.apache.maven.model.Model readMavenModel(InputStream reader) throws Co
...
-   * Read the model from the provided pom file (might not exits) pointer
+   * Shortcut for {@link #readMavenModel(File, IProgressMonitor) with 2nd argument being {@code null}.
    * 
-   * @param pomFile the file pointer
-   * @return a maven model, or <code>null</code> if the pointer do not point to any valid maven directory
-   * @throws CoreException if the file points to a valid maven directory but the pom could not be read
+   * @deprecated Use {@link #readMavenModel(File, IProgressMonitor)} instead.
    */
+  @Deprecated
   public org.apache.maven.model.Model readMavenModel(File pomFile) throws CoreException {
```
kwin: @laeubi Are you suggesting to keep this as is (i.e. raw model) and rather introduce a new method for returning the effective model? The only consumer for the raw model I see within m2e is probably org.eclipse.m2e.tests.common/src/org/eclipse/m2e/tests/common/AbstractMavenProjectTestCase.java...

PR #2115, opened 2026-01-02
Add Option to Rewrite Existing Manifest for Maven Artifacts

weemve:
**Problem**

Currently, the .target platform supports generating an OSGi manifest when a Maven artifact is missing one, using the option generate. However, there is no way to modify or regenerate an existing manifest for Maven-based libraries.

When migrating third-party libraries to Maven locations in .target files, many artifacts already include a manifest that does not meet our requirements (e.g., incomplete OSGi metadata). We need a mechanism to rewrite these manifests without manually editing the JARs.

**Proposed Solution**

Introduce a new option to rewrite a manifest file in the _Maven Artifact Target Entry Dialog_:

image

The new option _generate_rewrite_ behaves like the current generate mode for missing manifests, but applies even when a manifest already exists. So, if _generate_rewrite_ is enabled, the manifest will be regenerated based on the same logic used for missing manifests. This allows consistent handling of Maven artifacts during migration without manual intervention.

**Benefits**

- Simplifies migration of third-party libraries to Maven-based .target definitions.
- Ensures uniform OSGi metadata across all artifacts.
- Reduces manual work and risk of errors when modifying manifests.

This Pull Request is related to [https://github.com/eclipse-pde/eclipse.pde/pull/2164](https://github.com/eclipse-pde/eclipse.pde/pull/2164)

### DIFF
--- a/org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/MavenSourceBundle.java
+++ b/org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/MavenSourceBundle.java
@@ -35,13 +35,14 @@ public class MavenSourceBundle extends TargetBundle {
 		try (JarFile jar = new JarFile(sourceFile)) {
 			manifest = Objects.requireNonNullElseGet(jar.getManifest(), Manifest::new);
 		}
+		File sourceBundleFile = MavenBundleWrapper.getEclipseSourceBundle(sourceFile, manifest, symbolicName, version);
 		if (MavenBundleWrapper.isValidSourceManifest(manifest)) {
-			fInfo.setLocation(sourceFile.toURI());
+			fInfo.setLocation(sourceBundleFile.toURI());
 		} else {
 			File generatedSourceBundle = cacheManager.accessArtifactFile(artifact, file -> {
-				if (CacheManager.isOutdated(file, sourceFile)) {
+				if (CacheManager.isOutdated(file, sourceBundleFile)) {
 					MavenBundleWrapper.addSourceBundleMetadata(manifest, symbolicName, version);
-					MavenBundleWrapper.transferJarEntries(sourceFile, manifest, file);
+					MavenBundleWrapper.transferJarEntries(sourceBundleFile, manifest, file);
 				}
 				return file;
 			});
--- a/org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/shared/MavenBundleWrapper.java
+++ b/org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/shared/MavenBundleWrapper.java
@@ -92,22 +92,26 @@ public class MavenBundleWrapper {
 	/**
 	 * Wraps an artifact (and possible its dependents if required) to produce a
 	 * manifest with OSGi metadata.
-	 * 
-	 * @param artifact           the artifact to wrap
-	 * @param instructionsLookup a lookup for bnd instructions
-	 * @param repositories       the repositories that should be used to resolve
-	 *                           dependencies
-	 * @param repoSystem         the repository system for lookup dependent items
-	 * @param repositorySession  the session to use
-	 * @param syncContextFactory the sync context factory to acquire exclusive
-	 *                           access to the wrapped artifact and its dependencies
+	 *
+	 * @param artifact               the artifact to wrap
+	 * @param instructionsLookup     a lookup for bnd instructions
+	 * @param repositories           the repositories that should be used to resolve
+	 *                               dependencies
+	 * @param repoSystem             the repository system for lookup dependent
+	 *                               items
+	 * @param repositorySession      the session to use
+	 * @param syncContextFactory     the sync context factory to acquire exclusive
+	 *                               access to the wrapped artifact and its
+	 *                               dependencies
+	 * @param ignoreExistingMetadata if true, the manifest file is forced to be
+	 *                               overridden
 	 * @return the wrapped artifact
 	 * @throws Exception if wrapping the artifact fails for any reason
 	 */
 	public static WrappedBundle getWrappedArtifact(Artifact artifact,
 			Function<DependencyNode, Properties> instructionsLookup, List<RemoteRepository> repositories,
 			RepositorySystem repoSystem, RepositorySystemSession repositorySession,
-			SyncContextFactory syncContextFactory) throws Exception {
+			SyncContextFactory syncContextFactory, boolean ignoreExistingMetadata) throws Exception {
 		CollectRequest collectRequest = new CollectRequest();
 		collectRequest.setRoot(new Dependency(artifact, null));
 		collectRequest.setRepositories(repositories);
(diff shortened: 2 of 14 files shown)


laeubi (changes requested):
Thank you for taking the time to create this pull request and for the detailed explanation of the use case!  I appreciate the effort you've put into this solution.

However, after careful consideration, I don't think we can accept this change. Let me explain the reasoning behind this decision:

## The Challenge with Modifying Existing OSGi Manifests

While I understand the convenience this feature would provide, allowing modifications to existing OSGi manifests goes against some core principles we try to maintain:

### 1. **Encouraging Upstream Fixes**
When an OSGi manifest contains incorrect or incomplete metadata, the right solution is to fix it at the source. Providing an easy workaround risks making these temporary fixes permanent, meaning the upstream library never gets properly corrected.  This ultimately affects the entire ecosystem, not just individual projects.

### 2. **Artifact Identity and Consistency**
A fundamental principle in dependency management is that a given GAV (Group, Artifact, Version) coordinate should represent the same artifact everywhere. When we allow manifest modifications, the same GAV could exist "in the wild" with different OSGi metadata depending on where and how it was processed.  This leads to: 
- Hard-to-diagnose dependency conflicts
- Reproducibility issues across different environments
- Confusion when sharing artifacts or debugging issues with others

### 3. **Hidden Technical Debt**
Making it easy to modify manifests can mask underlying problems rather than solving them. What appears to be a quick fix today can become a maintenance burden that's difficult to track and understand months or years later.

## What We Recommend Instead

For cases where you absolutely need to work with problematic artifacts: 

- **Short-term/prototyping**: The current approach of wrapping artifacts *without* existing OSGi manifests can serve as a temporary solution while working toward proper upstream support
- **Long-term solution**: Consider helping the library maintainers adopt proper OSGi metadata, or if necessary, create a properly forked/wrapped version under your own GAV using tools like [tycho-wrap: wrap](https://tycho.eclipseprojects.io/doc/latest/tycho-wrap-plugin/wrap-mojo.html)

This way, you have full control and transparency about what's been modified, and you're not creating confusion about what the original GAV represents. 

## Final Thoughts

I realize this might not be the answer you were hoping for, and I genuinely appreciate the thought you put into this feature. The goal isn't to make things harder, but to maintain principles that help keep the OSGi ecosystem healthy and predictable in the long run.

If you'd like to discuss alternative approaches or have questions about the recommended solutions, I'm happy to continue the conversation!

org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/MavenSourceBundle.java
```diff
@@ -24,8 +24,9 @@
 
 public class MavenSourceBundle extends TargetBundle {
 
-	public MavenSourceBundle(BundleInfo sourceTarget, Artifact artifact, CacheManager cacheManager) throws Exception {
-		this.fSourceTarget = sourceTarget;
+	public MavenSourceBundle(BundleInfo sourceTarget, Artifact artifact, CacheManager cacheManager,
+			boolean manifestOverride) throws Exception {
```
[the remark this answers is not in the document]
weemve: In our opinion, Manifest override is not irrelevant for source bundles. When regenerating artifacts in cases where no manifest exists yet, the binary bundle and its source bundle are created with symmetrical Bundle‑SymbolicNames, ensuring consistent naming from the start. If we apply manifestOverride (for already existing bundles) only to the binary bundle, we introduce a name asymmetry in the target platform: the binary receives a custom BSN while the source bundle keeps the original one. This breaks the expected 1:1 mapping. To preserve the same symmetry as in the “missing manifest: generate” scenario, the override must be mirrored for source bundles as well, ideally by deriving the source bundle’s BSN from the overridden BSN.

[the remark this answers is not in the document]
laeubi: I didn't say that the source bundle itself does not need to be different. But weather or not it has a rewritten manifest should not matter. The `BundleInfo sourceTarget` (if remembering correctly) should carry the information to what attach the source too. Also currently the flag is not considered but simply disables the caching as a whole while the cache key itself should already contain whats needed here (e.g. bsn+version).
Otherwise different target locations might compete for the same file. Also the check if a manifest is valid or not should not matter if it was generated or not.

org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/shared/MavenBundleWrapper.java
```diff
@@ -101,13 +101,15 @@ private MavenBundleWrapper() {
 	 * @param repositorySession  the session to use
 	 * @param syncContextFactory the sync context factory to acquire exclusive
 	 *                           access to the wrapped artifact and its dependencies
+	 * @param manifestOverride   if true, the manifest file is forced to be
+	 *                           overridden
```
laeubi: Reading through the PR, I think in general `manifestOverride` is a bad naming - we do not really override something, a better wording would be `ignoreExistingMetadata` because that's what effectively is done here.

org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/shared/MavenBundleWrapper.java
```diff
@@ -207,7 +210,7 @@ private static WrappedBundle getWrappedNode(DependencyNode node,
 		List<DependencyNode> children = node.getChildren();
 		List<WrappedBundle> depends = new ArrayList<>();
 		for (DependencyNode child : children) {
-			depends.add(getWrappedNode(child, instructionsLookup, visited));
+			depends.add(getWrappedNode(child, instructionsLookup, visited, manifestOverride));
```
laeubi:
```suggestion
			depends.add(getWrappedNode(child, instructionsLookup, visited, false));
```

We should not override (or as mentioned better ignore) the metadata for any dependencies as we have said this should only be valid for the one and only artifact wrapped here!

laeubi (commented):
Thanks for you patience!
I have added some review comments and think we are quite close now to complete this.

org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/shared/MavenBundleWrapper.java
weemve: Good point - I adjusted all locations in the code

laeubi (commented):
I think it looks good in general (except the thing with the exception), so if you can adjust this, then squash everything into one commit and rebase it on main I think we are fine.

laeubi (approved):
It all now looks good to me!

laeubi:
The jenkins warnings just changed file numbers due to the change but are pre-existing!

weemve:
> @weemve thanks for your patience and responsiveness.
> 
> The change should appear shortly in the snapshot builds. Please now also supply the Tycho side of the change here so we have the same support on the CI as well!

Thank you, too! Yes, we will provide the change on the Tycho side very soon.

PR #2116, opened 2026-01-02
Add Surefire Provider Classpath to Eclipse test classpath

### LINKED ISSUE #2112: JUnit Vintage Engine not added to test classpath if only part of maven-surefire-plugin's plugin classpath
In https://maven.apache.org/surefire/maven-surefire-plugin/examples/junit.html there are several examples where the [JUnit engine](https://docs.junit.org/6.0.1/advanced-topics/engines.html) is given via the m-surefire-plugin's *plugin classpath* (i.e. via `dependencies` below `plugin`). However those are never added to Eclipse's test classpath, therefore executing JUnit4 tests with JUnit5 runner fails. Only Maven's [test classpath](https://maven.apache.org/surefire/maven-surefire-plugin/examples/configuring-classpath.html) is considered to populate Eclipse's test classpath (which notably does not include m-s-p's plugin class ;

Compare also with https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2713.

OTOH the Eclipse built-in JUnit5 library consists of the following engines:
1. junit-platform-suite-engine_1.14.1.jar
2.  junit-vintage-engine_1.14.1.jar
 
In order to allow execution of JUnit4 and JUnit5 tests in the same Eclipse Test execution I would suggest to also add at least the engine dependencies (from the plugin's classpath) to the Eclipse test class path as well.

kwin:
Deprecate duplicate constants (defined in
org.eclipse.jdt.core.IClasspathAttribute)

This closes #2112

### DIFF
--- a/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/launch/MavenRuntimeClasspathProvider.java
+++ b/org.eclipse.m2e.jdt/src/org/eclipse/m2e/jdt/internal/launch/MavenRuntimeClasspathProvider.java
@@ -289,12 +289,12 @@ public class MavenRuntimeClasspathProvider extends StandardClasspathProvider {
       try {
         Version version = new GenericVersionScheme().parseVersion(mojoExecution.getPlugin().getVersion());
         if(version.compareTo(MIN_SUREFIRE_VERSION) < 0) {
-          log.debug(
+          log.error(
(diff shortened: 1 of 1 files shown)


laeubi:
>  I cannot reproduce the test issues locally. 

You mean in the IDE or when running maven?

PR #2123, opened 2026-01-13
Build with Tycho 5.0.2

HannesWell:


HannesWell:
> Therefore the following files need a version increment:
> 
> ```
> org.eclipse.m2e.sdk.feature/feature.xml
> ```

@laeubi it looks like that the removal of sources is still considered a breaking change.

So the skip of the baseline check for the sdk feature cannot be reverted yet.

PR #2139, opened 2026-02-18
Fix NPE when Model.getLocation() returns null

### LINKED ISSUE #1310: extensions in extensions.xml cause m2e not to update project in Eclipse
Observed on Eclipse 4.26/m2e 2.2.1 on MacOS (but this issue has been present for several Eclipse/M2E releases)

I work on a project which has a `.mvn/extensions.xml` configuration file and is causing Eclipse to fail importing and/or updating the project with the following error message:

```
Could not lookup required component
com.google.inject.ProvisionException: Unable to provision, see the following errors:

1) Error injecting: public void org.apache.maven.eventspy.internal.EventSpyDispatcher.setEventSpies(java.util.List)
  at ClassRealm[maven.ext., parent: ClassRealm[plexus.core, parent: null]] (via modules: org.eclipse.sisu.wire.WireModule -> org.eclipse.sisu.plexus.PlexusBindingModule)
  while locating org.apache.maven.eventspy.internal.EventSpyDispatcher
    for field at org.apache.maven.internal.aether.DefaultRepositorySystemSessionFactory.eventSpyDispatcher(Unknown Source)
  while locating org.apache.maven.internal.aether.DefaultRepositorySystemSessionFactory
  while locating org.apache.maven.DefaultMaven
  at ClassRealm[maven.ext., parent: ClassRealm[plexus.core, parent: null]] (via modules: org.eclipse.sisu.wire.WireModule -> org.eclipse.sisu.plexus.PlexusBindingModule)
  while locating org.apache.maven.Maven

1 error
      role: org.apache.maven.Maven
  roleHint: 
```

If I comment the content of the file or remove it, Eclipse has no problem importing the project.

Attaching a small test project to reproduce the issue: [m2e-lifecycle-issue.tar.gz](https://github.com/eclipse-m2e/m2e-core/files/10935313/m2e-lifecycle-issue.tar.gz)

EricBlanquer:
## Summary

Maven core extensions such as `maven-git-versioning-extension` modify the Maven model at runtime without preserving `InputLocation` metadata. This causes `Model.getLocation("")` to return `null`, leading to `NullPointerException` during project import and update.

**Stack trace:**
```
java.lang.NullPointerException: Cannot invoke "org.apache.maven.model.InputLocation.getSource()"
  because the return value of "org.apache.maven.model.Model.getLocation(Object)" is null
    at o.e.m2e.core.internal.lifecyclemapping.AnnotationMappingMetadataSource.parsePIs(AnnotationMappingMetadataSource.java:174)
```
```
java.lang.NullPointerException: Cannot invoke "org.apache.maven.model.InputLocation.getSource()"
  because "inputLocation" is null
    at o.e.m2e.core.internal.markers.SourceLocationHelper.getSourceLocation(SourceLocationHelper.java:243)
```

## Changes

- `AnnotationMappingMetadataSource`: add null checks in constructor and `parsePIs()` for `Model.getLocation()` and `InputLocation.getSource()`
- `SourceLocationHelper`: add null checks in `getSourceLocation()`, `findLocation()` and `getMavenDependency()` for all `getLocation(SELF)` dereferences

## Reproducer

Use any Maven core extension declared in `.mvn/extensions.xml` (e.g. `maven-git-versioning-extension`) and try to import or update a Maven project in Eclipse.

Fixes #1310

### DIFF
--- a/org.eclipse.m2e.core.tests/resources/projects/gitVersioningExtension/.mvn/extensions.xml
+++ b/org.eclipse.m2e.core.tests/resources/projects/gitVersioningExtension/.mvn/extensions.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<extensions>
+	<extension>
+		<groupId>me.qoomon</groupId>
+		<artifactId>maven-git-versioning-extension</artifactId>
+		<version>9.11.0</version>
+	</extension>
+</extensions>
--- a/org.eclipse.m2e.core.tests/resources/projects/gitVersioningExtension/.mvn/maven-git-versioning-extension.xml
+++ b/org.eclipse.m2e.core.tests/resources/projects/gitVersioningExtension/.mvn/maven-git-versioning-extension.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration xmlns="https://github.com/qoomon/maven-git-versioning-extension"
+               xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+               xsi:schemaLocation="https://github.com/qoomon/maven-git-versioning-extension https://qoomon.github.io/maven-git-versioning-extension/configuration.xsd">
+	<refs>
+		<ref type="branch">
+			<pattern>.*</pattern>
+			<version>${version}</version>
+		</ref>
+	</refs>
+</configuration>
--- a/org.eclipse.m2e.core.tests/resources/projects/gitVersioningExtension/pom.xml
+++ b/org.eclipse.m2e.core.tests/resources/projects/gitVersioningExtension/pom.xml
@@ -1,0 +1,6 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<groupId>test</groupId>
+	<artifactId>git-versioning-test</artifactId>
+	<version>1.0.0-SNAPSHOT</version>
+</project>
--- a/org.eclipse.m2e.core.tests/src/org/eclipse/m2e/core/ExtensionsTest.java
+++ b/org.eclipse.m2e.core.tests/src/org/eclipse/m2e/core/ExtensionsTest.java
@@ -13,6 +13,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
+import java.io.File;
 import java.io.IOException;
 import java.util.Collection;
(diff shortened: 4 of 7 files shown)


laeubi:
@EricBlanquer can you try to address the possible NPE marked by the compiler? The code itself is rather complex already, can you probably try if it is possible to extract some parts into individual methods what the would make an early exit in case of null maybe more easier.

EricBlanquer:
@laeubi Addressed in the latest push (force-pushed commit `a2cfdfa2`):

**NPE warning fix** — `getMavenDependency()` now uses `Objects.equals(projectId, depId)` with a `depId != null` guard before calling `findDependencyImport()`, which also protects the inner `depId.equals(importId)`.

**Extracted helpers in `SourceLocationHelper`** for clearer null early-exits:
- `getModelId(InputLocationTracker)` — replaces the verbose `getLocation(SELF).getSource().getModelId()` chain, returns null safely
- `isInCurrentPom(InputLocation, File)` — deduplicates the pom-match check used in both `findLocation(MojoExecutionKey)` and `findLocation(Dependency)`
- `resolveParentPomLocation(MavenProject, SourceLocation)` — unifies parent pom navigation with early-exits for both `parent == null` and `parent.getLocation(SELF) == null`

Both `findLocation` methods are now shorter and structurally symmetric.

Unrelated question: would you be open to replacing the test's `ProcessBuilder("git", ...)` calls with the JGit API? It would remove the dependency on a `git` binary being on `PATH` and the risk of an unbounded `waitFor()`. JGit is already in the target platform; it would only require adding `org.eclipse.jgit` to the test bundle's `Require-Bundle`. I can do that here or in a separate PR — whichever you prefer.

EricBlanquer:
Status of the latest push (`fffc7e74`):

**4/7 checks green** — `Tests / Build` runs and passes (`skipped: 16, passed: 679`), so `testGitVersioningExtension` exercises the original NPE path and confirms the fix. `Maven`, `JavaDoc` and `eclipsefdn/eca` are all green.

**Compiler check is still red with 4 new warnings**, but they're not in files this PR touches — all 4 are pre-existing `Discouraged access: 'DependencyDepth' is not API` warnings in `org.eclipse.m2e.pde.ui/src/.../MavenTargetLocationWizard.java` at lines 235 and 348. The previous push had 6 new warnings, of which 2 came from the new test (accessing `workspace` and `monitor` fields of `AbstractMavenProjectTestCase`); those are now fixed by using `ResourcesPlugin.getWorkspace()` and `new NullProgressMonitor()` instead. The remaining 4 appear to be a baseline-comparison artifact of `warnings-ng` rather than anything introduced here.

I also discovered while iterating that the test needed an additional resource file (`.mvn/maven-git-versioning-extension.xml`) for the extension to actually activate — added in this push. Without it the test fails at import time before reaching the NPE path.

Let me know if you'd like me to address those `MavenTargetLocationWizard` warnings as part of this PR or in a separate one.

laeubi (approved):
Looks all fine and code structure is much better now!

laeubi:
@EricBlanquer thanks for the fix and your patience. The change will show up soon in the snapshot builds.

Issue #966, opened 2022-10-05
 Cannot invoke "org.eclipse.m2e.core.embedder.IMaven.lookup(java.lang.Class)" because the return value of "org.eclipse.m2e.core.internal.MavenPluginActivator.getMaven()" is null


sratz:
Consider the following:

```
<project
  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0">
  <modelVersion>4.0.0</modelVersion>
  <groupId>foo.bar</groupId>
  <artifactId>baz</artifactId>
  <version>0.1.0-SNAPSHOT</version>
  <packaging>pom</packaging>

  <build>
    <plugins>
      <plugin>
        <groupId>org.eclipse.tycho.extras</groupId>
        <artifactId>tycho-eclipserun-plugin</artifactId>
        <version>3.0.0</version>
        <executions>
          <execution>
            <id>run-director</id>
            <goals>
              <goal>eclipse-run</goal>
            </goals>
            <phase>package</phase>
            <configuration>
              <applicationsArgs>
                <args>-consoleLog</args>
                <args>-application</args>
                <args>org.eclipse.equinox.p2.artifact.repository.mirrorApplication</args>
                <args>-source</args>
                <args>mvn:org.eclipse.jetty:jetty-p2:11.0.7:zip:p2site</args>
                <args>-destination</args>
                <args>-destination</args>
                <args>dummy</args>
              </applicationsArgs>
              <jvmArgs>
                <jvmArg>-Declipse.p2.mirrors=false</jvmArg>
              </jvmArgs>
              <repositories>
                <repository>
                  <layout>p2</layout>
                  <url>https://download.eclipse.org/releases/2022-06</url>
                </repository>
              </repositories>
              <dependencies>
                <dependency>
                  <artifactId>org.apache.felix.scr</artifactId>
                  <type>p2-installable-unit</type>
                </dependency>
                <dependency>
                  <artifactId>org.eclipse.equinox.p2.repository.tools</artifactId>
                  <type>p2-installable-unit</type>
                </dependency>
                <dependency>
                  <artifactId>org.eclipse.equinox.p2.core.feature.feature.group</artifactId>
                  <type>p2-installable-unit</type>
                </dependency>
                <dependency>
[...]

PR #2168, opened 2026-05-07
Drop unused IWorkspace @Reference from PlexusContainerManager

wenytang-ms:
The IWorkspace field on PlexusContainerManager was added in aecf30541 (Jun 2022) when the multi-module-root lookup was implemented inline in this class. That logic was later moved to
MavenProperties.computeMultiModuleProjectDirectory(File), which calls ResourcesPlugin.getWorkspace() statically and lazily. Since then the field has had no readers — its only effect is to make SCR delay the activation of PlexusContainerManager (and transitively MavenImpl / IMaven) until the IWorkspace service from org.eclipse.core.resources is registered.

On a cold OSGi cache (e.g. first launch after an upgrade) the resources bundle / IWorkspace DS publication races with early callers of MavenPlugin.getMaven(), causing the IMaven ServiceTracker to return null and producing NPEs in embedders that touch m2e during their own initialization, e.g.

  Cannot invoke "org.eclipse.m2e.core.embedder.IMaven.getSettings()"
  because the return value of "org.eclipse.m2e.core.MavenPlugin.getMaven()"
  is null

Reported downstream as redhat-developer/vscode-java#3469 and matching the pattern described in #966. Removing the unused @Reference takes org.eclipse.core.resources off the IMaven activation critical path; the remaining static use of ResourcesPlugin.getWorkspace() in MavenProperties.computeMultiModuleProjectDirectory(File) is invoked lazily on actual project lookups, by which point the workspace is always available.

relateive issue 
- https://github.com/redhat-developer/vscode-java/issues/3469
- https://github.com/microsoft/vscode-java-pack/issues/1613

### DIFF
--- a/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/PlexusContainerManager.java
+++ b/org.eclipse.m2e.core/src/org/eclipse/m2e/core/internal/embedder/PlexusContainerManager.java
@@ -35,7 +35,6 @@ import org.slf4j.LoggerFactory;
 import com.google.inject.AbstractModule;
 
 import org.eclipse.core.resources.IResource;
-import org.eclipse.core.resources.IWorkspace;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.ILog;
 import org.eclipse.core.runtime.Platform;
@@ -88,9 +87,6 @@ public class PlexusContainerManager {
 
   @Reference
   private IMavenConfiguration mavenConfiguration;
-
-  @Reference
-  private IWorkspace workspace;
 
   @Deactivate
   void dispose() {


wenytang-ms:
@HannesWell please help review this PR.
the failure test case has no relative to this change, no matter from code layer or scenario.

fbricon (approved):

laeubi (changes requested):
> its only effect is to make SCR delay the activation of PlexusContainerManager (and transitively MavenImpl / IMaven) until the IWorkspace service from org.eclipse.core.resources is registered.

This is exactly the intend, as otherwise m2e starts to early and the IDE possibly crashes when a workspace chose dialog is used.


> On a cold OSGi cache (e.g. first launch after an upgrade) the resources bundle / IWorkspace DS publication races with early callers of MavenPlugin.getMaven()

Then these callers must be fixed to use the `IMaven` service instead. `MavenPlugin.getMaven()` is only for internal callers and only *after* the plugin has activated (during its lazy activation policy)

wenytang-ms:
@laeubi 
 Thanks for the review! Could you help me understand the activation-time dependency a bit better? A few observations that make
 me unsure the field is actually carrying the load you describe:
 
 
  1. The field was shadowed from day one. In aecf30541 (the commit that introduced both the @Reference and the only consumer at the 
time), the actual workspace use inside `computeMultiModuleProjectDirectory` was written as a local variable `IWorkspace workspace = 
ResourcesPlugin.getWorkspace();`, never reading the injected field. If the intent had been to use the DS-injected reference, the 
natural write would have been `this.workspace.getRoot()….`
  2. There is no @Activate / @Modified method on `PlexusContainerManager`. SCR therefore only assigns the reference and does no work at 
activation. The class's only direct IWorkspace use today (after 8351fdee moved the loop to 
`MavenProperties.computeMultiModuleProjectDirectory`) is reached lazily from public methods invoked by user code — by which point the 
workspace is already registered for any caller that has a File to look up.
  3. m2e itself has a precedent for the opposite policy in a component that actually uses `IWorkspace`. 
`ProjectRegistryManager#setWorkspace` (line 1179) is @Reference(policy=DYNAMIC, cardinality=OPTIONAL) even though it really does call 
`workspace.addSaveParticipant(...)` — explicitly opting out of the activation barrier you describe.
  4. The workspace-chooser-dialog scenario: that dialog is RCP UI, displayed after the resources bundle has registered IWorkspace. A 
mandatory @Reference IWorkspace would not gate activation against that scenario — it only gates against the small window before 
resources is up, which is exactly the window where headless callers (vscode-java, m2e-cli, etc.) hit the race in #3469.
 
 If keeping a barrier is preferred, would you be open to converting it to the same DYNAMIC/OPTIONAL pattern as ProjectRegistryManager?
 That would still let m2e bind the workspace when present, while not putting org.eclipse.core.resources on the IMaven activation
 critical path. Happy to amend the PR that way if you agree.

laeubi:
As this problem does not arise anywhere in the intended use (as a Eclipse Plugin), I would first look at the call stack in jdt-ls. Most likely the accessing internal API or bypass lazy-activation. The internal implementation of m2e is not API and nothing one can rely on, it might change at any time in any way.

Eclipse works "lazy" on many parts and especiall JDT is sensitive if it is activated to early therefore we must strictly avoid having *anything* from m2e to be loaded to early and a DS component of any kind will activate as soon as it is satisfied.

I'm also not really convinced to be bound to specific things of what I would call off-label use ... so unless you can reproduce the bug in a plain Eclipse Installation I would classify this not a bug of m2e.

wenytang-ms:
Thanks @laeubi - on reflection your point about avoiding early activation in RCP is fair, especially during the workspace-chooser window where IWorkspace is genuinely not yet registered. 
I underestimated the second-order effects of moving PlexusContainerManager off the IWorkspace barrier (other m2e components transitively reachable from IMaven do depend on a ready workspace). 
The right place for this fix is on the jdt.ls side, by treating `MavenPlugin.getMaven()` as nullable per your contract clarification. Closing this PR.
Thanks for the time spent reviewing.

PR #2183, opened 2026-05-30
Add version rule set for updating Maven target dependencies #2179

### LINKED ISSUE #2179: The "Update" operation in the target location should be configurable
This is something I'd like to invest some time in, over the next few weeks. Assuming that there is interest, of course. But first the use case:

Currently one can update Maven dependencies to their latest version via the `Update` button in the dependency dialog.

... which works fine, expect when projects decide to also publish milestone builds and release candidates, which I usually don't want to depend on in a productive environment. But because they have the highest version number, they are automatically chosen by this operation.

Image

Additionally, some projects also disregard semantic versioning, where the version with the "highest" version number isn't actually the "latest" version. (Below an example where `1.1.3.4.O` is semantically higher than `1.1.4c`, despite being older).

Image

Idea:

The `update` method of the `MavenTargetLocation` is adapted to accept a `Predicate<Version>` filter. Rather than returning the last entry of all available version, the update operation now returns the last entry of a filtered list of available versions.

https://github.com/eclipse-m2e/m2e-core/blob/01af4d55ac9745d625ac5cdb36a36e1db466e946/org.eclipse.m2e.pde.target/src/org/eclipse/m2e/pde/target/MavenTargetLocation.java#L454-L481

Via the preferences, the user can configure a list of regular expressions. The filter is then the union of all expressions and any version that matches at least one expression is ignored. For example `\d+\.\d+\.\d+-M\d` to filter out all milestone builds. Thoughts?

ptziegler:
This change introduces a new preference page with which users can import a rule set compatible with the "Versions Maven Plugin". When updating the Maven dependencies in the target definition, these rules are applied to filter out undesirable versions.

Currently, the update operation will always pick the latest version, which causes problems when the maintainers publish milestone builds and release candidates, or if they don't use semantic versioning.

Note that this change doesn't provide a way to edit the rule set directly in the IDE. Instead users have to modify the XML file, which then has to be re-imported.

Closes https://github.com/eclipse-m2e/m2e-core/issues/2179


ptziegler:
As a simple example, consider the following rule set:

```xml
<ruleset xmlns="https://www.mojohaus.org/VERSIONS/RULE/3.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
		 xsi:schemaLocation="https://www.mojohaus.org/VERSIONS/RULE/3.0.0 https://www.mojohaus.org/versions/versions-model/xsd/rule-3.0.0.xsd">
	<rules>
		<rule groupId="org.eclipse.platform" artifactId="org.eclipse.core.runtime">
			<ignoreVersions>
				<ignoreVersion type="range">[3.34.0, 4.0.0)</ignoreVersion>
			</ignoreVersions>
		</rule>
	</rules>
</ruleset>
```

When imported into the IDE, the new preference page will look like this:

image

If I now go to the target editor and click on the "Update" button, the version will be set to 3.33.100, as opposed to 3.34.200, because that's the latest version that is not excluded by this rule set.

image

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/ruleset/RegexUtils.java
```diff
@@ -0,0 +1,147 @@
...
+import static java.util.Optional.ofNullable;
+
+
+/**
+ * Utility methods to help with regex manipulation.
+ * <p>
+ * This class is a modified version of {@code org.codehaus.mojo.versions.utils.RegexUtils}.
+ * </p>
+ */
+public final class RegexUtils {
```
ptziegler: This and the RuleService class are for the most part copied over from the version plugin to make sure that the evaluation of the rule-set is identical. The project is using the Apache 2.0 license, so that should be fine. I'm not sure if there is a "better" solution to this.

laeubi:
@ptziegler Can we have a link in the editor to jump to the preference page like we have for the maven configuration?
Also I wonder if it would be possible to just specify a file directly where it is loaded from?

> This and the RuleService class are for the most part copied over from the version plugin to make sure that the evaluation of the rule-set is identical. The project is using the Apache 2.0 license, so that should be fine. I'm not sure if there is a "better" solution to this.

Have you tried if instead of copy the class just add the required jar to the target platform so we can use the original? As we already do similar for the archetype plugin it seems natural to do it here as well as Tycho is doing that as well it should even be vetted already.

ptziegler:
> Can we have a link in the editor to jump to the preference page like we have for the maven configuration?

Makes sense. I put it on my to-do list.

> Also I wonder if it would be possible to just specify a file directly where it is loaded from?

That's a good idea. And probably a lot less tedious than explicitly re-/importing them to keep both in-sync.

> Have you tried if instead of copy the class just add the required jar to the target platform so we can use the original? 

The artifact in question would be this, which lacks OSGi metadata. Should I add the instructions directly to the target platform or should the artifact be added to Orbit?

```xml
<dependency>
    <groupId>org.codehaus.mojo.versions</groupId>
    <artifactId>versions-common</artifactId>
    <version>2.21.0</version>
</dependency>
```

It'll then add a new dependency to the M2E plugin. Is that an issue?

laeubi:
> The artifact in question would be this, which lacks OSGi metadata. Should I add the instructions directly to the target platform or should the artifact be added to Orbit?

Just directly to the target platform.

> It'll then add a new dependency to the M2E plugin. Is that an issue?

Not in general I think (I assume you are talking about the `m2e.pde.target`?

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/MavenRuleSetPreferencePage.java
laeubi, line 1: I would expect this to reside in the `pde.target.ui` any particular reason we have it here?

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/util/M2EUIUtils.java
laeubi, line 1: This part seems more suitable to be placed in the `m2e.pde.target` plugin.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/MavenRuleSetPreferencePage.java
ptziegler: No reason other than the other preference pages also being there.

laeubi: As this is specific to the target support I think all code belongs to either the core or ui target bundle.

ptziegler:
I've updated the screenshots for the preference page and made it so that the user can only set the path for the rule-set file. I've also restructured everything so that the implementation is located in the PDE plugins and added the Versions artifacts to the target platform.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/MavenRuleSetPreferencePage.java
ptziegler: I've moved the preference page to `org.eclipse.m2e.pde.ui` because `org.eclipse.m2e.pde.target` doesn't depend on any UI classes.

org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/util/M2EUIUtils.java
ptziegler: This logic is now implemented via the `RuleSetMatcher`, which is located in `org.eclipse.m2e.pde.target`.

ptziegler:
The test failures seem unrelated. I'm not sure about the Maven issues:

> [WARNING] Mirror tool: Problems resolving provisioning plan.:
> Unable to satisfy dependency from org.eclipse.m2e.runtimes.category 1.0.0 to org.eclipse.equinox.p2.iu; org.eclipse.m2e.maven.runtime.38.iu [1.0.0,1.0.0].
> Unable to satisfy dependency from org.eclipse.m2e.runtimes.category 1.0.0 to org.eclipse.equinox.p2.iu; org.eclipse.m2e.maven.runtime.39.iu [1.0.0,1.0.0].

laeubi:
That warning is likely unrelated. I have seen this in the past but haven't had time to take a deeper look.

laeubi (approved):

laeubi:
@ptziegler very nice addition! I also like the viewer for the rule set you might want to consider making this an own editor that is able to display rule files in general but this is a completely different topic!

PR #2194, opened 2026-06-18
Ignore all mojo executions outside interesting phases

### LINKED ISSUE #2193: Plugin's lifecycle-mapping-metadata.xml applies irrespective of phase
A lifecycle-mapping-metadata.xml shipped with a Maven plugin like https://github.com/diffplug/spotless/blob/main/plugin-maven/src/main/resources/META-INF/m2e/lifecycle-mapping-metadata.xml is considered no matter to which phase the configured goal is bound, e.g.:

```
<lifecycleMappingMetadata>
    <pluginExecutions>
        <pluginExecution>
            <pluginExecutionFilter>
                <goals>
                    <goal>apply</goal>
                    <goal>check</goal>
                </goals>
            </pluginExecutionFilter>
            <action>
                <execute>
                    <runOnIncremental>true</runOnIncremental>
                    <runOnConfiguration>false</runOnConfiguration>
                </execute>
            </action>
        </pluginExecution>
    </pluginExecutions>
</lifecycleMappingMetadata>
```

would execute goal `apply` even if bound to phase `verify` in the POM.xml (which is usually not relevant for incremental builds).

As the default execution of maven plugins introduced in https://github.com/eclipse-m2e/m2e-core/blob/main/RELEASE_NOTES.md#mojos-without-a-mapping-are-now-executed-by-default-in-incremental-builds only ever execute by default up to phase `process-test-classes` the same should be replicated for plugins carrying an explicit mapping for consistency reasons.

Image

kwin:
This closes #2193

### DIFF
--- a/org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.m2e.core.ui;singleton:=true
-Bundle-Version: 2.4.201.qualifier
+Bundle-Version: 2.4.301.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-21
 Bundle-Name: %Bundle-Name
 Bundle-Vendor: %Bundle-Vendor
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/LifecycleMappingsViewer.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/preferences/LifecycleMappingsViewer.java
@@ -449,8 +449,8 @@ public class LifecycleMappingsViewer {
   }
 
   boolean isIgnoreMapping(MojoExecutionKey execution, List<IPluginExecutionMetadata> mappings) {
-    if(mappings == null || mappings.isEmpty()) {
-      return !LifecycleMappingFactory.isInterestingPhase(execution.lifecyclePhase());
+    if(!LifecycleMappingFactory.isInterestingPhase(execution.lifecyclePhase())) {
+      return true;
     }
     for(IPluginExecutionMetadata mapping : mappings) {
       if(PluginExecutionAction.ignore != mapping.getAction()) {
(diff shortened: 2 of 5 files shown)


kwin:
Still requires adjusting some tests.

laeubi:
I'm not completely sure this is right to skip just all executions that are "not interesting" here... Especially if tests needs adjustments it more seems that this behavior is actually intended here.

kwin:
There is no possibility for plugin specific lifecycles to filter for phases. I would argue that Eclipse should only try to replicate up to phase process-test-resources because other executions should only be performed on demand. The phase is determined by the consuming POM for good reasons and this should be considered.

laeubi:
> There is no possibility for plugin specific lifecycles to filter for phases. I would argue that Eclipse should only try to replicate up to phase process-test-resources because other executions should only be performed on demand.

This heavily depends on the use-case, e.g. many downstream mojos require at best `package` phase to be performed. Given the high volume of failing tests this likely breaks others and it does not aligns with your description of a mojo phase binding was changed! Instead it unconditionally skips executions now.

I would suggest to first provide a test that shows *what* actually is the faulty behavior so we have one failing test. The we might add code to fix this one problem (if we agree it actually is one) and it should not break existing test unless they are clearly fall into the same category.

For cases where one just want to skip one execution there are already ways for the user to do so.

kwin:
Often mojos are not bound by default (via package lifecycle) but only via explicit execution. Also not all Mojos have a default phase.

kwin:
> For cases where one just want to skip one execution there are already ways for the user to do so.

I tried with syntax https://eclipse.dev/m2e/documentation/release-notes-17.html#new-syntax-for-specifying-lifecycle-mapping-metadata and also via https://eclipse.dev/m2e/documentation/m2e-execution-not-covered.html#ignore-plugin-goal. Neither did overwrite the explicit plugin configuration from https://github.com/diffplug/spotless/blob/58e96f324e712d536c9c6a9622a5a9ddd7788c19/plugin-maven/src/main/resources/META-INF/m2e/lifecycle-mapping-metadata.xml#L26

laeubi:
> > For cases where one just want to skip one execution there are already ways for the user to do so.
> 
> I tried with syntax https://eclipse.dev/m2e/documentation/release-notes-17.html#new-syntax-for-specifying-lifecycle-mapping-metadata and also via https://eclipse.dev/m2e/documentation/m2e-execution-not-covered.html#ignore-plugin-goal. Neither did overwrite the explicit plugin configuration from https://github.com/diffplug/spotless/blob/58e96f324e712d536c9c6a9622a5a9ddd7788c19/plugin-maven/src/main/resources/META-INF/m2e/lifecycle-mapping-metadata.xml#L26

Than you should add a testcase for this so we can fix that instead first. Actually I would expect that this takes precedence over whatever is supplied elsewhere.

PR #2221, opened 2026-08-13
Add IMavenProjectFacade#getMojoParameterValue

laeubi:
Currently we require consumers to go through IMaven for getting Mojo parameters with a maven project. This is both cumbersome and binds us to the MavenProject API from maven-core what complicates Maven 4 migration.

This removes the roundtrip by adding the methods for getting a mojo parameter directly to the facade what later will allow us to exchange the backing implementation much easier.

### DIFF
--- a/org.eclipse.m2e.apt.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.apt.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.m2e.apt.core;singleton:=true
-Bundle-Version: 2.3.100.qualifier
+Bundle-Version: 2.3.200.qualifier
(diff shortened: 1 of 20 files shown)


HannesWell:
This reminds me of
- https://github.com/eclipse-m2e/m2e-core/pull/996

Maybe we should add the supporting methods to `IMavenProjectFacade` too instead of `IMaven`?

laeubi:
Yes I played around with this and pushed a draft now here:

- https://github.com/eclipse-m2e/m2e-core/pull/2223

The problem is the amount of legacy that one can not easily get around, but I think the `IMaven` is actually something that could/should die on the long term.

My "vision" here is that we actually have

- `IMavenProjectFacade` (+ supporting facades)
- `IMavenExecutionContext` (+ supporting facades)

what wrap the actual maven objects internally and where we have access to these with a service that is provided by a concrete runtime (e.g. Maven 3.10, Maven 3.9, Maven 4, ...) and one can install multiple of those in parallel. 

m2e would then select the runtime e.g. based on the model version or things like enforcer rules (like we do already for JVMs)

PR #2235, opened 2026-09-02
Split up RuleSetViewer into two tab items

ptziegler:
## What does this PR do?

The current implementation of the `RuleSetViewer` introduced with 3b36a6710696cafedcb1bbd1d5ac9511e7832e10 wastes a lot of available space.

- When a rule set contains almost exclusively global rules, half the preference page will be empty because this table and the table for the artifact rules are always equally sized. These tables are now split between two tab items. For each item, the viewer takes up all available space. The names for these viewers has been changed to "Global Rules" and "Artifact Rules".

- When an artifact rule only contains a single element, two items are created in the viewer. This combination is now combined into a single item, so that users don't have to expand the artifact item anymore to see the "ignore version" rule.

- Both viewers now also provide a quick filter. Note that both use a `FilteredList`, even though one viewer is technically a table, because the Eclipse 3.x UI plugin doesn't provide a `FilteredTable` class.

## How was this tested?

Locally using the following ruleset:

```xml
<ruleset xmlns="https://www.mojohaus.org/VERSIONS/RULE/3.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="https://www.mojohaus.org/VERSIONS/RULE/3.0.0 https://www.mojohaus.org/versions/versions-model/xsd/rule-3.0.0.xsd">
    <rules>
        <rule groupId="net.bytebuddy" artifactId="*">
            <ignoreVersions>
                <ignoreVersion type="regex">.+-jdk\d</ignoreVersion>
            </ignoreVersions>
        </rule>
    </rules>
    <ignoreVersions>
        <ignoreVersion type="regex">(.+-SNAPSHOT|.+-M\d)</ignoreVersion>
        <ignoreVersion type="regex">.+A</ignoreVersion>
        <ignoreVersion type="regex">.+B</ignoreVersion>
        <ignoreVersion type="regex">.+C</ignoreVersion>
        <ignoreVersion type="regex">.+D</ignoreVersion>
        <ignoreVersion type="regex">.+E</ignoreVersion>
        <ignoreVersion type="regex">.+F</ignoreVersion>
        <ignoreVersion type="regex">.+G</ignoreVersion>
        <ignoreVersion type="regex">.+H</ignoreVersion>
        <ignoreVersion type="regex">.+I</ignoreVersion>
        <ignoreVersion type="regex">.+J</ignoreVersion>
        <ignoreVersion type="regex">.+K</ignoreVersion>
        <ignoreVersion type="regex">.+L</ignoreVersion>
    </ignoreVersions>
</ruleset>
```

## Checklist

- [x] I have searched existing PRs/issues and this is not a duplicate.
- [ ] `mvn clean verify` (add `-Pits` for integration tests) passes locally.
- [x] Bundle versions are bumped where required (see "Version bump" in [CONTRIBUTING.md](../CONTRIBUTING.md)).
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](http://www.eclipse.org/legal/ECA.php).

### DIFF
--- a/org.eclipse.m2e.pde.ui/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.pde.ui/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: M2E PDE Integration UI
 Bundle-SymbolicName: org.eclipse.m2e.pde.ui;singleton:=true
-Bundle-Version: 2.1.200.qualifier
+Bundle-Version: 2.1.300.qualifier
 Export-Package: org.eclipse.m2e.pde.ui.target.editor;x-friends:="org.eclipse.m2e.swtbot.tests"
 Automatic-Module-Name: org.eclipse.m2e.pde.ui
 Bundle-RequiredExecutionEnvironment: JavaSE-21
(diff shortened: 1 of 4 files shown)


ptziegler:
This is how it currently looks. For the rule-set in our product, it's even worse. All of the enclosed area is effectively dead space and you also can't easily see the rule I have for ByteBuddy.

image

With this change, the page now looks like this:

image

image

Issue #2008, opened 2025-07-09
Deadlock during eclipse startup


ahoehma:
Hello,

I don't know why but suddenly I found this deadlock here ... happens only in a specific workspace. Was working for months without major problems.

`Found one Java-level deadlock:
=============================
"main":
  waiting to lock monitor 0x000001d48eedbf10 (object 0x000000066d1017d0, a org.eclipse.m2e.core.internal.embedder.MavenImpl$MavenSettings),
  which is held by "org.eclipse.jdt.internal.ui.text.JavaReconciler"

"org.eclipse.jdt.internal.ui.text.JavaReconciler":
  waiting to lock monitor 0x000001d490cf8ec0 (object 0x000000060881d570, a java.util.HashMap),
  which is held by "main"

Java stack information for the threads listed above:
===================================================
"main":
        at org.eclipse.m2e.core.internal.embedder.MavenImpl$MavenSettings.getSettings(MavenImpl.java:1083)
        - waiting to lock <0x000000066d1017d0> (a org.eclipse.m2e.core.internal.embedder.MavenImpl$MavenSettings)
        at org.eclipse.m2e.core.internal.embedder.MavenImpl.getSettings(MavenImpl.java:279)
        at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.createExecutionRequest(MavenExecutionContext.java:207)
        at org.eclipse.m2e.core.internal.embedder.PlexusContainerManager.loadCoreExtensions(PlexusContainerManager.java:350)
        at org.eclipse.m2e.core.internal.embedder.PlexusContainerManager.newPlexusContainer(PlexusContainerManager.java:225)
        at org.eclipse.m2e.core.internal.embedder.PlexusContainerManager.aquire(PlexusContainerManager.java:179)
        - locked <0x000000060881d570> (a java.util.HashMap)
        at org.eclipse.m2e.core.internal.embedder.PlexusContainerManager.getComponentLookup(PlexusContainerManager.java:207)
        at org.eclipse.m2e.core.internal.project.registry.ProjectRegistryManager.getPom(ProjectRegistryManager.java:245)
        at org.eclipse.m2e.core.internal.project.registry.ProjectRegistryManager.create(ProjectRegistryManager.java:204)
        at org.eclipse.m2e.core.internal.project.registry.MavenProjectManager.create(MavenProjectManager.java:110)
[...]

Issue #1749, opened 2024-04-26
Deadlock during startup


mtheiss:
I have some editors open and m2e gets deadlocked.

Short threaddump
```
"main":
	at org.eclipse.m2e.core.internal.embedder.MavenImpl$MavenSettings.getSettings(MavenImpl.java:1083)
	- waiting to lock <0x000000060d417400> (a org.eclipse.m2e.core.internal.embedder.MavenImpl$MavenSettings)
	at org.eclipse.m2e.core.internal.embedder.MavenImpl.getSettings(MavenImpl.java:279)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.createExecutionRequest(MavenExecutionContext.java:207)
	at org.eclipse.m2e.core.internal.embedder.PlexusContainerManager.loadCoreExtensions(PlexusContainerManager.java:347)
	at org.eclipse.m2e.core.internal.embedder.PlexusContainerManager.newPlexusContainer(PlexusContainerManager.java:222)
	at org.eclipse.m2e.core.internal.embedder.PlexusContainerManager.aquire(PlexusContainerManager.java:176)
	- locked <0x0000000600d5c388> (a java.util.HashMap)
	at org.eclipse.m2e.core.internal.embedder.PlexusContainerManager.getComponentLookup(PlexusContainerManager.java:204)
"org.eclipse.jdt.internal.ui.text.JavaReconciler":
	at org.eclipse.m2e.core.internal.embedder.PlexusContainerManager.aquire(PlexusContainerManager.java:146)
	- waiting to lock <0x0000000600d5c388> (a java.util.HashMap)
	at org.eclipse.m2e.core.internal.embedder.MavenImpl.getPlexusContainer(MavenImpl.java:934)
	at org.eclipse.m2e.core.internal.embedder.MavenImpl.lookup(MavenImpl.java:991)
	at org.eclipse.m2e.core.internal.embedder.MavenImpl$MavenSettings.getSettings(MavenImpl.java:1117)
	- locked <0x000000060d417400> (a org.eclipse.m2e.core.internal.embedder.MavenImpl$MavenSettings)
	at org.eclipse.m2e.core.internal.embedder.MavenImpl.getSettings(MavenImpl.java:279)

```
[threaddump-m2e-deadlock.txt](https://github.com/eclipse-m2e/m2e-core/files/15125886/threaddump-m2e-deadlock.txt)

PR #2238, opened 2026-09-04
Do not read the Maven settings while activating the m2e UI bundle

vogella:
The *Update Project* contribution in the Project menu carries `forcePluginActivation="true"`, so the workbench starts `org.eclipse.m2e.core.ui` on the UI thread while it computes menu visibility during startup. The activator then cleaned up the caches of m2e 1.x, and the `getLocalRepository()` call needed for that entered the Maven settings lock that `RepositoryRegistryUpdateJob` holds while Guice builds the Plexus container. Sampled thread dumps of an SDK startup show the UI thread blocked there for about 600 ms.

The cleanup was added in October 2022 with a TODO to remove it once workspaces that need it are unlikely, which four years later it is. Removing it makes the activation cheap and takes the Maven settings out of the startup path, without changing when the bundle starts.

This is separate from #1749, #2008 and #2217, which address the order in which the two locks are taken rather than who takes them during startup.

`Bundle-Version` is bumped to 2.4.301 because the baseline check asks for it after the inner classes of the deleted job disappear.

### DIFF
--- a/org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF
+++ b/org.eclipse.m2e.core.ui/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.m2e.core.ui;singleton:=true
-Bundle-Version: 2.4.201.qualifier
+Bundle-Version: 2.4.301.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-21
 Bundle-Name: %Bundle-Name
 Bundle-Vendor: %Bundle-Vendor
--- a/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/M2EUIPluginActivator.java
+++ b/org.eclipse.m2e.core.ui/src/org/eclipse/m2e/core/ui/internal/M2EUIPluginActivator.java
@@ -14,11 +14,6 @@
 
 package org.eclipse.m2e.core.ui.internal;
 
-import java.io.File;
-import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.ServiceReference;
 import org.osgi.util.tracker.ServiceTracker;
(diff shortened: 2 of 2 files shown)


vogella:
@HannesWell please review

laeubi (changes requested):
@vogella please follow https://github.com/eclipse-m2e/m2e-core/blob/main/CONTRIBUTING.md#-using-ai--genai-tools-to-contribute

This is a perfect example of an AI generate junk we want to avoid. The PR description/Commit title has nothing to do with the actual change (what simply removes some code) and violates the "Prefer signal over prose" rule by being exhaustively elaborating about things like `forcePluginActivation="true"` what is 

1. Just here by accident - any other activation of the plugin would trigger the activator
2. It was already mentioned at the issue that the first priority would be to find out **why** platform actually calls this codepath **on start-up** when no menu is show at all.

Please also be aware that in the future we might close such PRs without further comment if they violate the contribution rules here.

As a nit-pick the version increment would be 201 -> 300... as I'm sure **you** (not the agent that impersonates you) are well aware off...

laeubi:
Also in general your behavior to just ignore concerns raised by other (me in this case) by simply asking someone else (usually @HannesWell ) for a review about a topic we already have discussed in some length quite feels quite unfriendly. 

Neither me nor Hannes like to be wasting time reading lengthy AI generated messages that don't come to a point and then feel like you are more seeking for confirmation in the hope that the concerns are "overruled" than for a honest review.

vogella:
The code is from @HannesWell that is why I ask him. forcePluginActivation="true" slip into this PR accidently from my m2e fork which I use in my Eclipse build.

Sorry for that, I redo the deletion for the 2022 migration code again at some time in a change, and if Hannes want I can also place the migration into a job. The forcePluginActivation="true" item is IMHO independent of this and I do not plan to investigate. 600ms of ~3800ms in Eclipse for an empty workspace is in my opinion a lot.

laeubi:
> The code is from @HannesWell that is why I ask him. 

So why not "ask him" in the PR description directly if he has concerns about the *deletion* in a first place instead of something with 90% usless information and then adding another comment? Just to make clear, the change itself is likely fine, the form it is presented here is not.

> The forcePluginActivation="true" item is IMHO independent of this and I do not plan to investigate

Again `forcePluginActivation="true"` is not related here (and was not even part of the change its just hallucination of the AI from the context-window), so there is nothing that needs "investigation".

> 600ms of ~3800ms in Eclipse for an empty workspace is in my opinion a lot.

I do not know how to make it clearer but trying once again:

If you have an empty workspace **and** a fresh start of Eclipse, m2e should *never* be activated so the activator should *never run* (regardless of `forcePluginActivation="true"`) this is a bug (or regression) of Eclipse platform, because ther is nothing there where it would require to evaluate the `PropertyTester` at all!
The only thing where it _might_ block would be if you open the menu what then will possibly activate m2e (as intended!).

PR #2241, opened 2026-09-12
Update dependencies

merks:
This just uses Orbit generated update.  Maybe there are good reasons some things aren't updated...


HannesWell:
Thanks for the heads up. Some changes indeed should not yet be applied, i.e. the archetype update since it's part of another change.
I've in-cooperated this into a greater change in
- https://github.com/eclipse-m2e/m2e-core/pull/2242

but that needs investigation since it cause a lot of failures in the refactoring area.

